PR #26, opened 2015-01-21
Add a contribution guide.

Raynos:
This is the same guideline that is used in multiple places.

 - https://github.com/request/request/blob/master/CONTRIBUTING.md
 - https://github.com/rvagg/node-levelup/blob/master/CONTRIBUTING.md
 - sirvice

This communicates to contributors how to work with the project.

More comments inline

reviewers: @mranney @jwolski 

cc: @sectioneight

### DIFF
--- a/CONTRIBUTION.md
+++ b/CONTRIBUTION.md
@@ -1,0 +1,48 @@
+# This is an OPEN Open Source Project
+
+## What?
+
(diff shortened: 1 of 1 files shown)


mranney:
I updated the contribution file with what I think it should be. If it makes sense to add external contributors, I'm fine with that. In this file, I'd rather focus on the basics of what makes a good contribution.

mranney:
LGTM

PR #29, opened 2015-02-02
Add missing run-parallel package

ynachiket:
What:
Adds run-parallel to package.json.

Why:
Package is missing and is needed by test/send.js


Raynos:
lgtm. thanks :)

PR #31, opened 2015-02-03
Add benchmark comparison tooling

jcorbin:
- refactor benchmark to support collecting results
- add compare scripts
- add package.json "compare_to_master" target

### DIFF
--- a/benchmarks/README.md
+++ b/benchmarks/README.md
@@ -1,0 +1,77 @@
+# How to read comparison output
+
+The comparison script runs the benchmark suite against two different git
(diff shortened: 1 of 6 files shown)


Raynos:
lgtm. Can we do explanation of how to use it ?

We run it and it prints positive or negative ?

If it prints a regression how do we find out what regressed ? 

Can we add sample output or screenshots to the README or something ? Although maybe a seperate markdown document if thats preferred.

jcorbin:
Okay I:
- added a README about comparison with an example
- iterated on the math / setup

PR #32, opened 2015-02-08
Decoupled tchannel config and creation and added unit tests

ynachiket:
What:

1. Moved serverSocket.listen out of the constructor to decouple configuration from connection.
2. Added this.listening option TChannel which would be true by default so existing tests would work. Decoupling config and connection creation can be achieved by setting this to false.
3. Added unit tests for methods on TChannel testing both positive as well as throwing scenarios.

Why:

1. While trying to understand the code flow, I usually try adding unit tests. I realized that as we were creating a connection (which would persist), unit testing was unnecessarily  hard.
2. From a design point of view, I think it makes sense to move establishing the connection out of the constructor.
3. Improved code coverage: 

Master:
screen shot 2015-02-07 at 4 46 43 pm

This branch:
screen shot 2015-02-07 at 4 46 16 pm

### DIFF
--- a/index.js
+++ b/index.js
@@ -64,15 +64,15 @@ function TChannel(options) {
 
 	this.endpoints = Object.create(null);
 	this.destroyed = false;
-
-	// TODO do not create a tcp server in the constructor
-	// Doing IO in a constructor is bad. Move into a method.
+	// to provide backward compatibility. 
+	this.listening = this.options.listening === false ?
+	  false : true;
 	this.serverSocket = new net.createServer();
-	// TODO do not call listen by default. We might want to
-	// seperate allocating a server from calling listen on it.
-	// We can module.exports a `createTchannel()` helper that
-	// calls listen() by default.
-	this.serverSocket.listen(this.port, this.host);
+
+	if (this.listening) {
+		this.createTchannel();
+	}
+
 	this.serverSocket.on('listening', function () {
 		self.logger.info(self.name + ' listening');
 		if (!self.destroyed) {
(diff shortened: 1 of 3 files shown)


index.js
```diff
@@ -158,10 +176,6 @@ TChannel.prototype.addPeer = function (name, connection) {
 		connection = this.makeOutConnection(name);
 	}
 
-	if (!connection) {
```
ynachiket: This seemed like a copy pasta.

index.js
```diff
@@ -94,6 +94,24 @@ function TChannel(options) {
 }
 require('util').inherits(TChannel, require('events').EventEmitter);
 
+// Decoulping config and creation from the constructor.
+// This also allows us to better unit test the code as the test process
+// is not blocked by the listening connections
+TChannel.prototype.createTchannel = function () {
```
Raynos: Let's call this `.createServer()` or something to imply that it is creating a TCP server.

index.js
```diff
@@ -64,15 +64,15 @@ function TChannel(options) {
...
-	// seperate allocating a server from calling listen on it.
-	// We can module.exports a `createTchannel()` helper that
-	// calls listen() by default.
-	this.serverSocket.listen(this.port, this.host);
+
+	if (this.listening) {
+		this.createTchannel();
+	}
+
 	this.serverSocket.on('listening', function () {
```
Raynos: if `this.listening` is false then `this.serverSocket` is undefined and this will throw. We need to refactor this as well.

Raynos: Wait, you've only moved the listen() call out not anything else.

test/tchannel.js
```diff
@@ -0,0 +1,169 @@
+var test = require('tape');
+var TChannel = require('../index.js');
+
+var serverOptions = {host: '127.0.0.1', port: 4040, listening: false},
+  clientOptions = {host: '127.0.0.1', port: 4041, listening: false},
```
Raynos: style nitpicks:

For consistency it would be great if we had:

 - one `var` statement per variable
 - four spaces or tabs.

index.js
Raynos: Actually since this only listens we should just call this method `.listen()`

ynachiket: @Raynos  changed in  05549f0

test/tchannel.js
ynachiket: @Raynos changed in  05549f0

index.js
ynachiket: @Raynos yes I figured , the event listeners could be left here. Someone looking at the constructor can still understand what's in store and we still achieve the decoupling by just moving the listen out.
Thoughts?

mranney:
@Raynos @ynachiket I'm working on tchannel v2, which involves some significant changes. Can you guys wrap this PR up and get it merged so my branch doesn't get too far out of whack?

Raynos:
lgtm. Will merge.

Raynos:
@mranney master is ready for you to rebase on :)

Issue #33, opened 2015-02-18
memory leak when server callback is never called.


Raynos:
Line 575 allocates a `TChannelServerOp` ( https://github.com/uber/tchannel/blob/master/index.js#L575 ).

We only a de-allocate it from the `inOps` object when the callback get's called.

If you implement a tchannel endpoint that never calls the callback then it will leak memory as `delete inOps[...]` never happens.

 - We need more logging in ( https://github.com/uber/tchannel/blob/master/index.js#L577 )
 - We need to somehow time the server out and de-allocate the `inOps` object or have a periodic garbage collector (i.e. every minute, clear out inOps). Maybe update `onTimeoutCheck` to also purge the `inOps` object.

cc @jwolski @mranney @Matt-Esch

Raynos:
Not calling the callback is also a programmer error.

We should get statsd instrumentation on tchannel or our usage of tchannel in ringpop to identify how many inOps timeout or how many callbacks do not fire.

Raynos:
We also need to de-allocate the `inOps` in that `sendResponse` function in the error case. Currently we just early return and leak an `inOps`.

We should clear `inOps` in `self.closing` for good measure.

Raynos:
The ringpop proxy implementation probably does not call the callback in all cases.

PR #34, opened 2015-02-18
do not leak in operations

Raynos:
This fixes a subset of https://github.com/uber/tchannel/issues/33

It should reduce the memory leak, but does not fix the fact
that something caused it in the first place.

We also have to get insight in who doesnt call the callback.

However at least the application will be stable and isolated
from that one bad tchannel req/res pair.

reviewers: @iproctor @jwolski

cc: @sh1mmer

### DIFF
--- a/index.js
+++ b/index.js
@@ -578,9 +578,8 @@ TChannelConnection.prototype.handleReqFrame = function (reqFrame) {
 
 	this.inPending++;
 	var opCallback = responseFrameBuilder(reqFrame, sendResponse);
-	var op = this.inOps[id] = new TChannelServerOp(this, handler, reqFrame, {
-		
-	}, this.channel.now(), opCallback);
+	var op = this.inOps[id] = new TChannelServerOp(this,
+		handler, reqFrame, this.channel.now(), {}, opCallback);
 
 	function sendResponse(err, handlerErr, resFrame) {
 		if (err) {
--- a/index.js
+++ b/index.js
@@ -419,15 +419,16 @@ TChannelConnection.prototype.onTimeoutCheck = function () {
 		return;
 	}
 
-	checkOpsForTimeout(this, this.outOps, 'outPending');
-	checksInOpsForTimeout(this, this.inOps);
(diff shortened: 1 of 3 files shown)


Raynos:
reviewers: @mranney

index.js
```diff
@@ -572,7 +578,9 @@ TChannelConnection.prototype.handleReqFrame = function (reqFrame) {
 
 	this.inPending++;
 	var opCallback = responseFrameBuilder(reqFrame, sendResponse);
-	var op = this.inOps[id] = new TChannelServerOp(this, handler, reqFrame, opCallback);
+	var op = this.inOps[id] = new TChannelServerOp(this, handler, reqFrame, {
+
```
iproctor: looks like you're passing options (why this whitespace) then start, but the constructor takes start then options

index.js
```diff
@@ -418,37 +419,64 @@ TChannelConnection.prototype.onTimeoutCheck = function () {
...
 
-	var opKeys = Object.keys(this.outOps);
-	var now = this.channel.now();
+	checkOpsForTimeout(this, this.outOps, 'outPending');
+	checksInOpsForTimeout(this, this.inOps);
+
+	this.startTimeoutTimer();
+};
+
+function checksInOpsForTimeout(tconn, ops) {
```
jcorbin: def method(self) in any language, please make these methods

index.js
```diff
@@ -418,37 +419,64 @@ TChannelConnection.prototype.onTimeoutCheck = function () {
 		return;
 	}
 
-	var opKeys = Object.keys(this.outOps);
-	var now = this.channel.now();
+	checkOpsForTimeout(this, this.outOps, 'outPending');
```
jcorbin: this as first arg -> method

jcorbin: also call it "checkOutOps" since the other one is "checkInOps"

Raynos:
I'm sorry they were not methods. They are now methods.

iproctor:
lgtm

PR #37, opened 2015-02-22
Factor out identify routines

jcorbin:
Half of the structure needed for v2 init (completes {send,handle} X {Request,Response})


Raynos:
lgtm.

PR #38, opened 2015-02-22
Refactor frame type dispatch

jcorbin:
Straight up trivial refactor to make changes clearer as we iterate on types and their handlers.


Raynos:
lgtm.

PR #41, opened 2015-02-23
More refactoring ahead of v2

jcorbin:
Mostly tweaking encapsulation barriers and code flow so that the eventual "use v2" diff is cleaner.

### DIFF
--- a/index.js
+++ b/index.js
@@ -605,16 +605,10 @@ TChannelConnection.prototype.handleReqFrame = function handleReqFrame(reqFrame) 
     }
 
     self.inPending++;
-    var opCallback = responseFrameBuilder(reqFrame, sendResponse);
     var op = self.inOps[id] = new TChannelServerOp(self,
-        handler, reqFrame, self.channel.now(), {}, opCallback);
+        handler, reqFrame, self.channel.now(), {}, sendFrame);
 
-    function sendResponse(err, handlerErr, resFrame) {
-        if (err) {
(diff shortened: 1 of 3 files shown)


index.js
```diff
@@ -385,6 +391,23 @@ function TChannelConnection(channel, socket, direction, remoteAddr) {
...
 require('util').inherits(TChannelConnection, require('events').EventEmitter);
 
+TChannelConnection.prototype.onParserError = function onParserError(err) {
+    var self = this;
+    self.channel.logger.error('tchannel parse error', {
+        remoteName: self.remoteName,
+        localName: self.channel.name,
+        error: err
+    });
+    // TODO should we close the connection?
```
Raynos: yes we should.

v1/frame.js
```diff
@@ -98,6 +98,19 @@ TChannelFrame.prototype.checksum = function () {
     return csum;
 };
 
+TChannelFrame.prototype.verifyChecksum = function () {
+    var self = this;
+    var err = null;
+    var actual = self.checksum();
+    var expected = self.header.csum;
+    if (expected !== actual) {
+        err = new Error('tchannel checksum validation failed');
```
Raynos: we should set something like `err.type = 'tchannel.frame.failed-checksum'`.

In the future we may want to handle checksum failures differently from other parser failures.

Raynos:
lgtm.

PR #42, opened 2015-02-23
Improve tests

jcorbin:
Some log tweaks, but most concerningly, some ordering changes; while developing towards v2, I kept hitting non-determinism in the tests:
- sometimes but not always the first connection coming into a newly listening server would be lost
- these changes _seemed_ to help, but the problem never fully went away
- I only got the problem to reliably go away by deferring run of primary test logic for 1ms after alloc cluster returned

I don't feel confident yet that the non-determinism wasn't a bug introduced in v2 since I've not got it past the full suite yet.

So while I'm certainly not going to propose adding magic delays, these changes seem sensible.

### DIFF
--- a/index.js
+++ b/index.js
@@ -64,11 +64,12 @@ function TChannel(options) {
     // to provide backward compatibility.
     self.listening = self.options.listening === false ?
       false : true;
-    self.serverSocket = new net.createServer();
-
-    if (self.listening) {
-        self.listen();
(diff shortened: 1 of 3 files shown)


Raynos:
lgtm.

Raynos:
Adding a connection listener before calling `.listen()` is a good idea.

PR #43, opened 2015-02-23
Correct frame id overflow logic

jcorbin:


index.js
```diff
@@ -406,8 +406,7 @@ TChannelConnection.prototype.onParserError = function onParserError(err) {
 
 TChannelConnection.prototype.nextFrameId = function nextFrameId() {
     var self = this;
-    // TODO this needs to wrap around with 32-bit uint space
-    ++self.lastSentFrameId;
+    self.lastSentFrameId = (self.lastSentFrameId + 1) % 0xffffffff;
```
Raynos: octal literals are banned in strict mode. I recommend `4294967295` and a comment about 32bits

Raynos: Actually `Math.pow(2, 32) - 1` might be even nicer.

jcorbin: hex !== octal...

this form isn't ambiguous, so if it's linted against, we should fix the linter

also these files are using strict mode already so that's not it

Raynos: Apparently this is a hex literal, not an octal literal.

Raynos:
lgtm.

PR #44, opened 2015-02-23
Drop the auto-JSON confusion

jcorbin:
It's easy to drop this in the v2 implementation, so let's make the breaking change.


jcorbin:
Note: even if we decide not to break semantics, we should move this concern out of the parsing modules and into the edge (e.g. TChannel.send) as much as possible.

Raynos:
lgtm. I'm :+1: for no magic in tchannel, your sending strings or buffers. Just like a TCP socket.

jcorbin:
Done into v2 in #59

PR #45, opened 2015-02-23
V2 support

jcorbin:
In v2, what we used to call a "name" is now very solidly the "hostPort" and we've added a new "processName".

I still don't have a great story on the processName yet, but this can work.

Plus having a field called name alongside "processName", let alone forbid "serviceName" if we ever have it at the channel layer is heavily confusing.

### DIFF
--- a/index.js
+++ b/index.js
@@ -42,7 +42,7 @@ function TChannel(options) {
     self.host = self.options.host || '127.0.0.1';
     // TODO do not default the port.
     self.port = self.options.port || 4040;
-    self.name = self.host + ':' + self.port;
+    self.hostPort = self.host + ':' + self.port;
     self.random = self.options.random ?
         self.options.random : globalRandom;
     self.setTimeout = self.options.timers ?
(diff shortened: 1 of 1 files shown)


jcorbin:
I could by the way see calling it "peerName" if we must shed, but I think "hostPort" does for now.

jcorbin:
Also I'm not sure how much of a "breaking" change this is, since external users may depend on `.name`... so we either shim it back in to land early, or defer landing this until right before 2.0 tag cut.

Raynos:
`.hostPort` seems fine to me.

Raynos:
My general breaking change heuristic is that if it's in the README or the type definition ( https://github.com/uber/tchannel/blob/master/docs.jsig#L54 ) then a breaking change requires a new major version.

Since its not in the README and only in the type definition its a grey area.

Raynos:
lgtm.

jcorbin:
branch pocalypse

PR #46, opened 2015-02-23
Add hex dump of any buffer from parser errors

jcorbin:


jcorbin:
We may not want this as is since it jams a pretty printed string into a log message, but the utility of it is invaluable for debugging.

index.js
```diff
@@ -396,7 +397,12 @@ require('util').inherits(TChannelConnection, require('events').EventEmitter);
 
 TChannelConnection.prototype.onParserError = function onParserError(err) {
     var self = this;
-    self.channel.logger.error('tchannel parse error', {
+    var mess = 'tchannel parse error';
+    if (err.buffer) {
+        mess += '\n' + hexer(err.buffer);
```
Raynos: consider adding a second `logger.debug()` message for the buffer.

Remember that errors go to sentry.

index.js
```diff
@@ -28,6 +28,7 @@ var globalNow = Date.now;
 var globalRandom = Math.random;
 var net = require('net');
 var inspect = require('util').inspect;
+var hexer = require('hexer');
```
kriskowal: reviewed hexer. legitimate.

jcorbin:
obsolete, will be redoing as needed

PR #47, opened 2015-02-23
More visibility to chase test race condition

jcorbin:
- logs from resetAll
- logger out of allocCluster for tests to use themselves

### DIFF
--- a/index.js
+++ b/index.js
@@ -524,6 +524,18 @@ TChannelConnection.prototype.resetAll = function resetAll(err) {
     var self = this;
     if (self.closing) return;
     self.closing = true;
+
+    var inOpKeys = Object.keys(self.inOps);
+    var outOpKeys = Object.keys(self.outOps);
+
+    self.logger[err ? 'warn' : 'info']('resetting all connections', {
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -524,6 +524,18 @@ TChannelConnection.prototype.resetAll = function resetAll(err) {
     var self = this;
     if (self.closing) return;
     self.closing = true;
+
+    var inOpKeys = Object.keys(self.inOps);
+    var outOpKeys = Object.keys(self.outOps);
+
+    self.logger.info('resetting all connections', {
```
jcorbin: may should be a warn

Raynos: consider warning if there is an err argument and assume that no err === clean shutdown and err === not clean shutdown.

Raynos:
lgtm.

jcorbin:
hacked in a level switch for now, would be better on a `logger.log(level, msg, extra)` api, but the null / echo loggers we hacked into TChannel don't suffice.  So leaving like this pending proper logtron-pocalypse.

PR #48, opened 2015-02-23
Make cluster tests deterministic

jcorbin:
Remove non-determinism by only running test once all channels are listening.


Raynos:
lgtm.

PR #51, opened 2015-02-23
Fix and better test subtle inOps edge case

jcorbin:
- fix inOps leak when handler calls its callback synchronously (as part of RHS of inOps assign)
- test against assumed behavior underscored by the recent regression-inops-leak test by @Raynos 
- always mutate inOps and inPending, even in closing case
- add a marked TODO warn log


Raynos:
lgtm. partially reviewed in person.

kriskowal:
verified, fixes tests locally

PR #53, opened 2015-02-23
emit an endpoint event

### LINKED ISSUE #52: emit request event on tchannel
It would be nice to emit an event on `tchannel` itself before we call an endpoint handler, at least with arg1.

This allows implementing an access log

Raynos:
This closes #52

Emitting an endpoint event allows us to write a tchannel
access logger.

reviewers: @jcorbin @jwolski

cc: @jsu1212

### DIFF
--- a/index.js
+++ b/index.js
@@ -625,7 +625,7 @@ TChannelConnection.prototype.handleReqFrame = function handleReqFrame(reqFrame) 
             cb(err, null, null);
         };
 
-        self.channel.emit('missingEndpoint', {
+        self.channel.emit('endpoint.missing', {
             name: name
         });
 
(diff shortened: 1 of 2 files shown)


jwolski:
wonderful. lgtm

index.js
```diff
@@ -613,6 +613,10 @@ TChannelConnection.prototype.handleReqFrame = function handleReqFrame(reqFrame)
 
     var handler = self.localEndpoints[name] || self.channel.endpoints[name];
 
+    self.channel.emit('endpoint', {
```
jcorbin: Let's please not emit for:
- invalid handler
- local handler

In fact you should probably emit misses specially.

Let's for now do something like:
- inside the typeof not function branch, emit something (maybe same with diff args?)
- add an `else if (handler !== self.localEndpoints[name]) { /* emit */` branch; in v2 this becomes simply an `else`

Raynos:
Addressed feedback from @jcorbin

jcorbin:
lgtm; but let's nail the semantics down more in v2 since this is part of the public api

PR #57, opened 2015-02-23
Handler refactor

jcorbin:
Trivial handler refactor ahead of v2.

I plan to shape up a proper `vX/handler` structure to decouple these concerns away from the public API and set a path for future compat work.


Raynos:
lgtm.

PR #58, opened 2015-02-23
V2 support

jcorbin:
In v2, what we used to call a "name" is now very solidly the "hostPort" and we've added a new "processName".

I still don't have a great story on the processName yet, but this can work.

Plus having a field called name alongside "processName", let alone forbid "serviceName" if we ever have it at the channel layer is heavily confusing.


jcorbin:
merging per review in #45

PR #60, opened 2015-02-24
V2 parser

jcorbin:
First round at a V2 parser:
- relies on a monadic lib/{read,write} to solve for correctness first and minimize duplication
- parts implemented:
  - frame
  - init req / res
  - call req / res

Tested within the dev.josh branch, works well enough to merge and go forward from here.

### DIFF
--- a/test/v2/frame.js
+++ b/test/v2/frame.js
@@ -1,5 +1,5 @@
 // Copyright (c) 2015 Uber Technologies, Inc.
-//
+
 // Permission is hereby granted, free of charge, to any person obtaining a copy
 // of this software and associated documentation files (the "Software"), to deal
 // in the Software without restriction, including without limitation the rights
@@ -20,12 +20,20 @@
 
 'use strict';
 
-require('./safe-quit.js');
-require('./timeouts.js');
-require('./send.js');
-require('./register.js');
-require('./identify.js');
-require('./tchannel.js');
-require('./regression-inOps-leak.js');
-require('./emits-endpoint.js');
-require('./v2/index.js');
+var Frame = require('../../v2/frame.js');
+var TestBody = require('./test_body.js');
+var testRead = require('../lib/readTest.js');
+
+TestBody.testWith('read a cat frame', function t(assert) {
+    testRead(assert, Frame.read, TestBody.catBuffer, function s(frame) {
+        assert.equal(frame.id, 0x02030405, 'expected frame id');
+        assert.equal(String(frame.body.payload), 'cat', 'expected body payload');
+    });
+});
+
+TestBody.testWith('write a cat frame', function t(assert) {
+    var frame = Frame(0x01020304, TestBody(Buffer('doge')));
+    var buf = frame.toBuffer();
+    assert.deepEqual(buf, TestBody.dogeBuffer, 'expected buffer output');
+    assert.end();
+});
--- a/v2/call.js
+++ b/v2/call.js
@@ -1,0 +1,176 @@
+// Copyright (c) 2015 Uber Technologies, Inc.
+
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+'use strict';
+
+var read = require('../lib/read');
+var write = require('../lib/write');
+var Checksum = require('./checksum');
+var header = require('./header');
+
+module.exports.Request = CallRequest;
+module.exports.Response = CallResponse;
+
+var emptyBuffer = new Buffer(0);
+var emptyTracing = new Buffer(25);
+emptyTracing.fill(0);
+
+/* jshint maxparams:9 */
+
+// flags:1 ttl:4 tracing:24 traceflags:1 service~2 nh:1 (hk~1 hv~1){nh} csumtype:1 (csum:4){0,1} arg1~2 arg2~2 arg3~2
+function CallRequest(flags, ttl, tracing, service, headers, csum, arg1, arg2, arg3) {
+    if (!(this instanceof CallRequest)) {
+        return new CallRequest(flags, ttl, tracing, service, headers, csum, arg1, arg2, arg3);
+    }
+    var self = this;
+    // TODO Checksum-like class for tracing
+    self.type = CallRequest.TypeCode;
+    self.flags = flags || 0;
+    self.ttl = ttl;
+    self.tracing = tracing || emptyTracing;
+    self.service = service || emptyBuffer;
+    self.headers = headers;
+    self.csum = Checksum.objOrType(csum);
+    self.arg1 = arg1 ? write.bufferOrString(arg1) : emptyBuffer;
+    self.arg2 = arg2 ? write.bufferOrString(arg2) : emptyBuffer;
+    self.arg3 = arg3 ? write.bufferOrString(arg3) : emptyBuffer;
+}
+
+CallRequest.TypeCode = 0x03;
+
+CallRequest.Flags = {
+    Fragment: 0x01
+};
+
+CallRequest.read = read.chained(read.series([
+    read.UInt8,     // flags:1
+    read.UInt32BE,  // ttl:4
+    read.fixed(25), // tracing:24 traceflags:1
+    read.buf2,      // service~2
+    header.read,    // nh:1 (hk~1 hv~1){nh}
+    Checksum.read,  // csumtype:1 (csum:4){0,1}
+    read.buf2,      // arg1~2
+    read.buf2,      // arg2~2
+    read.buf2       // arg3~2
+]), function buildCallReq(results, buffer, offset) {
+    var flags = results[0];
+    if (flags & CallRequest.Flags.Fragment) {
+        throw new Error('streaming call request not implemented');
+    }
+    var ttl = results[1];
+    var tracing = results[2];
+    var service = results[3];
+    var headers = results[4];
+    var csum = results[5];
+    var arg1 = results[6];
+    var arg2 = results[7];
+    var arg3 = results[8];
+    var err = csum.verify(arg1, arg2, arg3);
+    if (err) return [err, offset, null];
+    var req = new CallRequest(flags, ttl, tracing, service, headers, csum, arg1, arg2, arg3);
+    return [null, offset, req];
+});
+
+CallRequest.prototype.write = function writeCallReq() {
+    var self = this;
+    self.csum.update(self.arg1, self.arg2, self.arg3);
+    return write.series([
+        write.UInt8(self.flags, 'CallRequest flags'),         // flags:1
+        write.UInt32BE(self.ttl, 'CallRequest ttl'),          // ttl:4
+        write.fixed(25, self.tracing, 'CallRequest tracing'), // tracing:24 traceflags:1
+        write.buf2(self.service, 'CallRequest service'),      // service~2
+        header.write(self.headers),                           // nh:1 (hk~1 hv~1){nh}
+        self.csum.write(),                                    // csumtype:1 (csum:4){0,1}
+        write.buf2(self.arg1, 'CallRequest arg1'),            // arg1~2
+        write.buf2(self.arg2, 'CallRequest arg2'),            // arg2~2
+        write.buf2(self.arg3, 'CallRequest arg3')             // arg3~2
+    ]);
+};
+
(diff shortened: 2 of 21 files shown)


v2/parser.js
```diff
@@ -0,0 +1,108 @@
...
+            case States.PendingLength:
+                self.expecting = self.buffer.readUInt32BE(0);
+                self.state = States.Seeking;
+                break;
+            case States.Seeking:
+                var frameChunk = self.shift();
+                self.handleFrame(frameChunk);
+                self.expecting = 4;
+                self.state = States.PendingLength;
+                break;
```
Raynos: add default case ?

v2/parser.js
```diff
@@ -0,0 +1,108 @@
...
+    var self = this;
+    self.push(chunk);
+    while (self.buffer.length >= self.expecting) {
+        switch (self.state) {
+            case States.PendingLength:
+                self.expecting = self.buffer.readUInt32BE(0);
+                self.state = States.Seeking;
+                break;
+            case States.Seeking:
+                var frameChunk = self.shift();
```
Raynos: how do we know that this is a whole frameChunk ? what if its a partial one ?

Raynos: oh because of the buffer.length >= expecting thing. cool story.

v2/parser.js
```diff
@@ -0,0 +1,108 @@
...
+        self.buffer = self.buffer.slice(self.expecting);
+    } else {
+        chunk = self.buffer;
+        self.buffer = Buffer(0);
+    }
+    return chunk;
+};
+
+ChunkParser.prototype.flush = function flush() {
+    // TODO: do we need this?
```
Raynos: This one does not because you always seek to the end. Other protocols like seeking to the next `\n` need a flush() if the last byte of the stream is not a \n.

v2/parser.js
jcorbin: sure defense in depth for hilarity sake... good luck testing that branch ;-)

v2/parser.js
jcorbin: yeah I was more thinking to note any mid-way state on channel close as like a warn log or some such

v2/init.js
```diff
@@ -0,0 +1,106 @@
...
+        return new InitRequest(version, hostPort, processName);
+    }
+    var self = this;
+    self.type = InitRequest.TypeCode;
+    self.version = version;
+    self.hostPort = hostPort;
+    self.processName = processName;
+}
+
+InitRequest.TypeCode = 0x01;
```
Raynos: Consider putting all TypeCodes in a single place, would make detecting holes or duplicates easier.

v2/init.js
```diff
@@ -0,0 +1,106 @@
...
+    var req = new Type(version, hostPort, processName);
+    return [null, offset, req];
+});
+
+// version:2 (key~2 value~2){2}
+InitRequest.prototype.write = function encode() {
+    var self = this;
+    return write.series([
+        write.UInt16BE(self.version),
+        write.buf2('host_port'),
```
Raynos: suggestion: for consistency add a description argument like `'init hostPort key'`

v2/errors.js
```diff
@@ -0,0 +1,90 @@
...
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+'use strict';
+
+var TypedError = require('error/typed');
+
+module.exports.ShortChunkRead = TypedError({
+    type: 'tchannel.short-chunk-read',
```
Raynos: suggestion: consider `tchannel.v2-parser.short-chunk-read`

This allows us to tell what part of tchannel the error came from.

v2/error_response.js
```diff
@@ -0,0 +1,59 @@
...
+module.exports = ErrorResponse;
+
+var emptyBuffer = new Buffer(0);
+
+function ErrorResponse(code, message) {
+    if (!(this instanceof ErrorResponse)) {
+        return new ErrorResponse(code, message);
+    }
+    var self = this;
+    self.code = code;
```
Raynos: might be nice to check for unknown code and warn somewhere.

v2/checksum.js
```diff
@@ -0,0 +1,119 @@
...
+module.exports = Checksum;
+
+// csumtype:1 (csum:4){0,1}
+function Checksum(type, val) {
+    if (!(this instanceof Checksum)) {
+        return new Checksum(type, val);
+    }
+    var self = this;
+    self.type = type;
+    self.val = val || 0;
```
Raynos: consider a `self.optional` field. There is a difference between the value literally checksumed to `0` and the checksum was not provided.

Raynos: Never mind. That's just `self.type === Checksum.Types.None`

Raynos: The `self.val` field should probably still not default to zero though. Maybe null.

test/v2/frame.js
```diff
@@ -0,0 +1,96 @@
...
+    var frame = res[2];
+    assert.error(err, 'should not fail to read');
+    assert.ok(offset === TestBody.catBuffer.length, 'read entire buffer');
+    assert.equal(frame.id, 0x01020304, 'expected frame id');
+    assert.equal(frame.flags, 0x05, 'expected flags');
+    assert.equal(String(frame.body.payload), 'cat', 'expected body payload');
+    assert.end();
+});
+
+TestBody.test('write a cat frame', function t(assert) {
```
Raynos: typo: `write a doge frame`

v2/header.js
```diff
@@ -0,0 +1,63 @@
...
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+'use strict';
+
+var errors = require('./errors');
+var read = require('../lib/read');
+var write = require('../lib/write');
+
+// nh:1 (hk~1 hv~1){nh}
```
Raynos: It might be nice to have a `Headers` prototype.

This gives consistency in all these files. It also gives it a Headers prototype that we can track in heapdumps. It will make the frame more structured instead of having a plain object of key/value pairs

Don't really know what good naming conventions are. `frame.headers.headers[key]` might be weird. `frame.header.values[key]` might be nice.

v2/header.js
```diff
@@ -0,0 +1,63 @@
...
+module.exports.read = read.chained(read.UInt8, function readHeaders(numHeaders, buffer, offset) {
+    var headers = {};
+    for (var i=0; i<numHeaders; i++) {
+        var res = read.pair2(buffer, offset);
```
Raynos: Do we want `read.pair1` here ?

v2/parser.js
```diff
@@ -0,0 +1,122 @@
...
+        self.emit('error', err);
+    } else {
+        self.emit('frame', frame);
+    }
+};
+
+ChunkParser.prototype.push = function push(chunk) {
+    var self = this;
+    if (self.buffer.length) {
+        self.buffer = Buffer.concat([self.buffer, chunk], self.buffer.length + chunk.length);
```
kriskowal: There might be a performance win here by multiplying the buffer capacity only when necessary and distinguishing length used from capacity.

v2/header.js
Raynos: you should write a test that has headers ;)

v2/header.js
```diff
@@ -0,0 +1,63 @@
...
+    var keys = Object.keys(headers);
+    var length = keys.length; // nh:1
+    var parts = new Array(1 + 2 * keys.length);
+    parts[0] = write.UInt8(keys.length);
+    for (var i=0, j=1; i<keys.length; i++) {
+        var key = write.buf1(keys[i], 'header key');
+        var val = write.buf1(headers[keys[i]], 'header val');
+        parts[++j] = key;
+        parts[++j] = val;
+        length += 1 + key.length + 1 + val.length; // hk~1 hv~1
```
Raynos: Why do we track length ?

v2/init.js
jcorbin: No the codes are very much part of the protocol document and not chosen by this repository.

We could of course do the dup check in index... not sure there's a big win there.

v2/error_response.js
jcorbin: Yes the entire error response needs to be revamped and is out of date.

v2/checksum.js
jcorbin: rather have it default numeric for type hinting

v2/header.js
jcorbin: I thought about this a long while, I'd rather not move to an entire object in this pr, maybe later.

You're asking for a false consistency, the unifying notion here is the read/write monad and not the shape of the objects they happen to operate on/under.

v2/parser.js
jcorbin: Good point, keeping a static 2*64KiB buffer should suffice, let's note that as a future point for investigation.

v2/header.js
jcorbin: it's wholly half of the entire point of our write monad, tracking length that is.

jcorbin:
Now with test coverage for everything but handler and error.

Next up I need to:
- correct and test error
- test checksum
- test headers
- consider whether testing handler is useful

checksum and headers have been minimally tested en passant by call, but I plan to cover them better in their own tests.

I'll almost certainly punt on a handler test for now and instead rely on the pre-existing tests as they are.

v2/header.js
Raynos: but you do not return length or do anything with it.

Raynos:
there is no handler in this PR yet.

v2/frame.js
```diff
@@ -0,0 +1,101 @@
...
+
+module.exports = Frame;
+
+function Frame(id, body) {
+    if (!(this instanceof Frame)) {
+        return new Frame(id, body);
+    }
+    var self = this;
+    self.id = id;
+    self.body = body;
```
Raynos: should we store `self.type` on the frame ? What about `self.size` ?

v2/call.js
```diff
@@ -0,0 +1,176 @@
...
+function CallRequest(flags, ttl, tracing, service, headers, csum, arg1, arg2, arg3) {
+    if (!(this instanceof CallRequest)) {
+        return new CallRequest(flags, ttl, tracing, service, headers, csum, arg1, arg2, arg3);
+    }
+    var self = this;
+    // TODO Checksum-like class for tracing
+    self.type = CallRequest.TypeCode;
+    self.flags = flags || 0;
+    self.ttl = ttl;
+    self.tracing = tracing || emptyTracing;
```
Raynos: why are some fields optional ? i.e. tracing is optional but ttl is not.

v2/checksum.js
```diff
@@ -0,0 +1,136 @@
...
+}
+
+Checksum.objOrType = function objOrType(arg) {
+    if (arg instanceof Checksum) {
```
Raynos: be wary of instanceof. If an application server is using tchannel@2.3 and they have a module doge-client using tchannel@2.2 you will run into a situation where there are two tchannels in node_modules and thus two Checksum prototypes.

In this situation `instanceof` fails to work.

The best known work arounds are detecting by string constants i.e. `arg && arg.type === 'Checksum'`. There are other type branding strategies that include globally caching uuids ( https://github.com/mmalecki/rust-result.js/blob/master/index.js#L4-L20 ) and check for the existance of `arg && arg.type === KNOWN_UUID`.

This may or may not be out of scope depending on usage of this method. Expect to have multiple copies of tchannel in production though because of how each client package will have it's own tchannel copy.

v2/call.js
```diff
@@ -0,0 +1,176 @@
...
+    self.service = service || emptyBuffer;
+    self.headers = headers;
+    self.csum = Checksum.objOrType(csum);
```
Raynos: it seems strange that csum is either a type or an object. I would have expected a checksum instance or the numeric checksum.

v2/call.js
```diff
@@ -0,0 +1,176 @@
...
+    Fragment: 0x01
+};
+
+CallRequest.read = read.chained(read.series([
+    read.UInt8,     // flags:1
+    read.UInt32BE,  // ttl:4
+    read.fixed(25), // tracing:24 traceflags:1
+    read.buf2,      // service~2
+    header.read,    // nh:1 (hk~1 hv~1){nh}
+    Checksum.read,  // csumtype:1 (csum:4){0,1}
```
Raynos: checksum happens after arg1, arg2, arg3.

v2/header.js
jcorbin: actually that's a great point, this is vestige of a time before write.series

v2/frame.js
jcorbin: body carries type, and size is computed JIT during write.

v2/call.js
jcorbin: will provide a ttl default, I've been trying to provide type-appropriate defaults everywhere

v2/checksum.js
jcorbin: hmm that being the case, I'd rather drop the convenience constructor pattern throughout then.

in this particular case, callers shouldn't be instantiating Checksum directly, merely passing a checksum type code.  only the call module should instantiate a checksum

v2/call.js
jcorbin: no callers do not get to specify the checksum themselves, merely the type. the design here is that the actual checksum value gets computed JIT during write.

v2/call.js
```diff
@@ -0,0 +1,176 @@
...
+    read.buf2,      // arg1~2
+    read.buf2,      // arg2~2
+    read.buf2       // arg3~2
+]), function buildCallReq(results, buffer, offset) {
+    var flags = results[0];
+    if (flags & CallRequest.Flags.Fragment) {
+        throw new Error('streaming call request not implemented');
+    }
+    var ttl = results[1];
+    var tracing = results[2];
```
Raynos: do we want to split this into tracing and traceflag ?

v2/call.js
jcorbin: not anymore, check the protocol doc

jcorbin:
going to abandon this PR and decompose into a series of more unital PRs that we can move forward on.

v2/checksum.js
```diff
@@ -0,0 +1,136 @@
...
+    }
+    return csum;
+};
+
+Checksum.prototype.update = function update(arg1, arg2, arg3) {
+    var self = this;
+    self.val = self.compute(arg1, arg2, arg3);
+};
+
+Checksum.prototype.verify = function verify(arg1, arg2, arg3) {
```
Raynos: verify relies on the fact that when the checksum type is 0, i.e. no checksum the value will be zero. It also relies on compute returning zero when the type is zero.

I recommend adding a if (type === 0) { return null } branch

Raynos:
I finished reviewing callrequest and callresponse

PR #61, opened 2015-02-24
Allow listening on port 0

kriskowal:
Note that this is a backward incompatible change.  addPeer depends on
the channel to know its own name, and the name is now only knowable
after the channel is listening.  I have removed support for
options.listening and server.listening is now a flag indicating that the
server is listening.  Tests have been adjusted to listen, and those that
required a random, available port, now use port 0 and need not wonder
whether the port was actually available.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,25 +6,34 @@ network multiplexing and framing protocol for RPC
 
 ```js
 var TChannel = require('tchannel');
+var after = require('after');
 
-var server = new TChannel({host: '127.0.0.1', port: 4040});
-var client = new TChannel({host: '127.0.0.1', port: 4041});
+var server = new TChannel();
+var client = new TChannel();
 
-// normal response
-server.register('func 1', function (arg1, arg2, peerInfo, cb) {
-    console.log('func 1 responding immediately 1:' + arg1.toString() + ' 2:' + arg2.toString());
-    cb(null, 'result', 'indeed it did');
+var listening = after(2, function (err) {
+    // ...forward or handle err
+
+    // normal response
+    server.register('func 1', function (arg1, arg2, peerInfo, cb) {
+        console.log('func 1 responding immediately 1:' + arg1.toString() + ' 2:' + arg2.toString());
+        cb(null, 'result', 'indeed it did');
+    });
+    // err response
+    server.register('func 2', function (arg1, arg2, peerInfo, cb) {
+        cb(new Error('it failed'));
+    });
+    client.send({host: '127.0.0.1:4040'}, 'func 1', "arg 1", "arg 2", function (err, res1, res2) {
+        console.log('normal res: ' + res1.toString() + ' ' + res2.toString());
+    });
+    client.send({host: '127.0.0.1:4040'}, 'func 2', "arg 1", "arg 2", function (err, res1, res2) {
+        console.log('err res: ' + err.message);
+    });
+
 });
-// err response
-server.register('func 2', function (arg1, arg2, peerInfo, cb) {
-    cb(new Error('it failed'));
-});
-client.send({host: '127.0.0.1:4040'}, 'func 1', "arg 1", "arg 2", function (err, res1, res2) {
-    console.log('normal res: ' + res1.toString() + ' ' + res2.toString());
-});
-client.send({host: '127.0.0.1:4040'}, 'func 2', "arg 1", "arg 2", function (err, res1, res2) {
-    console.log('err res: ' + err.message);
-});
+
+server.listen(4040, '127.0.0.1', listening);
+client.listen(4041, '127.0.0.1', listening);
 ```
 
 This example registers two functions on the "server". "func 1" always works and "func 2" always 
@@ -124,6 +133,7 @@ The host name and port must be a unique identifier for your
 #### `options.port`
 
 The port for which `TChannel` will open a TCP server on.
+If the port is 0, the operating system will select an available port for you.
 
 #### `options.logger`
 
--- a/README.md
+++ b/README.md
@@ -121,20 +121,6 @@ var channel = TChannel({
 });
 ```
 
-#### `options.host`
-
-You must specify a local host name. This local host name will
-    be used the remote server to identify you.
-
-The host name and port must be a unique identifier for your
-    TChannel server and its strongly recommended that this host
-    is the publicly addressable IP address.
-
-#### `options.port`
-
-The port for which `TChannel` will open a TCP server on.
-If the port is 0, the operating system will select an available port for you.
-
 #### `options.logger`
 
 ```ocaml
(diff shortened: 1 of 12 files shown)


kriskowal:
review @jcorbin, @raynos

index.js
```diff
@@ -92,9 +95,7 @@ function TChannel(options) {
         self.logger.warn('server socket close');
     });
 
-    if (self.listening) {
```
Raynos: why did we remove this flag ? I'm using it in autobahn. Seperating out allocating and effects is nice.

index.js
```diff
@@ -109,13 +110,17 @@ TChannel.prototype.listen = function listen() {
...
     }
-    if (!self.port) {
+    if (self.port === undefined) {
         throw new Error('Missing server port.');
     }
 
     self.serverSocket.listen(self.port, self.host);
 };
 
+TChannel.prototype.address = function address() {
```
Raynos: :+1:

index.js
```diff
@@ -109,13 +110,17 @@ TChannel.prototype.listen = function listen() {
     if (!self.host) {
         throw new Error('Missing server host.');
     }
-    if (!self.port) {
+    if (self.port === undefined) {
```
Raynos: suggestion: `typeof self.port !== 'number'`

index.js
```diff
@@ -265,6 +275,7 @@ TChannel.prototype.makeOutConnection = function makeOutConnection(dest) {
     return connection;
 };
 
+// TODO why 'quit' not 'close'?
```
Raynos: either `close` or `destroy` would make sense.

package.json
```diff
@@ -33,7 +33,8 @@
     "tape": "^3.0.3",
     "time-mock": "^0.1.2",
     "uber-licence": "^1.1.0",
-    "xtend": "^4.0.0"
+    "xtend": "^4.0.0",
+    "lint-trap": "^1.0.1"
```
Raynos: why add lint-trap ?

Raynos:
Let's add `options.listening` back in but keep `this.listening` to mean whether its actually listening.

package.json
kriskowal: accident. will remove.

index.js
kriskowal: I'm in favor of opt-in listen() anyway. I can do that in a follow-up.

kriskowal:
Pushed new commit to address @raynos feedback.

index.js
jcorbin: can we please leave the flag until we remove the .listen() call itself from the constructor?

index.js
```diff
@@ -265,7 +280,12 @@ TChannel.prototype.makeOutConnection = function makeOutConnection(dest) {
     return connection;
 };
 
-TChannel.prototype.quit = function quit(callback) {
+// to provide backward compatibility.
+TChannel.prototype.quit = function close(callback) {
+    this.close(callback);
```
jcorbin: please stick with the self convention

jcorbin:
can we just go all the way in this diff please:
- drop the host and port options to the constructor, including the listening or not option
- self.listening starts out false
- add TChannel.prototype.listen(port, host) with both args required
  - sets listening state bit
  - sets .hostPort, maybe also .host and .port if anything else uses those pieces, but just drop them if not
  - does the listening
  - maybe takes an optional 3rd arg that gets once'd to the listening event

kriskowal:
@jcorbin done.

index.js
```diff
@@ -265,7 +280,13 @@ TChannel.prototype.makeOutConnection = function makeOutConnection(dest) {
...
 };
 
-TChannel.prototype.quit = function quit(callback) {
+// to provide backward compatibility.
+TChannel.prototype.quit = function close(callback) {
+    var self = this;
+    self.close(callback);
+};
+
+TChannel.prototype.close = function close(callback) {
```
Raynos: :+1:

index.js
```diff
@@ -91,37 +97,41 @@ function TChannel(options) {
...
+        throw new Error('TChannel can only listen once');
     }
-    if (!self.host) {
-        throw new Error('Missing server host.');
+    if (typeof host !== 'string') {
+        throw new Error('TChannel requires host argument');
     }
-    if (!self.port) {
-        throw new Error('Missing server port.');
+    if (host === '0.0.0.0') {
```
Raynos: should we gaurd against 127.0.0.1 or localhost ?

index.js
```diff
@@ -91,37 +97,41 @@ function TChannel(options) {
...
+        throw new Error('TChannel must listen with externally visible host');
     }
-
-    self.serverSocket.listen(self.port, self.host);
+    // Default port to 4040 if null or undefined. 0 is a valid port, indicating
+    // that the system should assign an available port.
+    port = port === undefined || port === null ? 4040 : port;
```
Raynos: can we remove the default port in this pr as well. having a default port is something we wanted to remove .

index.js
```diff
@@ -91,37 +97,41 @@ function TChannel(options) {
...
+        throw new Error('TChannel requires host argument');
     }
-    if (!self.port) {
-        throw new Error('Missing server port.');
+    if (host === '0.0.0.0') {
+        throw new Error('TChannel must listen with externally visible host');
     }
-
-    self.serverSocket.listen(self.port, self.host);
+    // Default port to 4040 if null or undefined. 0 is a valid port, indicating
```
Raynos: we need to set `self.listened = true;`

Raynos:
Reviewed. We also need to update the README.md, examples, benchmarks & docs.jsig.

index.js
```diff
@@ -91,37 +97,41 @@ function TChannel(options) {
...
 require('util').inherits(TChannel, require('events').EventEmitter);
 
 // Decoulping config and creation from the constructor.
-// This also allows us to better unit test the code as the test process
-// is not blocked by the listening connections
-TChannel.prototype.listen = function listen() {
+TChannel.prototype.listen = function listen(port, host, callback) {
     var self = this;
-    if (!self.serverSocket) {
-        throw new Error('Missing server Socket.');
```
jcorbin: add TODO typed error on all these please

index.js
jcorbin: yes that should happen immediately after the section imo, so a line up before we call the lower level listen

index.js
jcorbin: default it to 0 if anything, but yeah I say drop it.

index.js
kriskowal: No, they’re used for testing. It’s a bit of a slippery slope anyway.

kriskowal:
Feedback addressed. @jcorbin @Raynos

README.md
```diff
@@ -124,6 +133,7 @@ The host name and port must be a unique identifier for your
 #### `options.port`
 
 The port for which `TChannel` will open a TCP server on.
+If the port is 0, the operating system will select an available port for you.
```
Raynos: there is no `options.port` or `options.host` we have to change the documentation for the `TChannel` constructor and add documentation for `listen()` method.

Raynos:
Still need to update the files in examples/ and benchmark/

Raynos:
lgtm. #shipit

PR #62, opened 2015-02-24
Factor out v1 handler module

jcorbin:


v1/handler.js
```diff
@@ -0,0 +1,175 @@
...
+    resFrame.header.id = id;
+    resFrame.header.seq = 0;
+    resFrame.set(arg1, self.conn.channel.hostPort, null);
+    resFrame.header.type = v1.Types.resCompleteMessage;
+    self.writeFrame(resFrame);
+};
+
+function isError(obj) {
+    return typeof obj === 'object' && (
+        Object.prototype.toString.call(obj) === '[object Error]' ||
```
kriskowal: This line alone should suffice for all cases.

Raynos: This is a copy of https://github.com/joyent/node/blob/master/lib/util.js#L564-L568

kriskowal: Appeal to tradition.

kriskowal: Sorry, that was terse.

typeof obj === 'object' is not the same as isObject(obj) for obj === null, which is the only case it's actually guarding against. Object.prototype.toString.call(null) and Object.prototype.toString.call(undefined) do not throw. If obj instanceof Error is true, then Object.prototype.toString.call(obj) is guaranteed to return '[object Error]', so it is superfluous, unless someone has replaced the global Error. The reason Object.prototype.toString.call(obj) is better than instanceof Error is because of Error from another context/realm.

index.js
```diff
@@ -116,6 +116,25 @@ TChannel.prototype.listen = function listen() {
     self.serverSocket.listen(self.port, self.host);
 };
 
+TChannel.prototype.getEndpointHandler = function getEndpointHandler(name) {
+    var self = this;
+    var handler = self.endpoints[name];
+    if (typeof handler !== 'function') {
+        handler = function noSuchHandler(arg2, arg3, remoteAddr, cb) {
+            var err = new Error('no such operation');
```
Raynos: leave todo about typed error ?

index.js
```diff
@@ -720,30 +607,67 @@ TChannelConnection.prototype.send = function send(options, frame, callback) {
...
+            });
+            return;
+        }
+        delete self.inOps[id];
+        self.inPending--;
+        self._writeFrame(resFrame);
+    }
+};
+
+TChannelConnection.prototype._writeFrame = function _writeFrame(frame, callback) {
```
Raynos: so much win.

index.js
```diff
@@ -720,30 +607,67 @@ TChannelConnection.prototype.send = function send(options, frame, callback) {
...
+    var self = this;
+    if (self.closing) {
```
Raynos: consider adding logging about write after close.

index.js
```diff
@@ -763,40 +687,12 @@ TChannelServerOp.prototype.sendResponse = function sendResponse(err, res1, res2)
...
-    return typeof obj === 'object' && (
-        Object.prototype.toString.call(obj) === '[object Error]' ||
-        obj instanceof Error);
-}
-
-function TChannelClientOp(options, frame, start, callback) {
+function TChannelClientOp(options, start, callback) {
     var self = this;
     self.options = options;
-    self.frame = frame;
```
Raynos: why remove the frames from ops ?

index.js
```diff
@@ -720,30 +607,67 @@ TChannelConnection.prototype.send = function send(options, frame, callback) {
     //  throw new Error('duplicate frame id in flight');
     // }
 
-    frame.header.id = id;
-    frame.header.seq = 0;
+    var frame = self.handler.buildRequestFrame({
```
Raynos: Might be nicer as `sendCalllRequest()`. This will move the `_writeFrame` call into the `writeFrame` function of the handler.

index.js
```diff
@@ -680,37 +595,9 @@ TChannelConnection.prototype.completeOutOp = function completeOutOp(id, err, arg
...
-    self.remoteName = remote;
-    self.channel.emit('identified', remote);
-};
-
 // send a req frame
 /* jshint maxparams:5 */
-TChannelConnection.prototype.send = function send(options, frame, callback) {
+TChannelConnection.prototype.send = function send(options, arg1, arg2, arg3, callback) {
     var self = this;
     var id = self.nextFrameId();
```
Raynos: can we move this into the handler ?

v1/handler.js
```diff
@@ -0,0 +1,175 @@
...
+    var id = errFrame.header.id;
+    var message = errFrame.arg1;
+    var err = new Error(message);
+    self.conn.completeOutOp(id, err, null, null);
+};
+
+TChannelHandler.prototype.sendInitRequest = function sendInitRequest() {
+    var self = this;
+    var reqFrame = new v1.Frame();
+    var id = self.conn.nextFrameId();
```
Raynos: can we put this method on the `TChannelHandler` ?

v1/handler.js
jcorbin: I have a later diff in the new v2/handler that does new things, I'm not changing this here

index.js
jcorbin: To force an encapsulation barrier, otherwise you'd have to write version specific code either within ops themselves or their users.  This forces us to create options conventions for all semantic pieces of all supported versions (essential the semantic union of all supported versions within the library).  Currently you see this as passing id, arg1, arg2, arg3, and timeout.

I'd love to see the options become ephemeral and fully decomposed onto properties on the ops themselves rather than carrying around op.options all the time.

Raynos:
Finished initial review.

kriskowal:
Prior commit introduced a bug, emptyBuffer is not defined in v1/frame.js.

jcorbin:
Great catch on emptybuffer @kris, fixed separately in #64.

Did a lot of updating after initial round from @raynos, less coupled now.

index.js
```diff
@@ -720,30 +624,85 @@ TChannelConnection.prototype.send = function send(options, frame, callback) {
...
+        arg1: arg1,
+        arg2: arg2,
+        arg3: arg3
+    }, function requestSent(err) {
+        if (err) {
+            self.logger.warn('failed to send request', {
+                id: id,
+                error: err
+            });
+            // TODO: retry?
```
Raynos: does tchannel do any network failure retries yet ? Does that belong in tchannel or tchannel_pool ?

index.js
```diff
@@ -720,30 +624,85 @@ TChannelConnection.prototype.send = function send(options, frame, callback) {
...
+            });
+            return;
+        }
+        sendResponseFrame(err, res1, res2, function responseSent(err) {
+            if (err) {
+                self.logger.warn('failed to send response', {
+                    id: id,
+                    error: err
+                });
+                // TODO: retry
```
Raynos: a response frame id is scoped to a socket. in an error case the socket is dead. so we cant really retry.

index.js
```diff
@@ -372,20 +394,21 @@ function TChannelConnection(channel, socket, direction, remoteAddr) {
         }
     });
 
-    self.localEndpoints['TChannel identify'] = function identifyEndpoint(arg1, arg2, hostInfo, cb) {
-        cb(null, self.channel.hostPort, null);
-    };
+    self.handler.on('error', function onHandlerError(err) {
```
Raynos: :+1: nice seperation.

v1/handler.js
```diff
@@ -0,0 +1,203 @@
...
+TChannelHandler.prototype.sendRequestFrame = function sendRequestFrame(options, callback) {
+    var self = this;
+    var id = self.nextFrameId();
+    var reqFrame = new v1.Frame();
+    reqFrame.header.id = id;
+    reqFrame.header.seq = 0;
+    reqFrame.set(options.arg1, options.arg2, options.arg3);
+    reqFrame.header.type = v1.Types.reqCompleteMessage;
+    self.writeFrame(reqFrame, callback);
+    return id;
```
Raynos: In terms of interface design we may not want to return the id.

A common thing we want to do as well as pass the `callback` to `socket.write()` is to return the boolean result of `socket.write()` which is the "should I write more" backpressure boolean.

If we want the `TChannelHandler` interface to expose more information about the stream state we might want to return "should I write more" backpressure information in some fashion.

v1/handler.js
```diff
@@ -0,0 +1,207 @@
...
+
+TChannelHandler.prototype.handleInitRequest = function handleInitRequest(reqFrame) {
+    var self = this;
+    if (self.remoteName !== null) {
+        self.emit('error', new Error('duplicate init request')); // TODO typed error
+        return;
+    }
+    var hostPort = reqFrame.arg2.toString();
+    self.remoteName = hostPort;
+    self.emit('ident.in', hostPort);
```
Raynos: consider naming this event `identify.in`

Raynos:
Finished another round of review.

v1/handler.js
jcorbin: This is part of the runInOp / completeOutOp boundary concern, so this will stay like it is until we get a better abstraction there

Raynos:
#land

Issue #63, opened 2015-02-24
Lint comply ignored code, remove ignores


jcorbin:

Raynos:
Fixed in #64

PR #64, opened 2015-02-24
Such broken

jcorbin:
Ignore what isn't lint compliant for now (#63).

Lack of this already let a programming bug slip.


Raynos:
lgtm. As long as index and the main production passes lint then #coolstory

PR #65, opened 2015-02-25
Add lib/{read,write}

jcorbin:


kriskowal:
:ok:

Raynos:
this bit is probably important to read. I'll read it later.

PR #66, opened 2015-02-25
Add v2 chunk parser

jcorbin:
Parser is now loosely coupled to framing concerns, hopefully meaning that this particular chunk parser will be usable even beyond v2.

### DIFF
--- a/test/v2/test_frame.js
+++ b/test/v2/test_frame.js
@@ -20,9 +20,28 @@
 
 'use strict';
 
-module.exports.VERSION = 2;
+var read = require('../../lib/read.js');
+var write = require('../../lib/write.js');
 
-var Types = {};
-module.exports.Types = Types;
+module.exports = TestFrame;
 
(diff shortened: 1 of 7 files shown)


test/v2/test_frame.js
```diff
@@ -0,0 +1,47 @@
...
+'use strict';
+
+var read = require('../../lib/read.js');
+var write = require('../../lib/write.js');
+
+module.exports = TestFrame;
+
+function TestFrame(size, payload) {
+    var self = this;
+    self.size = size || 1 + 1 + self.payload.length;
```
kriskowal: Perhaps comment on what 1 and 1 represent respectively. A byte for length prefix, a byte for a terminator?

kriskowal:
:ok: comment stands on buffer allocation churn.

jcorbin:
I'd like to address buffer allocation optimization later and do actual performance testing on it.

PR #67, opened 2015-02-25
Add v2 frame

jcorbin:


v2/frame.js
```diff
@@ -0,0 +1,138 @@
...
+        var frame = new Frame(id, body);
+        return [null, offset, frame];
+    }
+});
+
+// size:2: type:1 reserved:1 id:4 reserved:8 ...
+Frame.prototype.write = function writeFrame() {
+    var self = this;
+    var body = self.body.write();
+    if (body.length > Frame.MayBodySize) {
```
Raynos: `Frame.MaxBodySize` #needsmoretest

v2/frame.js
```diff
@@ -0,0 +1,138 @@
...
+    if (id === null || id === undefined) {
+        self.id = Frame.NullId;
+    } else {
+        self.id = id;
+    }
+    self.body = body;
+}
+
+Frame.Overhead = 0x10;
+Frame.MayBodySize = 0xffff - Frame.Overhead;
```
Raynos: 0xffff is the same as two bytes right? basically saying body must be `body~2` ?

v2/frame.js
```diff
@@ -0,0 +1,138 @@
...
+
+module.exports = Frame;
+
+function Frame(id, body) {
+    if (!(this instanceof Frame)) {
+        return new Frame(id, body);
+    }
+    var self = this;
+    if (id === null || id === undefined) {
+        self.id = Frame.NullId;
```
Raynos: noice.

v2/frame.js
Raynos: the body can be bigger then 2 bytes worth because its arg1~2 + arg2~2 etc.

jcorbin: yes, but it can't, total frame max is 64KiB, if args sum to more, you have to fragment.

Raynos: Yeah just re-read. We might want to figure out how to deal with the fact that this is less then arg2 + arg3. but thats someone elses problem.

jcorbin: yes streaming support will require a whole new layer in the tchannel library as yet unstarted

jcorbin:
typo fixed

Raynos:
lgtm.

v2/frame.js
```diff
@@ -0,0 +1,138 @@
...
+    var type = head[1];
+    var id = head[3];
+
+    var BodyType = Frame.Types[type];
+    if (!BodyType) {
+        return [InvalidFrameTypeError({typeNumber: type}), offset, null];
+    }
+
+    // don't use read.len here to avoid doing a potentially large slice
+    var bodyLength = length - 16;
```
kriskowal: 16 is Frame.Overhead?

jcorbin: good point

jcorbin:
updated to use Frame.Overhead consistently

v2/frame.js
```diff
@@ -0,0 +1,138 @@
...
+    var err = read.want(bodyLength, buffer, offset);
+    if (err) return [err, offset, null];
+
+    var end = offset + bodyLength;
+
+    return read.chain(function(buffer, offset) {
+        return BodyType.read(buffer, offset);
+    }, buffer, offset, buildFrame);
```
kriskowal: This closure has no free variables. You can take it out of scope. I expect you might be able to factor the chain such that nothing gets allocated here.

kriskowal: Got it. Closes over BodyType. In person discussed building a book of readers by BodyType.

jcorbin:
Turns out I don't even need to bind, that was originally a hack to get the upcoming init req / res readers to be bound to their class types.  If a given body type needs this, than it can bind its own readers internally.

kriskowal:
looks good to me.

PR #69, opened 2015-02-26
Add v2/init

jcorbin:


kriskowal:
Looks good to me. I recall that you argue that the type codes for each of these request and response frame types should be with the constructors of those frames. I think we will need an index of type code to frame constructor somewhere, expressed in a fashion that demands that it is kept up-to-date, so that maintainers can look them up. That may mean a DRY violation, but it seems worthwhile.

PR #71, opened 2015-02-26
V2 checksum

jcorbin:


v2/checksum.js
```diff
@@ -0,0 +1,144 @@
...
+    return csum;
+};
+
+Checksum.prototype.update = function update(arg1, arg2, arg3) {
+    var self = this;
+    self.val = self.compute(arg1, arg2, arg3);
+};
+
+Checksum.prototype.verify = function verify(arg1, arg2, arg3) {
+    var self = this;
```
Raynos: Explicitely returning `null` when type is zero would be nice.

Raynos:
lgtm.

PR #74, opened 2015-02-26
Add v2 protocol doc as PDF

mranney:
I know it's not cool to use PDF or anything other than markdown, but this
process is happening mostly on Google Docs, and exporting to PDF is the only
way I could figure to publish this.


jcorbin:
I'd really rather not add a large PDF if at all possible:
- we could do a one-time conversion to markdown and then iterate in repo
- if google fu is really still useful to the process, we could use something like StackEdit maybe

If a PDF really is the best we can do, then let's put it on a separate and disjoint branch from the code.  I'm not 100% on exactly how npm packaging works, but will having this next to code cause it to get baked into all installed node_modules?

Finally if this really has to go on same branch as code, this PR should be targeting the v2 branch not master.  We'll eventually merge v2 into master and do away with it, but master is still v1.

Raynos:
@jcorbin there is an `.npmignore` file you can add to ignore the pdf.

jcorbin:
Discussed offline with @mranney : 
- we'll be doing a markdown or textile conversion
- and comitting that to this or another repo
- tracking that on https://github.com/uber/tchannel/issues/78 with other v2 priorities

kriskowal:
And there’s package.json "files" if you prefer a whitelist. Frontend devs often favor the whitelist to publish approach.

PR #76, opened 2015-02-26
Buf speedup

jcorbin:
From a preliminary test, this proves to be 30% and 50% faster for trivial strings (random sampling from /usr/share/dict/words).

### DIFF
--- a/lib/utf8len.js
+++ b/lib/utf8len.js
@@ -1,0 +1,56 @@
+// Copyright (c) 2015 Uber Technologies, Inc.
+
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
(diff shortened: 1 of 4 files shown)


Raynos:
This looks ok. I would recommend testing a bunch of strings, maybe use these ones ( https://github.com/joyent/node/blob/master/test/simple/test-string-decoder.js#L28-L53 ).

lib/utf8len.js
```diff
@@ -0,0 +1,55 @@
...
+        return 5;
+    } else if (c < 0x80000000) {
+        // 1111110b 10bbbbbb 10bbbbbb 10bbbbbb 10bbbbbb 10bbbbbb
+        return 6;
+    } else {
+        throw new Error('codepoint out of range');
+    }
+}
+
+function numBytesForUtf8String(s) {
```
Raynos: lets leave a comment about this being faster then `Buffer.byteLength`

jcorbin:
Interestingly using my utf8len function seems to be 10% - 35% faster than using the `Buffer.byteLength(source, 'utf-8')`

jcorbin:
afaik, correctness of utf8len is down to correctness of `String.prototype.charCodeAt`

kriskowal:
@jcorbin bear in mind that String.prototype.charCodeAt is ucs2, not unicode in general. If folks express code points outside the Basic Multilingual Plane, they have to resort to surrogate pairs. Not sure whether utf8len takes that into account, but we may be able to safely punt the issue. https://mathiasbynens.be/notes/javascript-unicode

jcorbin:
v2 about to dead land, will reboot if needed

PR #79, opened 2015-02-26
throw exception on double close

Raynos:
I spend a bunch of time debugging something that was
a double close bug.

Throwing a hard exception will make this a lot easier
to see.

cc: @jsu1212


kriskowal:
:+1: ship it

PR #81, opened 2015-02-27
ensure the socket is always closed

Raynos:
The `resetAll()` method does not call socket.destroy().
There are only 3 callsites for it.

 - when the socket closes or errors, resetAll()
 - when the server quits, end the socket then resetAll()
 - when the handler has an error resetAll()

When the handler has an error we do not destroy the socket
which would leak a socket until the net timeout cleaned it
up.

cc: @jsu1212

### DIFF
--- a/index.js
+++ b/index.js
@@ -428,6 +428,8 @@ function TChannelConnection(channel, socket, direction, remoteAddr) {
 
     self.handler.on('error', function onHandlerError(err) {
         self.resetAll(err);
+        // resetAll() does not close the socket
+        self.socket.destroy();
     });
 
     if (direction === 'out') {


kriskowal:
Looks good to me.

index.js
```diff
@@ -428,6 +428,8 @@ function TChannelConnection(channel, socket, direction, remoteAddr) {
 
     self.handler.on('error', function onHandlerError(err) {
         self.resetAll(err);
+        // resetAll() does not close the socket
```
jcorbin: should it? if not why not?

Raynos: It cant because the server path calls `socket.end(); resetAll()`. We cannot call destroy in the `resetAll()` method because its used for both clean exit and shutdown.

It would also be weird to call `socket.destroy()` in the close event listener on the socket.

The fact that `resetAll()` is used for both "destroy" and "clean close of socket" is what makes it difficult.

PR #83, opened 2015-02-27
V2 handler

jcorbin:


v2/handler.js
```diff
@@ -0,0 +1,234 @@
...
+// THE SOFTWARE.
+
+'use strict';
+
+var TypedError = require('error/typed');
+var EventEmitter = require('events').EventEmitter;
+var util = require('util');
+
+var v2 = require('./index');
+module.exports = TChannelV2Handler;
```
Raynos: nitpick: it's nice to put the `module.exports` in its own "block"; i.e. new line above and new line below

v2/handler.js
```diff
@@ -0,0 +1,234 @@
...
+
+function TChannelV2Handler(channel, options) {
+    if (!(this instanceof TChannelV2Handler)) {
+        return new TChannelV2Handler(channel, options);
+    }
+    EventEmitter.call(this);
+    var self = this;
+    self.channel = channel;
+    self.writeFrame = options.writeFrame;
+    // TODO: may be better suited to pull out an operation collection
```
Raynos: :+1:

v2/handler.js
```diff
@@ -0,0 +1,234 @@
...
+        case v2.Types.InitResponse:
+            return self.handleInitResponse(frame);
+        case v2.Types.CallRequest:
+            return self.handleCallRequest(frame);
+        case v2.Types.CallResponse:
+            return self.handleCallResponse(frame);
+        case v2.Types.Error:
+            return self.handleError(frame);
+        default:
+            self.logger.error('unhandled frame type', {
```
Raynos: we should probably emit an error on unhandled frame so that the socket dies

v2/handler.js
```diff
@@ -0,0 +1,234 @@
...
+
+    // TODO: couple reqFrame.body.tracing for all downstream requests
+    // TODO: implement service name?
+
+    self.runInOp(handler, {
+        id: id,
+        arg1: reqFrame.body.arg1,
+        arg2: reqFrame.body.arg2,
+        arg3: reqFrame.body.arg3,
+        headers: headers
```
Raynos: can we serviceName from `reqFrame.body` and pass it here ?

v2/handler.js
```diff
@@ -0,0 +1,234 @@
...
+    var id = reqFrame.id;
+    var name = String(reqFrame.body.arg1);
+
+    if (self.remoteHostPort === null) {
+        self.emit('error', new Error('call request before init request')); // TODO typed error
+        return;
+    }
+
+    var handler = self.channel.getEndpointHandler(name);
+    var headers = {};
```
Raynos: this can be read from reqFrame.body ?

v2/handler.js
```diff
@@ -0,0 +1,234 @@
...
+    // TODO: implement service name?
+
+    self.runInOp(handler, {
+        id: id,
+        arg1: reqFrame.body.arg1,
+        arg2: reqFrame.body.arg2,
+        arg3: reqFrame.body.arg3,
+        headers: headers
+    }, function sendResponseFrame(err, res1, res2, callback) {
+        self.sendResponseFrame(reqFrame, headers, err, res1, res2, callback);
```
Raynos: why pass request headers to response frame ? Is this to enable threading of tracing ?

v2/handler.js
```diff
@@ -0,0 +1,234 @@
...
+    var resFrame = v2.Frame(id, body);
+    self.writeFrame(resFrame);
+};
+
+/* jshint maxparams:6 */
+TChannelV2Handler.prototype.sendRequestFrame = function sendRequestFrame(options, arg1, arg2, arg3, callback) {
+    var self = this;
+    var id = self.nextFrameId();
+    var flags = 0; // TODO: streaming
+    var ttl = options.timeout || 1; // TODO: better default, support for dynamic
```
Raynos: ttl is milliseconds, did you mean `1000` ?

v2/handler.js
```diff
@@ -0,0 +1,234 @@
...
+    var tracing = options.tracing || null; // TODO: generate
+    var service = options.service || null; // TODO: what even is this? default it to some sort of "my" name?
```
Raynos: this is the name of the remote service I want to talk to. 

It will be used as an implementation detail of autobahn client as part of `autobahn.send({ service, arg1, arg2, arg3 }, cb)`

v2/handler.js
```diff
@@ -0,0 +1,234 @@
...
+    // TODO: refactor this all the way back out through the op handler calling convention
+    var self = this;
+    var id = reqFrame.id;
+    var flags = 0; // TODO: streaming
+    var arg1 = reqFrame.body.arg1;
+    var checksumType = reqFrame.body.csum.type;
+    var resBody;
+    if (err) {
+        var errArg = isError(err) ? err.message : JSON.stringify(err); // TODO: better
+        resBody = v2.CallResponse(flags, v2.CallResponse.Codes.Error, headers, checksumType, arg1, res1, errArg);
```
Raynos: why are we sending request headers back ?

v2/handler.js
```diff
@@ -0,0 +1,234 @@
...
+    return id;
+};
+
+TChannelV2Handler.prototype.sendResponseFrame = function sendResponseFrame(reqFrame, headers, err, res1, res2, callback) {
+    // TODO: refactor this all the way back out through the op handler calling convention
+    var self = this;
+    var id = reqFrame.id;
+    var flags = 0; // TODO: streaming
+    var arg1 = reqFrame.body.arg1;
+    var checksumType = reqFrame.body.csum.type;
```
Raynos: does the checksum have to be the same as the one from the incoming request.

What if we always want to send farmhash back on the res because #speed

v2/handler.js
```diff
@@ -0,0 +1,234 @@
...
+    var resBody;
+    if (err) {
+        var errArg = isError(err) ? err.message : JSON.stringify(err); // TODO: better
```
Raynos: error serialization should probably be configurable somehow.

v2/handler.js
```diff
@@ -0,0 +1,234 @@
...
+    }
+    var id = resFrame.id;
+    var code = resFrame.body.code;
+    var arg1 = resFrame.body.arg1;
+    var arg2 = resFrame.body.arg2;
+    var arg3 = resFrame.body.arg3;
+    if (code === v2.CallResponse.Codes.OK) {
+        self.completeOutOp(null, id, arg2, arg3);
+    } else {
+        self.completeOutOp(TChannelApplicationError({
```
Raynos: should we be casting remote error arg3 into an error object at the tchannel level ?

In http you get back `null, { statusCode: 500, body: ... }` and the `err` arg in the callback is left for actual tcp failures.

I think casting an error res2 into an `err` arg is a bit too opinionated for `tchannel`. It should probably reserve `err` for network, io & protocol errors.

v2/handler.js
```diff
@@ -0,0 +1,234 @@
...
+    }
+};
+
+TChannelV2Handler.prototype.handleError = function handleError(errFrame) {
+    var self = this;
+    var id = errFrame.id;
+    var code = errFrame.body.code;
+    var message = errFrame.body.message;
+    var err = v2.ErrorResponse.CodeErrors[code]({message: message});
+    self.completeOutOp(err, id, null, null);
```
Raynos: this makes sense as an `err` because the error frame type is a protocol/tchannel error of the remote tchannel implementation.

Raynos:
Finished review.

v2/handler.js
jcorbin: No these are the responseHeaders, I'll rename to make that clearer

v2/handler.js
jcorbin: nope, if you don't specify you get insanely tight limit, I'd almost like to default it to 0, thoughts?

v2/handler.js
jcorbin: sure, but how should we fill it in for just basic tchannel library is the question I think.

it'd be great to have normal tchannel client for say a gossip ring doing low level service name checking e.g.

v2/handler.js
jcorbin: they're not request headers as mentioned above

v2/handler.js
jcorbin: passing through requesting checksum means higher chance of caller understanding it, if say they only grok crc or set it to none

v2/handler.js
jcorbin: or just be down to the Arg Scheme, I've got a future commit which does different thing here, but that make more sense to separate into -as-json

v2/handler.js
jcorbin: note I'm currently bundling all 3 args into this error object, not "casting" any of them.

yes this exactly one of the touch points for an as yet unstarted as-handler api

jcorbin:
Updates and responded to comments @Raynos

v2/handler.js
Raynos: oh we should leave a comment about how the endpointHandler needs to populate response headers

v2/handler.js
Raynos: oh i see, if you have a ring of tchannels for a single service. I guess you do want the service to be yourself by default.

You also want to check incoming call requests and double check the incoming service is for yourself and otherwise return "i am not a proxy" error.

v2/handler.js
Raynos: Oh I see.

v2/handler.js
Raynos: True your not casting them.

But you are turning a "non-error" network callback into an "error" tchannel callback. That smells wrong to me.

v2/handler.js
```diff
@@ -0,0 +1,241 @@
...
+    var self = this;
+    if (self.remoteHostPort !== null) {
+        self.emit('error', new Error('duplicate init request')); // TODO typed error
+        return;
+    }
+    var hostPort = reqFrame.body.hostPort;
+    var processName = reqFrame.body.processName;
+    self.remoteHostPort = hostPort;
+    // TODO: use processName
+    self.emit('identify.in', hostPort, processName);
```
kriskowal: perhaps 'identify sender' and 'identify receiver'?

kriskowal: or 'identify local' and 'identify remote'

Raynos:
LGTM.

kriskowal:
Yes, let’s move forward.

kriskowal:
I mean, looks good to me.

v2/handler.js
jcorbin: these names were already chosen when we factored out v1/handler, changing them could happen in future

PR #85, opened 2015-02-27
throw an exception on double registration

Raynos:
Currently if someone registers the same endpoint twice
we just silently overwrite. This makes debugging a lot
harder.

My use case is that I want my application and ringpop to
share a single tchannel server. It's a lot easier if
to do this if i know that any conflict between ringpop
and my application results in a programmer error.

reviewers: @jwolski @jcorbin @kriskowal

cc: @jsu1212


jcorbin:
lgtm

PR #86, opened 2015-02-28
V2 parse streams

jcorbin:
Refactor:
- factor textExpectations primitive out of parserTest
- factor out a parse buffer boundary for future experimentation
- v2 parser to be a transform stream

The parse buffer factor out was useful since `Readable.prototype.push` conflicted with our prior `ChunkParser.prototype.push` for the chunk buffering.

reviewers: @kriskowal @Raynos

### DIFF
--- a/index.js
+++ b/index.js
@@ -408,11 +408,6 @@ function TChannelConnection(channel, socket, direction, remoteAddr) {
 
     self.socket.setNoDelay(true);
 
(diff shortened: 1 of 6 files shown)


kriskowal:
Looks good to me. Step in a good direction.

Raynos:
lgtm.

PR #88, opened 2015-02-28
V2 init headers

jcorbin:
reviewers: @Raynos @kriskowal

### DIFF
--- a/test/v2/init.js
+++ b/test/v2/init.js
@@ -24,38 +24,46 @@ var test = require('tape');
 var testRead = require('../lib/read_test.js');
 var Init = require('../../v2/init.js');
 
+/* jshint camelcase:false */
+
 var testInitReq = Buffer([
     0x00, 0x02,             // version:2
-
+    0x00, 0x03,             // nh:2
     0x00, 0x09, 0x68, 0x6f, // key~2 -- host_port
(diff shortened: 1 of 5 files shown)


test/v2/init.js
```diff
@@ -24,38 +24,46 @@ var test = require('tape');
 var testRead = require('../lib/read_test.js');
 var Init = require('../../v2/init.js');
 
+/* jshint camelcase:false */
+
 var testInitReq = Buffer([
     0x00, 0x02,             // version:2
-
+    0x00, 0x03,             // nh:2
```
sectioneight: 0x03 == 2?

jcorbin: nope, the comment is to remind that it's a 2-byte field, it's just a line-wise decomposition of the comment banner above each class in e.g. v2/{init,call,frame}

sectioneight: ah right. i was assuming you were only sending the "supported" 2 headers and have been staring at hex all week. carry on.

jcorbin: no worries, we only added this yesterday, you may have rightly assumed we had it all along since not having it is frankly insane ;-)

Raynos:
lgtm.

PR #90, opened 2015-02-28
add NODE_DEBUG=tchannelhexer support

Raynos:
This instruments `tchannel` itself with both hexer and `debuglog`.

I've copied the `debuglog` to add an `enabled` true/false flag.

This means we will only call out to hexer when `enabled` is set to true.
I wanted to write the frames, so i've instrumented the v2 parser as that was the nicest place to put it.

This allows us to run any program, like `tcurl` with `NODE_DEBUG=tchannelhexer` and it will pretty print all the frames coming in & out.

cc: @jsu1212

### DIFF
--- a/lib/debug-log.js
+++ b/lib/debug-log.js
@@ -1,0 +1,48 @@
+// Copyright (c) 2015 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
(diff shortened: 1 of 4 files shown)


kriskowal:
The name `hexer` reveals an implementation detail that is unimportant to the user and constraining to the implementer. `hex` or `hexdump` would be more appropriate.

lib/debug-log.js
```diff
@@ -0,0 +1,48 @@
...
+// THE SOFTWARE.
+
+'use strict';
+// Modified to expose an enabled boolean
+// https://github.com/joyent/node/blob/fbfe562d71ae8d8f8bbf627808c755e513e4e905/lib/util.js#L96-L114
+var util = require('util');
+var debugEnv = process.env.NODE_DEBUG || '';
+var cache = {};
+
+module.exports = function (set) {
```
kriskowal: Function needs a name.

kriskowal:
Apart from advice, looks good to me.

jcorbin:
v2 branch is about to be dead landed, plus redo this after we finish refactoring handler to streaming

PR #91, opened 2015-02-28
Integrate node v2

jcorbin:
This is just a subtree graft of the v2 branch into the newly rebooted master plus a README update on instability.

cc @Raynos @kriskowal


Raynos:
lgtm.

PR #92, opened 2015-02-28
Initial round of converting the protocol doc

jcorbin:
cc @mranney @Raynos @kriskowal @sectioneight @mmihic @jwolski @eklitzke  ... others?

For review I'd suggest also looking at the rendered version: https://github.com/uber/tchannel/blob/import_protocol_doc/protocol.md

Errata:
- I took liberties with the formatting and wording in places in an attempt to clarify
- I reordered the sections such that
  - all payloads are defined in one go
  - then a section on transport headers
  - then a section on fragmentation
- I changed the as=json definition to say that there can be app headers encoded in arg2 as json
- I added future-safety language to the call-res code table: basically do the unix thing, and consider any non-zero "not ok"... whether we want to say "applications can define..." or just leave room for future protocol, I feel this is the right option

I didn't change the as=http part, but it seems to me like json-encoded headers are very odd and that we should just use our binary header encoding as for as=thrift.


protocol.md
```diff
@@ -0,0 +1,748 @@
...
+1. match incoming message M1 with existing "call req"
+2. deliver args to application
+
+## Framing
+
+All frames of all types use this structure:
+
+```
+----------------------------------------------
+| 0-7  | size:2 | type:1 | reserved:1 | id:4 |
```
sectioneight: Might just be me, but I find these "tables" easier to read if we add a separating lines of `----`s between the rows.

protocol.md
```diff
@@ -0,0 +1,748 @@
...
+used for interop with other systems.
+
+The following table describes the valid `as` values and their corresponding
+arg1, arg2, and arg3 forms:
+
+value    | arg1                  | arg2                                     | arg3
+---------|-----------------------|------------------------------------------|---------------
+`thrift` | string method name    | application headers `nh:2 (k~2 v~2){nh}` | thrift payload
+`json`   | method name as string | application headers as JSON              | body as JSON
+`http`   | method + URI          | headers as JSON array                    | raw body
```
sectioneight: It seems kind of weird to me that we spend a bunch of time making short header names and single-byte codes to define their values, but then go full-on string for `as`. I like the readability, but it feels inconsistent. Perhaps we should consider an `as` code table?

jcorbin: These strings are short enough to not be a concern, plus they're used by the forwarding layer.

At any rate, this PR isn't the forum for changing these values anyhow.

sectioneight: Yea I'm not saying we should change it in this PR, but I am forgetful and wanted to make a not of it while I was reading this.

sectioneight:
Looks great. A couple random thoughts.

sectioneight:
:shipit: -- the build failure is because you're missing the .travis.yml that I added yesterday

sectioneight:
Also don't forget to `git rm node/docs/protocol.md` so we don't have confusing information laying around :)

protocol.md
mranney: I agree that it feels a bit awkward, but let's leave any more debate or optimizations for v3.

mranney:
The retry flags section looks a bit corrupted. Maybe you forgot to finish that part?

protocol.md
```diff
@@ -0,0 +1,752 @@
...
+`0x02` | init res          | Remote response to init req
+`0x03` | call req          | RPC method request
+`0x04` | call res          | RPC method response
+`0x13` | call req continue | RPC request continuation fragment
+`0x14` | call res continue | RPC response continuation fragment
+`0xc0` | cancel            | Cancel an outstanding call req / forward req (no body)
+`0xc1` | claim             | Claim / cancel a redundant request
+`0xd0` | ping req          | Protocol level ping req (no body)
+`0xd1` | ping res          | Ping res (no body)
+`0xff` | error             | Protocol level error.
```
abhinav: The table doesn't render correctly right now. I think GitHub Markdown requires headers for tables.

protocol.md
```diff
@@ -0,0 +1,752 @@
...
+`version` is a 16 bit number. The currently specified protocol version is 2. If
+new versions are required, this is where a common version can be negotiated.
+
+### headers
+
+There are a variable number of key/value pairs. For version 2, the following
+are required:
+
+`host\_port`    | address:port where this process can be reached
+`process\_name` | additional identifier for this instance, used for logging
```
abhinav: `\` before the `_`.

Also, table rendering.

protocol.md
```diff
@@ -0,0 +1,752 @@
...
+
+Schema: `version:2 nh:2 (key~2 value~2){nh}`
+
+The initiator requests a version number, and the server responds with the
+actual version that will be used for the rest of this connection. The
+name/values are the same, but identify the server.
+
+## Payload: call req (type 0x03)
+
+Schema: `flags:1 ttl:4 tracing:24 traceflags:1 service~2 nh:1 (hk~1 hv~1){nh} csumtype:1 (csum:4){0,1} arg1~2 arg2~2 arg3~2`
```
abhinav: This is too long. This can probably be changed into a code block split with the schema split across multiple lines.

protocol.md
```diff
@@ -0,0 +1,752 @@
...
+of the args triple.
+
+A "call req" may be fragmented across multiple frames. If so, the first frame
+is a "call req", and all subsequent frames are "call req continue" frames.
+
+### flags:1
+
+Used to control fragmentation. Valid flags:
+
+`0x01` | more fragments follow
```
abhinav: Table rendering

protocol.md
```diff
@@ -0,0 +1,752 @@
...
+arg1, arg2, and arg3 is unspecified at the transport level. These are opaque
+binary blobs as far as tchannel is concerned.
+
+Future versions will likely allow callers to specify specific service instances
+on which to run this request, or a mechanism to route a certain percentage of
+all traffic to a subset of instances for canary analysis.
+
+## Payload: call res (0x04)
+
+Schema: `flags:1 code:1 tracing:24 traceflags:1 nh:1 (hk~1 hv~1){nh} csumtype:1 (csum:4){0,1} arg1~2 arg2~2 arg3~2`
```
abhinav: Consider converting to code block split across multiple lines

protocol.md
```diff
@@ -0,0 +1,752 @@
...
+Headers described below in the "Transport Headers" section.
+
+### code:1
+
+Response code:
+
+code | name  | description
+-----|-------|------------
+0x00 | OK    | everything is great and we value your contribution.
+0x01 | Error | application error, details are in the args.
```
abhinav: Nit: Inconsistent with tables above where items in the `code` column were fixed-width.

protocol.md
```diff
@@ -0,0 +1,752 @@
...
+0x02 | cancelled
+0x03 | busy
+0x04 | declined
+0x05 | unexpected error
+0x06 | bad request
+0xFF | fatal protocol error
+
+#### Error code `0x00` -- invalid
+
+Not a valid value for `code`. Do not use.
```
abhinav: Personally, I would move the descriptions into the table instead of adding separate sections for them another level deeper.

protocol.md
```diff
@@ -0,0 +1,752 @@
...
+`n`   | no retry
+`ct`  | retry on connection error or timeout
+`tc`  | retry on connection error or timeout
+
+### Transport Header `se` -- Speculative Execution
+
+Speculative execution count, encoded as a single digit (therefore <10) ascii
+digit. Indicates the number of nodes on which to run the request.
+
+The only valid value for `se` in protocol version "2" (ascii `0x32`).
```
abhinav: There's something missing in this statement

jcorbin:
Great feedback @abhinav, I'm going to try some other formatting on the schema section next, I addressed all of your other points.

Note for reviewers: I'm keeping the commits additive and small for now to be clear about which feedback has been addressed in this PR, but I'll squash things down before merging.

kriskowal:
Note that `npm install greadme -g` gives you a `greadme` you can use to check Github markdown before committing.

abhinav:
lgtm

PR #94, opened 2015-03-01
node: rename v2/{parser => reader}

jcorbin:
More idiomatic, really it's a "tokenizer" and less a "parser" in a pedantic sense. Plus this will make a nice dual with an eventual chunk writer.

### DIFF
--- a/node/test/v2/reader.js
+++ b/node/test/v2/reader.js
@@ -21,9 +21,10 @@
 'use strict';
 
 var util = require('util');
-var Parser = require('../../v2/parser.js');
+var Reader = require('../../v2/reader.js');
 var TestFrame = require('./test_frame');
-var parserTest = require('../lib/parser_test');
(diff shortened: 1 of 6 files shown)


node/test/v2/reader.js
```diff
@@ -21,9 +21,10 @@
 'use strict';
 
 var util = require('util');
-var Parser = require('../../v2/parser.js');
+var Reader = require('../../v2/reader.js');
 var TestFrame = require('./test_frame');
-var parserTest = require('../lib/parser_test');
```
Raynos: delete this parser_test file ?

Raynos:
lgtm.

node/test/v2/reader.js
jcorbin: it got renamed to `test/lib/test_expectations`, I already had the primitive factored out internally, just not exposed yet.

PR #95, opened 2015-03-01
node: drop callback from send{Request/Response}Frame

jcorbin:
Helps to prepare for handler becoming a Writable since `self.push` takes no callback.

I'm open to other options if there's a good way to preserve write flush notification.


Raynos:
The only way to preserve the write flush notification is to duplex two seperate streams together.

Which effectively gives you two `.write()` calls, one used by `frames.pipe(handler);` and once used internally by `handler.sendFrame...()`. 

See https://github.com/deoxxa/duplexer2

Raynos:
lgtm. I'll leave the duplexer decision upto you.

PR #96, opened 2015-03-01
node: better error objects from handleError

jcorbin:
- expose code number in `.errorCode`
- pass along error res's original message id field in `.originalId`


Raynos:
lgtm.

Maybe add tests assertion for originalId & code existance.

PR #97, opened 2015-03-01
node: emit handler error on non-message-associated error frame

jcorbin:


Raynos:
Can we add a test for this case ?

Raynos:
lgtm.

jcorbin:
well test it eventually in #99 once we get done refactoring the boundary between it and TChannelConnection

PR #100, opened 2015-03-01
node: factor out chunk writer

jcorbin:


node/index.js
```diff
@@ -391,9 +391,10 @@ function TChannelConnection(channel, socket, direction, remoteAddr) {
     self.closing = false;
 
     self.reader = new v2.Reader(v2.Frame);
+    self.writer = new v2.Writer();
```
Raynos: pass self or logger in ?

node/index.js
```diff
@@ -688,22 +693,6 @@ TChannelConnection.prototype.runInOp = function runInOp(handler, options, sendRe
     }
 };
 
-TChannelConnection.prototype._writeFrame = function _writeFrame(frame, callback) {
-    var self = this;
-    if (self.closing) {
-        self.logger.warn('write frame after close', {
```
Raynos: we've lost the self.closing check

jcorbin: Yup and that's okay I think, since close should propagate throughout the pipe chain, eventually unified over handler.

Any ops settling after socket close, the best we can do is just proceed for now... is there any easy way to get the warn back from our current pipeline setup? I'm not too concerned if we can't.

Raynos:
We can resetAll() on socket close.

node/index.js
Raynos: Really this check should be in `send()` and friends. `_writeFrame` probably got called synchronously anyway from `.send()`

Raynos:
lgtm.

PR #101, opened 2015-03-01
node: make v2/handler a readable stream

jcorbin:


node/index.js
```diff
@@ -418,9 +418,7 @@ function TChannelConnection(channel, socket, direction, remoteAddr) {
 
     // TODO: refactor handler to be objectMode Writable
     self.reader.on('data', function onReaderFrame(frame) {
-        if (!self.closing) {
```
Raynos: we should add the `socket.destroy()` to resetAll() before we remove these checks.

Raynos: lol this moved. #nevermind

Raynos:
lgtm.

PR #102, opened 2015-03-01
node: refactor handler to be a Duplex stream

jcorbin:


node/v2/handler.js
```diff
@@ -90,6 +89,24 @@ TChannelV2Handler.prototype._write = function _write(frame, encoding, callback)
...
+    var self = this;
+    if (!callback) {
+        callback = function emitIt(err) {
+            if (err) self.emit('error', err);
+        };
+    }
+    if (self.push(frame)) {
+        callback();
+    } else {
+        self.once('drain', callback);
```
Raynos: you are pushing into your readable side and waiting for your writable side to drain. That's not how it works.

if `write()` returns `false` you should listen to `'drain'`.

I don't know what to listen for if `push()` returns `false`. Maybe `readable` ?

Raynos: Having read the source code, there is no good way to have a drain event.

The best you can do is listen to readable and check `self._readableState.length < self._readableState.highWaterMark`

There is no garantuee that the buffer is below the hwm after a readable in case the user calls `.read(2)` and only reads two bytes but we overfilled the buffer by 2MB.

Raynos:
My suggestion is to remove the callback from `sendFrame` until we have duplexer or something akin to it.

node/v2/handler.js
jcorbin: interesting, I copied exactly what duplexer2 does internally, what did I misunderstand?

jcorbin: actually rather, I copied the suggested `_write` logic from duplexer2's README example in this case.

Raynos: The duplexer2 readme is just wrong.

Raynos: Apparently there is a node core issue for this ( https://github.com/iojs/readable-stream/issues/111 )

jcorbin:
Gotcha, okay reeling back from the callback for now then.

Raynos:
lgtm. (tm)

PR #103, opened 2015-03-01
node: add debug hex dumping of all incoming / outgoing frame data

jcorbin:


node/index.js
```diff
@@ -445,10 +448,27 @@ function TChannelConnection(channel, socket, direction, remoteAddr) {
 
     socket.once('close', clearTimer);
 
-    self.socket
+    var stream = self.socket;
+
+    if (dumpEnabled) {
+        stream = stream.pipe(Spy(process.stdout, {
```
Raynos: this is going to give us nkb of tcp packets. instead of printing nice tchannel frame boundaries.

Is there a way of hooking this up so we hex can show us the frame boundaries ?

node/index.js
```diff
@@ -445,10 +448,27 @@ function TChannelConnection(channel, socket, direction, remoteAddr) {
...
+    }
+
+    stream = stream
         .pipe(self.reader)
         .pipe(self.handler)
         .pipe(self.writer)
+        ;
+
+    if (dumpEnabled) {
+        stream = stream.pipe(Spy(process.stdout, {
```
Raynos: The readable stream for `self.writer` does coalescing upto the high water mark so this might also not be whole frames.

Raynos:
lgtm. Let's figure out the framing problem later.

We might have to not use the streams for the debugging and just instrument the reader and writer directly

jcorbin:
Try it out sometime:
```
$ NODE_DEBUG=tchannel_dump npm test
```

Observe that it does break up the dumps on chunk (aka frame) boundaries.

That's why I had to write v2/spy rather than using hexer's spy directly.

node/v2/spy.js
```diff
@@ -0,0 +1,127 @@
...
+    } else {
+        self.buffer = chunk;
+    }
+    while (self.buffer.length) {
+        switch (self.state) {
+            case States.PendingLength:
+                if (self.buffer.length >= self.expecting) {
+                    self.expecting = self._readLength.call(self.buffer, 0);
+                    self.state = States.Seeking;
+                    self.hex.reset();
```
jcorbin: see this logic right here? it's where we reset hex dup session between each chunk boundary

node/v2/spy.js
```diff
@@ -0,0 +1,127 @@
...
+    function emitIt(err) {
+        self.emit('error', err);
+    }
+};
+
+ChunkSpy.prototype._flush = function _flush(callback) {
+    var self = this;
+    if (self.buffer.length) {
+        self.push(self.buffer);
+        self.hex.write(self.buffer);
```
jcorbin: hmm nuts I should close self.hex here... missed that before merge

node/v2/spy.js
Raynos: i completely forgot to read v2/spy :p Cool.

Raynos:
Nice work josh :)

PR #104, opened 2015-03-01
node: refactor handler <-> channel init boundary

jcorbin:


node/index.js
```diff
@@ -429,15 +429,29 @@ function TChannelConnection(channel, socket, direction, remoteAddr) {
 
     if (direction === 'out') {
         self.handler.sendInitRequest();
-        self.handler.once('identify.out', function onOutIdentified(hostPort) {
+        self.handler.once('init.response', function onOutIdentified(headers) {
+            /* jshint camelcase:false */
+            var hostPort = headers.host_port;
```
Raynos: this is adding v2 coupling to index.js

As long as we are ok with that, then thats fine.

Raynos:
lgtm.

jcorbin:
refactored for less coupling:
- snake_case -> camelCase now contained in v2/handler
- passes out limited object rather than all headers

Raynos:
lgtm.

kriskowal:
woot

PR #106, opened 2015-03-02
Move `docs` out to top-level directory

sectioneight:
This is part of the overall protocol, not specific to the Node.js stuff.

@jcorbin @Raynos


jcorbin:
No we should just remove the old v1 protocol doc, see #92 for the import of the v2 doc.

Raynos:
:+1:

PR #107, opened 2015-03-02
Splice Python implementation into python/ subdirectory

sectioneight:
I did some weird git stuff to rewrite the commits from uber/tchannel.py into this repo. There is definitely some stuff we'll want to cleanup/change (e.g. the repo URL in setup.py) but I figured it would be saner to do the splice first, then the cleanup.

#77

@uber/soap @mranney @jcorbin

PR #108, opened 2015-03-02
Add python subdirectory from tchannel.py.git

sectioneight:
Like #107 but with a less weird way of pulling the commits in.


sectioneight:
I don't think the outcome is any different @jcorbin but here it is with a subtree

jcorbin:
:+1: thanks for keeping things consistent and our history / metadata clean :-)

sectioneight:
Alright, hold on to your butts, I'm going to merge this thing.

Issue #109, opened 2015-03-02
Travis CI / coveralls strategy?


sectioneight:
I have Travis/Coveralls set up nicely for the tchannel.py repo. I think we can get travis to do the Right Thing here by running all the tests for every PR, but I'm less clear about coveralls. Should we just turn off coveralls? Or try to merge the two coverage reports together?

Raynos:
I'd be good to have two travis projects somehow. Or at least have two build statuses.

sectioneight:
I'm not super familiar with how/if that would work. I think to start we should just get a top-level .travis.yml file and populate it with the bits to run a Node test job and a Python test job.

https://github.com/travis-ci/travis-ci/issues/538

Raynos:
sounds good to me.

sectioneight:
Cool. The Python stuff is up and running.

PR #111, opened 2015-03-02
Add .travis.yml for testing

sectioneight:
Right now it just does the Python tests. Eventually we will want to add
a Node runner here as well.

References #109


Raynos:
lgtm.

.travis.yml
```diff
@@ -0,0 +1,11 @@
+---
+language: python
```
sectioneight: We may need to revisit this when we add node tests. I don't think it matters, but then I'm not sure why it asks...

PR #112, opened 2015-03-02
Add Travis build badge to readme

sectioneight:
It's green and shiny.

Preview: https://github.com/uber/tchannel/blob/build-badge/README.md

Not in love with the markdown syntax/sooper-long line but this seems to be the way people do this.


Raynos:
lgtm.

PR #113, opened 2015-03-02
Refactor calls to take a Context object

sectioneight:
I want to remove the knowledge of Framing from any library/layer outside
of TChannel, similar to how TCP just exposes a stream. Part of this
includes returning a Context object instead of a Frame.

I attached the Message into the context to avoid the awkward tuple unpacking.

@uber/soap

### DIFF
--- a/python/docs/api.rst
+++ b/python/docs/api.rst
@@ -28,3 +28,10 @@ Connection
 
 .. automodule:: tchannel.connection
     :members:
+
+
+Context
+-------
+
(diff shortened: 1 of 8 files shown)


jcorbin:
Remarking from the node side, we're chasing exactly this as an objective from the node side this week.  It'd be interesting to compare notes on API plans at some point.

abhinav:
:shipit: 

Yeah, we should try to go for a consistent API between languages.

PR #114, opened 2015-03-02
First pass at multi-language test matrix

sectioneight:
I think this could be cleaner if we put a ``Makefile`` in each
subdirectory, or maybe just a single top-level Makefile that runs all
the tests, but this is a start.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,16 +2,14 @@
 language: python
 
 env:
-  - LANGUAGE=node
-  - LANGUAGE=python
+  - SUBDIR=node
+  - SUBDIR=python
 
 install:
-  - if [ "$LANGUAGE" == "python" ] ; then pushd python && pip install tox coveralls && python setup.py develop && popd ; fi
-  - if [ "$LANGUAGE" == "node" ] ; then pushd node && npm install && popd ; fi
+  - pushd $SUBDIR && make install && popd
 
 script:
-  - if [ "$LANGUAGE" == "python" ] ; then cd python && tox ; fi
-  - if [ "$LANGUAGE" == "node" ] ; then cd node && npm test ; fi
+  - cd $SUBDIR && make test_ci
 
 after_success:
   - coveralls
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,11 +1,17 @@
 ---
 language: python
 
+env:
+  - LANGUAGE=node
+  - LANGUAGE=python
+
 install:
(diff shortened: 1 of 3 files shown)


kriskowal:
Good idea.

jcorbin:
:+1: on having a Makefile for this, especially:
- as we add more languages
- as we add cross-language integration testing, we'll need some demarcation point for "run a client, run a server, ..."

If you're down for it, let's start with something like the following:
```
# node/Makefile
PHONY: install test

install:
  npm install

test:
  npm test
```

.travis.yml
```diff
@@ -1,11 +1,15 @@
 ---
 language: python
 
+env:
+  - SUBDIR=node
```
sectioneight: I have no idea what node version this is using. What versions do we want to target for CI? In python, we're using `tox` to do multi-language (2.7 and 3.4 currently)

jcorbin: There's not a great story for solving node version skew afaik, we baked our own internally to solve this problem. +@raynos for feedback on how this is usually dealt with wrt travis.

Raynos: just configure travis to be 0.10

node/Makefile
```diff
@@ -0,0 +1,7 @@
+.PHONY: install test_ci
+
+install:
+	npm install
+
+test_ci:
```
Raynos: can we also add `test:` that runs `npm test`.

If we start a `cd folder && make test` pattern for local vs `cd folder && make test_ci` pattern for CI that might be nice.

jcorbin: yeah can we just have:
```
test_ci: test

test:
  npm test
```

since there's no difference for node

sectioneight: I did it the other way around, but same effect.

node/Makefile
```diff
@@ -0,0 +1,11 @@
...
+
+.DEFAULT_GOAL := test
+
+install:
+	npm install
+
+test_ci:
+	npm test
+
+test: test_ci
```
sectioneight: like this

jcorbin:
:+1: merge at will

.travis.yml
Raynos: Can we just try setting

```
node_js:
  - "0.10"
  - "0.12"
```

and see what it dose ?

Raynos:
This PR looks good to ship to me.

We can try iterating on multiple node languages in a seperate PR ?

sectioneight:
Yea let's experiment with that next.

PR #115, opened 2015-03-02
Add node linting and dev/release branch updating targets

jcorbin:
Added git-subtree automating targets which update `dev_node` and `release_node` targets respectively.

We can then cut npm release tags off the `release_node` branch, and have internal projects install github references like `uber/tchannel#dev_node`.

Oh plus: we've got lint too, let's expose it.

When the future comes that we're ready to cut a new release tag, we can add tooling in place of `npm version` that writes `node-v*` tags instead.

cc @sectioneight @Raynos

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,10 +6,10 @@ env:
   - SUBDIR=python
 
 install:
-  - pushd $SUBDIR && make install && popd
+  - make -C $SUBDIR install
 
 script:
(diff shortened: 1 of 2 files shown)


Raynos:
If this works then cool story.

jcorbin:
It works, I used the updated command to update dev_node and it installs as `npm install uber/tchannel#dev_node`

PR #116, opened 2015-03-03
Add error code table and cleanup call_res

sectioneight:
Call response now matches the fields in the most recent protocol
document. Error codes are added and translated in ErrorMessage.

@uber/soap


abhinav:
lgtm

PR #117, opened 2015-03-03
Fix Node.js examples and readme

kriskowal:
Since we moved to using ready-signal instead of after.

### DIFF
--- a/node/README.md
+++ b/node/README.md
@@ -16,12 +16,13 @@ Status is being tracked in #78.
 
 ```js
 var TChannel = require('tchannel');
-var after = require('after');
+var CountedReadySignal = require('ready-signal/counted');
 
 var server = new TChannel();
 var client = new TChannel();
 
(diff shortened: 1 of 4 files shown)


kriskowal:
review @raynos @jcorbin

jcorbin:
:+1:

node/README.md
```diff
@@ -16,12 +16,13 @@ Status is being tracked in #78.
...
 ```js
 var TChannel = require('tchannel');
-var after = require('after');
+var CountedReadySignal = require('ready-signal/counted');
 
 var server = new TChannel();
 var client = new TChannel();
 
-var listening = after(2, function (err) {
+var ready = CountedReadySignal(3);
```
Raynos: why 3 ? should that be two ?

Raynos:
other then the 2/3 typo, lgtm.

kriskowal:
Fixed signal error in README

PR #120, opened 2015-03-05
Provide explicit request/response objects

jcorbin:
Reviewers: @Raynos @kriskowal 

- TChannelV2Handler is now coupled to TChannelConnection only by a loose event observer pattern
  - event `call.incoming.request` with a `req` object (`req :: TChannelIncomingRequest` defined in `reqres`)
  - event `call.incoming.response` with a `res` object (`res :: TChannelIncomingResponse` defined in `reqres`)
  - event `call.incoming.error` with an `err` object (`err :: {type, message, errorCode, originalId}` defined in `v2/error_response`)
- TChannelConnection then uses these events to drive completeOutOp and runInOp as before
- TChannelConnection uses `TChannelV2Handler.buildOutgoingResponse(req)` to get a `res :: TChannelOutgoingResponse` object

### DIFF
--- a/node/index.js
+++ b/node/index.js
@@ -411,7 +411,7 @@ function TChannelConnection(channel, socket, direction, remoteAddr) {
     self.handler.on('call.incoming.request', function onCallRequest(req) {
         var handler = self.channel.getEndpointHandler(req.name);
         var res = self.handler.buildOutgoingResponse(req);
-        self.runInOp(handler, req, res.send.bind(res));
+        self.runInOp(handler, req, res);
     });
 
     self.handler.on('call.incoming.response', function onCallResponse(res) {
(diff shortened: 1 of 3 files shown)


kriskowal:
What alternative names did you discard for this "ReqRes" object?

kriskowal:
Pardon, reqres module.

jcorbin:
I considered no others, I found no reason to make 4 files for 4 small classes that will only share more structure in the near (streaming) future

jcorbin:
In case it helps any reviewers, this PR is a non-breaking step on the way to https://github.com/uber/tchannel/compare/dev.josh and beyond.

kriskowal:
Looks correct.

PR #128, opened 2015-03-05
.send() -> .request().send()

jcorbin:
Non breaking move forward which uses TChannelOutgoingRequest to send client calls.

Deprecates the old 5-arg .send in lieu of `.request(opt, cb).send(a1, a2, a3)`.

reviewers: @Raynos @kriskowal

### DIFF
--- a/node/examples/example2.js
+++ b/node/examples/example2.js
@@ -21,6 +21,8 @@
 var TChannel = require('../index.js');
 var CountedReadySignal = require('ready-signal/counted');
 
+var async = require('async');
+
 var server = new TChannel();
 var client = new TChannel();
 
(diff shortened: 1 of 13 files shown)


kriskowal:
:+1:

node/examples/example2.js
```diff
@@ -21,6 +21,8 @@
 var TChannel = require('../index.js');
 var CountedReadySignal = require('ready-signal/counted');
 
+var async = require('async');
```
Raynos: nitpick: `async` as a dep for examples seems a bit heavy.

Raynos:
lgtm.

PR #131, opened 2015-03-06
create proper server socket error

Raynos:
TChannel used to eat server errors and much sadness would be
had when you tried to figure out why your tests do not work.

My emitting an error we have insight into what's going on.

```
AUTOBAHN 30809: error: null server socket error: { [Error: tchannel: listen EADDRINUSE]
  code: 'EADDRINUSE',
  errno: 'EADDRINUSE',
  syscall: 'listen',
  requestedPort: 36431,
  host: '10.32.163.144' } ~ null

events.js:72
        throw er; // Unhandled 'error' event
              ^
Error: tchannel: listen EADDRINUSE
    at Server.onServerSocketError (/home/raynos/uber/autobahn/node_modules/tchannel/index.js:98:27)
    at Server.EventEmitter.emit (events.js:95:17)
    at net.js:1046:12
    at process._tickDomainCallback (node.js:459:13)
```

Above is an example of the print statement of this branch.

Here we see what the failure is and because we create a new error we see a useful stack trace. Also by prefixing the EADDRINUSE with `tchannel` we know its a `listen()` failure from a tchannel server.

### DIFF
--- a/node/index.js
+++ b/node/index.js
@@ -28,9 +28,9 @@ var globalNow = Date.now;
 var globalRandom = Math.random;
 var net = require('net');
 var format = require('util').format;
-var inspect = require('util').inspect;
 var Spy = require('./v2/spy');
 var TypedError = require('error/typed');
+var WrappedError = require('error/wrapped');
 
 var dumpEnabled = /\btchannel_dump\b/.test(process.env.NODE_DEBUG || '');
 
@@ -41,6 +41,13 @@ var TChannelApplicationError = TypedError({
     arg1: null,
     arg2: null,
     arg3: null
+});
+
+var TChannelListenError = WrappedError({
+    type: 'tchannel.server.listen-failed',
+    message: 'tchannel: {origMessage}',
+    requestedPort: null,
+    host: null
 });
 
 function TChannel(options) {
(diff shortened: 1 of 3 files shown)


node/index.js
```diff
@@ -105,7 +105,15 @@ function TChannel(options) {
...
     self.serverSocket.on('error', function onServerSocketError(err) {
-        self.logger.error(self.hostPort + ' server socket error: ' + inspect(err));
+        var serverError = new Error('tchannel: ' + err.message);
+        serverError.code = err.code;
+        serverError.errno = err.errno;
+        serverError.syscall = err.syscall;
+        serverError.requestedPort = self.requestedPort;
+        serverError.host = self.host;
+
+        self.logger.error(self.hostPort + ' server socket error: ' + inspect(serverError));
```
kriskowal: Dynamic string. Pass serverError properties as second arg.

Raynos:
@kriskowal @jcorbin Updated this PR.

I fixed the `logger.error()` callsite and added a `WrappedError` base class to dry up wrapping node core IO errors in a typederror thingy.

node/index.js
```diff
@@ -105,7 +112,20 @@ function TChannel(options) {
         }
     });
     self.serverSocket.on('error', function onServerSocketError(err) {
-        self.logger.error(self.hostPort + ' server socket error: ' + inspect(err));
+        if (err.code === 'EADDRINUSE') {
```
Raynos: note: we should only wrap EADDRINUSE errors; any other errors will be bubbled up without wrapping context.

Maybe we should wrap them in a generic `TChannelServerError` instead ...

node/test/regression-listening-on-used-port.js
```diff
@@ -0,0 +1,34 @@
...
+
+test('listening on a used port', function t(assert) {
+    var otherServer = TChannel();
+    var server = TChannel();
+
+    otherServer.once('listening', onPortAllocated);
+    otherServer.listen(0, 'localhost');
+
+    function onPortAllocated() {
+        server.on('error', onError);
```
jcorbin: once?

node/test/regression-listening-on-used-port.js
```diff
@@ -0,0 +1,34 @@
...
+        assert.equal(err.requestedPort,
+            otherServer.address().port);
+        assert.equal(err.host, 'localhost');
+        assert.equal(err.code, 'EADDRINUSE');
+        assert.equal(err.errno, 'EADDRINUSE');
+        assert.equal(err.syscall, 'listen');
+        assert.notEqual(-1, err.origMessage
+            .indexOf('listen EADDRINUSE'));
+
+        otherServer.close();
```
jcorbin: does close take a callback? can it be `server.close(assert.end)` ?

jcorbin:
:+1: , feedback to taste

node/test/regression-listening-on-used-port.js
Raynos: `once('error')` causes uncaught exceptions. it's bad pattern

node/test/regression-listening-on-used-port.js
Raynos: sure. Doesn't really matter though.

PR #134, opened 2015-03-09
Rework the request sending api

jcorbin:
reviewers: @Raynos @kriskowal 

Old:
```
.send(opt, a1, a2, a3, cb)
```

New:
```
.request(opt)
.send(a1, a2, a3, cb)
```

New option:
```
.request(opt)
.on('error', onError)
.on('response', onResponse)
.send(a1, a2, a3)
```

### DIFF
--- a/node/MIGRATION.md
+++ b/node/MIGRATION.md
@@ -10,9 +10,22 @@ chan
     .send(options, a1, a2, a3, cb);
 ```
 
-After:
+After (simple):
 ```
 chan
-    .request(options, cb)
+    .request(options)
+    .send(a1, a2, a3, cb);
+```
+
+After (if needed/useful):
+```
+chan
+    .request(options)
+    .on('error', function onError(err) {
+        console.error('too bad:', err);
+    })
+    .on('response', function onResponse(res) {
+        console.log('got:', res.arg2, res.arg3);
+    })
     .send(a1, a2, a3);
 ```
--- a/node/index.js
+++ b/node/index.js
@@ -627,7 +627,8 @@ TChannelConnection.prototype.checkOutOpsForTimeout = function checkOutOpsForTime
 TChannelConnection.prototype.onReqTimeout = function onReqTimeout(op) {
     var self = this;
     op.timedOut = true;
-    op.callback(new Error('timed out'), null, null); // TODO typed error
+    op.req.emit('error', new Error('timed out')); // TODO typed error
+    // TODO: why don't we pop the op?
     self.lastTimeoutTime = self.channel.now();
 };
 
--- a/node/index.js
+++ b/node/index.js
@@ -750,9 +750,9 @@ TChannelConnection.prototype.runInOp = function runInOp(handler, req, res) {
 };
 
 /* jshint maxparams:6 */
-function TChannelServerOp(connection, handler, start, options, callback) {
+function TChannelServerOp(connection, handler, start, req, callback) {
     var self = this;
-    self.options = options;
+    self.req = req;
     self.connection = connection;
     self.logger = connection.logger;
     self.handler = handler;
@@ -761,7 +761,7 @@ function TChannelServerOp(connection, handler, start, options, callback) {
     self.callback = callback;
     self.responseSent = false;
     process.nextTick(function runHandler() {
-        self.handler(self.options.arg2, self.options.arg3, connection.remoteName, sendResponse);
+        self.handler(self.req.arg2, self.req.arg3, connection.remoteName, sendResponse);
     });
     function sendResponse(err, res1, res2) {
         self.sendResponse(err, res1, res2);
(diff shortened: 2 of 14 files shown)


node/MIGRATION.md
```diff
@@ -10,9 +10,22 @@ chan
...
+After (if needed/useful):
+```
+chan
+    .request(options)
+    .on('error', function onError(err) {
+        console.error('too bad:', err);
+    })
+    .on('response', function onResponse(res) {
+        console.log('got:', res.arg2, res.arg3);
+    })
```
kriskowal: And the even long story that we’re shooting for is streaming request and streaming response, so instead of request.send(a1, a2, a3), we’ll have request.arg1.end(name); request.arg2.end(headers); body.pipe(request.arg3), and similar for response.arg* streams. Is this correct?

jcorbin: Correct, that could well be what the end api looks like.

node/reqres.js
```diff
@@ -86,23 +106,39 @@ function TChannelOutgoingResponse(id, options, sendFrame) {
...
 module.exports.IncomingRequest = TChannelIncomingRequest;
 module.exports.IncomingResponse = TChannelIncomingResponse;
 module.exports.OutgoingRequest = TChannelOutgoingRequest;
 module.exports.OutgoingResponse = TChannelOutgoingResponse;
+
+function isError(obj) {
+    return typeof obj === 'object' && (
+        Object.prototype.toString.call(obj) === '[object Error]' ||
+        obj instanceof Error);
+}
```
kriskowal: The pervasive replication of this non-normalized logic needs to stop. I swear to you on a stack of Silmarillions, `Object.prototype.toString.call(obj) === '[object Error]'` is equivalent and sufficient for all values of `obj` including `null`, `undefined`, numbers, strings, custom objects, and instances of `Error` from other JavaScript contexts, unless someone decides to subvert Object.prototype.toString, which one would only ever do to deliberately alter the behavior of this function.

kriskowal:
Gorgeous. Please land.

node/index.js
```diff
@@ -614,7 +617,7 @@ TChannelConnection.prototype.checkOutOpsForTimeout = function checkOutOpsForTime
                 });
             continue;
         }
-        var timeout = op.options.timeout || self.channel.reqTimeoutDefault;
+        var timeout = op.req.ttl || self.channel.reqTimeoutDefault;
```
Raynos: are we populating the `ttl` from the `timeout` ?

Are `ttl` and `timeout` even the same thing ?

I think `ttl` is telling autobahn router "when should you stop retrying" where as `timeout` is telling a local tchannel client when you should you time out a pending operation.

Generally in production you want `ttl = timeout - fudgefactor` ?

node/reqres.js
```diff
@@ -75,9 +86,18 @@ function TChannelOutgoingRequest(id, options, sendFrame) {
...
 }
 
-TChannelOutgoingRequest.prototype.send = function send(arg1, arg2, arg3) {
+inherits(TChannelOutgoingRequest, EventEmitter);
+
+TChannelOutgoingRequest.prototype.send = function send(arg1, arg2, arg3, callback) {
     var self = this;
     self.sendFrame(arg1, arg2, arg3);
+    if (callback) {
+        callback = once(callback);
```
Raynos: suggestion: use a technique like ( https://github.com/tommymessbauer/hammock/blob/master/lib/response.js#L32-L39 )

This means you will remove listeners and not leave dangling error or response listeners on the `TChannelOutgoingRequest` instance.

node/reqres.js
```diff
@@ -86,23 +106,39 @@ function TChannelOutgoingResponse(id, options, sendFrame) {
     }
     options = options || {};
     var self = this;
+    EventEmitter.call(self);
     self.id = id || 0;
     self.code = options.code || 0;
     self.tracing = options.tracing || emptyTracing;
     self.headers = options.headers || {};
     self.checksumType = options.checksumType || 0;
+    self.ok = true;
```
Raynos: can we add a `self.ok` field to incoming response for symmetry.

node/reqres.js
Raynos: nope. `Object.prototype.toString.call(Object.create(new Error('foo')))`

Raynos: i dont know what the `obj === 'object'` check is for tbh, maybe performance ??

node/v2/handler.js
```diff
@@ -200,20 +200,19 @@ TChannelV2Handler.prototype.sendRequestFrame = function sendRequestFrame(req, ar
     return id;
 };
 
-TChannelV2Handler.prototype.sendResponseFrame = function sendResponseFrame(res, err, res1, res2) {
+TChannelV2Handler.prototype.sendResponseFrame = function sendResponseFrame(res, arg1, arg2, arg3) {
```
Raynos: noice. arg1, arg2, arg3 :)

Raynos:
Finished review. ttl vs timeout conflation feels iffy to me.

Issue #135, opened 2015-03-09
Better error tracking


kriskowal:
Resolve how to encode an application error with as-thrift and as-json. The scope of this issue does not extend to communicate transport level code.

Raynos:
Error encode; #muchhard

Raynos:
Application errors have well defined encoding formats for json and thrift

PR #136, opened 2015-03-09
Only allow outgoing req/res to be sent once

jcorbin:


node/reqres.js
```diff
@@ -122,18 +130,25 @@ function TChannelOutgoingResponse(id, options, sendFrame) {
...
 inherits(TChannelOutgoingResponse, EventEmitter);
 
 TChannelOutgoingResponse.prototype.send = function send(err, res1, res2) {
     var self = this;
-    if (err) {
-        self.ok = false;
-        var errArg = isError(err) ? err.message : JSON.stringify(err); // TODO: better
-        self.sendFrame(self.name, res1, errArg);
+    if (self.sent) {
+        self.emit('error', new Error('response already sent'));
```
kriskowal: Does this constitute a programming error or an operational error?

node/reqres.js
```diff
@@ -122,18 +130,25 @@ function TChannelOutgoingResponse(id, options, sendFrame) {
...
     } else {
-        self.sendFrame(self.name, res1, res2);
+        self.sent = true;
+        if (err) {
+            self.ok = false;
+            var errArg = isError(err) ? err.message : JSON.stringify(err); // TODO: better
```
kriskowal: For reference, #135

kriskowal: There is an Error frame type. Is that intended for this case?

jcorbin: no this is application error only, the error frame is protocol level error exclusively.

kriskowal:
Looks good to me.

PR #137, opened 2015-03-09
Use Fingerprint32 instead of Hash32.

mranney:
Getting a consistent result from Hash32 across different runtimes is harder than we thought. We should use the slightly slower but more consistent Fingerprint32 instead.


kriskowal:
Looks good to me.

Raynos:
lgtm

PR #139, opened 2015-03-09
Refactor handling wrt req/res

jcorbin:
Use req/res objects throughout the handling paths; sets the stage for other forms of handling.

reviewers: @Raynos @kriskowal

### DIFF
--- a/node/index.js
+++ b/node/index.js
@@ -417,9 +417,7 @@ function TChannelConnection(channel, socket, direction, remoteAddr) {
 
     // TODO: refactor op boundary to pass full req/res around
     self.handler.on('call.incoming.request', function onCallRequest(req) {
-        var handler = self.channel.getEndpointHandler(req.name);
-        var res = self.handler.buildOutgoingResponse(req);
-        self.runInOp(handler, req, res);
+        self.handleCallRequest(req);
     });
 
     self.handler.on('call.incoming.response', function onCallResponse(res) {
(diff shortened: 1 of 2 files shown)


node/index.js
```diff
@@ -417,9 +419,7 @@ function TChannelConnection(channel, socket, direction, remoteAddr) {
 
     // TODO: refactor op boundary to pass full req/res around
     self.handler.on('call.incoming.request', function onCallRequest(req) {
-        var handler = self.channel.getEndpointHandler(req.name);
-        var res = self.handler.buildOutgoingResponse(req);
-        self.runInOp(handler, req, res);
+        self.handleRequest(req);
```
Raynos: how does this work without res argument ?

Raynos: same name on different objects with different signatures.. fair enough

node/index.js
```diff
@@ -731,42 +731,50 @@ TChannelConnection.prototype.request = function request(options) {
...
+
+    var done = false;
+    process.nextTick(function runHandler() {
+        self.channel.handleRequest(req, res);
+    });
 
-    function opDone(err, res1, res2) {
+    res.on('end', opDone);
+    res.on('error', function onResError(err) {
+        self.logger.error(err);
```
Raynos: `self.logger.error('unexpect response failure??', err)` First arg is a string

node/index.js
```diff
@@ -731,42 +731,50 @@ TChannelConnection.prototype.request = function request(options) {
...
+        self.channel.now(), req, res);
+
+    var done = false;
+    process.nextTick(function runHandler() {
+        self.channel.handleRequest(req, res);
+    });
 
-    function opDone(err, res1, res2) {
+    res.on('end', opDone);
+    res.on('error', function onResError(err) {
```
Raynos: why do we handle error here ? feels like the wrong place to handle. `channel.handleRequest()` should really handle it

Raynos: the answer is nobody emits error so just remove this thingy.

Raynos:
lgtm.

PR #140, opened 2015-03-10
Refactor request handling to be injectable with a TRouter default

jcorbin:
TODO before merge:
- [ ] update README, MIGRATION, etc
- [ ] provide a compatibility shim

reviewers: @Raynos @kriskowal


node/index.js
```diff
@@ -77,7 +78,16 @@ function TChannel(options) {
 
     self.peers = Object.create(null);
 
-    self.endpoints = Object.create(null);
+    if (self.options.handleRequest) {
+        self.router = null;
+        self.handleRequest = self.options.handleRequest;
+    } else {
+        self.router = TRouter();
```
kriskowal: <3 Trouter

kriskowal: Perhaps else if options.router, curry handleRequest on options.router. So provide handleRequest, provide router, or fall back to default.

node/index.js
```diff
@@ -77,7 +78,16 @@ function TChannel(options) {
 
     self.peers = Object.create(null);
 
-    self.endpoints = Object.create(null);
+    if (self.options.handleRequest) {
+        self.router = null;
+        self.handleRequest = self.options.handleRequest;
```
Raynos: Let's `null` out the `handleRequest` field in the constructor.

Raynos: Same for `self.router`

node/reqres.js
```diff
@@ -42,7 +42,7 @@ function TChannelIncomingRequest(id, options) {
     self.remoteAddr = null;
     self.headers = options.headers || {};
     self.checksumType = options.checksumType || 0;
-    self.name = options.name || '';
+    self.arg1 = options.arg1 || emptyBuffer;
```
Raynos: noice.

node/trouter.js
```diff
@@ -0,0 +1,151 @@
...
+
+function TRouter() {
+    if (!(this instanceof TRouter)) {
+        return new TRouter();
+    }
+    var self = this;
+    self.services = Object.create(null);
+}
+
+TRouter.prototype.register =
```
Raynos: If possible; let's pick a single canonical name for this method instead of two.

node/trouter.js
```diff
@@ -0,0 +1,151 @@
...
+TRouter.prototype.registerEndpoint =
+function registerEndpoint(service, endpoint, handler) {
+    var self = this;
+    var svc = self.services[service];
+    if (!svc) {
+        svc = self.registerService(service, TRouterService(service));
+        throw ServiceAlreadyDefinedError({
```
Raynos: should this err be in the `else` case ?

Raynos: Ah looks like a bad refactor; this belongs in the other method

node/trouter.js
```diff
@@ -0,0 +1,151 @@
...
+            service: service,
+            oldHandler: self.services[service],
+            newHandler: handler
+        });
+    }
+    return svc.register(endpoint, handler);
+};
+
+TRouter.prototype.registerService = function registerService(service, handler) {
```
Raynos: consider naming the args `serviceName, handleOrService`

node/trouter.js
```diff
@@ -0,0 +1,151 @@
...
+    }
+    return handler;
+};
+
+TRouter.prototype.handleRequest = function handleRequest(req, res) {
+    var self = this;
+    var svc = self.services[req.service];
+    if (!svc) {
+        res.send(NoSuchServiceError({
+            service: req.service
```
Raynos: maybe pass in `Object.keys(self.services)` to advertise available services

node/trouter.js
```diff
@@ -0,0 +1,151 @@
...
+    var svc = self.services[service];
+    if (!svc) {
+        svc = self.registerService(service, TRouterService(service));
+        throw ServiceAlreadyDefinedError({
+            service: service,
+            oldHandler: self.services[service],
+            newHandler: handler
+        });
+    }
+    return svc.register(endpoint, handler);
```
Raynos: `svc` might not have a `register()` method and might be `{ handleRequest: Function }`

In this case we should throw a "cannot register with service conflict bla bla" error

kriskowal:
Muy bueno.

node/trouter.js
```diff
@@ -0,0 +1,151 @@
...
+TRouter.prototype.registerService = function registerService(service, handler) {
+    var self = this;
+    if (self.services[service] !== undefined) {
+        throw ServiceAlreadyDefinedError({
+            service: service,
+            oldHandler: self.services[service],
+            newHandler: handler
+        });
+    }
+    if (typeof handler === 'function') {
```
Raynos: this shortcut is probably going to hurt us.

node/trouter.js
```diff
@@ -0,0 +1,151 @@
...
+    var self = this;
+    if (self.endpoints[name] !== undefined) {
+        throw EndpointAlreadyDefinedError({
+            service: this.name,
+            endpoint: name,
+            oldHandler: self.endpoints[name],
+            newHandler: handler
+        });
+    }
+    if (typeof handler !== 'function') {
```
Raynos: :+1:

node/trouter.js
```diff
@@ -0,0 +1,151 @@
...
+        };
+    } else {
+        self.services[service] = handler;
+    }
+    return handler;
+};
+
+TRouter.prototype.handleRequest = function handleRequest(req, res) {
+    var self = this;
+    var svc = self.services[req.service];
```
Raynos: coerce `req.service` to a string ? or is it already a string ?

node/trouter.js
```diff
@@ -0,0 +1,151 @@
...
+    }
+    svc.handleRequest(req, res);
+};
+
+function TRouterService(name) {
+    if (!(this instanceof TRouterService)) {
+        return new TRouterService(name);
+    }
+    var self = this;
+    self.name = name;
```
Raynos: consider `self.service` instead

node/trouter.js
```diff
@@ -0,0 +1,151 @@
...
+    self.endpoints = Object.create(null);
+}
+
+TRouterService.prototype.register = function register(name, handler) {
+    var self = this;
+    if (self.endpoints[name] !== undefined) {
+        throw EndpointAlreadyDefinedError({
+            service: this.name,
```
Raynos: Having `name` and `this.name` is confusing.

Raynos:
Looks like the interface is 

```js
tchannel.router.register('app', 'arg1', handler);
```

Really I would like an interface like

```js
var ringpopRouter = tchannel.router.service('ringpop');

ringpopRouter.register('v1_proxy_req', handler)
```

This probably the cleanest I can think off. It allows us to avoid the "two string arguments and what is their order again?" issue.

It also makes the tree nature of the routing a bit more explicit.

Raynos:
This is ready for review. 

Raynos:
Wait a second...

 - [x] fixed examples
 - [x] fixed README & jsig
 - [x] fixed tests
 - [x] fixed benchmarks
 - [x] Updated implementation to add `handler()`
 - [ ] Update the MIGRATION.md
 - [ ] Provide a compatibility shim

This is still good for review though! I will add some more commits with MIGRATION.md documentation and a compatibility shim >_<

kriskowal:
Looks good to me. Compatibility does appear to be broken in more than just the interface. I’m seeing the following type of message a lot when examples are run against tcap with the last tchannel release.

```
session=4 127.0.0.1:63122 --> 127.0.0.1:4040 frame parse error TchannelShortBufferError
buffer smaller than 1638 bytes, got 17
```

These errors do not occur anymore if I link tcap to this tchannel branch. It isn’t clear to me that this branch has backward incompatible changes on the wire by looking at the code, but it is worth noting in migration.

Raynos:
@kriskowal that is the `service~2` -> `service~1` change that is already in master.

We need to update all things that depend on tchannel to the latest alpha tag once a new one is cut.

Raynos:
This now does everything I said it owuld do.

jcorbin:
lgtm

kriskowal:
lgtm

PR #141, opened 2015-03-10
tchannel golang initial implementation

mmihic:
Initial golang tchannel implementation.  See [README](golang/README.md) for high level description, and  [TODO](golang/TODO.md) for list of TODO items.


golang/README.md
```diff
@@ -0,0 +1,152 @@
...
+
+NOTE: `master:golang` is **not yet stable**
+
+## Example
+
+```golang
+import (
+    "encoding/json"
+    "time"
+    "github.com/uber/tchannel"
```
abhinav: Shouldn't this be tchannel/golang?

golang/README.md
```diff
@@ -0,0 +1,152 @@
...
+
+    // Make a call to ourselves, with a timeout of 10s
+    ctx, cancel := context.WithTimeout(context.Background(), time.Second * 10)
+    defer cancel()
+
+    var responseHeaders Headers
+    var pong Pong
+    if err := client.RoundTrip("localhost:8050", "PingService", "ping",
+        NewJSONOutput(Headers{}),  NewJSONOutput(Ping{Message: "Hello world"}),
+        NewJSONInput(&responseHeaders), NewJSONInput(&pong)); err != nil {
```
abhinav: These calls probably need to be prepended with `tchannel.`

golang/Godeps/Godeps.json
```diff
@@ -0,0 +1,31 @@
+{
+	"ImportPath": "code.uber.internal/personal/mmihic/tchannel-go",
```
sectioneight: We probably want to scrub this

mmihic: oops, yes, need to cleanup all of these imports.

golang/Makefile
```diff
@@ -0,0 +1,58 @@
...
+	mkdir -p $(BUILD)
+	go test $(PKGS) $(TEST_ARG) -parallel=4
+
+# target: cover - Runs tests under code coverage
+cover: clean
+	echo Testing packages:
+	mkdir -p $(BUILD)
+	go test ./ $(TEST_ARG)  -coverprofile=$(BUILD)/coverage.out
+	go tool cover -html=$(BUILD)/coverage.out
+	# TODO(mmihic): Temporarily disabled while working out false positives
```
sectioneight: is this still a TODO?

golang/arguments.go
```diff
@@ -0,0 +1,103 @@
+package tchannel
+
+import (
+	"encoding/json"
+	"io"
+	"io/ioutil"
+)
+
+// An Input is able to read an argument from a call body
```
sectioneight: Suggest making comments golint-compatible (e.g. start the comment with the type name you're describing):

```
// Input describes a type...
```

golang/channel.go
```diff
@@ -0,0 +1,160 @@
...
+}
+
+// A TChannel is a bi-directional connection to the peering and routing network.  Applications
+// can use a TChannel to make service calls to remote peers via BeginCall, or to listen for incoming calls
+// from peers.  Once the channel is created, applications should call the ListenAndHandle method to
+// listen for incoming peer connections.  Because channels are bi-directional, applications should call
+// ListenAndHandle even if they do not offer any services
+type TChannel struct {
+	log               *logging.Logger
+	hostPort          string
```
sectioneight: Do we want any of these fields to be exported? e.g. `HostPort`?

golang/Makefile
mmihic: yes annoyingly go-logging results in false positives from vet due to having an Error method.  There's a TODO to remove the go-logging dependency which will resolve; I held off doing that in this branch but can do so.

golang/channel.go
mmihic: Prefer not to export immutable fields.

mmihic: I can add a method for this if we think it we want access.

golang/checksum.go
```diff
@@ -0,0 +1,82 @@
...
+	switch t {
+	case ChecksumTypeNone:
+		return NullChecksum{}
+	case ChecksumTypeCrc32:
+		return NewCrc32Checksum()
+	case ChecksumTypeFarmhash:
+		return nil
+	}
+
+	return nil
```
sectioneight: Might make more sense as the `default:` case

golang/checksum.go
```diff
@@ -0,0 +1,82 @@
...
+
+// Creates a new Checksum of the given type
+func (t ChecksumType) New() Checksum {
+	switch t {
+	case ChecksumTypeNone:
+		return NullChecksum{}
+	case ChecksumTypeCrc32:
+		return NewCrc32Checksum()
+	case ChecksumTypeFarmhash:
+		return nil
```
sectioneight: What does this mean for callers? Should there be a TODO about support? Or should we make this a NullChecksum if we're never going to support this?

mmihic: There's a TODO for this.

golang/connection.go
```diff
@@ -0,0 +1,475 @@
...
+// Information about a TChannel peer
+type PeerInfo struct {
+	// The host and port that can be used to contact the peer, as encoded by net.JoinHostPort
+	HostPort string
+
+	// The logical process name for the peer, used for only for logging / debugging
+	ProcessName string
+}
+
+func (p PeerInfo) String() string {
```
sectioneight: nice touch

golang/arguments.go
mmihic: This is already golint-compatible (comments for types can start with an article).  But a bunch of other stuff is not (such as the lack of godoc comments on the ReadFrom/WriteTo implementations).  Will update.

golang/TODO.md
```diff
@@ -0,0 +1,105 @@
+h1. Functionality
+
+[ ] Cancel outbound requests
```
sectioneight: This doesn't render properly in github. If you want it to, I think you need to make this a list, e.g. prefix with `- `

golang/connection.go
```diff
@@ -0,0 +1,475 @@
...
+		case MessageTypeCallRes:
+			c.outbound.handleCallRes(frame)
+		case MessageTypeCallResContinue:
+			c.outbound.handleCallResContinue(frame)
+		case MessageTypeInitReq:
+			c.handleInitReq(frame)
+		case MessageTypeInitRes:
+			c.handleInitRes(frame)
+		case MessageTypeError:
+			c.handleError(frame)
```
sectioneight: Looking at the TODO list, I don't see anything about protocol-level ping/pong handling. Am I missing where that is implemented?

golang/arguments.go
sectioneight: Ah good call. And yea doesn't have to be now, just something we should strive for in the long term.

golang/connection.go
mmihic: Nope, you didn't miss it.  Added to the TODO list.

golang/arguments.go
mmihic: I'll do it now; I'd rather have the build enforce linting.

sectioneight:
This looks great. I can't do a full review given the total lines of code but I say ship this and then let's follow up on the TODOs. Maybe see if you can get TravisCI running?

golang/messages.go
```diff
@@ -0,0 +1,487 @@
...
+	if err != nil {
+		return err
+	}
+
+	m.Headers = callHeaders{}
+	if err := m.Headers.read(r); err != nil {
+		return err
+	}
+
+	serviceNameLen, err := r.ReadUint16()
```
abhinav: This is being changed to `service~1` so `ReadByte` here instead.

golang/messages.go
```diff
@@ -0,0 +1,487 @@
...
+
+	if err := w.WriteByte(m.TraceFlags); err != nil {
+		return err
+	}
+
+	if err := m.Headers.write(w); err != nil {
+		return err
+	}
+
+	if err := w.WriteUint16(uint16(len(m.Service))); err != nil {
```
abhinav: `WriteByte`

Raynos:
Is this already in master ? can we close this ?

PR #142, opened 2015-03-10
Fix bug for null response

kriskowal:
This may be masking a larger systemic issue. The polling example drives
this issue, but nothing weird is going on that would cause a response to
arrive after it has been popped. Further investigation required.

I created the polling example to exercise tcap tracking frames from an already open connection.

### DIFF
--- a/node/examples/polling.js
+++ b/node/examples/polling.js
@@ -1,0 +1,51 @@
+// Copyright (c) 2015 Uber Technologies, Inc.
+
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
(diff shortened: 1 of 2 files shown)


kriskowal:
review @raynos @jcorbin

node/examples/polling.js
```diff
@@ -0,0 +1,51 @@
...
+    }, 1000)
+});
+
+var ready = CountedReadySignal(2);
+var listening = ready(function (err) {
+    if (err) {
+        throw err;
+    }
+
+    setInterval(function () {
```
jcorbin: can we hoist this function to toplevel and name it like `function pollIt` ?

node/index.js
```diff
@@ -444,6 +444,9 @@ function TChannelConnection(channel, socket, direction, remoteAddr) {
 
     self.handler.on('call.incoming.response', function onCallResponse(res) {
         var op = self.popOutOp(res.id);
+        if (!op) {
+            return;
```
jcorbin: Hmm, if you hit this then I'd ask please implement the pre-existing TODO in popOutOp:
```
        // TODO else case. We should warn about an incoming response for an
        // operation we did not send out.  This could be because of a timeout
        // or could be because of a confused / corrupted server.
```

jcorbin:
I just realized that I don't understand your bug, can you please explain more?

Also if this really was a bug, we should right a test more so than an advisory example.

node/index.js
```diff
@@ -720,9 +723,13 @@ TChannelConnection.prototype.popOutOp = function popOutOp(id) {
...
     var op = self.outOps[id];
     if (!op) {
-        // TODO else case. We should warn about an incoming response for an
-        // operation we did not send out.  This could be because of a timeout
-        // or could be because of a confused / corrupted server.
+        // We warn about an incoming response for an operation we did not send
+        // out.  This could be because of a timeout or could be because of a
+        // confused / corrupted server.
+        self.logger.warn('dropped response for forgotten or lost request', {
+            hostPort: self.channel.hostPort,
```
jcorbin: Please also log self.remoteAddr here

node/index.js
```diff
@@ -444,6 +444,9 @@ function TChannelConnection(channel, socket, direction, remoteAddr) {
 
     self.handler.on('call.incoming.response', function onCallResponse(res) {
         var op = self.popOutOp(res.id);
+        if (!op) {
```
Raynos: please add regression tests instead of regression examples

PR #144, opened 2015-03-10
add parse and serialize methods in the call_request message.

junchaowu:
for dev purpose

### DIFF
--- a/python/tchannel/messages/call_request.py
+++ b/python/tchannel/messages/call_request.py
@@ -94,12 +94,11 @@ class CallRequestMessage(BaseMessage):
         )
 
         self.checksum_type = read_number(payload, self.CSUMTYPE_SIZE)
-        csum_size = self.CHECKSUM[self.checksum_type]
         if self.checksum_type:
+            csum_size = self.CHECKSUM[self.checksum_type]
             self.checksum = read_number(payload, csum_size)
 
         self.parse_args(payload)
-
         self.extra_space_check(payload)
 
     def extra_space_check(self, payload):
@@ -140,7 +139,7 @@ class CallRequestMessage(BaseMessage):
         self._write_headers(out, self.headers, self.NH_SIZE, self.HEADER_SIZE)
 
         out.extend(write_number(self.checksum_type, self.CSUMTYPE_SIZE))
-        if self.checksum_type:
+        if self.checksum_type and self.checksum:
             out.extend(write_number(self.checksum,
                                     self.CHECKSUM[self.checksum_type]))
 
--- a/python/tchannel/messages/call_request.py
+++ b/python/tchannel/messages/call_request.py
@@ -1,5 +1,8 @@
 from __future__ import absolute_import
 
+import logging
+import os
+
 from .base import BaseMessage
 from .types import Types
 from ..parser import read_number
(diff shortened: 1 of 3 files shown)


python/tchannel/messages/call_request.py
```diff
@@ -36,6 +36,7 @@ class CallRequestMessage(BaseMessage):
 
     FLAGS_SIZE = 1
     TTL_SIZE = 4
+    # span_id + parent_id + trace_id = 3*8 = 24
```
abhinav: ?

python/tchannel/messages/call_response.py
```diff
@@ -27,3 +32,87 @@ class CallResponseMessage(CallRequestMessage):
         'arg_2',
         'arg_3',
     )
+
+    FLAGS_SIZE = 1
+    CODE_SIZE = 1
+    # span_id + parent_id + trace_id = 3*8 = 24
```
abhinav: I don't understand what this is for

python/tchannel/messages/call_response.py
```diff
@@ -27,3 +32,87 @@ class CallResponseMessage(CallRequestMessage):
...
+
+        self.headers, _ = self._read_headers(
+            payload,
+            self.NH_SIZE,
+            self.HEADER_SIZE,
+        )
+
+        self.checksum_type = read_number(payload, self.CSUMTYPE_SIZE)
+        if self.checksum_type:
+            self.checksum = read_number(payload, self.CSUM_SIZE)
```
abhinav: Checksum size is not exactly to 4 bytes. It just so happens that all supported checksum types are 32-bit. use a mapping from checksum_type to checksum size (might as well update call req while you're at it).

python/tchannel/messages/call_response.py
```diff
@@ -27,3 +32,87 @@ class CallResponseMessage(CallRequestMessage):
...
+
+        # TODO check if we're at the end of the stream by counting bytes
```
abhinav: What?

python/tchannel/messages/call_response.py
```diff
@@ -27,3 +32,87 @@ class CallResponseMessage(CallRequestMessage):
...
+        self.arg_1, _ = read_variable_length_key(
+            payload,
+            self.ARG_SIZE,
```
abhinav: This is not really arg size. It's arg length size.

python/tchannel/messages/call_response.py
junchaowu: done

python/tchannel/messages/call_request.py
```diff
@@ -34,17 +37,24 @@ class CallRequestMessage(BaseMessage):
...
     TTL_SIZE = 4
+    # sizeOf(span_id) + sizeOf(parent_id) + sizeOf(trace_id) = 3*8 = 24
     TRACE_SIZE = 8
     TRACEFLAGS_SIZE = 1
     SERVICE_SIZE = 2
     NH_SIZE = 1
     HEADER_SIZE = 1
     CSUMTYPE_SIZE = 1
-    CSUM_SIZE = 4
+    CSUM_SIZE = CHECKSUM[CSUMTYPE_SIZE]
```
sectioneight: I don't think you can do this at import time -- it's going to vary based on the message.

python/tchannel/messages/call_response.py
```diff
@@ -27,3 +35,95 @@ class CallResponseMessage(CallRequestMessage):
...
+    def parse(self, payload, size):
+        """Parse a call request message from a payload."""
+        self.flags = read_number(payload, self.FLAGS_SIZE)
+        self.code = read_number(payload, self.CODE_SIZE)
+
+        self.span_id = read_number(payload, self.TRACE_SIZE)
+        self.parent_id = read_number(payload, self.TRACE_SIZE)
+        self.trace_id = read_number(payload, self.TRACE_SIZE)
+
+        self.traceflags = read_number(payload, self.TRACEFLAGS_SIZE)
```
sectioneight: This looks copy-pasted from above. We should avoid copying code at all costs. It's generally an indication that some refactoring / generalization is in order.

python/tchannel/messages/call_response.py
```diff
@@ -27,3 +35,95 @@ class CallResponseMessage(CallRequestMessage):
...
+
+    def serialize(self, out):
+        """Write a call request message out to a buffer."""
+        out.extend(write_number(self.flags, self.FLAGS_SIZE))
+        out.extend(write_number(self.code, self.code_SIZE))
+
+        out.extend(write_number(self.span_id, self.TRACE_SIZE))
+        out.extend(write_number(self.parent_id, self.TRACE_SIZE))
+        out.extend(write_number(self.trace_id, self.TRACE_SIZE))
+        out.extend(write_number(self.traceflags, self.TRACEFLAGS_SIZE))
```
sectioneight: Some comment: please don't copy paste code around

python/tchannel/messages/call_request.py
junchaowu: oh you are right.

python/tchannel/messages/call_response.py
junchaowu: this is just the method in the subclass overrides the base class method. There are some difference between subclass one and base class one.

python/tchannel/messages/call_response.py
junchaowu: same comments.

python/tchannel/messages/call_response.py
junchaowu: These two method did almost same thing but for different payload format.

sectioneight: I get that there are some parts that are different, which is why I comment on this part -- reading the zipkin and traceflags info. I feel pretty strongly that outside of tests, we should not have the exact same code copy/pasted into multiple files.

sectioneight: In other words -- extract the span_id, parent_id, trace_id, traceflags stuff (and whatever else is in common) into a method.

python/tchannel/messages/call_response.py
sectioneight: Same response: extract code that is performing the exact same calls into a re-usable method.

python/tchannel/messages/call_response.py
```diff
@@ -27,3 +35,95 @@ class CallResponseMessage(CallRequestMessage):
...
+            self.arg_2,
+            self.ARG_LENGTH,
+            encode=False,
+        )
+        write_variable_length_key(
+            out,
+            self.arg_3,
+            self.ARG_LENGTH,
+            encode=False,
+        )
```
sectioneight: This also looks extractable

blampe:
lgtm

PR #147, opened 2015-03-11
make service a `service~1`

Raynos:
At request from @jcorbin the service field in CallRequest
should be a `service~1`, not a `service~2`.

This probably breaks back compat with the python implementation

cc: @jsu1212 @abhinav


jcorbin:
lgtm, cc @abhinav

abhinav:
:shipit: 

@junchaowu we should update the python version

mranney:
FWIW, this change was discussed elsewhere and universally agreed to be a Good Idea. Thanks for following through.

Issue #124, opened 2015-03-05
node: Implement ping support


jcorbin:
- [ ] ping req/res frame parsers
- [ ] internal handler support
- [ ] API through TChannelConnection?
- [ ] API through TChannel?

jcorbin:
To get started on the frames, look in:
- `node/v2/error_response.js` main point is the `ErrorResponse.RW`
- `node/test/v2/error_response.js` which is a unit test for `ErrorResponse.RW`
- `docs/protocol.md` for frame definitions

PR #148, opened 2015-03-11
add ping payloads

blampe:
tcap isn't able to show these frames because tchannel doesn't know how to parse type codes `0xd0` and `0xd1`. These don't have bodies and so I think `skip(0)` is right but I'm not sure how the `write` method should look. @jcorbin thoughts?

cc @kriskowal

### DIFF
--- a/node/lib/read.js
+++ b/node/lib/read.js
@@ -92,6 +92,11 @@ module.exports.UInt32BE = readWidth(4, function UInt32BE(buffer, offset) {
     return buffer.readUInt32BE(offset);
 });
 
+module.exports.null = naught;
+function naught(buffer, offset) {
+    return [null, offset, null];
+};
+
 module.exports.skip = skip;
 
 function skip(n) {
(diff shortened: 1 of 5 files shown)


jcorbin:
Needs a test, but otherwise this could be a step towards #124

kriskowal:
I’ve taken this a step further, though sideways, perhaps not toward landing but toward exploring. I’ve added lib/read,write "null" frame reader and writer. These have already been implemented with the "null" reader and writer in bufrw, which will replace lib/read,write.js. I also wrote out some scaffolding and TODOs for the ping request and response handler, which raise some questions. Does the incoming ping request handler have enough on hand to construct a ping response and send it back on the channel from which it came? Do we need to generalize the incoming request and response handler methods such that they do not depend on a call request/response body, or should we split those methods into call and ping variations? Both rely on a frame identifier to handler mapping.

breerly:
Should this be merged or closed? Lets act :)

blampe:
Not worrying about this right now, we can re-visit when we start caring more about pings.

breerly:
Closing for now - will stay available here:

https://github.com/uber/tchannel/pulls?q=is%3Apr+is%3Aclosed

PR #151, opened 2015-03-12
destroy sockets on close

Raynos:
The current semantics of calling `.end()` on peers
is incorrect.

This only ends the Writable sends. This means existing
Incoming Peer objects for this server still have an
open readable side.

This means the server can still receive incoming
CallRequests from already accepted sockets.

To close() the server we have to destroy() these
sockets to end both the Writable and Readable side.

reviewers: @rf @jcorbin

cc: @jsu1212


jcorbin:
:ok:

PR #152, opened 2015-03-12
Add description of client-only init option.

mranney:
Synchronous python and various command line utilities find it difficult to support listening for new connections. This change outlines a convention to use for these implementations.

### DIFF
--- a/protocol.md
+++ b/protocol.md
@@ -229,7 +229,7 @@ these listed above.
 
 For connections where listening for new connections is not possible or doesn't make sense,
 implementations should send a `host_port` value of `0.0.0.0:0`. This special value tells
-receiving implementations to use the results of `getpeername` or equivalent API to uniquely
+receiving implementations to use the results of `getpeername(2)` or equivalent API to uniquely
 identify this connection. It also tells receivers that this address is not valid beyond this
 connection, so it should not be forwarded to other nodes.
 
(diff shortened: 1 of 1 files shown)


abhinav:
:shipit:

protocol.md
```diff
@@ -225,7 +225,13 @@ name            | format             | description
 `process_name` | *arbitrary string* | additional identifier for this instance, used for logging
 
 For forward compatibility, implementations should ignore headers other than
-these.
+these listed above.
+
+For connections where listening for new connections is not possible or doesn't make sense,
+implementations should send a `host_port` value of `0.0.0.0:0`. This special value tells
+receiving implementations to use the results of `getpeername` or equivalent API to uniquely
```
jcorbin: niggle, can we spell this `getpeername(2)` idiomatically?

jcorbin:
:+1:

Raynos:
There was another idea floating around to make `0.0.0.0:0` mean 

> Hello autobahn. I do not know what my IP or port is. Please figure it out and then tell me what my host & port are
> I am still a server though! I'm not an ephemeral tcp socket

Should we encode whether a connection is ephemeral or not in a specific flag as part of the init request & response protocol ?

mranney:
@Raynos sadly, the server has no way of knowing what the listen socket is, so the client really has to supply it.

PR #156, opened 2015-03-13
[python] tcurl.py and starting point for tchannel interface

blampe:


python/tchannel/tchannel.py
```diff
@@ -0,0 +1,135 @@
...
+
+log = logging.getLogger('tchannel')
+
+
+class TChannel(object):
+    """Manages inbound and outbound connections to various hosts."""
+
+    def __init__(self):
+        self.peers = {}
+        self.process_name = "%s[%s]" % (sys.argv[0], os.getpid())
```
abhinav: Make this an optional argument and generate this as the default value.

Also, it wouldn't be the worst idea to truncate `sys.argv[0]` to some length.

python/tchannel/tchannel.py
```diff
@@ -0,0 +1,135 @@
...
+    def get_peer(self, hostport):
+        if hostport in self.peers:
+            peer = self.peers[hostport]
+        else:
+            peer = yield self.add_peer(hostport)
+        raise tornado.gen.Return(peer)
+
+    @tornado.gen.coroutine
+    def make_out_connection(self, hostport):
+        host, port = hostport.split(":")
```
abhinav: IPv6 addresses contain colons.

Use `.rsplit(':', 1)` instead

python/tchannel/tchannel.py
```diff
@@ -0,0 +1,135 @@
...
+    def respond(self):
+        # TODO: handle inbound requests
+        raise NotImplementedError
+
+    @tornado.gen.coroutine
+    def request(self, hostport):
+
+        peer_connection = yield self.get_peer(hostport)
+
+        class TChannelClientOp(object):
```
abhinav: What's the advantage of this being a nested class instead of an external class which accepts the `peer_connection` in its constructor?

python/tchannel/tcurl.py
```diff
@@ -0,0 +1,122 @@
...
+
+log = logging.getLogger('tchannel')
+
+
+@tornado.gen.coroutine
+def tcurl(host, endpoint, headers, body):
+
+    client = TChannel()
+
+    def handle_call(context):
```
abhinav: It doesn't look like this is used?

python/tchannel/tcurl.py
```diff
@@ -0,0 +1,122 @@
...
+    print "Sending this to", host
+    print "*" * 80
+    print " arg1:", endpoint
+    print " arg2:", headers
+    print " arg3:", body
+    print
+
+    request = yield client.request(host)
+
+    response = yield request.send(
```
abhinav: I don't like that there are two yields involved in making this call.

Can `request()` be changed to return a simple object that makes the actual connection on `send()` if required?

python/tchannel/tchannel.py
blampe: Good point.

python/tchannel/tchannel.py
blampe: It's just a placeholder right now to mimic the node API. Needs to be flushed out with the response handling story.

python/tchannel/tcurl.py
blampe: `setup.py` installs it as a script.

abhinav: I meant the handle_call function

python/tchannel/tcurl.py
blampe: Totally.

breerly:
blaarrgg :shipit:

PR #157, opened 2015-03-13
[python] Blocking outgoing-only API

abhinav:
This supports outgoing connections only.

`TChannel` manages remote connections and for each connection, spins up two threads: one to send requests and one to ingest responses. Out-of-order responses are supported.

It's currently in the `tchannel.outgoing` because I wasn't sure where to put it.


python/tchannel/outgoing.py
```diff
@@ -0,0 +1,331 @@
...
+        return super(SettableFuture, self).set_result(result)
+
+    def set_exception(self, exception=None):
+        e, tb = sys.exc_info()[1:]
+        if not exception:
+            exception = e
+        super(SettableFuture, self).set_exception_info(exception, tb)
+
+
+class TChannelException(Exception):
```
sectioneight: We already have a base exception `TChannelException` defined in `exceptions.py`

python/tchannel/outgoing.py
```diff
@@ -0,0 +1,342 @@
...
+        self._on_close(self)
+
+    def __del__(self):
+        try:
+            self.close()
+        except:
+            pass
+
+
+class OutgoingTChannel(object):
```
blampe: Both `TChannel` interfaces support outgoing so I wonder if we could call this something more descriptive. Otherwise lgtm, I look forward to the merge conflicts :)

abhinav: I'm hoping that namespacing them by module will be enough.

abhinav: Merging this for now. We can clear out the naming later.

PR #160, opened 2015-03-16
Update protocol.md

breerly:
Fixing typo

### DIFF
--- a/protocol.md
+++ b/protocol.md
@@ -174,7 +174,7 @@ decode some part of the payload.
 An identifier for this message that is chosen by the sender of a request. This
 id is only valid for this sender on this connection. This is similar to how TCP
 has a sequence number in each direction. Each side of a connection may happen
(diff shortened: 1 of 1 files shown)


kriskowal:
This spelling seems deliberate to me, to indicate that the uniqueness of an identifier is a combination of its value and the side it came from.

breerly:
Aaah - that makes sense :)

protocol.md
```diff
@@ -174,7 +174,7 @@ decode some part of the payload.
 An identifier for this message that is chosen by the sender of a request. This
 id is only valid for this sender on this connection. This is similar to how TCP
 has a sequence number in each direction. Each side of a connection may happen
-to select overlapping message ids, which is fine because they are directional.
+to select overlapping message ids, which is fine because they are bi-directional.
```
mmihic: Message ids are uni-directional (scoped to the side initiating the message exchange) not bi-directional.

PR #161, opened 2015-03-16
Remove `resp.send()` in OutgoingCallResponse.

Raynos:
This PR removes the automatic handling of errors
for the application developer.

It adds two new methods.

 - `resp.sendOk(arg2, arg3)`
 - `resp.sendNotOk(arg2, arg3)`

and removes `resp.send(err?, arg2, arg3)`

Since we are no longer having an opinion about error serialization
we also do not convert `IncomingCallResponse` into an error if
its `ok` boolean is set to `false`.

This means, just like with HTTP, the application developer
gets a response when making an outgoing `request` and has
to check the `ok` boolean on the response (similar to how you check
`statusCode` for HTTP).

The outgoing `request()` can still result in error objects though.
For example

 - time outs
 - network (TCP) issues
 - TChannel error frames

cc: @jsu1212

### DIFF
--- a/node/MIGRATION.md
+++ b/node/MIGRATION.md
@@ -30,10 +30,10 @@ chan.handler.register('my-endpoint', function (req, res) {
     // req.arg2, req.arg3, req.remoteAddr
 
     // either
-    res.send(new Error('oops'));
+    res.sendNotOk(null, 'oops');
 
     // or
-    res.send(null, 'res1', 'res2');
+    res.sendOk(null, 'res1', 'res2');
 });
 ```
 
@@ -42,14 +42,23 @@ chan.handler.register('my-endpoint', function (req, res) {
 Before:
 ```
 chan
-    .send(options, a1, a2, a3, cb);
+    .send(options, a1, a2, a3, function (err, res1, res2) {
+        // CallResponse error is in `err`
+        // CallResponse ok is in `res1`, `res2`
+    });
 ```
 
 After (simple):
 ```
 chan
     .request(options)
-    .send(a1, a2, a3, cb);
+    .send(a1, a2, a3, function (err, res) {
+        // CallResponse is in `res.arg2`, `res.arg3`
+        // res has an `.ok` field that is tells whether it is
+        // an error or not.
+
+        // `err` is for IO & timeouts. Also tchannel error frames
+    });
 ```
 
 After (if needed/useful):
--- a/node/MIGRATION.md
+++ b/node/MIGRATION.md
@@ -57,7 +57,11 @@ chan
         // res has an `.ok` field that is tells whether it is
         // an error or not.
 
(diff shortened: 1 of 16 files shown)


kriskowal:
sendNotOk should probably carry a nonzero response code as well.

node/MIGRATION.md
```diff
@@ -30,10 +30,10 @@ chan.handler.register('my-endpoint', function (req, res) {
     // req.arg2, req.arg3, req.remoteAddr
 
     // either
-    res.send(new Error('oops'));
+    res.sendNotOk(null, 'oops');
 
     // or
-    res.send(null, 'res1', 'res2');
+    res.sendOk(null, 'res1', 'res2');
```
kriskowal: The above are still taking null arguments.

node/MIGRATION.md
```diff
@@ -42,14 +42,23 @@ chan.handler.register('my-endpoint', function (req, res) {
...
 
 After (simple):
 ```
 chan
     .request(options)
-    .send(a1, a2, a3, cb);
+    .send(a1, a2, a3, function (err, res) {
+        // CallResponse is in `res.arg2`, `res.arg3`
+        // res has an `.ok` field that is tells whether it is
+        // an error or not.
```
kriskowal: Perhaps explicate that error means application error in this case, and that the "err" argument implies an error carried by an Error frame indicating a failure to deliver the request.

node/examples/legacy_example.js
```diff
@@ -0,0 +1,41 @@
+'use strict';
```
jcorbin: blurb this if you must add it

jcorbin: also why did license check not stop you?

node/reqres.js
```diff
@@ -134,19 +134,29 @@ function TChannelOutgoingResponse(id, options, sendFrame) {
...
-        self.sendFrame(self.arg1, res1, errArg);
-    } else {
-        self.sendFrame(self.arg1, res1, res2);
+    self.ok = true;
+
+    self.sendFrame(self.arg1, res1, res2);
+    self.emit('end');
+};
+
+TChannelOutgoingResponse.prototype.sendNotOk = function sendNotOk(res1, res2) {
```
jcorbin: hmm.... it'd be nice if we could pass in application error codes through this... it's poorly / not spec'd, but applications could benefit form defining an error code space >0.

node/examples/legacy_example.js
Raynos: pre commit stuff doesnt work if your package.json is not at root. Bug in pre-commit library.

node/reqres.js
Raynos: Agreed. However that also requires a change to the v2/handler as its hard coded to return 1.

I'll consider a seperate PR for this.

PR #162, opened 2015-03-16
Speed up Travis.

breerly:
Travis is taking a long time. This is for many reasons.

* We are not caching anything installed  by npm & pip.
* Tox runs in serial for every version of Python.
* We run python tests even when we dont edit that directory (same for node).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,25 @@
 ---
-language: python
 
+#
+# use container-based infrastructure
+# @see http://docs.travis-ci.com/user/workers/container-based-infrastructure/
+# @see http://blog.travis-ci.com/2014-12-17-faster-builds-with-container-based-infrastructure/
+#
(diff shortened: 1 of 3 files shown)


kriskowal:
This is good magic. The same magic could probably be performed for the ~/.npm cache.

breerly:
I think our biggest gains, in order, will be:

1. Allowing Tox to run in parallel.
2. Add caching for NPM & pip.
3. Figuring out if we can not run a certain languages test suite if there were no changes.

breerly:
Upgraded concurrent jobs from 0 to 100 in settings:

screen shot 2015-03-16 at 5 31 24 pm

Raynos:
lgtm.

breerly:
This is ready to go - build takes around ~1 min now.

Raynos:
<3

PR #168, opened 2015-03-17
Implement `resp.sendErrorFrame()`

Raynos:
This implements and test the ability to send error frames
as a response to an incoming call request.

This interface is not publically documented in purpose.
I believe it should only be used by autobahn and by
tchannel-as-*.

cc: @jsu1212

### DIFF
--- a/node/reqres.js
+++ b/node/reqres.js
@@ -129,7 +129,7 @@ function TChannelOutgoingResponse(id, options, senders) {
     self.arg1 = options.arg1 || emptyBuffer;
     self.arg2 = options.arg2 || emptyBuffer;
     self.arg3 = options.arg3 || emptyBuffer;
-    self.sendCallResponseFrame = senders.callResponseFrame
+    self.sendCallResponseFrame = senders.callResponseFrame;
     self.sendErrorFrame = senders.errorFrame;
     self.sent = false;
 }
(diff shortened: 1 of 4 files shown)


kriskowal:
:ok:

node/reqres.js
```diff
@@ -112,9 +112,13 @@ TChannelOutgoingRequest.prototype.hookupCallback = function hookupCallback(callb
     return self;
 };
 
-function TChannelOutgoingResponse(id, options, sendFrame) {
+function TChannelOutgoingResponse(
+    id, options, sendCallResponseFrame, sendErrorFrame
```
jcorbin: 3 -> 4 -> many! ... how about:
```
sendFrame :: {
  resFrame: (a1, a2, a3) => void,
  errFrame: (msg) => void
}
```

node/v2/handler.js
```diff
@@ -219,6 +226,23 @@ TChannelV2Handler.prototype.sendResponseFrame = function sendResponseFrame(res,
 };
 /* jshint maxparams:4 */
 
+TChannelV2Handler.prototype.sendErrorFrame = function sendErrorFrame(
+    res, codeString, message
```
jcorbin: should take req as first arg not res

node/v2/handler.js
```diff
@@ -219,6 +226,23 @@ TChannelV2Handler.prototype.sendResponseFrame = function sendResponseFrame(res,
...
+) {
+    var self = this;
+
+    var code = v2.ErrorResponse.Codes[codeString];
+    if (typeof code !== 'number') {
```
jcorbin: `=== undefined`

PR #163, opened 2015-03-17
Node: convert to bufrw

jcorbin:
- converts frame and all its sub pieces
- updates all tests
- drops lib/{read,write}
- reader/writer refactored out into bufrw itself
- -1KLOC

reviewers: @Raynos @kriskowal 
cc: @leizha @abhinav

PR #173, opened 2015-03-17
Fully document all the tchannel client errors

Raynos:
I was planning to update the code to turn a bunch of
these errors into a set of well defined typed errors.

However, as a first pass (its late) I'm just auditing
and documenting all the errors that can happen
as far as an outgoing request is concerned.

We should figure out what combination of `TypedError`
and `WrappedError` we want to use to have a set of
whitelisted errors being returned to the callback
given to `.request()`

Ideally we have

 - 'tchannel.socket-reset.handler-failure'
 - 'tchannel.socket-reset.reader-failure'
 - 'tchannel.error-frame.{frameType}'
 - 'tchannel.socket-reset.net-socket-failure'
 - 'tchannel.socket-reset.socket-closed'
 - 'tchannel.socket-reset.server-closed'
 - 'tchannel.socket-reset.net-server-failure'
 - 'tchannel.client-timeout'

The main classes of failure here are

  - The tchannel connection needs to reset because
    the other side does not speak tchannel properly.
  - The call request got back a tchannel error frame
  - The call request timed out in the client
  - A TCP network error occurred and the id is invalid.

reviewers: @jcorbin @rf

### DIFF
--- a/node/README.md
+++ b/node/README.md
@@ -366,6 +366,86 @@ The callback will either get called with `cb(err)` or with
  - `resp.arg2` will be the `head` response from the server as a buffer
  - `resp.arg3` will be the `body` response from the server as a buffer
 
+### TChannel client Errors. Errors from `.request().send()`
+
+When making an `OutgoingRequest` there are multiple edge cases
+that can go happen. There are multiple operational errors that
+can occur.
+
+#### Writing errors.
+
+When we write `TChannelFrame` objects to the
+socket, we can get exceptions serializing these frames. The
+current implementation will only throw these exceptions and
+thus are considered fatal bugs.
+
+#### Handler errors
+
+When something goes wrong with the frames or the order we 
+receive them in we will get a handler error.
+
+For example, receiving an unknown frame type or receiving multiple
+init requests.
+
+When an error occurs on the handler we reset the connection with
+the error; thus error out all pending outgoing requests.
+
+We also destroy the TCP socket
+
+#### Reading errors
+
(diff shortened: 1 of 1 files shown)


kriskowal:
lgtm

jcorbin:
- let's take out the -v2 tokens from those types if they're meant literally
- fwiw the reader is leaving in #163 replaced by `require('bufrw/stream/chunk_reader')`; by this I mean, it's not even "in" v2 anymore real soon now
- can you please comment on how you see these types mapping to our [error response codes](https://github.com/uber/tchannel/blob/master/protocol.md#code1-1) where appropriate?

Raynos:
Note; this PR changes zero code, just documentation, didn't get around to changing code.

> let's take out the -v2 tokens from those types if they're meant literally

Will do.

> can you please comment on how you see these types mapping to our error response codes where appropriate?

These do not relate to error frames. This is just documenting tchannel client errors ( will update PR title ). The majority of these failures are in the client. However

 `'tchannel.error-frame.{frameType}'` maps cleanly onto the error frames in the protocol. A outgoing client request can get an error based on an incoming error frame.

When a `'tchannel.socket-reset.v2-reader-failure'` or `'tchannel.socket-reset.v2-handler-failure'` failure happens; its because we closed the connection and send an error frame to the other side. Because of the bidirectional nature; if you had any inflight outgoing requests to that connection, their callbacks will be called with such an error.

rf:
lgtm

PR #174, opened 2015-03-17
Fix protocol error code

jcorbin:


jcorbin:
After #163 lands, it'll be trivial to add a slew of test cases to exhaustively cover our implementation of the protocol in unit tests.

breerly:
:shipit:

PR #175, opened 2015-03-17
allocCluster: support specifying listen ports

jcorbin:
This makes it easier to tcap a unit test, by temporarily adding something like `listen: [4040, 4041]` to a test's options.

reviewers: @Raynos @kriskowal


kriskowal:
:ok:

Raynos:
lgtm.

PR #176, opened 2015-03-17
Always treat args as buffers after req/res

jcorbin:
This allows the frame reading/writing layer to be tighter

reviewers: @Raynos @kriskowal


kriskowal:
This is fine, but I have mixed feelings about the approach. For any function, there should be more call sites than definition sites, so I feel like this kind of convention should be handled inside the function instead of at the call sites.

Raynos:
lgtm.

jcorbin:
Realize that this uncomfortable hack will go away when these are "just" streams

PR #178, opened 2015-03-17
bufrw tracing

rf:
`reqres.js` had to require the `emptyTracing` out of the v2 folder which is probably not what you want.


Raynos:
lgtm.

kriskowal:
:ok:

jcorbin:
Didn't need it in reqres anyhow:
- it was just there to pass to a frame constructor
- which is in the v2 layer and can provide the right default
- so I made reqres default to null

Landing so I can deconflict with my streaming work.

PR #179, opened 2015-03-17
[python] out-of-order response handling and tcurl support for multiple requests

blampe:
* Allow `tcurl.py` to issue multiple requests on the same connection (e.g., `-d {1..10000}`).
* Add `profile` option to `tcurl.py`.
* Stop creating new connections for every request in `add_peer`.
* Keep track of pending responses so that we can receive them out of order on that single connection.
* Add a `message_id` option to `frame_and_write` so we can respond with the correct id.

@uber/dsg

### DIFF
--- a/python/tchannel/tcurl.py
+++ b/python/tchannel/tcurl.py
@@ -39,7 +39,7 @@ def multi_tcurl(hostports, headers, bodies, profile=False):
         profiler.disable()
         profiler.create_stats()
         stats = pstats.Stats(profiler)
-        stats.strip_dirs().sort_stats('cumulative').print_stats(15)
+        stats.strip_dirs().sort_stats('tot').print_stats(15)
 
     stop = time.time()
 
--- a/python/tchannel/tcurl.py
+++ b/python/tchannel/tcurl.py
@@ -132,7 +132,8 @@ def parse_args(args=None):
 
     args = parser.parse_args(args)
 
-    # Allow a body/header to specified once and shared across multiple requests.
+    # Allow a body/header to specified once and shared across multiple
+    # requests.
     if args.headers and args.body and len(args.headers) != len(args.body):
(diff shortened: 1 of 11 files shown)


python/tchannel/tcurl.py
```diff
@@ -13,10 +17,46 @@
...
 
+    futures = [
+        tcurl(client, hostport, header, body)
+        for hostport, header, body
+        in getattr(itertools, 'izip', zip)(hostports, headers, bodies)
+    ]
+
+    start = time.time()
+
+    profiler = cProfile.Profile()
```
breerly: Seems like you shouldn't incur the cost of instantiating this is `profile` is `None`.

python/tchannel/tcurl.py
```diff
@@ -13,10 +17,46 @@
...
+
+    if profile:
+        profiler.disable()
+        profiler.create_stats()
+        stats = pstats.Stats(profiler)
+        stats.strip_dirs().sort_stats('tot').print_stats(15)
+
+    stop = time.time()
+
+    log.debug(
```
breerly: Nice.

python/tchannel/tcurl.py
```diff
@@ -81,7 +124,43 @@ def parse_args():
         action="store_true"
     )
 
-    return parser.parse_args()
+    parser.add_argument(
```
breerly: Suggest moving all `parser.add_argument` calls to top of file.

breerly:
Had some nits. Still - :shipit:

python/tchannel/tcurl.py
blampe: Fixed.

python/tchannel/tcurl.py
blampe: Word, done.

PR #180, opened 2015-03-17
add inbound server to the python tchannel

junchaowu:


python/tchannel/inbound_server.py
```diff
@@ -0,0 +1,36 @@
...
+class InboundServer(tornado.tcpserver.TCPServer):
+
+    def handle_stream(self, stream, address):
+        tchannel_connection = TornadoConnection(
+            connection=stream
+        )
+
+        print("Received request from %s:%d" % address)
+
+        print("Waiting for TChannel handshake...")
```
abhinav: You probably want to remove the print statements.

python/tchannel/inbound_server.py
```diff
@@ -0,0 +1,36 @@
...
+        tchannel_connection.await_handshake(headers={
+            'host_port': '%s:%s' % address,
+            'process_name': sys.argv[0],
```
abhinav: Don't hardcode this. Make it an optional argument with this as the default.

python/tchannel/inbound_server.py
```diff
@@ -0,0 +1,36 @@
...
+        print(
+            "Successfully completed handshake with %s" %
+            connection.remote_process_name
+        )
+        # TODO add dispatch handler after handshake complete
+        connection.handle_calls(self.handle_call)
+
+    def handle_call(self, context, connection):
+        # TODO replace this function with request handlers
+        connection.handle_calls(self.handle_call)
```
abhinav: Won't this just loop infinitely?

python/tchannel/tchannel.py
```diff
@@ -10,6 +10,7 @@
 
 from .exceptions import InvalidMessageException
 from .messages import CallRequestMessage
+from tchannel.inbound_server import InboundServer
```
blampe: Use relative imports.

python/tchannel/tcurl.py
```diff
@@ -101,10 +109,9 @@ def main():
 
     host, endpoint = args.host.split('/', 1)
 
-    tornado.ioloop.IOLoop.instance().run_sync(
-        lambda: tcurl(host, endpoint, args.headers, args.body)
-    )
+    tcurl(host, endpoint, args.headers, args.body)
 
+    tornado.ioloop.IOLoop.instance().start()
```
blampe: This should still `run_sync` if we're not listening for inbound requests.

python/tchannel/tcurl.py
```diff
@@ -81,6 +82,13 @@ def parse_args():
         action="store_true"
     )
 
+    parser.add_argument(
+        "-in", "--inport",
```
blampe: Single-dash arguments should one letter. Also let's call this `--listen`.

python/tchannel/tchannel.py
```diff
@@ -81,12 +83,8 @@ def make_out_connection(self, hostport, sock=None):
         raise tornado.gen.Return(connection)
 
     @tornado.gen.coroutine
-    def make_in_connection(self):
-        raise NotImplementedError
-
-    def listen(self):
-        # TODO: listen to inbound handshakes
-        raise NotImplementedError
+    def make_in_connection(self, port):
```
blampe: Just do this inline.

python/tchannel/tchannel.py
```diff
@@ -20,13 +21,14 @@
...
-    def __init__(self, process_name=None):
+    def __init__(self, process_name=None, in_port=None):
         self.peers = {}
         self.process_name = (
             process_name or "%s[%s]" % (sys.argv[0], os.getpid())
         )
 
-        # TODO: self.server_socket = self.make_in_connection()
+        self.inbound_server = InboundServer()
+        self.make_in_connection(in_port)
```
blampe: Can we consolidate the inbound/outbound stuff is some `Connection` class methods, e.g. `Connection.inbound(listen_port)` and `Connection.outbound(host, port)`? The single-threaded case wouldn't implement `inbound`.

PR #182, opened 2015-03-18
Golang tracing

mmihic:
Implements Zipkin span creation and propagation.  All outbound calls are automatically assigned new Spans, spans are propagated from caller to receiver, the current span is available from receiver via the context object.

R: @sectioneight

### DIFF
--- a/golang/tracing.go
+++ b/golang/tracing.go
@@ -23,8 +23,13 @@ package tchannel
 import (
 	"fmt"
 	"math/rand"
+	"time"
 
 	"golang.org/x/net/context"
+)
+
+var (
+	rng = rand.New(rand.NewSource(time.Now().UnixNano()))
 )
 
 // Span represents Zipkin-style span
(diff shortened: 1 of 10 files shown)


golang/tracing.go
```diff
@@ -0,0 +1,76 @@
+package tchannel
+
```
sectioneight: Do we need a license in this file as well?

mmihic: yeah, updating.

golang/tracing.go
```diff
@@ -0,0 +1,76 @@
...
+}
+
+const (
+	tracingFlagEnabled byte = 0x01
+	tracingKey              = "tracing"
+)
+
+// NewRootSpan creates a new top-level Span for a call-graph within the provided context
+func NewRootSpan() *Span {
+	return &Span{traceID: uint64(rand.Int63())}
```
sectioneight: Do we need to seed `rand` before we use it? I think otherwise it generates the same numbers every time:

```go
~ ❯ cat test.go
package main

import (
	"fmt"
	"math/rand"
)

func main() {
	fmt.Println(rand.Int63())
}
~ ❯ go build test.go
~ ❯ ./test
5577006791947779410
~ ❯ ./test
5577006791947779410
~ ❯ ./test
5577006791947779410
~ ❯ go run test.go
5577006791947779410
```

sectioneight:
Looks good other than possible issue with PRNG

golang/tracing.go
mmihic: yeah good catch

PR #185, opened 2015-03-18
Allow extending logger with additional context

mmihic:
Use this context to specify the localPeerInfo and remotePeerInfo on
connections


prashantv:
Merged #967 which allows adding map[string]interface{} fields to entries.

PR #186, opened 2015-03-18
[python] Rework serialization

abhinav:
Also add default constructors for most messages.

CC: @blampe @breerly


python/tchannel/frame.py
```diff
@@ -1,101 +1,62 @@
...
+    size_rw = rw.number(2)  # size:2
 
-    @classmethod
-    def decode(cls, stream, message_length=None):
-        """Decode a sequence of bytes into a frame and message.
+    # ReadWriter for FrameHeaders
+    header_rw = rw.instance(
+        FrameHeader,
+        ('message_type', rw.number(1)),         # type:1
+        ('_', rw.constant(rw.number(1), 0)),    # reserved:1
```
blampe: Can you just define `rw.skip` as '_' or something so this is a little more explicit.

breerly:
Go go go!

blampe:
Go for it!

PR #190, opened 2015-03-19
Jank Tracing agent

rf:
It's janky but it works. The points where we create spans and stuff should be in better places, in constructors, etc. Also had to pass around a ref to the tracer which is ugly. Works for my basic test cases, more test cases needed. 

PRed on the bufrw-tracing branch since it needs that

@Raynos @kriskowal @jcorbin


Raynos:
This is two features in one pull request.

 - setting up the tracing agent to generate tracing data and span ids.
 - using the same tracing agent to report incoming and outgoing messages including host ports and service names.

I would recommend doing the latter in a seperate PR and having the first or be the minimum code to implement "setupNewSpan()" and "span.getTracing()"

This PR is looking good though

rf:
I'm OOO until next Wednesday. I'll probably be able to get a chunk of a few hours this weekend at the hackathon to address review comments so I think it makes more sense to keep this in one PR and do one larger code review than split it up.

@jcorbin: I think the first thing will be to point out the best place to create the spans and add the tracing data. Then I can put that where it belongs. For this I just hacked at it until I found a spot that worked, and passed the `tracer` to whatever needed it. It might be ok to leave it in a dumb place for now, but this is going to impact performance.

node/reqres.js
```diff
@@ -105,12 +124,26 @@ TChannelOutgoingRequest.prototype.hookupCallback = function hookupCallback(callb
...
     function onError(err) {
+        // TODO: better annotations
+        self.span.annotate('cr'); // client recv
+        self.tracer.report(self.span);
         self.removeListener('response', onResponse);
+
+        var prev = self.tracer.getCurrentSpan();
+        self.tracer.setCurrentSpan(self.span);
         callback(err, null);
+        self.tracer.setCurrentSpan(prev);
```
jcorbin: dangerous: assumes callback is synchronous

node/reqres.js
```diff
@@ -105,12 +124,26 @@ TChannelOutgoingRequest.prototype.hookupCallback = function hookupCallback(callb
...
     function onResponse(res) {
+        // TODO: better annotations
+        self.span.annotate('cr');
+        self.tracer.report(self.span);
         self.removeListener('error', onError);
+
+        var prev = self.tracer.getCurrentSpan();
+        self.tracer.setCurrentSpan(self.span);
         callback(null, res);
+        self.tracer.setCurrentSpan(prev);
```
jcorbin: dangerous: assumes callback is synchronous

node/reqres.js
```diff
@@ -76,8 +77,10 @@ function TChannelOutgoingRequest(id, options, sendFrame) {
     self.id = id || 0;
     self.ttl = options.ttl || 0;
     self.tracing = options.tracing || emptyTracing;
+    self.tracer = options.tracer; // tracing agent
```
jcorbin: dangerous: a shared mutable tracer between multiple ongoing req / res objects in the same process (later mutated by `tracer.setFoo` methods)

jcorbin:
Note: TChannelOutgoingRequest.prototype.send and TChannelOutgoingRequest.prototype.hookupCallback are optional paths that may even become deprecated depending on where we go with the API.  At the very least, relying on them means that you won't be tracing non-trivial streaming use cases which don't use .send or .hookupCallback.

node/test/trace/index.js
```diff
@@ -0,0 +1,3 @@
+require('./basic_server');
```
jcorbin: missing copyright blurb

node/test/trace/basic_server_fixture.js
```diff
@@ -0,0 +1,123 @@
+var validators = require('../lib/simple-validators');
```
jcorbin: missing copyright blurb

node/test/lib/simple-validators.js
```diff
@@ -0,0 +1,66 @@
+module.exports.timestamp = function (assert, value) {
```
jcorbin: missing copyright blurb

node/test/trace/server_2_requests.js
```diff
@@ -0,0 +1,143 @@
+
```
jcorbin: this weird extra newline may screw with the copyright checker, not sure if...

node/test/trace/server_2_requests_fixture.js
```diff
@@ -0,0 +1,180 @@
+var validators = require('../lib/simple-validators');
```
jcorbin: missing copyright blurb

node/trace/agent.js
```diff
@@ -0,0 +1,114 @@
+if (!process.addAsyncListener) {
```
jcorbin: missing copyright blurb

node/trace/agent.js
```diff
@@ -0,0 +1,114 @@
...
+    });
+
+    process.addAsyncListener(self.asyncListener);
+
+    // TODO: options validation
+
+    self.logger = options.logger || NullLogtron();
+    self.reporter = options.reporter || null;
+
+    self.logger.info('tracing enabled');
```
jcorbin: should get removed before production, maybe keep it as a .debug if you prefer.

node/trace/agent.js
```diff
@@ -0,0 +1,114 @@
...
+Agent.prototype.destroy = function destroy() {
+    var self = this;
+
+    process.removeAsyncListener(self.asyncListener);
+};
+
+Agent.prototype.setCurrentSpan = function setCurrentTracing(tracing) {
+    var self = this;
+
+    self.currentSpan = tracing;
```
jcorbin: if you returned the old one here, you could simplify consumers leaving less room for error.

node/trace/agent.js
```diff
@@ -0,0 +1,114 @@
...
+    return span;
+};
+
+Agent.prototype.destroy = function destroy() {
+    var self = this;
+
+    process.removeAsyncListener(self.asyncListener);
+};
+
+Agent.prototype.setCurrentSpan = function setCurrentTracing(tracing) {
```
jcorbin: reconcile functions names please, "...Span" vs "...Tracing"

node/trace/agent.js
```diff
@@ -0,0 +1,114 @@
...
+    return self.currentSpan;
+};
+
+Agent.prototype.report = function report(span) {
+    var self = this;
+
+    // TODO: actual reporting
+
+    span.ready(function spanReady() {
+        self.logger.info('got span: ' + span.toString());
```
jcorbin: how about we provide this as default `.reporter`, and then eliminate the possibility of it being optional?

node/trace/span.js
```diff
@@ -0,0 +1,203 @@
+var Ready = require('ready-signal');
```
jcorbin: missing copyright blurb

node/trace/span.js
```diff
@@ -0,0 +1,203 @@
...
+module.exports = Span;
+module.exports.Endpoint = Endpoint;
+module.exports.Annotation = Annotation;
+module.exports.BinaryAnnotation = BinaryAnnotation;
+
+function mathRng(cb) {
+    cb(null, Math.random() * 100000000000000000);
+}
+
+function cryptoRng(cb) {
```
jcorbin: Do we actually need cryptographically secure randomness here? It doesn't seem to me like we should...

If you just need 8 bytes of random, then you could use a simple Linear Congruential Generator, which is fast to compute and not async... I'm adding one to generate test data for streaming: https://github.com/uber/tchannel/commit/7b76ccb67ded605dad2384323ccd6e586d7403f2 , but you don't even need all the overhead of being a stream... just a function that returns 8 bytes will do, which can probably be done by choosing some decent 32-bit constants from http://en.wikipedia.org/wiki/Linear_congruential_generator and then turning the crank twice (fits inside JS's double-math better).

node/trace/span.js
```diff
@@ -0,0 +1,203 @@
...
+    if (!(this instanceof Span)) {
+        return new Span(options);
+    }
+    var self = this;
+
+    // TODO: options validation
+
+    self.logger = options.logger || NullLogtron();
+
+    self._idReady = Ready();
```
jcorbin: building on all the above then, if we use inlined sync PRNGs, there's no need for all this ready signalling

node/v2/handler.js
```diff
@@ -278,10 +281,13 @@ TChannelV2Handler.prototype.buildOutgoingResponse = function buildOutgoingRespon
 };
 
 TChannelV2Handler.prototype.buildIncomingRequest = function buildIncomingRequest(reqFrame) {
+    var self = this;
+
     var req = TChannelIncomingRequest(reqFrame.id, {
+        tracer: self.tracer,
         id: reqFrame.id,
         ttl: reqFrame.ttl,
-        tracing: reqFrame.tracing,
```
jcorbin: this was a bug fixed in 6ad13e4 on master

node/reqres.js
rf: `async-listener` takes care of setting and restoring the context during any async operations. Note that the tests use `setTimeout` and `process.nextTick` to test this.

node/endpoint-handler.js
```diff
@@ -43,12 +43,13 @@ var NoSuchEndpointError = TypedError({
...
 });
 
-function TChannelEndpointHandler(serviceName) {
+function TChannelEndpointHandler(serviceName, hostPort) {
     if (!(this instanceof TChannelEndpointHandler)) {
         return new TChannelEndpointHandler(serviceName);
     }
     var self = this;
     self.serviceName = serviceName;
+    self.hostPort = hostPort;
```
jcorbin: after reading through, I'm still not sure why we're adding this?

jcorbin:
It seems to me like we're going to great lengths to allow consumers of tchannel to not have to pass span objects around greatly.

However the cost we seem to be incurring for that seems to be a shared mutable  "tracer" object.

Unless I've missed something about how this tracer works, which is easily possible due to its use of shared mutable state, we still haven't solved the case of a downstream request on a wholly different TChannel.

How about we pivot this feature to be just some reporting facility (which arguably should be wholly external to TChannel, supported only in TChannel(Connection) by an observer interface of some form (maybe an event emit of some sort?).

In fact, I see all of your span work as pretty close to landable if we disentangle it from reporting agent(s) and drop the use of the async crypto rng.

jcorbin:
also please note:
- `npm run check-licence`
- `npm run add-licence`

node/index.js
```diff
@@ -127,6 +129,18 @@ function TChannel(options) {
 
             self.logger.info(self.hostPort + ' listening');
             self.emit('listening');
+
+            // TODO: make this in constructor and not listening event
+            self.tracer = new TracingAgent({
```
jcorbin: At least initialize self.tracer to null during construction, the definition of a new field eventually will cause a change of type for self currently.

Even better if agent could support having host/port optionally filled in later.

Note that ephemeral clients, like TCurl, don't necessarily listen on a socket at all.

rf:
`test/trace/server_2_requests.js` is a test that has 2 listening tchannel instances; the client requests from one which then makes two requests to the other. If you look at the callback, you'll see it's not synchronous, that is, the outgoing requests occur in callbacks. This is because `async-listener` takes care of the state propagation for us: https://github.com/othiym23/async-listener

rf:
Have to close and reopen because I rebased on master

PR #192, opened 2015-03-19
fix buildIncomingRequest()

Raynos:
We were not copying fields over properly.

This means I could not read `req.service`.

cc: @kriskowal @jcorbin

### DIFF
--- a/node/v2/handler.js
+++ b/node/v2/handler.js
@@ -279,11 +279,10 @@ TChannelV2Handler.prototype.buildOutgoingResponse = function buildOutgoingRespon
 
 TChannelV2Handler.prototype.buildIncomingRequest = function buildIncomingRequest(reqFrame) {
     var req = TChannelIncomingRequest(reqFrame.id, {
-        id: reqFrame.id,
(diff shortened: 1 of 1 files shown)


node/v2/handler.js
```diff
@@ -279,11 +279,10 @@ TChannelV2Handler.prototype.buildOutgoingResponse = function buildOutgoingRespon
...
-        id: reqFrame.id,
-        ttl: reqFrame.ttl,
-        tracing: reqFrame.tracing,
-        service: reqFrame.service,
-        headers: reqFrame.headers,
+        ttl: reqFrame.body.ttl,
+        tracing: reqFrame.body.tracing,
+        service: reqFrame.body.service,
+        headers: reqFrame.body.headers,
         checksumType: reqFrame.body.csum.type,
```
kriskowal: I don’t know whether csum.val is interesting. Let’s try to approach using the same terms, e.g., csum or checksum but not both throughout.

jcorbin:
:+1:

kriskowal:
:ok:

PR #193, opened 2015-03-19
Test work

jcorbin:
- fixed bug in testExpectations
- explicated the send test cases, streaming test will have much more similar structure when it hits


jcorbin:
cc @kriskowal @Raynos

PR #195, opened 2015-03-19
add inbound server in the tchannel python client using tornado

junchaowu:
All functions are not fully supported yet. Currently only support basic http request and response

### DIFF
--- a/python/tchannel/tcurl.py
+++ b/python/tchannel/tcurl.py
@@ -11,7 +11,8 @@ import sys
 import time
 
 import tornado.ioloop
-
+import tornado.web
+from .tornado.http_request import HttpRequest
 from .tornado import TChannel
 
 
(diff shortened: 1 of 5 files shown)


python/tchannel/tcurl.py
```diff
@@ -115,17 +123,33 @@ def parse_args(args=None):
...
     bodies,
+    in_port=4040,
     profile=False,
     rps=None,
     quiet=False,
 ):
-    client = TChannel()
-
+    app = make_app()
+    client = TChannel(app=app)
```
blampe: Eventually this should become a more general `handler` interface.

python/tchannel/tornado/connection.py
```diff
@@ -24,6 +25,7 @@ def __init__(self, connection):
         self.remote_host = None
         self.remote_process_name = None
         self.requested_version = None
+        self.response = CallResponseMessage()
```
blampe: Eventually this will need to use `awaiting_responses`.

blampe:
go for it

PR #198, opened 2015-03-19
[python] consolidate common outgoing connection stuff

blampe:


breerly:
Sweet. Lets get those tests passing then :shipit:

python/tchannel/tornado/connection.py
```diff
@@ -1,18 +1,25 @@
...
 from ..messages.common import PROTOCOL_VERSION
 
-from tornado import gen
-from tornado.iostream import StreamClosedError
-from tchannel.messages import CallResponseMessage
+
+log = logging.getLogger('tchannel')
 
 
 class TornadoConnection(object):
```
junchaowu: shall we make two subclass of Tornadoconnection.  One is for outgoingconnection and the other one is for inboundconnection. both can implement their own logic and share the common part

blampe: I don't think we need that if we have a good `Handler` interface. `TChannel`, being the `Connection` manager might have 2 buckets for inbound/outbound, but a `Connection` itself just needs to listen to the wire and take action when it sees messages. That would happen via these `Handlers`, so inbound/outbound would make more sense there. None of this API stuff is finalized though so whatever works right now ;)

PR #199, opened 2015-03-20
Fix Tornado perf regression

abhinav:
`gen.coroutine` has more overhead than first believed.


blampe:
sheep it

breerly:
This is dope

PR #204, opened 2015-03-20
[python] make timeouts only kill the request not the connection

blampe:
cc @breerly


breerly:
Looks good. Let's ship the darn thing :+1:

PR #208, opened 2015-03-23
Add search-based test modules

jcorbin:
- TestSearch itself is useful for exhaustive searches
- TestIsolateSearch is useful for exploration / isolation, my first use coming soon is exponential exploration with bisect isolation

### DIFF
--- a/node/test/lib/test_isolate_search.js
+++ b/node/test/lib/test_isolate_search.js
@@ -1,0 +1,267 @@
+// Copyright (c) 2015 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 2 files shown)


jcorbin:
cc @kriskowal @Raynos would love to get feedback on these, but you'll probably need to see the coming use case examples to do them justice... anyhow going to land so I can put out two more PRs that add new passing tests based on these modules.

Raynos:
Just scanned it. Would need to see use cases and interface to understand.

jcorbin:
The primary use case (the bisection test I set out to make at first): https://github.com/uber/tchannel/blob/dev.josh/node/test/streaming_bisect.js

The secondary use case (the exhaustive test I knew I needed to write before I even started on bisection): https://github.com/uber/tchannel/blob/dev.josh/node/test/argstream.js

PR #209, opened 2015-03-23
Add argstream module

jcorbin:
First step towards implementing streaming.

This module is the core piece that knows how to split/combine arg chunks from/to arg streams.

Test is an exhaustive integration test, the current search space covered is 649 cases, however it still gets to run fast since `setImmediate` deferral is sufficient "asynchrony" for this test.

reviewers: @Raynos @kriskowal

### DIFF
--- a/node/argstream.js
+++ b/node/argstream.js
@@ -1,0 +1,272 @@
+// Copyright (c) 2015 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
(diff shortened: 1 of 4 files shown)


jcorbin:
Currently the test hangs when run from `test/index.js` ala `npm test` even tho running `node test/argstream.js` directly works fine.

node/argstream.js
```diff
@@ -0,0 +1,276 @@
...
+    self.arg3.on('error', passError);
+    function passError(err) {
+        self.emit('error', err);
+    }
+
+    self.arg2.on('start', function onArg2Start() {
+        if (!self.arg1._writableState.ended) self.arg1.end();
+    });
+    self.arg3.on('start', function onArg3Start() {
+        if (!self.arg2._writableState.ended) self.arg2.end();
```
Raynos: Is this a programmer error ?

It seems like cutting arg2 mid stream when someone writes to arg3 is going to lead to corrupted data somewhere.

node/argstream.js
```diff
@@ -0,0 +1,276 @@
...
+        if (++self._numFinished === 3) {
+            self.finished = true;
+            self.emit('finish');
+        }
+    }
+}
+
+inherits(InArgStream, ArgStream);
+
+InArgStream.prototype.handleFrame = function handleFrame(parts) {
```
Raynos: Is `parts` a triplet of buffers, i.e. `[arg1, arg2, arg3]` ?

kriskowal:
I like how compartmentalized this commit is. I expect that with this factoring we will be able to use the arg stream object in the Autobahn clients without imposing it upon the Autobahn service.

:ok:

kriskowal:
:ok:

PR #211, opened 2015-03-23
add request handler injection to the Tchannel

junchaowu:


python/examples/tchannel_handler_server.py
```diff
@@ -0,0 +1,78 @@
+from __future__ import absolute_import
+
+import argparse
+import logging
+import sys
+
+import tornado.ioloop
+import tornado.web
```
abhinav: You're not using this import

python/examples/tornado_handler_server.py
```diff
@@ -0,0 +1,78 @@
...
+        type=int,
+        help="Port for inbound connections"
+    )
+
+    args = parser.parse_args(args)
+    return args
+
+
+class MainHandler(tornado.web.RequestHandler):
+    def get(self):
```
abhinav: Are you sure you want to do this? `tornado.web` seems like it's coupled with HTTP.

python/tchannel/req_handler.py
```diff
@@ -0,0 +1,67 @@
+class RequestHandler(object):
+
+    def __init__(self):
+        pass
+
+    def handle_request(self, context, conn):
+        raise NotImplementedError
```
abhinav: `NotImplementedError()`.

Without the brackets, you're trying to throw the class itself. Not sure if that's a throwable.

python/tchannel/req_handler.py
```diff
@@ -0,0 +1,67 @@
...
+        self.endpoints = {}
+
+    def handle_request(self, context, conn):
+        """Handle incoming request
+
+        :param context: incoming message context
+        :param conn: incoming connection
+        """
+        request = TChannelRequest(context, conn)
+        endpoint = self._find_endpoint(request.method)
```
abhinav: `request.method`? Won't that be `GET`/`POST`?

python/tchannel/req_handler.py
```diff
@@ -0,0 +1,67 @@
...
+            self.register_handler(rule, handler, **opts)
+            return handler
+
+        return decorator
+
+    def register_handler(self, rule, handler, **opts):
+        self.endpoints[rule] = {
+            "handler": handler,
+            "opts": opts
+        }
```
abhinav: Use a `namedtuple` to represent endpoints instead of a dict like this.

python/tchannel/req_handler.py
junchaowu: not sure about this one. It depends on how we define the request class

abhinav:
Overall I'm not a fan of this "Translate-from-HTTP" approach. Do we have to use the HTTP-specific handlers and request objects?

junchaowu:
translate-from-http is an approach to be compatible with existing tornado services

abhinav:
Resolved in person. Approach makes sense.

python/tchannel/req_handler.py
```diff
@@ -0,0 +1,69 @@
...
+EndPoint = collections.namedtuple('EndPoint', ['handler', 'opts'])
+
+
+class RequestHandler(object):
+
+    def __init__(self):
+        pass
+
+    def handle_request(self, context, conn):
+        raise NotImplementedError()
```
abhinav: Document this class. Especially things like,

- How implementations are expected to behave.
- What the arguments for `handle_request` are.

python/tchannel/req_handler.py
```diff
@@ -0,0 +1,69 @@
+import collections
+
+EndPoint = collections.namedtuple('EndPoint', ['handler', 'opts'])
+
+
+class RequestHandler(object):
+
+    def __init__(self):
+        pass
```
abhinav: If you're not doing anything in `__init__`, you can remove it.

python/tchannel/req_handler.py
```diff
@@ -0,0 +1,69 @@
...
+
+    def handle_request(self, context, conn):
+        raise NotImplementedError()
+
+
+class TChannelRequestHandler(RequestHandler):
+    def __init__(self):
+        self.endpoints = {}
```
abhinav: Add `super(.., self).__init__()`

python/tchannel/req_handler.py
```diff
@@ -0,0 +1,69 @@
...
+
+        :param context: incoming message context
+        :param conn: incoming connection
+        """
+        request = TChannelRequest(context, conn)
+        endpoint = self._find_endpoint(request.method)
+        if endpoint is not None:
+            response = TChannelResponse(conn)
+            endpoint.handler(request, response, endpoint.opts)
+            response.finish()
```
abhinav: How should the system behave when a matching endpoint is not found?

python/examples/tchannel_handler_server.py
```diff
@@ -0,0 +1,79 @@
...
+from tchannel.tornado import TChannel
```
blampe: Feel free to replace the existing tornado example server with this stuff, it's nice.

python/tchannel/req_handler.py
```diff
@@ -0,0 +1,69 @@
...
+        self.endpoints[rule] = EndPoint(handler=handler, opts=opts)
+
+    def _find_endpoint(self, rule):
+        return self.endpoints.get(rule, None)
+
+
+class TChannelRequest(object):
+    """ TChannel Request Wrapper """
+
+    def __init__(self, context, conn):
```
abhinav: Add `__slots__` listing

python/tchannel/req_handler.py
```diff
@@ -0,0 +1,69 @@
...
+
+    def register_handler(self, rule, handler, **opts):
+        self.endpoints[rule] = EndPoint(handler=handler, opts=opts)
+
+    def _find_endpoint(self, rule):
+        return self.endpoints.get(rule, None)
+
+
+class TChannelRequest(object):
+    """ TChannel Request Wrapper """
```
abhinav: Nit: Generally, don't leave an extra space between the quotes and the description.

python/examples/tchannel_handler_server.py
```diff
@@ -0,0 +1,79 @@
...
+    logging.basicConfig(
+        format="%(name)s[%(process)s] %(levelname)s: %(message)s",
+        stream=sys.stdout,
+        level=logging.INFO,
+    )
+
+    if args.verbose:
+        log.setLevel(logging.DEBUG)
+
+    client = TChannel()
```
blampe: `client` is confusing, `tchannel = TChannel()`

python/tchannel/tornado/connection.py
```diff
@@ -226,6 +224,10 @@ def set_close_callback(self, callback):
 
     def finish(self):
         """ write response """
-        self.response.arg_1 = "from inbound"
+        self.response.arg_1 = "response from inbound server"
```
abhinav: I don't understand. Why is this hard-coded here?

python/tchannel/tornado/tchannel.py
```diff
@@ -46,28 +44,22 @@ def get_peer(self, hostport):
...
-        self.endpoints[rule] = {
-            "handler": handler,
-            "opts": opts
-        }
+class TChannelServerOperation(object):
+
+    def __init__(self, port, tchannel, handler, options):
+        self.inbound_server = InboundServer(handler, options)
+        self.port = port
+        self.tchannel = weakref.ref(tchannel)
```
blampe: You don't need `tchannel` right?

python/examples/tchannel_handler_server.py
```diff
@@ -0,0 +1,79 @@
+from __future__ import absolute_import
```
blampe: `tchannel_server.py`

python/examples/tchannel_handler_server.py
blampe: nvm just realized this is the tchannel example :P

python/examples/tornado_handler_server.py
```diff
@@ -0,0 +1,73 @@
+from __future__ import absolute_import
```
blampe: Replace `tornado_server.py` with *this* one.

python/tchannel/req_handler.py
```diff
@@ -0,0 +1,69 @@
+import collections
```
blampe: `handler.py`

python/tchannel/tornado/connection.py
```diff
@@ -226,6 +224,10 @@ def set_close_callback(self, callback):
...
         self.response.arg_2 = "inbound"
-        return self.frame_and_write(self.response)
+        self.frame_and_write(self.response)
+
+        # reset response message
+        self.response = CallResponseMessage()
```
abhinav: Why is all this being tracked on the same instance? Aren't we expecting the same TornadoConnection to handle multiple possibly out-of-order requests?

python/tchannel/tornado/tornado_req_handler.py
```diff
@@ -0,0 +1,92 @@
+import json
```
blampe: `tornado_handler.py`

python/tchannel/tornado/inbound_server.py
```diff
@@ -12,77 +9,25 @@
...
-
-
 class InboundServer(tornado.tcpserver.TCPServer):
-    def __init__(self, request_callback=None):
-        self.request_callback = request_callback
+    def __init__(self, req_handler, options=None):
+        assert req_handler is not None
+
+        self.req_handler = req_handler
+        self.options = options
```
abhinav: You don't appear to be using `options` anywhere. Remove it. We can add a `**options` for options later if necessary.

Also, my guess is that options was going to be a dict? That's not the standard way of doing this in Python. `**options` is preferred intsead.

python/tchannel/tornado/inbound_server.py
```diff
@@ -12,77 +9,25 @@
...
+        self.peers = {}
         super(InboundServer, self).__init__()
```
abhinav: Prefer doing `super` init before setting attributes

python/tchannel/tornado/inbound_server.py
```diff
@@ -12,77 +9,25 @@
...
 
     def build_stream(self):
```
abhinav: Where is this used? And why is it tracking an outgoing connection as an instance variable? Won't it have multiple outgoing connections?

python/tchannel/tornado/inbound_server.py
```diff
@@ -12,77 +9,25 @@
...
         self.sock.connect(self.server_address)
 
-    def start_serving(self, request_conn):
-        return _ServerRequestAdapter(self, request_conn)
-
     def handle_stream(self, stream, address):
         tchannel_connection = TornadoConnection(
             connection=stream
         )
+        self.peers[tchannel_connection] = self.req_handler
```
abhinav: `TornadoConnection` is not hashable. It contains a `socket` object in it which is not hashable either. Don't use it as a key in a dictionary.

python/tchannel/tornado/tchannel.py
```diff
@@ -46,28 +44,22 @@ def get_peer(self, hostport):
 
         raise tornado.gen.Return(peer)
 
-    @tornado.gen.coroutine
-    def make_in_connection(self, port):
-        self.inbound_server.listen(port)
-
     def request(self, hostport):
-        return TChannelClientOperation(hostport, self)
+        return TChannelClientOperation(hostport, self,)
```
abhinav: Unnecessary trailing comma. Generally they're added if the call is spread across multiple lines (with one parameter per line)

python/tchannel/tornado/tornado_req_handler.py
```diff
@@ -0,0 +1,92 @@
...
+    def handle_request(self, context, conn):
+        """Handle incoming request
+
+        :param context: incoming message context
+        :param conn: incoming connection
+        """
+        request_delegate = self.start_serving(conn)
+        message = context.message
+        # process http message
+        if message.headers["as"] == "http":
```
abhinav: What's the behavior when `as` is not `http`? You need to have some sort of feedback instead of silently ignoring the error.

python/tchannel/tornado/tornado_req_handler.py
```diff
@@ -0,0 +1,92 @@
...
+            method = "GET"
+            if (hasattr(message, "arg_3") and
```
abhinav: Assert that the message is a `CallRequestMessage` intsead?

python/tchannel/tornado/tornado_req_handler.py
```diff
@@ -0,0 +1,92 @@
...
+                    message.arg_3 is not None and
+                    message.arg_3 != ""):
```
abhinav: If you assert that the message is a `CallRequestMessage` then the `hasattr` check isn't needed and the other two are consolidated into `if message.arg_3:`

python/tchannel/tornado/tornado_req_handler.py
```diff
@@ -0,0 +1,92 @@
...
+                method = "POST"
+
+            start_line = RequestStartLine(method, message.arg_1, 'HTTP/1.1')
```
abhinav: The protocol doc says that `arg_1` is "method + URI". So maybe you don't need to guess the method based on the presence of `arg_3`

python/tchannel/tornado/tornado_req_handler.py
```diff
@@ -0,0 +1,92 @@
...
+            try:
+                headers = json.loads(message.arg_2)
+            except:
+                headers = {}
```
abhinav: In case of `http`, `arg_2` contains headers as a JSON array, not a dict. Make this an empty list.

python/tchannel/tornado/tornado_req_handler.py
```diff
@@ -0,0 +1,92 @@
...
+            method = "GET"
+            if (hasattr(message, "arg_3") and
+                    message.arg_3 is not None and
+                    message.arg_3 != ""):
+                method = "POST"
+
+            start_line = RequestStartLine(method, message.arg_1, 'HTTP/1.1')
+            try:
+                headers = json.loads(message.arg_2)
+            except:
```
abhinav: I'm not sure you should be doing a catch-all like this. If `arg_2` is set, it must always be parsed, and if it fails to parse, the request was invalid. Throw an error. Avoid silent failure.

python/tchannel/tornado/tornado_req_handler.py
```diff
@@ -0,0 +1,92 @@
...
+                headers = {}
+
+            body = message.arg_3 if hasattr(message, "arg_3") else ""
+            request_delegate.headers_received(start_line, headers)
+            request_delegate.data_received(body)
+            request_delegate.finish()
+
+
+class _ServerRequestAdapter():
```
abhinav: inherit `object`

python/tchannel/tornado/tornado_req_handler.py
```diff
@@ -0,0 +1,92 @@
...
+    """Adapts the `TChannelMessageDelegate` interface to the interface expected
+    by our clients.
+    """
+    def __init__(self, server, request_conn, server_conn=None):
```
abhinav: Document the arguments

python/tchannel/tornado/tornado_req_handler.py
```diff
@@ -0,0 +1,92 @@
...
+                headers = {}
+
+            body = message.arg_3 if hasattr(message, "arg_3") else ""
+            request_delegate.headers_received(start_line, headers)
+            request_delegate.data_received(body)
+            request_delegate.finish()
+
+
+class _ServerRequestAdapter():
+    """Adapts the `TChannelMessageDelegate` interface to the interface expected
```
abhinav: Explain

python/tchannel/tornado/tornado_req_handler.py
```diff
@@ -0,0 +1,92 @@
...
+    by our clients.
+    """
+    def __init__(self, server, request_conn, server_conn=None):
+        self.server = server
```
abhinav: Considering that you access the `server` only when you need to access the `request_callback`, maybe you should accept a reference to a callable representing the `request_callback` instead of the server?

python/tchannel/tornado/tornado_req_handler.py
```diff
@@ -0,0 +1,92 @@
...
+            self._chunks = []
+
+    def headers_received(self, start_line, headers):
+        # TODO implement xheaders
+        if self.delegate is None:
+            self.request = httputil.HTTPServerRequest(
+                connection=self.connection, start_line=start_line,
+                headers=headers)
+        else:
+            return self.delegate.headers_received(start_line, headers)
```
abhinav: This if-delegate-else behavior suggests this class should be split into two.

python/examples/tchannel_handler_server.py
junchaowu: done

python/examples/tchannel_handler_server.py
junchaowu: done

python/examples/tchannel_handler_server.py
junchaowu: done

python/examples/tornado_handler_server.py
junchaowu: done

python/tchannel/req_handler.py
junchaowu: done

python/tchannel/req_handler.py
junchaowu: add todo here

python/tchannel/req_handler.py
junchaowu: done

python/tchannel/tornado/connection.py
junchaowu: removed

python/tchannel/tornado/connection.py
junchaowu: when server gets a connection, it will generate a new tornadoconnection object. that's why I have a dict match in the inbound_server

python/tchannel/tornado/inbound_server.py
junchaowu: removed

python/tchannel/tornado/inbound_server.py
junchaowu: this is a function will be called from tornado iostream. when it accepts a new connection, it will call this method.
It also gives us the chance to save a new tornado connection instance

python/tchannel/tornado/tchannel.py
junchaowu: removed

python/tchannel/tornado/tchannel.py
junchaowu: I removed it for now. Highly doubt we may need in the future

python/tchannel/tornado/tornado_req_handler.py
junchaowu: done

python/tchannel/tornado/tornado_req_handler.py
junchaowu: add todo commect

python/tchannel/tornado/tornado_req_handler.py
junchaowu: I think add a try catch is okay here to avoid unexpected error

python/tchannel/tornado/tornado_req_handler.py
junchaowu: done

python/tchannel/tornado/tornado_req_handler.py
abhinav: I disagree. We shouldn't fail silently on a malformed request.

python/tchannel/tornado/tornado_req_handler.py
junchaowu: this is a class I move from tornado framework to here. I fake some of the implementation of some of the methods inside this class to work for our tchannel. I won't change any interface or structure of this class. I prefer to keep it same as tornado's.

python/tchannel/tornado/tornado_req_handler.py
junchaowu: explained above

python/tchannel/tornado/tornado_req_handler.py
junchaowu: explained above

python/tchannel/tornado/tornado_req_handler.py
junchaowu: I will put todo here. when we have formalize the http message, we change it

python/tchannel/tornado/inbound_server.py
blampe: I don't think you need this.

python/examples/tchannel_server.py
```diff
@@ -0,0 +1,60 @@
+from __future__ import absolute_import
+
+import argparse
+import sys
+
+import tornado.ioloop
+import tornado.web
```
abhinav: Unused import

python/tchannel/handler.py
```diff
@@ -0,0 +1,89 @@
+import collections
+
+EndPoint = collections.namedtuple('EndPoint', ['handler', 'opts'])
```
abhinav: Nit: Endpoint, not EndPoint

python/tchannel/handler.py
```diff
@@ -0,0 +1,89 @@
...
+
+    def route(self, rule, **opts):
+        def decorator(handler):
+            self.register_handler(rule, handler, **opts)
+            return handler
+
+        return decorator
+
+    def register_handler(self, rule, handler, **opts):
+        self.endpoints[rule] = EndPoint(handler=handler, opts=opts)
```
abhinav: Unless you're using opts somewhere, don't include it in the `Endpoint`. If and when you add extra options on the `Endpoint`, retrieve them from `opts` and set concrete attributes on the `Endpoint` instead of a formless dict.

python/tchannel/handler.py
```diff
@@ -0,0 +1,89 @@
...
+
+        :param context: context contains received CallRequestMessage
+        :param conn: An incoming TornadoConnection
+        """
+        request = TChannelRequest(context, conn)
+        endpoint = self._find_endpoint(request.method)
+        if endpoint is not None:
+            response = TChannelResponse(conn)
+            endpoint.handler(request, response, endpoint.opts)
+            response.finish()
```
abhinav: This should probably use a `try-finally` block. Otherwise an uncaught exception in the handler will result in us not closing the connection.

python/tchannel/handler.py
```diff
@@ -0,0 +1,89 @@
...
+        self.method = self.message.arg_1
+        self.connection = conn
+
+        # TODO fill up more attributes
+
+
+class TChannelResponse(object):
+    """TChannel Response Wrapper"""
+
+    __slots__ = ('_connection')
```
abhinav: Python needs a trailing comma for a tuple with just one item. Othewise, `(foo)` is the same as `foo`.

python/tchannel/tornado/connection.py
```diff
@@ -219,7 +218,7 @@ def handle_call_response(context, connection):
         raise gen.Return(connection)
 
     def write_headers(self, start_line, headers, chunk=None, callback=None):
-        self.response.headers = headers or {'currently': 'broken'}
+        self.response.arg_2 = json.dumps(headers) or ''
```
abhinav: No, you want to do

    json.dumps(headers) if headers else ''

python/tchannel/tornado/connection.py
```diff
@@ -230,7 +229,8 @@ def set_close_callback(self, callback):
...
     def finish(self):
-        """ write response """
-        self.response.arg_1 = "from inbound"
-        self.response.arg_2 = "inbound"
-        return self.frame_and_write(self.response)
+        """write response"""
+        self.frame_and_write(self.response)
+        # reset response message
+        self.response = CallResponseMessage()
+        return
```
abhinav: Don't need this return

python/tchannel/handler.py
junchaowu: done

python/tchannel/tornado/connection.py
junchaowu: done

python/tchannel/tornado/tchannel.py
```diff
@@ -46,28 +44,21 @@ def get_peer(self, hostport):
...
-        self.inbound_server.listen(port)
-
     def request(self, hostport):
         return TChannelClientOperation(hostport, self)
 
-    def route(self, rule, **opts):
-        def decorator(handler):
-            self.register_handler(rule, handler, **opts)
-            return handler
+    def host(self, port, handler):
```
abhinav: Probably call this `listen`?

python/tchannel/tornado/tornado_handler.py
```diff
@@ -0,0 +1,100 @@
...
+        """dispatch incoming request to particular endpoint
+
+        :param context: context contains received CallRequestMessage
+        :param conn: An incoming TornadoConnection
+        """
+        request_delegate = self.start_serving(conn)
+        message = context.message
+        # process http message
+        """TODO need a better way to figure out
+        message's method type
```
abhinav: Nit: Docstrings probably not the best way to specify a TODO. Use a `#` comment.

python/examples/tchannel_server.py
junchaowu: those modules are used in the file

python/tchannel/handler.py
junchaowu: done

python/tchannel/handler.py
junchaowu: tornado uses this way. I refers from tornado

python/examples/tchannel_server.py
abhinav: `tornado.web` is not

python/tchannel/handler.py
junchaowu: The error case has not covered in this code review. It is known that we will implement it later

python/tchannel/tornado/tornado_handler.py
```diff
@@ -0,0 +1,99 @@
...
+            start_line = RequestStartLine(method, message.arg_1, 'HTTP/1.1')
+            try:
+                headers = json.loads(message.arg_2)
+            except:
+                headers = {}
+            body = message.arg_3 if hasattr(message, "arg_3") else ""
+            request_delegate.headers_received(start_line, headers)
+            request_delegate.data_received(body)
+            request_delegate.finish()
+        # TODO process message in json/thrift format
```
abhinav: Generally when leaving TODOs like this, add a `raise NotImplementedError()` so that this isn't missed.

abhinav:
:ok:

PR #212, opened 2015-03-23
golang: consolidate frame writes...

mmihic:
...to coalesce packets for smaller messages

R: @kriskowal

### DIFF
--- a/golang/connection.go
+++ b/golang/connection.go
@@ -418,22 +418,10 @@ func (c *Connection) withStateRLock(f func() error) error {
 // send the incoming frame to a channel; the init handlers are a notable exception, since we cannot
 // process new frames until the initialization is complete.
 func (c *Connection) readFrames() {
-	fhBuf := typed.NewReadBufferWithSize(FrameHeaderSize)
-
 	for {
-		if _, err := fhBuf.FillFrom(c.conn, FrameHeaderSize); err != nil {
-			c.connectionError(err)
-			return
-		}
-
 		frame := c.framePool.Get()
-		if err := frame.Header.read(fhBuf); err != nil {
-			// TODO(mmihic): Should be a protocol error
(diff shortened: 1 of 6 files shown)


golang/connection.go
```diff
@@ -418,22 +418,10 @@ func (c *Connection) withStateRLock(f func() error) error {
...
 		frame := c.framePool.Get()
-		if err := frame.Header.read(fhBuf); err != nil {
-			// TODO(mmihic): Should be a protocol error
-			c.connectionError(err)
-			return
-		}
-
-		if _, err := c.conn.Read(frame.SizedPayload()); err != nil {
+		if err := frame.ReadFrom(c.conn); err != nil {
+			c.framePool.Release(frame)
```
abhinav: Does the frame not need to be released for the success case when the handlers are called?

mmihic: No, the frame and its buffer is still needed by the upstream handler.

golang/frame.go
```diff
@@ -57,11 +58,60 @@ type FrameHeader struct {
...
+	f.buffer = make([]byte, payloadCapacity+FrameHeaderSize)
+	f.Payload = f.buffer[FrameHeaderSize:]
+	f.headerBuffer = f.buffer[:FrameHeaderSize]
+	return f
+}
+
+// ReadFrom reads the frame from the given io.Reader
+func (f *Frame) ReadFrom(r io.Reader) error {
+	var rbuf typed.ReadBuffer
+	rbuf.Wrap(f.headerBuffer)
```
abhinav: What's the reason for not using `NewReadBuffer` here?

mmihic: unnecessary heap allocation in the middle of a core loop

abhinav:
:shipit:

kriskowal:
Thanks!

PR #213, opened 2015-03-23
Add v2/args

jcorbin:
Arguably this should be something like a `bufrw.Star(bufrw.buf2)`, but bufrw does not yet implement Star and I'm GTD focused wrt streaming.


jcorbin:
cc @kriskowal @Raynos

kriskowal:
:ok:

PR #214, opened 2015-03-23
Pivot reqres sending API to prepare for streaming

jcorbin:
- easier to add cont frame senders
- able to use setOk to set state bits without sending an initial frame


jcorbin:
cc @Raynos @kriskowal

Raynos:
lgtm

PR #215, opened 2015-03-23
v2/call: convert to ArgsRW

jcorbin:
Converts only within the call req/res bodies.

This will probably be a breaking change for things such as tcap.

reviewers: @kriskowal @Raynos

### DIFF
--- a/node/v2/call.js
+++ b/node/v2/call.js
@@ -21,6 +21,7 @@
 'use strict';
 
 var bufrw = require('bufrw');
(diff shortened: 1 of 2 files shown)


kriskowal:
Ah, yeah. Alright. This will make it a bit more complicated to provide deep argument inspection. Will have to introduce another layer that accumulates each arg. Thanks, lgtm.

Raynos:
lgtm.

PR #216, opened 2015-03-23
Drive checksumming from v2/handler

jcorbin:
Sets up so that we can carry prior checksum state between continuation frames.

reviewers: @Raynos @kriskowal

### DIFF
--- a/node/v2/call.js
+++ b/node/v2/call.js
@@ -83,6 +83,16 @@ CallRequest.RW = bufrw.Struct(CallRequest, [
     {call: {readFrom: readGuard}}
 ]);
 
+CallRequest.prototype.updateChecksum = function updateChecksum() {
+    var self = this;
+    return self.csum.update(self.args);
+};
+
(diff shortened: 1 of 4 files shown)


node/v2/call.js
```diff
@@ -126,13 +136,22 @@ CallResponse.RW = bufrw.Struct(CallResponse, [
...
+    return self.csum.verify(self.args);
+};
+
 function prepareWrite(body, buffer, offset) {
     if (body.flags & CallRequest.Flags.Fragment) {
         return WriteResult.error(
             new Error('streaming call not implemented'),
             offset);
     }
-    body.csum.update(body.args);
```
Raynos: where did the `updateChecksum` calls go ?

Raynos: We moved the `verifyChecksum` calls into v2/handler

jcorbin: great catch, fixed

Raynos:
lgtm.

PR #217, opened 2015-03-23
error handling for closed connections

blampe:
* cancel in-flight requests if our connection dies
* catch socket errors in `await`

We've still got a problem with callbacks in `await` where we hit an infinite recursion limit, looking into that.

### DIFF
--- a/python/tchannel/socket.py
+++ b/python/tchannel/socket.py
@@ -21,7 +21,7 @@ class _SocketIOAdapter(object):
         self._connection = connection
 
     def read(self, size):
-        result = self._recv(size)
+        result = self._connection.recv(size)
 
         remaining = size - len(result)
 
@@ -29,7 +29,7 @@ class _SocketIOAdapter(object):
         if remaining > 0:
             chunks = [result]
             while remaining > 0:
-                s = self._recv(remaining)
+                s = self._connection.recv(remaining)
 
                 if not s:  # end of stream reached
                     break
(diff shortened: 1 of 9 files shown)


python/tchannel/socket.py
```diff
@@ -85,7 +99,12 @@ def outgoing(cls, hostport):
         conn.await_handshake_reply()
         return conn
 
+    @classmethod
+    def incoming(cls, hostport, handler):
+        raise NotImplementedError
```
abhinav: `()`

python/tchannel/socket.py
```diff
@@ -62,6 +71,11 @@ def await(self):
             ctx = next(self.reader)
         except StopIteration:
             ctx = None
+        except socket.error as e:
```
abhinav: This will also catch timeouts. Shouldn't they have a different behavior?

blampe: yes but just trying to catch these for now

python/tchannel/tornado/connection.py
```diff
@@ -64,28 +73,58 @@ def extract_handshake_headers(self, message):
...
+        # TODO: track awaiting responses in here
         message_id = message_id or self.next_message_id()
+
+        if message.message_type in (
+            messages.Types.CALL_REQ,
+            messages.Types.INIT_REQ,
+            messages.Types.PING_REQ,
+        ):
+            log.debug("awaiting response for message %s", message_id)
+            self.awaiting_responses[message_id] = gen.Future()
```
abhinav: Where is this future consumed?

blampe: In `send`.

abhinav:
:shipit:

breerly:
:shipit:

PR #218, opened 2015-03-23
reset the socket on TChannel reader error

Raynos:
Currently when the server gets a reader error. For example
parsing a `service~2` when the protocol is `service~1` it
does not close the TChannelConnection object.

This means that the client thinks it's a timeout (i.e. server
is unresponsive) instead of thinking it's a BadRequest.

Ideally we terminate with a BadRequest error frame but just
closing the socket and turning the client error into an
ECONNRESET error is a significant improvement over a timeout.

This change will make my tests fail with a loud error instead
of a quiet timeout.

### DIFF
--- a/node/index.js
+++ b/node/index.js
@@ -590,10 +590,6 @@ TChannelConnection.prototype.onReaderError = function onReaderError(err) {
         localName: self.channel.hostPort
     });
 
-    self.channel.logger.error('tchannel read error', {
-        error: err
-    });
-
     // TODO instead of resetting send an error frame first.
     // and reset the socket after sending an error frame
     self.resetAll(readError);
--- a/node/index.js
+++ b/node/index.js
@@ -46,6 +46,13 @@ var TChannelListenError = WrappedError({
     host: null
 });
 
+var TChannelReadProtocolError = WrappedError({
+    type: 'tchannel.protocol.read-failed',
+    message: 'tchannel read failure: {origMessage}',
+    remoteName: null,
+    localName: null
+});
+
 var NoHandlerError = TypedError({
     type: 'tchannel.no-handler',
     message: 'no handler defined'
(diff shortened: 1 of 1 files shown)


kriskowal:
So when we establish that the peer is crazy and not speaking the lingo, or corrupted and out of sync, we throw away the connection. Sounds good to me. :ok:

node/index.js
```diff
@@ -582,6 +582,12 @@ TChannelConnection.prototype.onReaderError = function onReaderError(err) {
         localName: self.channel.hostPort,
         error: err
     });
+
+    // TODO instead of resetting send an error frame first.
```
jcorbin: I think this is a TODO for resetAll in general: when resetting with an error, send an appropriate error frame first?

node/index.js
```diff
@@ -582,6 +582,12 @@ TChannelConnection.prototype.onReaderError = function onReaderError(err) {
...
+    // and reset the socket after sending an error frame
+    self.resetAll(new Error('tchannel read error'));
```
jcorbin: why not pass in a wrapped error and remove the log site above (doesn't resetAll do an error log for us?)

Raynos: I'll make it a wrapped error.

There might be value in having a seperate log call site for the tchannel read error instead of inferring it from the tchannel socket reset error where `err.type === 'tchannel.read-error'`

node/index.js
Raynos: Not necessarily. There are two uses of `resetAll()`

 - If the socket closes or errors then `resetAll()` the connection to clear any pending out / in ops.
 - If something goes wrong in tchannel then `resetAll()` the connection to destroy the socket and clear any pending out / in ops.

These need to be either two seperate methods or we need `options` passed into `resetAll()`

PR #219, opened 2015-03-23
Eliminate explicit arg{1,2,3}

jcorbin:
- in call req/res frames
- in handler
- in req/res objects' frame senders


jcorbin:
cc @Raynos @Raynos

kriskowal:
:ok:

Raynos:
lgtm.

PR #220, opened 2015-03-24
Make req/res states more extensible

jcorbin:
As you might guess by the gap between `0` and `2` this is nothing but setting the stage for a new `States.Streaming`.

reviewers: @kriskowal @Raynos


kriskowal:
Looks fast. :ok:

Raynos:
lgtm.

PR #224, opened 2015-03-24
Add call & cont fragmentation

jcorbin:
reviewers: @kriskowal @Raynos


jcorbin:
I'm intentionally leaving the fixes unsquashed in this branch for future demoing around the bug

PR #226, opened 2015-03-24
Document Thrift interaction

abhinav:
Also moved protocol.md to docs/

CC: @leizha @Raynos @jcorbin


docs/thrift.md
```diff
@@ -0,0 +1,205 @@
...
+
+To do this, the server code responsible for registering endpoints must first
+denormalize the Thrift file into a set of "leaf" services that contain all
+methods -- including inherited ones.
+
+Uncaught Exceptions
+-------------------
+
+For uncaught server-side exceptions that are not defined in the Thrift IDL,
+server implementations must respond with a TChannel `error` message with error
```
Raynos: This is not a `must`. Servers can crash in which case the response is an ECONNRESET or equivelant.

Also is it possible for unknown operational errors that are not defined in the thrift IDL to be create a CallResponse whose code is `1` or non-zero.

docs/thrift.md
```diff
@@ -0,0 +1,205 @@
...
+must be set to `thrift`. Requests will be made using `call req` messages and
+responses will be sent using `call res` messages, with values for `arg{1,2,3}`
+as defined in [Arguments][].
+
+For `call res`,
+
+-   In case of success, the Response Code (`code:1`) must be set to `0x00`.
+-   In case of failure, the Response Code (`code:1`) must be set to `0x01`.
+
+Arguments
```
Raynos: Should we add a note for what the `service:1` field should be ?

We should probably make it clear that it does not have to be thrift service but should be set (i.e. not an empty string).

docs/thrift.md
mmihic: CallResponse with non-zero code was (at least at one point) intended to be reserved for application-level errors not operational errors.  Operational errors should return an Error message, not a CallResponse.

Raynos:
Looks reasonable.

docs/thrift.md
abhinav: Added

docs/thrift.md
abhinav: Changed must-vs-should on the error.

I agree with @mmihic that only application errors should be propagated through call response. As @Raynos suggested in-person, we could theoretically add an application error with the field identifier `0xff` that will hold information on operational errors if we decide to send them as a call response.

breerly:
lgtm

PR #227, opened 2015-03-24
Prepare reqres for streaming

jcorbin:
Final code rotation around reqres to prepare for streaming.

Reviewers: @kriskowal @Raynos

### DIFF
--- a/node/reqres.js
+++ b/node/reqres.js
@@ -111,10 +111,9 @@ TChannelOutgoingRequest.prototype.sendCallRequestFrame = function sendCallReques
 TChannelOutgoingRequest.prototype.send = function send(arg1, arg2, arg3, callback) {
     var self = this;
     if (callback) self.hookupCallback(callback);
-    if (self.sent) {
+    if (self.state !== States.Initial) {
         throw new Error('request already sent');
     }
-    self.sent = true;
     self.sendCallRequestFrame([
         arg1 ? Buffer(arg1) : null,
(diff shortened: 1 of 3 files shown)


node/reqres.js
```diff
@@ -47,13 +48,29 @@ function TChannelIncomingRequest(id, options) {
     self.headers = options.headers || {};
     self.checksum = options.checksum || null;
     self.checksumType = options.checksumType || 0;
-    self.arg1 = options.arg1 || emptyBuffer;
```
Raynos: Any reason to not pre-define these as `null` in the constructor ?

node/reqres.js
```diff
@@ -134,7 +205,13 @@ TChannelOutgoingRequest.prototype.hookupCallback = function hookupCallback(callb
     }
     function onResponse(res) {
         self.removeListener('error', onError);
-        callback(null, res);
+        if (callback.canStream) {
```
Raynos: I dont understand the difference between streaming and non streaming here.

Why would one need zalgo protection and not the other ?

Raynos:
lgtm.

node/reqres.js
jcorbin: it's a stub for later use, it shouldn't make a ton of sense yet

node/reqres.js
jcorbin: Because they're going away real soon at this layer, to be provided by the upcoming use of argstream as base class in a future pr.

PR #228, opened 2015-03-24
add checksum for the python tchannel. 

junchaowu:
It doesn't cover streaming message yet.

### DIFF
--- a/python/tchannel/messages/common.py
+++ b/python/tchannel/messages/common.py
@@ -1,9 +1,13 @@
 from __future__ import absolute_import
 
+import zlib
+
 from collections import namedtuple
 from enum import IntEnum
 
 from .. import rw
+from ..exceptions import InvalidChecksumException
+from .types import Types
 
 PROTOCOL_VERSION = 0x02
 
@@ -24,6 +28,11 @@ class ChecksumType(IntEnum):
     crc32 = 0x01
     farm32 = 0x02
 
+    @staticmethod
+    def standardize(checksum):
+        return (ChecksumType(checksum[0]), checksum[1])
+
+
 checksum_rw = rw.switch(
     rw.number(1),   # csumtype:1
     {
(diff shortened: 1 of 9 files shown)


python/tchannel/messages/base.py
```diff
@@ -25,3 +30,59 @@ def __str__(self):
             str(self.__class__.__name__),
             ", ".join(attrs)
         )
+
+    @staticmethod
+    def compute_checksum(checksum_type, args, csum=0):
```
abhinav: Given that these three related methods are static and don't really need to be in `BaseMessage`. Move to a separate module as top-level functions. `tchannel.messages.common` or `tchannel.checksum`.

python/tchannel/messages/base.py
```diff
@@ -25,3 +30,59 @@ def __str__(self):
...
+        if csum is None:
+            csum = 0
+
+        if checksum_type == ChecksumType.none:
+            return None
+        elif checksum_type == ChecksumType.crc32:
+            for arg in args:
+                csum = zlib.crc32(arg, csum)
```
abhinav: `& 0xffffffff`?

python/tchannel/messages/base.py
```diff
@@ -25,3 +30,59 @@ def __str__(self):
...
+        # TODO investigate farm32 cross platform issue
+        elif checksum_type == ChecksumType.farm32:
+            for arg in args:
+                csum = farmhash.hash32withseed(arg, csum)
```
abhinav: Investigation has already been done. Use `farmhash.fingerprint32`.

python/tchannel/messages/base.py
junchaowu: I discussed this with Bryce. Bryce suggests I put here. Since only message will use checksum

python/tchannel/messages/base.py
abhinav: Upon a second look, it appears that the Python implementation doesn't provide that method. Throw a `NotImplementedError` here for now.

python/tchannel/messages/common.py
```diff
@@ -32,3 +38,9 @@ class ChecksumType(IntEnum):
         ChecksumType.farm32: rw.number(4),  # csum:4
     }
 )
+
+
+checksum_msg_types = [Types.CALL_REQ,
```
abhinav: Use CAPS for top-level constants like this.

Also, make it a `frozenset` instead of a list.

    frozenset([...])

abhinav:
Everything else looks good.

abhinav:
:shipit: (provided that tests pass)

Issue #118, opened 2015-03-03
node crc module should use SSE42 version


mranney:
Modern Intel CPUs have instructions for computing CRC32. We should use them:

https://www.npmjs.com/package/sse4_crc32

jcorbin:
It's not clear from the README that it presents a function(data, prior) to
support our chaining needs? I hope that's just a README oversight.

mranney:
Yes, total readme fail. The second arg is the "prior" value:

```
mjr-uber-mbp:~$ node
> var c = 0;
undefined
> c = require("sse4_crc32").calculate("foo", c)
3485773341
> c = require("sse4_crc32").calculate("bar", c)
224353407
> c
224353407
```

mranney:
FIxed: https://github.com/Voxer/sse4_crc32/commit/0ac8a858e2fd17e645d7aedf1a35885d5389b18f

jcorbin:
As mentioned in #230, I did a brief try to use this:
- it didn't seem to have hardware support on my MBP
- the software version returned a wrong result

This was from all of ~5 minute of trying one example tho mind.

Raynos:
We now support crc32

PR #230, opened 2015-03-25
Golang protocol updates

mmihic:
Fixes drifts from the spec, mainly that service~1 should precede headers in call req/res, and that call-res should include an empty arg1.

### DIFF
--- a/golang/channel.go
+++ b/golang/channel.go
@@ -101,6 +101,7 @@ func NewChannel(hostPort string, opts *ChannelOptions) (*TChannel, error) {
 	ch.hostPort = l.Addr().String()
 	ch.connectionOptions.PeerInfo.HostPort = ch.hostPort
 	ch.connectionOptions.PeerInfo.ProcessName = ch.processName
+	ch.connectionOptions.ChecksumType = ChecksumTypeCrc32
 	ch.log.Infof("%s listening on %s", ch.processName, ch.hostPort)
 	return ch, nil
 }
--- a/golang/fragmentation.go
+++ b/golang/fragmentation.go
@@ -212,6 +212,10 @@ func (w *bodyWriter) Write(b []byte) (int, error) {
 		return 0, ErrWriteAfterComplete
 	}
 
+	if len(b) == 0 {
+		return 0, w.writeEmpty()
+	}
+
 	written := 0
 	for len(b) > 0 {
 		// Make sure we have a fragment and an open chunk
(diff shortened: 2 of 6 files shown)


jcorbin:
The frames now properly parse, but I'm getting invalid checksum errors with tcap.

This certainly is a step in the right direction so :+1: from me.

jcorbin:
Example dumps from tcap:
```
session=0 127.0.0.1:57097 --> 127.0.0.1:10500 frame parse error
TchannelChecksumError: invalid checksum
    00: 0054 0300 0000 0002 0000 0000 0000 0000  .T..............
    10: 0000 0075 2d00 0000 0000 0000 0000 0000  ...u-...........
    20: 0000 0000 0000 0000 0000 0000 0000 0b54  ...............T
    30: 6573 7453 6572 7669 6365 0001 9def 1ec0  estService......
    40: 0004 6563 686f 0005 6865 6c6c 6f00 0577  ..echo..hello..w
    50: 6f72 6c64                                orld

session=0 127.0.0.1:57097 <-- 127.0.0.1:10500 frame parse error
TchannelChecksumError: invalid checksum
    00: 0041 0400 0000 0002 0000 0000 0000 0000  .A..............
    10: 0000 0000 0000 0000 0000 0000 0000 0000  ................
    20: 0000 0000 0000 0000 0000 0000 01f9 eb20  ...............
    30: ad00 0000 0568 656c 6c6f 0005 776f 726c  .....hello..worl
    40: 64                                       d
```

The call req with arg1="echo" arg2="hello" arg3="world" claims to be a crc32 of `0x9def1ec0`, yet we have:
```
$ node
> crc = require('crc')
> h = crc.crc32
> ['echo', 'hello', 'world'].reduce(function(c, a) {return h(a, c);}, 0).toString(16)
'b3ab98f8'
```

But it turns out that node looks like the odd duck out:
```
$ python3
>>> import zlib
>>> import functools
>>> hex(functools.reduce(lambda c, a: zlib.crc32(a.encode(), c), ['echo', 'hello', 'world'], 0))
'0x9def1ec0'
```

So land this and let's sort out node's divergence.

jcorbin:
Perhaps apropos of #118

mmihic:
Yeah, node is definitely the odd man out.  Bizarrely, I get different results from node for the same byte streams depending on whether I run the crc directly from a REPL vs running it through tcap.  I pretty much abandoned investigation there;  I'll leave it to you guys to figure out. :)

When run via REPL (produces the correct result based on IEEE polynomial):

````node
> var crc32 = require('crc').crc32
> n = crc32("echo", 0)
Starting CRC is 0
Processing 101, 99, 104, 111
csum result: 386150450 
````

When run through tcap:
````node
Starting CRC is 0
Processing 101, 99, 104, 111
csum result: 3384742097
````

**Comment by jcorbin** [comment]
Found the problem, it's the zero initializer:
```
> h('world', h('hello', h('echo'))).toString(16)
'9def1ec0'
> h('world', h('hello', h('echo', 0))).toString(16)
'b3ab98f8'
```

Interestingly the SSE4 library linked by @mranney is even worse:
```
> h = ssecrc32.calculate
[Function: swCrc32c]
> h('world', h('hello', h('echo', undefined))).toString(16)
'26d7a49b'
> h('world', h('hello', h('echo', 0))).toString(16)
'26d7a49b'
```

But to be fair that's its software fallback, so not much point in it anyhow.

jcorbin:
fixed in #250, still punting on the fancy sse version

PR #231, opened 2015-03-25
Handler streaming prep

jcorbin:
Final handler code rotation before streaming implementation.

reviewers: @Raynos @kriskowal

### DIFF
--- a/node/reqres.js
+++ b/node/reqres.js
@@ -188,9 +188,9 @@ TChannelOutgoingRequest.prototype.send = function send(arg1, arg2, arg3, callbac
         throw new Error('request already sent');
     }
     self.sendCallRequestFrame([
-        arg1 ? Buffer(arg1) : null,
-        arg2 ? Buffer(arg2) : null,
-        arg3 ? Buffer(arg3) : null
+        arg1 ? Buffer(arg1) : Buffer(0),
+        arg2 ? Buffer(arg2) : Buffer(0),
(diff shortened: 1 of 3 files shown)


jcorbin:
The end of the rainbow comes after this PR:
https://github.com/uber/tchannel/compare/handler_streaming_prep...streaming

node/reqres.js
```diff
@@ -285,21 +287,22 @@ TChannelOutgoingResponse.prototype.sendCallResponseContFrame = function sendCall
...
 };
 
 TChannelOutgoingResponse.prototype.sendErrorFrame = function sendErrorFrame(codeString, message) {
     var self = this;
-    if (self.state === States.Done) {
+    if (self.state === States.Done || self.state === States.Error) {
         throw new Error('response already done'); // TODO: typed error
     } else {
         self.sendFrame.error(codeString, message);
-        self.state = States.Done;
```
Raynos: Does the protocol state that a call response which is not finished can be terminated by an error response with the same id ?

Should we check that we are not in the Streaming state here ?

mmihic: Yes you need to be able to fast terminate a response with an Error frame at any time.

Raynos:
Seems legit.

node/reqres.js
Raynos: Opened an issue to have protocol.md updated ( https://github.com/uber/tchannel/issues/234 ).

PR #235, opened 2015-03-25
Streaming

jcorbin:
What I set out to PR

reviewers: @kriskowal @Raynos

### DIFF
--- a/node/endpoint-handler.js
+++ b/node/endpoint-handler.js
@@ -20,6 +20,7 @@
 
 'use strict';
 
+var parallel = require('run-parallel');
 var TypedError = require('error/typed');
 
 var EndpointAlreadyDefinedError = TypedError({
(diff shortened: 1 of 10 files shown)


node/endpoint-handler.js
```diff
@@ -72,16 +73,30 @@ TChannelEndpointHandler.prototype.register = function register(name, handler) {
...
-    if (!handler) {
-        res.sendNotOk(null, NoSuchEndpointError({
-            service: self.serviceName,
-            endpoint: name
-        }).message);
-        return;
-    }
-    handler(req, res);
+    // TODO: waterfall
+    req.arg1.onValueReady(function arg1Ready(err, arg1) {
```
Raynos: I'll have to update the autobahn router to use this new interface.

node/endpoint-handler.js
```diff
@@ -72,16 +73,30 @@ TChannelEndpointHandler.prototype.register = function register(name, handler) {
...
+        if (err) throw err; // TODO: protocol error, respond with error frame
```
Raynos: This is scary to add to master.

Can we make a `logger.error()` call and a `res.writeErrorFrame('BadRequest', 'you done messed up');`

Raynos:
lgtm.

PR #238, opened 2015-03-25
remove the error event for outgoingresponse

Raynos:
This is a breaking change from v1 that causes ringpop integration tests to fail (error crashes the process).

There really is not much we can do with this error in the `CallResponse` object in the `EndpointHandler`.

The socket is gone so we cant really do anything.

I recommend we remove it and think about how to do stuff with it later.


jcorbin:
Rolled into #240

PR #239, opened 2015-03-25
do not set closing; set inside method

Raynos:
This is a breaking change from v1 to v2.

The resetAll() method now shortcircuits on `closing = true`.

We should not be setting it outside of the method. If we do
we leave a bunch of inOp and outOp in a pending state and
since we reset the connection the timeout sweeper is cancelled.

This is needed for ringpop. This also cannot land until #238
is landed.


jcorbin:
:+1: nice catch

jcorbin:
Rolled into #240

PR #240, opened 2015-03-25
Error end args

jcorbin:
r: @Raynos

### DIFF
--- a/node/reqres.js
+++ b/node/reqres.js
@@ -135,6 +135,9 @@ TChannelOutgoingRequest.prototype.sendParts = function sendParts(parts, isLast) 
             // TODO: could probably happen normally, like say if a
             // streaming request is canceled
             throw new Error('got frame in done state'); // TODO: typed error
+        case States.Error:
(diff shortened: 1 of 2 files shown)


node/reqres.js
```diff
@@ -284,8 +287,13 @@ TChannelOutgoingResponse.prototype.sendError = function sendError(codeString, me
     if (self.state === States.Done || self.state === States.Error) {
         throw new Error('response already done'); // TODO: typed error
     } else {
-        self.sendFrame.error(codeString, message);
+        // TODO: we could decide to flush any parts in a (first?) call res frame
+        self.finished = true;
         self.state = States.Error;
+        if (!self.arg1._readableState.ended) self.arg1.end();
```
Raynos: we should check _writableState instead `end()` is a method on writable.

Apparently double end() ( https://github.com/joyent/node/blob/v0.10.32-release/lib/_stream_writable.js#L342-L343 ) is idempotent.

PR #241, opened 2015-03-25
fix call response message id issue

junchaowu:
the inbound server should reply message with same message id

### DIFF
--- a/python/tchannel/handler.py
+++ b/python/tchannel/handler.py
@@ -1,4 +1,5 @@
 import collections
+from .messages import CallResponseMessage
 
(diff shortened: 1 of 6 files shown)


python/tchannel/handler.py
```diff
@@ -75,6 +77,8 @@ def __init__(self, context, conn):
         self.body = self.message.arg_3
         self.method = self.message.arg_1
         self.connection = conn
+        self.context = context
+        self.id = context.message_id
```
blampe: You don't need this if you already have the `context`. `id` also makes it look like it belongs to the request.

blampe:
lgtm, I have something that I think will test this.

PR #243, opened 2015-03-25
Correct deps

jcorbin:


kriskowal:
Ah, yeah. Those are being used outside examples now. :+1:

PR #244, opened 2015-03-25
Tracing agent

rf:
Tracing agent. Now behind `options.trace`, will do nothing if this flag is off. Does not support per-request tracing toggle. Now uses sync linear congruential generator to make ids.


node/index.js
```diff
@@ -127,6 +131,21 @@ function TChannel(options) {
 
             self.logger.info(self.hostPort + ' listening');
             self.emit('listening');
+
+            if (self.options.trace) {
+                console.log("making tracer");
```
jcorbin: self.logger or drop

node/trace/span.js
```diff
@@ -0,0 +1,174 @@
...
+
+    if (options.id) {
+        self.id = options.id;
+    }
+
+    if (options.traceid) {
+        self.traceid = options.traceid;
+    }
+
+    self.rng = new LCG();
```
jcorbin: could just have this as a module global, no need for one on every span

jcorbin:
:+1: would be great to add test/v2/tracing.js

node/trace/agent.js
```diff
@@ -0,0 +1,167 @@
...
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+if (!process.addAsyncListener) {
+    require('async-listener');
```
Raynos: Can we lazy load this.

This mutates global state and hits performance by at least 20%; lets make this opt in as part of the first time we get the singleton.

node/trace/agent.js
```diff
@@ -0,0 +1,167 @@
...
+
+    process.addAsyncListener(self.asyncListener);
+
+    self.logger = DebugLogtron('tchannelTrace');
+}
+
+Agent._instance = null;
+
+// Tracing agent is a singleton
+Agent.getInstance = function () {
```
Raynos: This is not how singletons work.

Because of duplication between multiple copies & node_modules you will need real global variables.

Raynos:
lgtm; Lets put async-listener behind the flag though.

PR #246, opened 2015-03-26
[python] add a fixture to test tcurl against tchannel servers

blampe:
This should hopefully prevent another bug where we totally kill our ability to make requests :)

This brings coverage from 81% up to 83% :[

@junchaowu `TCPServerManager` is still using its own request handler which makes things a little awkward.

### DIFF
--- a/python/tests/integration/server_manager.py
+++ b/python/tests/integration/server_manager.py
@@ -7,7 +7,11 @@ except ImportError:
     import socketserver as SocketServer
 from contextlib import contextmanager
 
-import tchannel.socket as tchannel
+import tornado
+
+from tchannel.handler import TChannelRequestHandler
+import tchannel.tornado.tchannel as tornado_tchannel
+import tchannel.socket as socket_tchannel
 import tchannel.messages as tmessage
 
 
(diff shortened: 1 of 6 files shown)


python/tests/integration/server_manager.py
```diff
@@ -103,29 +96,99 @@ def handle_call(self, context, connection):
...
+    def __enter__(self):
+        self.start()
+        return self
+
+    def __exit__(self, *args):
+        self.stop()
+
+    def start(self):
+        assert self.thread is None, 'server already started'
+        self.thread = threading.Thread(target=self.serve)
```
abhinav: Since this class is not meant to be used directly and expects child classes to implement `serve`, can you add a dummy `serve` method to this class that `raise`s `NotImplementedError`?

python/tests/integration/server_manager.py
```diff
@@ -103,29 +96,99 @@ def handle_call(self, context, connection):
...
+        self.thread.start()
+        while not self.ready:
+            pass
```
abhinav: This will spin really fast, taking up 100% CPU for that duration.

Consider using a [Condition](https://docs.python.org/2/library/threading.html#condition-objects) to let the other thread notify this one that it's ready. So, instead of setting `ready` to `True`, that thread can call `ready()` on this class, which will call `notify` on the condition (which this thread was `wait`ing on).

abhinav:
:ok: besides minor comments.

junchaowu:
lgtm

PR #249, opened 2015-03-26
[node] Fix lost operation crash error

kriskowal:
getOp returns undefined if the operation does not exist or was dropped.
This fixes a crash.

To reproduce the crash, start a service that registers over tchannel
before starting the tchannel service registry. When the service
discovery system comes up, the service will crash because op is
undefined.


jcorbin:
:+1:

PR #250, opened 2015-03-26
Fix crc32 initial round

jcorbin:


mmihic:
+1, thanks for fixing.

PR #252, opened 2015-03-26
add streaming logic to python tchannel inbound server

junchaowu:


python/tchannel/messages/call_request_continue.py
```diff
@@ -2,10 +2,48 @@
...
+        arg_1=None,
+        arg_2=None,
+        arg_3=None,
+        args=None
+    ):
+        self.flags = flags
+        if checksum is not None:
+            checksum = common.ChecksumType.standardize(checksum)
+        self.checksum = checksum or \
+            (common.ChecksumType.none, None)
```
abhinav: Wouldn't you rather do this?

    checksum = checksum or (..., ...)
    checksum = [..].standardize(checksum)

python/tchannel/rw.py
```diff
@@ -437,9 +437,13 @@ def read(self, stream):
                 if attr != skip:
                     kwargs[attr] = value
         except ReadException as e:
-            raise ReadException(
-                "Failed to read %s: %s" % (self._cls, e.message)
-            )
+            """ capture the read exception msg here but
```
abhinav: Don't catch this here. Instead of breaking all things that use this read writer, just override the behavior for specific instances.

python/tchannel/tornado/message_builder.py
```diff
@@ -0,0 +1,65 @@
+from ..messages import Types
+
+
+class MessageBuilder(object):
```
abhinav: Documentation

python/tchannel/tornado/message_builder.py
```diff
@@ -0,0 +1,65 @@
...
+
+    ARGS = ["arg_1", "arg_2", "arg_3"]
+
+    def __init__(self):
+        """key: message_id
+           value: incomplete messages
```
abhinav: What?

python/tchannel/tornado/message_builder.py
```diff
@@ -0,0 +1,65 @@
...
+        self.message_buffer = {}
+
+    @staticmethod
+    def _get_cur_pending_arg(message):
+        cur_arg = 0
+        if message.arg_3 != "":
+            cur_arg = 2
+        elif message.arg_2 != "":
+            cur_arg = 1
+        print "cur_arg" + str(cur_arg)
```
abhinav: Remove print statement

python/tchannel/tornado/message_builder.py
```diff
@@ -0,0 +1,65 @@
...
+        cur_arg = self._get_cur_pending_arg(call_msg)
+
+        arg_map = {"arg_1": "",
+                   "arg_2": "",
+                   "arg_3": ""}
+        t = 0
+
+        for i in range(cur_arg, 3):
+            arg_map[self.ARGS[i]] = self.ARGS[t]
+            t += 1
```
abhinav: Isn't all this just constructing `{"arg_1": "arg_1", "arg_2": "arg_2", "arg_3": "arg_3"}`?

python/tchannel/tornado/message_builder.py
```diff
@@ -0,0 +1,65 @@
...
+
+        return arg_map
+
+    def build(self, message_id, message):
+
+        if message.message_type is Types.CALL_REQ or\
+           message.message_type is Types.CALL_RES:
```
abhinav: Use

    message_type in [Types.CALL_REQ, Types.CALL_RES

python/tchannel/tornado/message_builder.py
```diff
@@ -0,0 +1,65 @@
...
+            if message.flags == 0x01:  # streaming message
+                message.flags = 0x00
+                self.message_buffer[message_id] = message
+                return None
+        elif message.message_type is Types.CALL_REQ_CONTINUE or\
+                message.message_type is Types.CALL_RES_CONTINUE:
+            call_msg = self.message_buffer.get(message_id)
+            if call_msg is None:
+                # missing call msg before continue msg
+                raise Exception()
```
abhinav: More descriptive exception

python/tchannel/tornado/message_builder.py
```diff
@@ -0,0 +1,65 @@
...
+    def __init__(self):
+        """key: message_id
+           value: incomplete messages
+        """
+        self.message_buffer = {}
+
+    @staticmethod
+    def _get_cur_pending_arg(message):
+        cur_arg = 0
+        if message.arg_3 != "":
```
abhinav: Do

    if not message.arg_3:

That will handle both, None and empty string cases.

python/tchannel/tornado/message_builder.py
```diff
@@ -0,0 +1,65 @@
...
+
+    def __init__(self):
+        """key: message_id
+           value: incomplete messages
+        """
+        self.message_buffer = {}
+
+    @staticmethod
+    def _get_cur_pending_arg(message):
+        cur_arg = 0
```
abhinav: Readability nit: this should be an else case

python/tchannel/tornado/message_builder.py
```diff
@@ -0,0 +1,65 @@
...
+
+    def compose_map(self, call_msg):
+        cur_arg = self._get_cur_pending_arg(call_msg)
+
+        arg_map = {"arg_1": "",
+                   "arg_2": "",
+                   "arg_3": ""}
+        t = 0
+
+        for i in range(cur_arg, 3):
```
abhinav: Add some comments explaining this logic

python/tchannel/tornado/message_builder.py
```diff
@@ -0,0 +1,65 @@
...
+
+    ARGS = ["arg_1", "arg_2", "arg_3"]
+
+    def __init__(self):
+        """key: message_id
+           value: incomplete messages
+        """
+        self.message_buffer = {}
+
+    @staticmethod
```
breerly: :+1: for using `@staticmethod` when not needing access to self.

python/tchannel/tornado/message_builder.py
```diff
@@ -0,0 +1,65 @@
...
+        t = 0
+
+        for i in range(cur_arg, 3):
+            arg_map[self.ARGS[i]] = self.ARGS[t]
+            t += 1
+
+        return arg_map
+
+    def build(self, message_id, message):
+
```
breerly: nit - no space here

python/tchannel/tornado/message_builder.py
```diff
@@ -0,0 +1,65 @@
...
+        if message.message_type is Types.CALL_REQ or\
+           message.message_type is Types.CALL_RES:
+            if message.flags == 0x01:  # streaming message
```
breerly: Seems like these values (`0x01`, `0x00`, etc...) should be in constants so that we can 1 change them and 2 reuse them - @abhinav @blampe waht do you guys think?

PR #256, opened 2015-03-26
Gracefully handle redundant error responses.

### LINKED ISSUE #254: node: Getting an error frame for a timed out operation
There is a race where we either get two error frames or a timeout and an error frame.

This causes us to have an undefined `op` which is not handled nicely.

```
/home/raynos/uber/ncar-performance-test/node_modules/tchannel/index.js:506
        op.req.emit('error', err);
          ^
TypeError: Cannot read property 'req' of undefined
```

kriskowal:
And downgrade redundant errors and responses to `info` log level since
these are expected in normal operation due to timeouts, retry,
speculative pretry, etc.

Fixes #254 

review: @jcorbin @Raynos

### DIFF
--- a/node/index.js
+++ b/node/index.js
@@ -489,7 +489,7 @@ function TChannelConnection(channel, socket, direction, remoteAddr) {
     self.handler.on('call.incoming.response', function onCallResponse(res) {
         var op = self.popOutOp(res.id);
         if (!op) {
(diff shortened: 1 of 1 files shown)


kriskowal:
TChannel does not presently thread a statsd client, which would be nice for these cases.

Raynos:
lgtm. #shipit

PR #260, opened 2015-03-27
better parity between tcp/tchannel request handling

blampe:
This makes our TCP server share `TChannelRequestHandler` logic with tornado.

This also fixes a long-standing issue where the TCP example couldn't respond to multiple call requests, but the `tcurl` test still fails with multiple requests and I still need to fix that.


abhinav:
:shipit:

PR #167, opened 2015-03-17
Update tracing definition for consistency

jcorbin:
Will block on #163.

Maybe roll into #164.

cc @mranney @rf

Issue #166, opened 2015-03-17
Why does the error frame have an `id:4` in the body?


Raynos:
There appears to be an `id` in the error frame; both in the frame id and the body id.

Not sure why that is.

cc @mranney

breerly:
@Raynos this still an issue?

Raynos:
I think its fixed.

jcorbin:
actually it's not been fixed yet, just tossed out #261 to get the ball rolling

PR #261, opened 2015-03-27
v2/error_response: correct id => tracing

jcorbin:
We changed this protocol slightly in #167 out of #166 late in the game, but never updated any implementations yet.

There aren't any claim/cancel implementations to update in node yet, so this PR is error frame only.

cc: @mmihic @abhinav @Raynos @kriskowal

### DIFF
--- a/node/test/v2/error_response.js
+++ b/node/test/v2/error_response.js
@@ -23,6 +23,14 @@
 var test = require('tape');
 var ErrorResponse = require('../../v2/error_response.js');
 var testRW = require('bufrw/test_rw');
+var Tracing = require('../../v2/tracing.js');
+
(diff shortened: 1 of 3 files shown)


mmihic:
I assume the intent is that this is the same value as the tracing element on the call-request that generated an error (or is empty, if the error is a protocol level error not corresponding to a specific call-request)

jcorbin:
Correct @mmihic, as currently written the node error body constructor will fill in an empty tracing (all 0s) when passed a null/undefined tracing.  Notably, the node side doesn't yet have any place where it sends any such protocol-level error frame.

Raynos:
lgtm.

Raynos:
lgtm.

PR #263, opened 2015-03-27
Allow service name in request

breerly:
Trouter.py needs to be able to specify the service parameter in a tchannel request.

Is there anywhere else I need to do something similar?


abhinav:
:shipit

PR #265, opened 2015-03-27
Error framing should use tracing instead of original_message_id

breerly:
Will fix tests once I know the direction is sane.

### DIFF
--- a/python/tchannel/messages/error.py
+++ b/python/tchannel/messages/error.py
@@ -41,10 +41,10 @@ class ErrorMessage(BaseMessage):
         ErrorCode.fatal: 'fatal protocol error'
     }
 
-    def __init__(self, code=None, original_message_id=None, message=None):
-        self.original_message_id = original_message_id
+    def __init__(self, code=None, tracing=None, message=None):
         self.code = ErrorCode(code) if code else ErrorCode.unexpected
         self.message = message or ''
+        self.tracing = tracing or common.Tracing(0, 0, 0, 0)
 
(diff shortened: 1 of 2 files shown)


python/tchannel/messages/error.py
```diff
@@ -53,6 +54,6 @@ def error_name(self):
 error_rw = rw.instance(
     ErrorMessage,
     ('code', error_code_rw),                            # code:1
-    ('original_message_id', rw.number(4)),              # id:4
+    ('tracing', common.tracing_rw),                     # tracing:24
```
breerly: I think this should be tracing:25? @abhinav is that right?

abhinav: I think tracing_rw only reads 24 bytes for the trace IDs right now. It needs to be modified to read the traceflags and explicit traceflags parsing needs to be removed from everywhere else.

Either that or this needs to read another byte for the traceflags.

junchaowu: in the common.py, we have tracing_rw = rw.instance(
    Tracing,
    ("span_id", rw.number(8)),      # span_id:8
    ("parent_id", rw.number(8)),    # parent_id:8
    ("trace_id", rw.number(8)),     # trace_id:8
    ("traceflags", rw.number(1)),   # traceflags:1
)

abhinav: Oh cool. Didn't realize that was fixed.

breerly:
:shipit:

PR #267, opened 2015-03-28
Peering hosts

jcorbin:
First steps towards peering changes:
- refactors the channel peering structure to:
  - by an explicit map-like
  - have explicit peer objects
- adds skeleton peer state objects, only used for selection in static ways presently
- adds support for hosts option
  - [ ] TODO document
  - [x] TODO tests
  - [ ] TODO example usage

reviewers: @kriskowal @Raynos

### DIFF
--- a/node/index.js
+++ b/node/index.js
@@ -869,7 +869,9 @@ function TChannelPeer(channel, hostPort, options) {
     self.options = options || {};
     self.hostPort = hostPort;
     self.isEphemeral = self.hostPort === '0.0.0.0:0';
+    self.state = null; // TODO
     self.connections = [];
+    self.setState(TChannelPeerHealthyState(self.channel));
 }
 
 inherits(TChannelPeer, EventEmitter);
@@ -902,6 +904,15 @@ TChannelPeer.prototype.close = function close(callback) {
             }
             callback();
         }
+    }
+};
+
+TChannelPeer.prototype.setState = function setState(state) {
+    var self = this;
+    if (!self.state || state.name !== self.state.name) {
+        var oldState = self.state;
+        self.state = state;
+        self.emit('stateChanged', oldState, state);
     }
 };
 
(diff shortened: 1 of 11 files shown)


jcorbin:
ready for review @kriskowal / @Raynos I want to get this in, then start deving sub channel support by integration with autobahn

kriskowal:
:+1:

node/index.js
```diff
@@ -867,4 +759,259 @@ function TChannelClientOp(req, start) {
...
+    if (threshold === undefined) threshold = self.options.peerScoreThreshold;
+    if (threshold === undefined) threshold = 0;
+
+    var selectedPeer = null, selectedScore = 0;
+    for (var i = 0; i < hosts.length; i++) {
+        var peer = self.add(hosts[i]);
+        var score = peer.state.shouldRequest(options, op);
+        var want = score > threshold &&
+                   (selectedPeer === null || score > selectedScore);
+        self.logger.debug('choose peer score', {
```
Raynos: This happens for every `request()`; consider using `.trace()`

node/index.js
```diff
@@ -867,4 +759,259 @@ function TChannelClientOp(req, start) {
...
+    var self = this;
+    self.channel = channel;
+    self.name = name;
+}
+
+TChannelPeerState.prototype.shouldRequest = function shouldRequest(/* options, op */) {
+    // TODO: op isn't quite right currently as a "TChannelClientOp", the
+    // intention is that the other (non-options) arg encapsulates all requests
+    // across retries and setries
+    return 0;
```
Raynos: What does `0` means ? I would guess that `0` means that it should never request.

Raynos:
#LEGIT

jcorbin:
Updated:
- renamed / implemented more map-like keys/values/entries/delete
- added addPeer, needed in upcoming sub-channel work
- commented out the score debugging for now, I don't think even trace is right

Raynos:
lgtm :D

PR #268, opened 2015-03-28
Pin version of node readable-stream

jcorbin:
cc @Raynos

### DIFF
--- a/node/package.json
+++ b/node/package.json
@@ -18,7 +18,7 @@
     "url": "git@github.com:uber/tchannel"
   },
   "dependencies": {
-    "bufrw": "^0.9.4",
+    "bufrw": "^0.9.8",
     "crc": "^3.2.1",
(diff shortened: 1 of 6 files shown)


node/package.json
```diff
@@ -18,11 +18,12 @@
     "url": "git@github.com:uber/tchannel"
   },
   "dependencies": {
-    "bufrw": "^0.9.4",
+    "bufrw": "^0.9.8",
     "crc": "^3.2.1",
     "error": "^5.1.1",
     "farmhash": "^0.2.0",
     "hexer": "^1.2.0",
+    "readable-stream": "^1.0.33",
```
Raynos: This is wrong. This will use `>= 1.0.33 && <= 2`

You dont want that; since `1.1.x` breaks (i.e. node0.12). You wan to remove the caret and depend on 1.0.33 exactly.

PR #269, opened 2015-03-28
Make streaming an optional mode

jcorbin:
What this does:
- unbreaks the benchmark
- restores unstreamed mode:
  - you now only need to use streaming if you don't know how much or have all of the data on hand
  - otherwise the old .send(...) path will send a call frame and 0 or more fragmented cont frames; l/w with res.send(Not)Ok
- points the way forward for even more efficient forwarding in TRouter:
  - either a new mode other than streamed (or not)
  - or a minor API pivot to the current unstreamed mode to provide per-frame semantics

How:
- restores old pre-streaming req/res semantics under a flag
- exposes streamed flag through req/res builders
- delay and delegate res building to the handler
- when we get an incoming message that's not fragmented, turn off streaming mode

cc: @Raynos @kriskowal

### DIFF
--- a/node/argstream.js
+++ b/node/argstream.js
@@ -81,6 +81,9 @@ function ArgStream() {
 inherits(ArgStream, EventEmitter);
 
 function InArgStream() {
+    if (!(this instanceof InArgStream)) {
+        return new InArgStream();
+    }
     var self = this;
     ArgStream.call(self);
     self.streams = [self.arg1, self.arg2, self.arg3];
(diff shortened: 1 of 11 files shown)


jcorbin:
Ready for review @kriskowal  / @Raynos

Raynos:
lgtm. This is a breaking change for autobahn but its easy to migrate.

kriskowal:
:ok:

PR #270, opened 2015-03-28
Examples should use 0.0.0.0 for portability

breerly:
This change is needed for the tchannel-testbed I'm working on - the examples get installed into containers, which need to be accessible outside of themselves.

### DIFF
--- a/node/examples/example1.js
+++ b/node/examples/example1.js
@@ -46,12 +46,12 @@ var listening = ready(function (err) {
     }
 
     client
-        .request({host: '127.0.0.1:4040'})
+        .request({host: '0.0.0.0:4040'})
         .send('func 1', "arg 1", "arg 2", function (err, res) {
             console.log('normal res: ' + res.arg2.toString() + ' ' + res.arg3.toString());
         });
     client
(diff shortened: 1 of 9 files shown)


breerly:
Looks like the node implementation doesn't like `0.0.0.0` as an address to listen on:

```
if (host === '0.0.0.0') {
     throw new Error('TChannel must listen with externally visible host'); // TODO typed error
}
```
@jcorbin @Raynos any idea how I can get the node examples accessible from outside the same box? With the python example, switching to `0.0.0.0` has the affect of allowing requests from anywhere.

Raynos:
@breerly `npm i my-local-ip` ( https://github.com/dominictarr/my-local-ip ).

This will grab the external; non localhost IP that you can listen to.

jcorbin:
@breerly listening on `0.0.0.0` is an anti pattern that we don't want to encourage, let's:
- [ ] drop the `.listen()` from all example clients now that we have ephemeral semantics
- [ ] use explicit IPs, either loopback or discovered, for example server

kriskowal:
Closing. Tracking follow-up on #505

PR #272, opened 2015-03-29
Update Error frame handling to follow...

mmihic:
...latest protocol changes.  Notably, get rid of originalMessageID and
add tracing.


abhinav:
LGTM. Something seems wrong with the `godep` restore?

mmihic:
yeah for some reason it timed out.  I haven't changed the godeps so I'm not concerned about it.

PR #281, opened 2015-03-30
Streamming

junchaowu:
implement streaming for both client and server
change old test cases to use new streaming methods


python/tchannel/exceptions.py
```diff
@@ -34,18 +34,21 @@ class InvalidChecksumException(TChannelException):
...
     """The remote application returned an exception.
 
     This is not a protocol error. This means a response was received with the
     ``code`` flag set to fail.
     """
-    def __init__(self, code, arg_1, arg_2, arg_3):
+    def __init__(self, code, args):
         super(TChannelException, self).__init__(
-            'TChannel application error (%s, %s, %s)' % (arg_1, arg_2, arg_3)
+            'TChannel application error (%s)' % (args)
```
abhinav: Is args a tuple? If so, this will fail because `(args)` just means `args`, not a tuple with `args`.

    In [1]: "foo %s" % ('a', 'b', 'c')
    ---------------------------------------------------------------------------
    TypeError                                 Traceback (most recent call last)
    <ipython-input-1-222a695f910b> in <module>()
    ----> 1 "foo %s" % ('a', 'b', 'c')

    TypeError: not all arguments converted during string formatting

You probably want to use the `(x,)` form to create a tuple with a single item:

    In [2]: "foo %s" % (('a', 'b', 'c'),)
    Out[2]: "foo ('a', 'b', 'c')"

python/tchannel/handler.py
```diff
@@ -38,23 +38,23 @@ def handle_request(self, context, conn):
...
-        elif context.message.message_type == PingRequestMessage.message_type:
+        if context.message.message_type == PingRequestMessage.message_type:
             response = TChannelResponse(request, conn)
             response.resp_msg = PingResponseMessage()
             response.finish()
 
         else:
-            # TODO error handling if endpoint is not found
-            raise NotImplementedError(request.method)
+            endpoint = self._find_endpoint(getattr(request, 'method', None))
```
abhinav: Shouldn't this happen only if the `message_type` is `CallRequestMessage`?

python/tchannel/exceptions.py
junchaowu: it's an array not tuple. it will pass

python/tchannel/handler.py
```diff
@@ -80,9 +80,14 @@ class TChannelRequest(object):
...
-        self.header = getattr(self.message, 'arg_2', None)
-        self.body = getattr(self.message, 'arg_3', None)
-        self.method = getattr(self.message, 'arg_1', None)
+
+        try:
+            self.method = self.message.args[0]
+            self.header = self.message.args[1]
+            self.body = self.message.args[2]
+        except:
+            pass
```
abhinav: This is extremely fragile. Check the length of `args` and handle the cases explicitly.

python/tchannel/messages/call_continue.py
```diff
@@ -0,0 +1,94 @@
...
+        if checksum is not None:
+            checksum = common.ChecksumType.standardize(checksum)
+        self.checksum = checksum or \
+            (common.ChecksumType.none, None)
+
+        self.args = args or []
+
+    def encode(self):
+        self.args = list(map(
+            lambda arg: common.encode_str(arg),
```
abhinav: Use list comprehension

python/tchannel/messages/base.py
```diff
@@ -25,3 +25,15 @@ def __str__(self):
             str(self.__class__.__name__),
             ", ".join(attrs)
         )
+
+    def encode(self):
+        """ Encode all the strings in the msg using
+        encode type from common.py
```
abhinav: What does this mean?

Also, for placeholders like this, use `raise NotImplementedError()` instead of `pass`.

python/tchannel/messages/call_continue.py
```diff
@@ -0,0 +1,94 @@
...
+        self,
+        flags=0,
+        checksum=None,
+        args=None,
+    ):
+        self.flags = flags
+        if checksum is not None:
+            checksum = common.ChecksumType.standardize(checksum)
+        self.checksum = checksum or \
+            (common.ChecksumType.none, None)
```
abhinav: Change to:

    if checksum is None:
        self.checksum = ...
    else:
        self.checksum = standardize(checksum)

python/tchannel/messages/call_continue.py
```diff
@@ -0,0 +1,94 @@
...
+
+        self.args = args or []
+
+    def encode(self):
+        self.args = list(map(
+            lambda arg: common.encode_str(arg),
+            self.args))
+
+    def decode(self):
```
abhinav: I don't understand why you're doing the UTF-8 encode/decode. As far as TChannel is concerned, arg1, 2, and 3 are binary blobs. If they are unicode strings, the caller is expected to encode them before passing them in.

python/tchannel/messages/call_continue.py
```diff
@@ -0,0 +1,94 @@
...
+
+    def fragment(self, space_left, fragment_msg):
+        """Streaming Message got fragmented based on
+        payload size. All the data within space_left
+        will be kept. All the rest will be shifted to
+        next fragment message.
+
+        :param space_left: space left for current frame
+        :param fragment_msg: the type is either
+        CallRequestMessage or CallResponseMessage
```
abhinav: For multi-line `:param:` entries, move the description to the next line and indent 4 spaces.

    :param fragment_msg:
        Multi-line description
        goes here

python/tchannel/rw.py
```diff
@@ -35,7 +35,11 @@ def number(width_bytes):
     return NumberReadWriter(width_bytes)
 
 
-def len_prefixed_string(length_rw, is_binary=False):
+def args(length_rw):
+    return ArgsReaderWriter(length_rw)
+
+
+def len_prefixed_string(length_rw, is_binary=True):
```
abhinav: Why was this default changed?

python/tchannel/messages/call_request.py
```diff
@@ -33,42 +28,65 @@ def __init__(
...
-            checksum = common.ChecksumType.standardize(checksum)
-        self.checksum = checksum or \
-            (common.ChecksumType.none, None)
 
-        self.arg_1 = arg_1 or ''
-        self.arg_2 = arg_2 or ''
-        self.arg_3 = arg_3 or ''
+    def encode(self):
+        super(CallRequestMessage, self).encode()
+        self.service = self.service.encode(common.ENCODE_TYPE)
```
abhinav: `service` is expected to be a UTF-8 string and is automatically encoded/decoded by `len_prefixed_string`.

python/tchannel/messages/call_response.py
```diff
@@ -30,40 +26,60 @@ def __init__(
...
-        self.checksum = checksum or \
-            (common.ChecksumType.none, None)
 
-        self.arg_1 = arg_1 or ''
-        self.arg_2 = arg_2 or ''
-        self.arg_3 = arg_3 or ''
+    def encode(self):
+        super(CallResponseMessage, self).encode()
+        self.headers = dict(map(
+            lambda (k, v): common.encode_pair(k, v),
```
abhinav: `dict` comprehension

python/tchannel/messages/call_response.py
```diff
@@ -30,40 +26,60 @@ def __init__(
...
+        size += 25      # tracing: 24 | traceflags: 1
+        size += 1       # nh: 1
+        for k, v in self.headers:
+            size += 1       # hk~1
+            size += len(k)
+            size += 1       # hv~1
+            size += len(v)
+
+        size += 1       # csumtype: 1
+        size += 1 if self.checksum[0] != ChecksumType.none else 0
```
abhinav: Is this for the actual checksum? That's 4 bytes, not 1.

python/tchannel/tornado/message_factory.py
```diff
@@ -0,0 +1,98 @@
...
+class MessageFactory(object):
+
+    # 64KB Max frame size
+    # 16B (size:2 | type:1 | reserved:1 | id:4 | reserved:8)
+    # 1 2 Bytes can represent 0~64KB-1
+    MAX_PAYLOAD_SIZE = 0xFFEF   # 64*1024 - 16 - 1
+
+    def __init__(self):
+        """key: message_id
+           value: incomplete messages
```
abhinav: This shouldn't be a docstring on the `__init__` method but a regular `#` comment on `message_buffer`.

python/tchannel/tornado/message_factory.py
```diff
@@ -0,0 +1,98 @@
...
+from ..messages.common import FlagsType
+from ..messages.call_continue import CallContinueMessage
+
+
+class MessageFactory(object):
+
+    # 64KB Max frame size
+    # 16B (size:2 | type:1 | reserved:1 | id:4 | reserved:8)
+    # 1 2 Bytes can represent 0~64KB-1
+    MAX_PAYLOAD_SIZE = 0xFFEF   # 64*1024 - 16 - 1
```
abhinav: Nit: This should probably be expressed as `(2^16- 1) - 16` in the comment?

Even better might be to calculate it at import time like so,

    2**16 - 1 - frame_rw.size_rw.width() - frame_rw.header_rw.width()

Thoughts?

python/tchannel/tornado/message_factory.py
```diff
@@ -0,0 +1,98 @@
+from ..messages import Types
+from ..exceptions import StreamingException
+from ..messages.common import FlagsType
+from ..messages.call_continue import CallContinueMessage
+
+
+class MessageFactory(object):
```
abhinav: Add a docstring explaining what this class is responsible for doing.

python/tchannel/tornado/message_factory.py
```diff
@@ -0,0 +1,98 @@
...
+        """
+        self.message_buffer = {}
+
+    def build(self, message_id, message):
+        """buffer all the streaming messages based on the
+        message id. Reconstruct all fragments together.
+
+        :param message_id: id
+        :param message: incoming message
+        :return: next complete message or None
```
abhinav: Explain the condition for when `None` is returned: ".. or None if [...]"

python/tchannel/tornado/message_factory.py
```diff
@@ -0,0 +1,98 @@
...
+            if message.flags == 0x01:  # streaming message
+                message.flags = 0x00
+                self.message_buffer[message_id] = message
+                return None
+        elif message.message_type in [Types.CALL_REQ_CONTINUE,
+                                      Types.CALL_RES_CONTINUE]:
+            call_msg = self.message_buffer.get(message_id)
+            if call_msg is None:
+                # missing call msg before continue msg
+                raise StreamingException()
```
abhinav: Add a more helpful error message to the exception

python/tchannel/tornado/message_factory.py
```diff
@@ -0,0 +1,98 @@
...
+        fragments = []
+        message.encode()
+        if message.message_type in [Types.CALL_RES,
+                                    Types.CALL_REQ]:
+
+            payload_space = self.MAX_PAYLOAD_SIZE -\
+                message.get_meta_size()
+
+            """split a call/request message into an array
+            with a call/request message and {0~n} continue
```
abhinav: Use `#` comments, not docstrings

python/tchannel/tornado/message_factory.py
```diff
@@ -0,0 +1,98 @@
...
+        :return: list of messages whose sizes <= max
+        payload size
+        """
+        fragments = []
+        message.encode()
+        if message.message_type in [Types.CALL_RES,
+                                    Types.CALL_REQ]:
+
+            payload_space = self.MAX_PAYLOAD_SIZE -\
+                message.get_meta_size()
```
abhinav: Use parens, not `\` for multi-line statements

python/tchannel/handler.py
blampe: It's confusing to call this `request.method` when it's really just `arg_1`.

python/tchannel/handler.py
```diff
@@ -94,21 +99,29 @@ class TChannelResponse(object):
     """TChannel Response Wrapper"""
 
     __slots__ = ('_connection', '_request',
-                 'resp_msg', 'id')
+                 'resp_msg', 'id', 'body',
+                 'headers')
```
blampe: Something's wrong here, we already have a `message` here, so any `headers`/`body` we write should be written to `message.arg_2` and `message.arg_3`, right?

python/tests/integration/server_manager.py
```diff
@@ -108,7 +108,8 @@ def client_connection(self):
             )
             yield conn
         finally:
-            conn.close()
+            if 'conn' in locals():
```
abhinav: No. Initialize it `None` outside the `try` block and check if it's `not None` here.

python/tests/conftest.py
```diff
@@ -1,8 +1,14 @@
 from __future__ import absolute_import
 
+from random import choice
 import socket
-
 import pytest
+from string import lowercase
+
+
+def big_arg():
```
abhinav: Make this a fixture with `@pytest.fixture`.

python/tchannel/handler.py
```diff
@@ -94,21 +99,29 @@ class TChannelResponse(object):
...
 
     def __init__(self, request, conn):
         self._connection = conn
         self._request = request
-        self.resp_msg = CallResponseMessage()
+        self.body = ""
+        self.headers = ""
         self.id = request.id
+        self.resp_msg = None
```
blampe: Or are we not using this anymore? If we're not using it we should get rid of it. It's a little confusion that there's two ways to build a response.

python/tests/integration/test_client_server.py
```diff
@@ -9,11 +9,12 @@
 from tchannel.outgoing import OutgoingTChannel
 from tchannel.tornado import TChannel
 from tchannel.tornado.connection import TornadoConnection
+from tests.conftest import big_arg
```
abhinav: Once `big_arg` is a fixture, instead of importing it, just use it in functions that need it.

python/tests/integration/test_client_server.py
```diff
@@ -111,27 +112,63 @@ def test_tchannel_call_request(tcp_server, call_response):
...
 
     response = yield tchannel.request(hostport).send(endpoint, None, None)
+    assert response.args[0] == call_response.args[0]
+    assert response.args[2] == call_response.args[2]
 
-    assert response.arg_1 == call_response.arg_1
-    assert response.arg_3 == call_response.arg_3
+
+@pytest.fixture
+def load_data_from_file():
```
abhinav: Why does this exist? You're not using it anywhere.

Also, this is a fixture that returns `None` so the caller wouldn't work anyway.

python/tchannel/messages/call_continue.py
```diff
@@ -0,0 +1,94 @@
...
+
+
+        :return: None if there is space left
+        or next fragment message
+        """
+        new_args = []
+        for i, arg in enumerate(self.args):
+            if space_left >= 2:     # 2bytes for size
+                space_left -= 2
```
blampe: Let's not hard-code these.

python/tests/integration/test_client_server.py
```diff
@@ -111,27 +112,63 @@ def test_tchannel_call_request(tcp_server, call_response):
...
+            file.seek(0)
+
+
+# TODO test case will fail due to StreamClosedError when
+# increase the LARGE_AMOUNT to even bigger
+@pytest.mark.gen_test
+@pytest.mark.parametrize('arg2, arg3', [
+    ("", big_arg()),
+    (big_arg(), ""),
+    ("test", big_arg()),
```
abhinav: Ah, given how you're using `big_arg`, leave it as a function but move it to a separate util module.

`conftest` is for configuration. Try not to import it directly.

python/tchannel/messages/init_request.py
```diff
@@ -44,7 +44,7 @@ def process_name(self, value):
     ('version', rw.number(2)),  # version:2
     ('headers', rw.headers(     # nh:2 (key~2 value~2){nh}
         rw.number(2),
-        rw.len_prefixed_string(rw.number(2)),
-        rw.len_prefixed_string(rw.number(2)),
+        rw.len_prefixed_string(rw.number(2), is_binary=False),
+        rw.len_prefixed_string(rw.number(2), is_binary=False),
```
blampe: Why `is_binary=False`?

python/tchannel/messages/init_response.py
```diff
@@ -14,7 +14,7 @@ class InitResponseMessage(InitRequestMessage):
     ('version', rw.number(2)),  # version:2
     ('headers', rw.headers(     # nh:2 (key~2 value~2){nh}
         rw.number(2),
-        rw.len_prefixed_string(rw.number(2)),
-        rw.len_prefixed_string(rw.number(2)),
+        rw.len_prefixed_string(rw.number(2), is_binary=False),
+        rw.len_prefixed_string(rw.number(2), is_binary=False),
```
blampe: Why `is_binary=False`?

python/tchannel/rw.py
```diff
@@ -332,12 +336,38 @@ def width(self):
...
+        return self.num * self._length_rw.width()
+
+
 class LengthPrefixedBlobReadWriter(ReadWriter):
     """See :py:func:`len_prefixed_string` for documentation."""
 
     __slots__ = ('_length', '_is_binary')
 
-    def __init__(self, length_rw, is_binary=False):
+    def __init__(self, length_rw, is_binary=True):
```
blampe: Ah I see nvm.

python/tchannel/socket.py
```diff
@@ -107,6 +108,12 @@ def close(self):
         self.closed = True
         return self.connection.close()
 
+    def message_and_write(self, message, message_id=None):
```
blampe: I just checked out your branch and we might have a similar performance degradation in the streaming case. So for now we probably want to do what @jcorbin is doing with node, where we allow both code paths (complex streaming, simplified unstreamed case) until we can figure out how to speed up this path.

Additionally I would make these functions names more representative of what they do, like `frame_and_write` and `frame_and_write_stream`, so it's more obvious when the caller is explicitly asking for streaming. It would be an error to `frame_and_write` a message over 64kb.

python/tchannel/handler.py
junchaowu: change the "method" to endpoint

python/tchannel/handler.py
junchaowu: the length of args have been check in the build method. Here just for safer consideration

abhinav: My concern is the catch-all exception. Check the length and assign attributes instead of a catch-all argument.

Also, you want to make sure these attributes exist regardless of whether the corresponding arg was set. So, you need to give them a default value of None or empty and then assign to them based on how many elements `message.args` contains.

python/tchannel/messages/call_continue.py
```diff
@@ -0,0 +1,94 @@
...
+        :param space_left: space left for current frame
+        :param fragment_msg: the type is either
+        CallRequestMessage or CallResponseMessage
+
+
+        :return: None if there is space left
+        or next fragment message
+        """
+        new_args = []
+        for i, arg in enumerate(self.args):
```
blampe: Eventually `self.args` will be initialized to something like `[EmptyBuffer(), EmptyBuffer(), EmptyBuffer()]`, and as we read `CallRequestContinue` messages those buffers will be written to (and closed once the arg is done).

Currently, this waits for all arguments to be available *and* blocks all other request handling in the meantime (a `for` loop that's doing a lot of work like this blocks the ioloop).

python/tchannel/tornado/message_factory.py
```diff
@@ -0,0 +1,98 @@
...
+            message
+            """
+            fragment_msg = message.fragment(payload_space)
+            fragments.append(message)
+            while fragment_msg is not None:
+                message = fragment_msg
+                payload_space = self.MAX_PAYLOAD_SIZE -\
+                    message.get_meta_size()
+                fragment_msg = message.fragment(payload_space)
+                fragments.append(message)
```
blampe: This should at least `yield` fragments.

python/tests/conftest.py
```diff
@@ -1,8 +1,14 @@
...
+    LARGE_AMOUNT = 64 * 1024 * 10
+    return "".join(choice(lowercase) for _ in range(LARGE_AMOUNT))
```
blampe: Why not truly random binary data?

```
def big_arg(n):
    return os.random(n)
```

python/tests/integration/test_client_server.py
blampe: Unused.

python/tchannel/messages/call_continue.py
```diff
@@ -0,0 +1,86 @@
...
+    ):
+        self.flags = flags
+        if checksum is None:
+            self.checksum = (common.ChecksumType.none, None)
+        else:
+            self.checksum = common.ChecksumType.standardize(checksum)
+
+        self.args = args or []
+
+    def get_meta_size(self):
```
abhinav: Unused. Delete.

python/tchannel/messages/call_response.py
```diff
@@ -30,40 +26,25 @@ def __init__(
...
     ("flags", rw.number(1)),    # flags:1
     ("code", rw.number(1)),     # code:1
-
     ("tracing", common.tracing_rw),     # tracing:24
                                         # traceflags: 1
-
     ("headers", rw.headers(             # nh:1 (hk~1 hv~1){nh}
         rw.number(1),
-        rw.len_prefixed_string(rw.number(1))
+        rw.len_prefixed_string(rw.number(1), is_binary=True)
```
abhinav: `is_binary` should be False. Headers are UTF8

python/tchannel/messages/init_request.py
abhinav: is_binary False is the default. You can remove these arguments.

python/tchannel/messages/call_request.py
```diff
@@ -33,42 +28,27 @@ def __init__(
...
+        rw.len_prefixed_string(rw.number(1), is_binary=True)
     )),
-
     ("checksum", common.checksum_rw),   # csumtype:1 (csum:4){0, 1}
-
-    ("arg_1", rw.len_prefixed_string(rw.number(2), is_binary=True)),  # arg1~2
-    ("arg_2", rw.len_prefixed_string(rw.number(2), is_binary=True)),  # arg2~2
-    ("arg_3", rw.len_prefixed_string(rw.number(2), is_binary=True)),  # arg3~2
+    ("args",
+     rw.args(rw.number(2), is_binary=True)),  # [arg1~2, arg2~2, arg3~2]
```
abhinav: is_binary True is default for args. Argument not needed.

python/tchannel/rw.py
```diff
@@ -35,7 +35,11 @@ def number(width_bytes):
     return NumberReadWriter(width_bytes)
 
 
-def len_prefixed_string(length_rw, is_binary=False):
+def args(length_rw, is_binary=True):
+    return ArgsReaderWriter(length_rw, is_binary=True)
```
abhinav: IMO you don't need `is_binary` to be customizable for `args`. It should always be True.

python/tchannel/rw.py
```diff
@@ -35,7 +35,11 @@ def number(width_bytes):
...
+
+
+def len_prefixed_string(length_rw, is_binary=True):
```
abhinav: `is_binary=False` should be the default for this

python/tchannel/rw.py
abhinav: Flip the default here

python/tchannel/messages/common.py
```diff
@@ -10,8 +10,14 @@
 from .types import Types
 
 PROTOCOL_VERSION = 0x02
+ENCODE_TYPE = 'utf-8'
+DECODE_TYPE = 'utf-8'
```
abhinav: Unused

abhinav:
:shipit:

PR #286, opened 2015-03-31
Instantiate socket server lazy

jcorbin:


kriskowal:
It occurs to me that this change introduces a hazard for anyone who wishes to grab the server socket before listening. Might be good to guard the address() method too. Things might exist that depend on the existence of the socket before listen.

Raynos:
lgtm.

I think that the public interface should only be `.address()` and `.listen()`; anyone grabbing `serverSocket` directly is doing the wrong thing.

PR #291, opened 2015-03-31
xtend should be a non-dev dependency

Raynos:
We are now using it in `index.js`

As an aside; There might be value in have all library
devDependencies be in depenendencies and using
`devDependencies` for things like istanbul and tape.

r: @kriskowal @jcorbin


kriskowal:
And async became a dependency of example2 while I wasn't looking.

kriskowal:
:ok:

PR #294, opened 2015-04-01
Node: add self peering support

jcorbin:
review: @kriskowal @Raynos


Raynos:
This looks good;

I would like to add a test for `.sendNotOk()` and `.error()`

PR #298, opened 2015-04-01
Tests for Thrift Tornado client integration.

abhinav:
@blampe @breerly


breerly:
I thought we were moving fast?

breerly:
:shipit:

PR #305, opened 2015-04-02
Add replr tchannel example

robskillington:
cc @Raynos @jcorbin


jcorbin:
I'm pretty happy with where this is at now, we could put the TermClient / TermServer in a new `/extras/term/{client,server}.js` for use as `require('tchannel/extras/term/...')` initially (later maybe a separate module once velocity has cooled off a bit).

jcorbin:
Handing off to @raynos as he's doing the first use case of this to restore repl functionality in our pilot project.

Raynos:
I already shipped the old repl which is good enough for now. Will get back to a tchannel repl later.

PR #310, opened 2015-04-02
RequestHandler should not be Tornado specific

abhinav:
Also, delete HTTP adapter layer. We don't need it.

@blampe @breerly @junchaowu


breerly:
:shipit:

PR #312, opened 2015-04-02
change the structure of thrift tornado transport class

junchaowu:


python/tchannel/thrift/transport/socket.py
```diff
@@ -58,7 +58,7 @@ def __init__(self, tchannel, hostport, service_name):
             name.
         """
         super(TChannelTransport, self).__init__(
-            tchannel, hostport, service_name
+            hostport, service_name
```
abhinav: Don't you need to record the tchannel object somewhere in this class?

junchaowu: you are right.

python/tchannel/thrift/transport/tornado_base.py
```diff
@@ -0,0 +1,95 @@
...
+            "read() must not be called directly. Use readFrame()."
+        )
+
+    def recv_call(self):
+        raise NotImplementedError(
+            "recv_call() not supported for Tornado. Use readFrame()."
+        )
+
+    @gen.coroutine
+    def flush(self):
```
abhinav: This doesn't need `coroutine` the decorator.

python/tchannel/thrift/transport/tornado_base.py
```diff
@@ -0,0 +1,95 @@
...
+
+
+class TChannelTornadoTransportBase(TChannelTransportBase):
+    """A Thrift Transport to send requests over an open TChannel connection.
+
+    Works with Thrift clients generated with the ``tornado`` option enabled.
+
+    .. code-block::
+
+        transport = TChannelTornadoTransport(
```
abhinav: Move the example back to the implementation. It's specific to that implementation.

python/tchannel/thrift/transport/socket.py
breerly: isn't he doing that on line 64?

abhinav: He is now. :)

python/tchannel/thrift/transport/tornado_base.py
```diff
@@ -0,0 +1,82 @@
...
+            "read() must not be called directly. Use readFrame()."
+        )
+
+    def recv_call(self):
+        raise NotImplementedError(
+            "recv_call() not supported for Tornado. Use readFrame()."
+        )
+
+    def flush(self):
+        raise NotImplementedError("Must be implemented.")
```
abhinav: Implementations shouldn't have to care that a queue is being used to send back responses. Add a method implementations can call to send a response back.

PR #313, opened 2015-04-02
add _writeframe method in the tornado_base.py

junchaowu:


abhinav:
LGTM but call it `_send_response` or something instead.

Framing is a low level concern. Tornado doesn't know about it.

PR #314, opened 2015-04-03
make tchannel a singleton

blampe:
@breerly

### DIFF
--- a/python/tchannel/tornado/tchannel.py
+++ b/python/tchannel/tornado/tchannel.py
@@ -21,6 +21,7 @@
 from __future__ import absolute_import
 
 import logging
+import threading
 import weakref
 
 import tornado.ioloop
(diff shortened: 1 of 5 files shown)


python/tchannel/tornado/tchannel.py
```diff
@@ -38,9 +39,20 @@
 class TChannel(object):
     """Manages inbound and outbound connections to various hosts."""
 
+    # We don't want to duplicate outgoing connections, so all instances of this
+    # class will be a signleton.
```
breerly: s/signleton/singleton/

breerly:
Not a huge fan of the non-explicit approach... prefer...

```
Tchannel::get_instance()
```

Or something that indicates more clearly that you're getting a singleton.

Whatevs. :shipit:

PR #324, opened 2015-04-03
Thrift Tornado TChannel Server Support

abhinav:
Tests pending

@breerly @blampe


python/tox.ini
```diff
@@ -13,3 +13,6 @@ commands = make lint
 [testenv:cover]
 commands =
   py.test --cov tchannel --cov-report=xml --cov-report=term-missing {posargs}
+
+[flake8]
+exclude = examples/*
```
blampe: don't do dis

python/tchannel/thrift/__init__.py
```diff
@@ -20,9 +20,98 @@
 
 from __future__ import absolute_import
```
blampe: put this somewhere other than `__init__.py`

python/tox.ini
abhinav: Changed to only ignore Thrift-generated code

python/tchannel/thrift/__init__.py
abhinav: Moved

blampe:
lgtm

PR #327, opened 2015-04-04
Improve operational errors:

jcorbin:
- don't create error objects or warn log for non-error conditions
- hack observability into debug logtron so that test failures are clearer

Because of the way resetAll works, a frequent test failure mode is a hang/timeout with a (previously silenced) warn log.  This certainly is what we want in production: connection errors get logged, and we move on.

However until we figure out someway to make such errors a hard failure in a test (say by registering some re-throwing error handlers) this provides us better visibility.  IMO we want this sort of visibility anyhow... maybe better done by hooking it up to arbitrary sink, such as `assert.comment` rather than `console.error`, but this works for now.

### DIFF
--- a/node/index.js
+++ b/node/index.js
@@ -736,21 +736,25 @@ TChannelConnection.prototype.start = function start() {
     var self = this;
     if (self.direction === 'out') {
         self.handler.sendInitRequest();
-        self.handler.once('init.response', function onOutIdentified(init) {
-            self.remoteName = init.hostPort;
-            self.emit('identified', {
-                hostPort: init.hostPort,
-                processName: init.processName
-            });
-        });
+        self.handler.once('init.response', onOutIdentified);
     } else {
-        self.handler.once('init.request', function onInIdentified(init) {
-            self.remoteName = init.hostPort;
-            self.channel.peers.add(self.remoteName).addConnection(self);
-            self.emit('identified', {
-                hostPort: init.hostPort,
-                processName: init.processName
-            });
+        self.handler.once('init.request', onInIdentified);
+    }
+
+    function onOutIdentified(init) {
+        self.remoteName = init.hostPort;
+        self.emit('identified', {
+            hostPort: init.hostPort,
+            processName: init.processName
+        });
+    }
+
+    function onInIdentified(init) {
+        self.remoteName = init.hostPort;
+        self.channel.peers.add(self.remoteName).addConnection(self);
+        self.emit('identified', {
+            hostPort: init.hostPort,
+            processName: init.processName
         });
     }
 };
--- a/node/index.js
+++ b/node/index.js
@@ -511,6 +511,9 @@ TChannelConnectionBase.prototype.onReqTimeout = function onReqTimeout(op) {
 // stumbles across this object in a core dump.
 TChannelConnectionBase.prototype.resetAll = function resetAll(err) {
     var self = this;
+
+    self.clearTimeoutTimer();
+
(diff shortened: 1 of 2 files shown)


jcorbin:
cc @Raynos especially on the logtron point

node/index.js
```diff
@@ -517,20 +514,21 @@ TChannelConnectionBase.prototype.resetAll = function resetAll(err) {
...
-        error: err,
-        remoteName: self.remoteName,
-        localName: self.channel.hostPort,
-        numInOps: inOpKeys.length,
-        numOutOps: outOpKeys.length,
-        inPending: self.inPending,
-        outPending: self.outPending
-    });
+    if (err) {
+        self.logger[err ? 'warn' : 'info']('resetting connection', {
```
Raynos: is this err ? redundant now ?

node/index.js
```diff
@@ -677,7 +675,7 @@ function TChannelConnection(channel, socket, direction, remoteAddr) {
         self.onSocketErr(err);
     });
     self.socket.on('close', function onSocketClose() {
-        self.onSocketErr(new Error('socket closed')); // TODO typed error
```
Raynos: consider removing onSocketErr() ?

Raynos:
This won't quite work. You have to do something about `op.req.emit('error', err);`

The error used in `resetAll()` was used to pass to all pending outgoing operations to tell it why the outgoing pending operation failed (i.e. socket closed or shutdown from quit).

Raynos:
As for logtron; I've long thought of a few features

 - throw on `logger.error()`. error() goes to sentry and everything in sentry is a bug; there should be no bugs in tests.
 - a mode to display `logger.warn()` callsites

I actually like the notion of debug-logtron being silent by default; but do want a way of seeing warnings.

What would probably be nice is two things; a boolean to logWarnings to the constructor and another environment variable. 

We would also need a boolean to downgrade thrown errors for `logger.error()` to log statements as we are transitioning to the new logtron.

jcorbin:
Updated @Raynos, pretty sure this should work per your comments.

jcorbin:
Slipped some more cleanups into this branch from why_slow, let me know if you'd rather them split out for clearer review or not.

node/index.js
```diff
@@ -47,6 +47,12 @@ var EndpointHandler = require('./endpoint-handler.js');
 var DEFAULT_OUTGOING_REQ_TIMEOUT = 2000;
 var dumpEnabled = /\btchannel_dump\b/.test(process.env.NODE_DEBUG || '');
 
+var SocketClosedError = TypedError({
+    type: 'socket-closed',
```
Raynos: suggestion: prefix type with `tchannel.`

node/index.js
```diff
@@ -392,13 +395,7 @@ inherits(TChannelConnectionBase, EventEmitter);
...
 TChannelConnectionBase.prototype.close = function close(callback) {
     var self = this;
-    self.clearTimeoutTimer();
-    self.logger.debug('destroy channel for', {
-        direction: self.direction,
-        peerRemoteAddr: self.remoteAddr,
-        peerRemoteName: self.remoteName
-    });
-    self.resetAll(new Error('shutdown from quit')); // TODO typed error
+    self.resetAll(SocketClosedError('local close'));
```
Raynos: Did you mean `SocketClosedError({ reason: 'local close' })`

jcorbin:
fixed

Raynos:
lgtm.

PR #329, opened 2015-04-06
Drop Constructor Conveniences

jcorbin:
Drop the "optional new" pattern from "hot" objects:
- in/out X req/res
- all frame parts
- bufrw result objects

This gains us around 6%-10% on our synthetic benchmark, in addition to just making things "simpler".


kriskowal:
:ok:

Raynos:
lgtm.

PR #331, opened 2015-04-06
Drop use of node core streams from socket -> handler -> socket path

jcorbin:
This is between a 20% and 30% speedup on our synthetic benchmark.

At the cost of dropping `NODE_DEBUG=tchannel_dump` convenience for now.

cc @Raynos @kriskowal

### DIFF
--- a/node/index.js
+++ b/node/index.js
@@ -32,7 +32,7 @@ var TypedError = require('error/typed');
 var WrappedError = require('error/wrapped');
 var extend = require('xtend');
 var bufrw = require('bufrw');
-var ChunkReader = require('bufrw/stream/chunk_reader');
+var ReadMachine = require('bufrw/stream/read_machine');
 var reqres = require('./reqres');
 
 var inherits = require('util').inherits;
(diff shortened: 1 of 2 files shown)


node/index.js
```diff
@@ -682,34 +684,43 @@ TChannelConnection.prototype.setupHandler = function setupHandler() {
...
+    function handleReadFrame(frame) {
         if (!self.closing) {
             self.lastTimeoutTime = 0;
         }
+        self.handler.handleFrame(frame, hanldedFrame);
     }
 
-    function onReaderError(err) {
-        self.onReaderError(err);
+    function hanldedFrame(err) {
```
kriskowal: mis-spell

kriskowal:
Great!

node/index.js
```diff
@@ -785,21 +790,6 @@ TChannelConnection.prototype.close = function close(callback) {
     self.socket.destroy();
 };
 
-TChannelConnection.prototype.onReaderError = function onReaderError(err) {
```
Raynos: this method is still used.

node/index.js
```diff
@@ -739,12 +750,6 @@ TChannelConnection.prototype.setupHandler = function setupHandler() {
         // TODO: should terminate corresponding inc res
     }
 
-    function onHandlerError(err) {
```
Raynos: function still used.

Raynos: I see; this function moved. fair enough.

Raynos:
seems legit.

PR #333, opened 2015-04-06
Refactor args fragmenting

jcorbin:
This one has something for everyone:
- [x] it's faster: ~30% on synthetic benchmark
- [x] it's simpler: 2 similar but slightly different splitArgs -> one writeFragmentInto
- [x] rather than synchronously compute all fragments, it's done one-at-a-time driver by _sendCallBodies; future work could defer the loop if useful
- [x] it's less coupled: we could now say "actually let's only send 32KiB or 16KiB" by capping the write buffer allocation; then argsrw would then naturally fragment at end of buffer

### DIFF
--- a/node/v2/args.js
+++ b/node/v2/args.js
@@ -23,6 +23,7 @@
 var TypedError = require('error/typed');
 var inherits = require('util').inherits;
 var bufrw = require('bufrw');
+var Checksum = require('./checksum');
 
 var LengthResult = bufrw.LengthResult;
 var WriteResult = bufrw.WriteResult;
@@ -50,6 +51,10 @@ ArgsRW.prototype.byteLength = function byteLength(body) {
     var length = 0;
     var res;
 
+    res = Checksum.RW.byteLength(body.csum);
+    if (res.err) return res;
+    length += res.length;
+
     if (body.args === null) {
         return LengthResult.just(length);
     }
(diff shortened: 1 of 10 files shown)


jcorbin:
Hmm, looks like tests are broken, needs more work.

jcorbin:
Okay fixed bug, please review the increase in complexity of v2.Frame.RW against how I plan to completely shift off bufrw.StructRW ala https://github.com/uber/tchannel/compare/custom_rws

node/v2/args.js
```diff
@@ -23,6 +23,12 @@
 var TypedError = require('error/typed');
 var inherits = require('util').inherits;
 var bufrw = require('bufrw');
+var Checksum = require('./checksum');
+
+var Flags;
+process.nextTick(function() {
+    Flags = require('./index').CallFlags;
+});
```
kriskowal: Would it be less smelly to just move this import to the bottom of the file, or even simply lazy-bind the index module?

node/v2/args.js
```diff
@@ -45,48 +51,117 @@ function ArgsRW(argrw) {
...
+    var lenres = Checksum.RW.byteLength(body.csum);
+    if (lenres.err) return WriteResult.error(lenres.err);
+    offset += lenres.length;
+
+    if (body.cont === null) {
+        res = self.writeFragmentInto(body, buffer, offset);
         if (res.err) return res;
         offset = res.offset;
+    } else {
+        // assume that something else already did the fragmentatino correctly
```
kriskowal: italian?

kriskowal:
:ok:

Raynos:
lgtm.

PR #334, opened 2015-04-06
Wait until terminal width to open REPL session and upgrade replr

robskillington:


node/examples/repl_server.js
```diff
@@ -89,19 +91,24 @@ function ReplSession(repler, options) {
...
+    });
 };
 
 ReplSession.prototype.control = function control(stream) {
     var self = this;
-    stream.on('data', function(obj) {
+    stream.on('data', function onControlData(obj) {
         if (obj.op === 'resize') {
             self.logger.info('resize to', obj);
+            self.emit('resize', obj);
```
jcorbin: I don't quite agree with the overly-loose coupling here, can we just have some `self._startReplr' utility method that gets called directly?

To be fair, I'm coming off a round of performance tuning whence EventEmitter (especially .once) showed to be abundantly slow, and this isn't a hot path... so I can live with this if you prefer.

robskillington: Sure thing, I'll refactor now =]

robskillington: Hmm, it makes the whole thing a little more stateful and cluttered, having to use a boolean and reset the boolean to only open it on the first resize.  Considering this is REPL and is invoked only when in use perhaps this setup is okay?

jcorbin: works for me, land at will

PR #335, opened 2015-04-06
v2/frame: switch to a custom RW

jcorbin:
As promised by #333:
- marginally faster
- simpler / easier to understand


kriskowal:
:+1:

PR #337, opened 2015-04-06
Custom tracing rw

jcorbin:
Another 5-10% synthetic benchmark boost


kriskowal:
:ok:

jcorbin:
Abandoning for now, makes tests too hard to fix at present for only marginal gain.

PR #344, opened 2015-04-08
[python] Track and re-use inbound connections.

abhinav:
Tests pending

CC @breerly @blampe @junchaowu


blampe:
cool

junchaowu:
lgtm

python/examples/tchannel_server.py
```diff
@@ -35,8 +35,8 @@ def main():  # pragma: no cover
 
     handler = get_example_handler()
 
-    server = client.host(args.port, handler)
-    server.listen()
+    server = client.host(handler)
+    server.listen(args.port)
```
breerly: Love this API change. Can you not specify the hostname?

python/requirements-test.txt
```diff
@@ -26,4 +26,5 @@ thrift==0.9.2
 pyfarmhash
 
 # Mocks
+mock
```
breerly: mock and doubles?

python/tchannel/dispatch.py
```diff
@@ -61,7 +61,7 @@ def handle_call(self, message_id, call, connection):
         endpoint = self.endpoints.get(request.endpoint, None)
 
         if endpoint is None:
-            return connection.send_error(
+            return connection.senderror(
```
breerly: Why did we `s/send_error/senderror/`?

python/requirements-test.txt
abhinav: I like to live dangerously.

(It's not actually being used anymore. Removed.)

python/tchannel/frame.py
```diff
@@ -63,7 +60,6 @@ def read(self, stream, size=None):
         header_body, payload = body[:header_width], body[header_width:]
 
         header = self.header_rw.read(BytesIO(header_body))
-        log.debug('decode frame for message %s', header.message_id)
```
breerly: Were these not helpful?

python/tchannel/tornado/connection.py
```diff
@@ -26,7 +26,12 @@
 import sys
 
 import tornado.gen
-import tornado.ioloop
+import tornado.iostream
+
+try:
+    import tornado.queues as queues  # included in 4.2
```
breerly: :+1:

python/examples/tchannel_server.py
abhinav: The hostname for peers that can connect to you? It should be trivial to add that later.

python/tchannel/tornado/connection.py
```diff
@@ -44,51 +51,97 @@
...
+    used depending on the direction of the connection.
+
+    ``initiate_handshake``
+        Perform a handshake with the remote host.
+    ``expect_handshake``
+        Expect a handshake request from the remote host.
+    """
+
+    # Message ID 0xffffffff is reserved
+    MAX_MESSAGE_ID = 0xfffffffe
```
breerly: What do you think about moving all constants to single module so that it can be shared between all implementations, not just tornado. Suggest `tchannel.glossary.MAX_MESSAGE_ID`.

python/tchannel/frame.py
abhinav: Too noisy. Do you want them back?

python/tchannel/tornado/connection.py
```diff
@@ -44,51 +51,97 @@
...
+    CALL_TYPES = frozenset([Types.CALL_REQ, Types.CALL_REQ_CONTINUE])
 
     def __init__(self, connection):
-        self.connection = connection
+        assert connection
```
breerly: If there isn't a way to add a helpful message to the `AssertionError` then lets use a regular exception here.

python/tchannel/frame.py
junchaowu: for debug purpose we want them

python/tchannel/tornado/connection.py
abhinav: Added a message.

Also, the trace will usually include the line at which the assertion failed so it's generally obvious for None checks why the assertion failed.

python/tchannel/tornado/connection.py
abhinav: Moved

PR #345, opened 2015-04-08
put all the legacy tchannel-as-x in here

Raynos:
This is step one to having a proper as interface. Putting
the legacy code here instead of copied between autobahn
and autobahn-node-client.

Next up are:

 - defining the as json semantics in markdown
 - defining the as node interface in markdown
 - re-implementing the as handlers.

r: @jcorbin @kriskowal


kriskowal:
:ok:

PR #346, opened 2015-04-08
document json over tchannel

Raynos:
This adds a small markdown document that documents what
the semantics of the arg scheme is for `json`.

r: @kriskowal @jcorbin

### DIFF
--- a/docs/json.md
+++ b/docs/json.md
@@ -1,0 +1,50 @@
+# JSON over TChannel
+
+This document outlines how we encode JSON over TChannel.
+
+For JSON call requests the `as` (arg scheme) transport header
+must be set to `json`. Requests will be made with `call req`
+messages and responses will be sent using `call res` messages,
+with values for `arg1`, `arg2` and `arg3` as defined in
+[Arguments][].
+
+For each `call req`, the service name (`service~1`) should be
+set to the TChannel service being called.
+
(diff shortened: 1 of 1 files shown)


docs/json.md
```diff
@@ -0,0 +1,50 @@
...
+
+This document outlines how we encode JSON over TChannel.
+
+For JSON call requests the `as` (arg scheme) transport header
+must be set to `json`. Requests will be made with `call req`
+messages and responses will be sent using `call res` messages,
+with values for `arg1`, `arg2` and `arg3` as defined in
+[Arguments][].
+
+For each `call req`, the service name (`service~1`) should be
```
kriskowal: Service name should be general, not specific to its k shard.

kriskowal:
:ok:

docs/json.md
Raynos: ?

kriskowal: ```For each `call req`, the service name (for example, `service`, not `service~1`) should be```

Raynos: `service~1` is a reference to the ~1 syntax of the binary protocol. i.e. 1byte length

kriskowal: Ah, to steer the next person away from my incorrect assumption, perhaps say

> service~1 (a 1 byte length-prefixed utf-8 service name string)

PR #347, opened 2015-04-08
Tracing agent rework

rf:
Just redid this on top of master rather than attempting the rebase


jcorbin:
:100: let's get it on master and go forward

Raynos:
lgtm.

PR #349, opened 2015-04-09
TChannel as Thrift

kriskowal:


node/as/thrift.js
```diff
@@ -0,0 +1,110 @@
...
+
+    function handleResponse(err, res) {
+        if (err) {
+            return callback(err);
+        }
+
+        var inBodyBuffer = res.arg3;
+
+        if (!res.ok) {
+            return callback(new Error(inBodyBuffer.toString('utf8')), null, null);
```
Raynos: need to support error structs. This will be junk.

node/as/thrift.js
```diff
@@ -0,0 +1,110 @@
...
+        var inBody = inBodyResult.value;
+
+        handle(inHeadBuffer, inBody, opts, handleThriftResponse);
+
+        function handleThriftResponse(error, outHeadBuffer, outBody) {
+            if (error) {
+                return onError(error);
+            }
+
+            // This will throw locally if the response body is malformed.
```
Raynos: we need to support error structs.

kriskowal:
r @raynos @jcorbin

node/as/thrift.js
```diff
@@ -0,0 +1,171 @@
...
+
+'use strict';
+
+var assert = require('assert');
+
+module.exports = TChannelAsThrift;
+
+function TChannelAsThrift(opts) {
+    var self = this;
+    assert(opts.spec, 'TChannelAsThrift expected spec');
```
jcorbin: opts &&

node/as/thrift.js
```diff
@@ -0,0 +1,171 @@
...
+function register(channel, name, opts, handle) {
+    var self = this;
+
+    var argsName = name + '_args';
+    var argsType = self.spec.getType(argsName);
+
+    var returnName = name + '_result';
+    var resultType = self.spec.getType(returnName);
+
+    channel.handler.register(name, handleThriftRequest);
```
jcorbin: future safe by adding a return here?

node/as/thrift.js
```diff
@@ -0,0 +1,171 @@
...
+        var inBody = inBodyResult.value;
+
+        // TODO process inHeadBuffer into inHead
+        var inHead = null;
+
+        handle(opts, inHead, inBody, handleThriftResponse);
+
+        function handleThriftResponse(error, response) {
+            if (error) {
+                return res.sendError('UnexpectedError', error.message);
```
jcorbin: this needs to be different:
- the case above that calls it for failure to parse body should be sending a 'BadRequest' error
- if this callback can get called with non malformed-request errors, then yes there should be a default to unexpected error

node/as/thrift.js
```diff
@@ -0,0 +1,171 @@
...
+            }
+
+            // TODO {head,body} or {arg2,arg3}?
+
+            var ok = response.ok;
+            var outBody = response.body;
+
+            if (typeof ok !== 'boolean') {
+                throw new Error('Expected true or false boolean on response object');
```
jcorbin: dislike... should at least send an unexpected error frame back to caller for faster failure in the happy path.

really why need we care so hard, just test it for truthiness and move on... seems a heavy cost (function that throws) for dubious gain

node/as/thrift.js
```diff
@@ -0,0 +1,171 @@
...
+            }
+
+            var outResult;
```
jcorbin: how about initialize it here..

node/as/thrift.js
```diff
@@ -0,0 +1,171 @@
...
+            if (ok) {
+                outResult = {success: outBody};
```
jcorbin: then mutate .success here

node/as/thrift.js
```diff
@@ -0,0 +1,171 @@
...
+            } else {
+                if (!outBody) {
```
jcorbin: ..then you can flatten this else/if ladder one level

node/as/thrift.js
```diff
@@ -0,0 +1,171 @@
...
+                    throw new Error('Error body required in the not ok response case'); // TODO TypedError
+                } else if (typeof outBody.nameAsThrift !== 'string') {
+                    throw new Error('Can\'t serialize error response that lacks nameAsThrift'); // TODO TypedError
```
jcorbin: again I dislike throwing for all these cases... misbehaving handlers shouldn't bring down an entire application, especially in the compound or aggregate application contexts

node/as/thrift.js
```diff
@@ -0,0 +1,171 @@
...
+                }
+                outResult = {};
+                outResult[outBody.nameAsThrift] = outBody;
+            }
+
+            // outBody must be a Thrift result, e.g., {success: value}, or
+            // {oops: {}}.
+
+            // This will throw locally if the response body is malformed.
+            var outBodyBuffer = resultType.toBuffer(outResult).toValue();
```
jcorbin: toValue throws, use result with explicit error handling as above

node/as/thrift.js
```diff
@@ -0,0 +1,171 @@
...
+    var resultType = self.spec.getType(endpoint + '_result');
+
+    // This will throw locally if the body is malformed.
+    var outBodyBuffer = argsType.toBuffer(outBody).toValue();
+
+    // TODO outHeadBuffer from outHead
+    var outHeadBuffer = null;
+
+    // Punch as=thrift into the transport headers
+    request.headers.as = "thrift";
```
jcorbin: this is cool, why not check for it above before even trying to parse?

node/as/thrift.js
```diff
@@ -0,0 +1,171 @@
...
+
+    request.send(endpoint, outHeadBuffer, outBodyBuffer, handleResponse);
+
+    function handleResponse(err, res) {
```
jcorbin: this is (err, res, arg2, arg3), stop reaching into res for arg2 and arg3 below

node/as/thrift.js
```diff
@@ -0,0 +1,171 @@
...
+        if (err) {
+            return callback(err);
+        }
+
+        var inBodyBuffer = res.arg3;
+
+        var inBodyResult = resultType.fromBuffer(inBodyBuffer);
+        if (inBodyResult.error) {
```
jcorbin: .err now, also // TODO wrapped error?

node/as/thrift.js
```diff
@@ -0,0 +1,171 @@
...
+    var returnName = name + '_result';
+    var resultType = self.spec.getType(returnName);
+
+    channel.handler.register(name, handleThriftRequest);
+
+    function handleThriftRequest(req, res, inHeadBuffer, inBodyBuffer) {
+
+        // Process incoming thrift body
+        var inBodyResult = argsType.fromBuffer(inBodyBuffer);
+        if (inBodyResult.error) {
```
jcorbin: .err

node/as/thrift.js
```diff
@@ -0,0 +1,171 @@
...
+
+        var response = new Response(res.ok, inHead, inBody);
+
+        callback(null, response);
+    }
+
+};
+
+function Response(ok, head, body) {
+    this.ok = ok;
```
jcorbin: s/this/self/
y

node/as/thrift.js
Raynos: @jcorbin 

I'm calling `channel.handler.register()` as well. This allows it to work cleanly with custom handlers; currently autobahn & ringpop use the default handler.

`channel.register()` is deprecated as far as I understand; if we make `channel.register()` an alias for `channel.handler.register()` then we might as well call handler.register() here.

node/as/thrift.js
Raynos: :+1: BadRequest for thrift parsing; UnexpectedError for any `cb(err)` in the application developers handler function.

node/test/as-thrift.js
```diff
@@ -0,0 +1,111 @@
...
+    });
+
+    var spec = thriftify.readSpecSync(path.join(__dirname, 'anechoic-chamber.thrift'));
+    var tchannelAsThrift = new TChannelAsThrift({spec: spec});
+    tchannelAsThrift.register(server, 'Chamber::echo', null, echo);
+
+    server.listen(0, '127.0.0.1', function () {
+        var request = client.request({host: server.hostPort});
+
+        tchannelAsThrift.send(request, 'Chamber::echo', null, {value: 10}, handleResponse);
```
jcorbin: can we encourage the idiomatic form:
```
tchannelAsFoo.send(channel.request({
  // yer options
}), /* args */, callback);
```

there's no need to locally bind the request object here that I can see.

node/as/thrift.js
Raynos: If the application developer responds with `ok: undefined` then it's simply ambigious whether its true or false.

Better to fail hard on what is a programmer error ( the application developer did not choose whether the call response is ok ) then to have undefined behaviour.

node/as/thrift.js
kriskowal: I’m in favor of accepting a "registerable" here, regardless.

node/as/thrift.js
kriskowal: All of these correspond to programmer errors.

node/as/thrift.js
Raynos: :+1: for checking in the handler.

node/as/thrift.js
Raynos: :+1:

node/as/thrift.js
Raynos: :+1: wrapped errors.

kriskowal:
Comments addressed except we intend to continue crashing on programmer errors including malformed responses.

Raynos:
GOGOGO SHIPIT. #lgtm

PR #350, opened 2015-04-09
proposed as json interface

Raynos:
I've tried to keep this minimal. I really liked
@jcorbin idea of passing in `tchannel.request()`
on the way out.

We should have a back and forth on this interface,
this is just a first pass. Then we can plow through
json and tchannel implementations.

r: @jcorbin @kriskowal


node/as/README.md
```diff
@@ -0,0 +1,215 @@
...
+```
+
+The `as` client interface can be made simpler then this.
+
+The proposed interface is modeled after `request/request` which
+is a popular HTTP client with a simpler interface.
+
+```js
+var TChannelJSON = require('tchannel/as/json');
+var tchannelJSON = TChannelJSON({
```
jcorbin: can we please not have to instantiate a thing? all this injected stuff can be had off the channel.... if req.channel isn't yet a thing, then it can be made so

node/as/README.md
```diff
@@ -0,0 +1,215 @@
...
+    logger: ...
+});
+
+tchannelJSON.send(tchannel.request({
+    timeout: 1000,
```
jcorbin: there's been a TODO in the code about renaming this to be ttl through the edge, thoughts?

node/as/README.md
```diff
@@ -0,0 +1,215 @@
...
+    serviceName; 'foo'
```
jcorbin: s/serviceName/service/ and s/;/:/

node/as/README.md
Raynos: we need an instantiate thing; at least for thrift; because spec files.

If we can get the logger out of req.channel then sure; cool.

node/as/README.md
Raynos: dont care either way; just consistent please.

node/as/README.md
Raynos: lets do that consistently everywhere;

node/as/README.md
kriskowal: timeout is idiomatic here.

node/as/README.md
kriskowal: :ok:

node/as/README.md
```diff
@@ -0,0 +1,217 @@
...
+
+            console.log('resp ok', {
+                arg2: Encoder.parse(resp.arg2),
+                arg3: Encoder.parse(resp.arg3)
+            });
+        }
+    );
+```
+
+The `as` client interface can be made simpler then this.
```
kriskowal: then/than

node/as/README.md
```diff
@@ -0,0 +1,217 @@
...
+                console.log('got error', {
+                    error: err
+                })
+                return;
+            }
+
+            if (!resp.ok) {
+                console.log('resp not ok', {
+                    arg2: Encoder.parse(resp.arg2),
+                    arg3: Encoder.parse(resp.arg3)
```
kriskowal: s/Encoder/Protocol or Encoding. Encoder would only have stringify, with a corresponding Decoder.parse.

node/as/README.md
```diff
@@ -0,0 +1,217 @@
...
+
+To make requests to a `json` or `thrift` service using the
+tchannel request interface looks like.
+
+```js
+tchannel
+    .request({
+        serviceName: 'foo',
+        timeout: 1000
+    })
```
kriskowal: Where do we thread the protocol level "as" header through?

node/as/README.md
kriskowal: We could technically thread the `spec` through createHandler and send functions as options or arguments, but I found it convenient to have the object around to effectively partially apply.

jcorbin:
also, this all looks sensible enough modulo shed, but how about the handler registration side?

Raynos:
@jcorbin theres an entire section called `## as handle interface` for handler registration side.

Raynos:
@jcorbin @kriskowal I updated this document based on what we talked about.

Please fact check me and get in any feedback; would love to ship as/json & as/thrift tomorrow :)

node/as/Overview.md
```diff
@@ -0,0 +1,137 @@
...
+The proposed interface is as follows:
+
+```js
+var TChannelJSON = require('tchannel/as/json');
+var tchannelJSON = TChannelJSON({
+    logger: ...
+});
+
+var handler = tchannelJSON.register(tchannel, 'arg1', opts,
+    function foo(opts, req, head, body, cb) {
```
kriskowal: @jcorbin’s argument that opts should come first because it serves as a "self" bag has merit. However, it is weird to receive it before arg{2,3} when it is sent after arg{1}. "opts" is coming in from our experience with sirvice and I think this goes against the grain.

kriskowal: And I should have said that while we were standing. Sorry.

Raynos: what if we do `.register(tchannel, opts, 'arg1', fn)`. I kind of like it being third.

kriskowal: `.registerWithContext(tchannel, opts, arg1, fn)` and `.register(tchannel, arg1, fn)`?

kriskowal: We appear to be in agreement face to face with `register(tchannel, opts, arg1, fn)` and `registerWithoutContext(tchannel, arg1, fn)`

jcorbin: :+1:

jcorbin:
lgtm

PR #354, opened 2015-04-09
Drop synchronous API

abhinav:
We're not using it and it's only slowing us down. We can add it back later if we decide we need it.

@blampe @junchaowu


junchaowu:
:ship:

PR #357, opened 2015-04-09
Drop debug logging

jcorbin:


Raynos:
Can has statsd instead ? 

lgtm.

PR #360, opened 2015-04-10
top level requests should error if no host.

Raynos:
Assuming that the top level tchannel only contains peers
for a single `service` is a bad assumption.

We add incoming peers to the top level tchannel so if
its a server we would randomly select incoming peers.

r: @rf @jcorbin @kriskowal

### DIFF
--- a/node/channel.js
+++ b/node/channel.js
@@ -372,6 +372,10 @@ TChannel.prototype.request = function channelRequest(options) {
         }
     }
 
+    if (!self.serviceName && !options.host) {
+        throw errors.TopLevelRequestError();
+    }
+
     // TODO: moar defaults
     if (self.destroyed) {
         throw new Error('cannot request() to destroyed tchannel'); // TODO typed error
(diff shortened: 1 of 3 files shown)


rf:
:+1:

node/index.js
```diff
@@ -367,6 +367,11 @@ TChannel.prototype.request = function channelRequest(options) {
     if (!options.service && self.serviceName) {
         options.service = self.serviceName;
     }
+
+    if (!self.topChannel && !options.host) {
```
jcorbin: Let's shift this to be slightly different please:
- key on `!self.serviceName` not `!self.topChannel`
- rename the error to taste if you want to

Semantically, what you really mean is "you must specify a host if this is a generic cross-service channel", however this is a potential use case:
```
var channel = TChannel({
  serviceName: "foo"
});
```

In which case there are no top/sub channels, but the channel is explicitly single-serviced.

I'd like to try to have strict *bahn clients use this mode unless there's a compelling reason not to.

PR #363, opened 2015-04-10
Use ref counting for trace agent

jcorbin:
- the trace is a singleton
- however its use in TChannel currently assumes that there will only ever be one TChannel...
- ...and that you never try to use another (traced) channel in the future after closing that first channel

So instead this diff shifts the singleton to:
- use reference counting
- add the async listener whenever ref count goes > 0
- remove the async listener whenever ref count goes to 0

r @Raynos @rf


rf:
+1

PR #368, opened 2015-04-10
Improve host req

jcorbin:
Fix typo from #360 and provide convenience so that users can specify a service option in lieu of a host.

r @Raynos, @kriskowal


Raynos:
noice. lgtm.

PR #369, opened 2015-04-10
Correctly handle errors for incoming requests

jcorbin:
- firstly, finish handling a server op when we send an error frame
- secondly, hookup req error events to send an error frame

r @kriskowal @Raynos

### DIFF
--- a/node/connection_base.js
+++ b/node/connection_base.js
@@ -292,7 +292,9 @@ TChannelConnectionBase.prototype.handleCallRequest = function handleCallRequest(
 
     function buildResponse(options) {
         if (op.res && op.res.state !== OutgoingResponse.States.Initial) {
-            throw new Error('response already built and started'); // TODO: typed error
+            throw errors.ResponseAlreadyStarted({
+                state: op.res.state
+            });
         }
         op.res = self.buildOutgoingResponse(req, options);
         op.res.on('finish', opDone);
--- a/node/connection_base.js
+++ b/node/connection_base.js
@@ -269,10 +269,17 @@ TChannelConnectionBase.prototype.handleCallRequest = function handleCallRequest(
     self.inPending++;
     var op = self.inOps[id] = new TChannelServerOp(self, self.timers.now(), req);
     var done = false;
+    req.on('error', onReqError);
     process.nextTick(runHandler);
 
     if (req.span) {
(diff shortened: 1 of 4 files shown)


jcorbin:
Depends on #370, review it first.

node/connection_base.js
```diff
@@ -269,12 +269,19 @@ TChannelConnectionBase.prototype.handleCallRequest = function handleCallRequest(
...
     process.nextTick(runHandler);
 
     if (req.span) {
         req.span.endpoint.serviceName = self.channel.serviceName;
     }
 
+    function onReqError(err) {
+        if (!op.res) buildResponse();
+        var errName = err.name || err.constructor.name;
+        op.res.sendError('UnexpectedError', errName + ': ' + err.message);
```
kriskowal: I must have misread the source. This is not supposed to be `'unexpected error'`?

node/connection_base.js
```diff
@@ -285,10 +292,13 @@ TChannelConnectionBase.prototype.handleCallRequest = function handleCallRequest(
...
     function buildResponse(options) {
         if (op.res && op.res.state !== OutgoingResponse.States.Initial) {
-            throw new Error('response already built and started'); // TODO: typed error
+            throw errors.ResponseAlreadyStarted({
+                state: op.res.state
+            });
         }
         op.res = self.buildOutgoingResponse(req, options);
         op.res.on('finish', opDone);
+        op.res.on('errored', opDone);
```
kriskowal: Avoiding a collision with 'error'?

node/connection_base.js
jcorbin: why not? we don't know anything about the error at this point, in the future we'll add some typed cases.

e.g. I've got one in queue to add timeout -> timed out coupling

node/connection_base.js
jcorbin: correct, it's simply noting that the response is over due to an error response... bit hack perhaps

kriskowal: :heavy_check_mark:

node/connection_base.js
kriskowal: Over the shoulder, 'UnexpectedError' it is. :heavy_check_mark:

kriskowal:
:ok:

PR #370, opened 2015-04-10
Fix self handled request error emission

jcorbin:
This matches semantics with the normal socket-borne path which emits error frame errors on the client side (aka outgoing) req.

r @kriskowal @Raynos


kriskowal:
:ok:

PR #374, opened 2015-04-11
Dedupe and better encapsulate timeout checknig

jcorbin:
- dedupe the similar but needlessly different paths
- shift timeout checking and state tracking concerns into the req objcets

r @kriskowal @Raynos


Raynos:
lgtm.

PR #375, opened 2015-04-11
Drop ops

jcorbin:
Ops have been redundant for a long time.

Now that I've drained them of all usefulness (timeouts), just drop them entirely.

Depends on #374 

r @kriskowal @Raynos


kriskowal:
Apart from the failing CI, this looks good to me. Did this get subsumed elsewhere?

PR #377, opened 2015-04-12
Trace default; better handling of service name

rf:
Defaults trace to true for outgoing requests when the tracer is loaded.

Added a `service` property to the tracer to be used as the service name on incoming requests. This is for strange cases like the service router where the service name specified on the request itself isn't the one we want in the span.

### DIFF
--- a/node/incoming_request.js
+++ b/node/incoming_request.js
@@ -80,9 +80,9 @@ function TChannelIncomingRequest(id, options) {
 
             // If a service hasn't been specified on the tracer, use the 
             // service on the incoming request. This is to handle the
-            // pathological case of the service router, which has a different
-            // service name than the one specified in the incoming request.
-            service: options.tracer.service || self.service,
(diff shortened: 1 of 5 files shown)


jcorbin:
:+1:

node/incoming_request.js
```diff
@@ -77,7 +77,12 @@ function TChannelIncomingRequest(id, options) {
             parentid: self.tracing.parentid,
             flags: options.tracer.forceTrace? 1 : self.tracing.flags,
             hostPort: options.hostPort,
-            serviceName: '', // the service in options.service is not what we want
+
+            // If a service hasn't been specified on the tracer, use the 
+            // service on the incoming request. This is to handle the
+            // pathological case of the service router, which has a different
+            // service name than the one specified in the incoming request.
+            service: options.tracer.service || self.service,
```
kriskowal: I believe we are converging on "serviceName" over "service".

PR #381, opened 2015-04-13
Streaming

junchaowu:
cc @blampe @abhinav @breerly


python/tchannel/dispatch.py
```diff
@@ -56,21 +56,32 @@ def _call_endpoint(self, endpoint, request, response):
...
+        # user still tries read from it, it will return empty.
+        chunk = yield request.argstreams[0].read()
+        while chunk:
+            request.endpoint += chunk
+            chunk = yield request.argstreams[0].read()
+
+        endpoint = self.endpoints.get(request.endpoint, None)
         if endpoint is None:
-            return connection.senderror(
+            connection.senderror(
```
abhinav: I would leave the `return`s here. That way if `senderror` or `call_endpoint` return a future (which they most likely do), that's propagated up to the caller.

python/tchannel/io.py
```diff
@@ -19,8 +19,169 @@
...
+
+from .messages import common
+
+
+class BaseArgStream(object):
+
+    def __init__(self):
+        """Base Arg Stream Class
+
+        _state is used to track the state of the arg stream.
```
abhinav: Don't need to mention this. It's obvious from the name. Also, `_state` is not part of the "public" interface of this class so it shouldn't be something exposed in the docstring.

python/tchannel/io.py
```diff
@@ -19,8 +19,169 @@
...
+
+        When object is created, it is in the "init" state.
+
+        When write func is called, it will transformed to "streaming" state.
```
abhinav: I appreciate the thought behind documenting this, but implementation details like this, if documented, should use comments (`#`), not docstrings.

Things part of the object's behavior may be included in the docstring without going into implementation details. Like, "Once the stream is exhausted, it will return an empty string on reads."

python/tchannel/io.py
```diff
@@ -19,8 +19,169 @@
...
+        """Async write to internal stream buffer
+
+        :return if stream has been closed, it will raise StreamingException
+        """
+        raise NotImplementedError()
+
+    def close(self):
+        """Close the stream
+
+        It updates the _state to completed.
```
abhinav: Again, state changes should not be in the docstring. Also, why does this class declare the state if it doesn't read or write to it at all. Maybe leave it up to the implementations to decide if they want to track state using StreamState? Some may decide they have more complicated needs and use something else to track state.

python/tchannel/io.py
```diff
@@ -19,8 +19,169 @@
...
+        """Async read from internal stream buffer
+
+        if it reaches the end of the stream, it will return empty(""), caller
+        will depend on the return value to tell if it reaches the end of the
+        stream.
+
+        It doesn't support seek functionality, which means it will only read
+        the stream data once in single direction.
+
+        NOTE: The data which has been read will be discarded in the stream.
```
abhinav: :+1:

python/tchannel/io.py
```diff
@@ -19,8 +19,169 @@
...
+
+        :return:chunk of bytes read from stream
+        """
+        raise NotImplementedError()
+
+    def write(self, chunk):
+        """Async write to internal stream buffer
+
+        :return if stream has been closed, it will raise StreamingException
```
abhinav: Use `:raises StreamingException: if the stream was closed` instead.

python/tchannel/io.py
```diff
@@ -19,8 +19,169 @@
...
+except ImportError:  # pragma: no cover
+    from toro import Condition
+
+from .messages import common
+
+
+class BaseArgStream(object):
+
+    def __init__(self):
+        """Base Arg Stream Class
```
abhinav: These words don't mean anything and don't provide any information that isn't already available from the class name. I would get rid of the whole __init__ docstring.

python/tchannel/io.py
```diff
@@ -19,8 +19,169 @@
...
+
+    def __init__(self, rpipe, wpipe=None, auto=False):
+        """Pipe-based arg stream
+
+        NOTE: if you want to pass a file descriptor of a file, you have to
+        use os.open to get the file descriptor instead of python's open.
+
+        :param rpipe: an integer file descriptor which supports read ops
+        :param wpipe: an integer file descriptor which supports write ops
+        :param auto: flag to indicate to close the stream automatically or not
```
abhinav: Call it `auto_close` or something similar.

python/tchannel/io.py
```diff
@@ -19,8 +19,169 @@
...
+    def read(self):
+        if self._state != StreamState.completed and self._stream == "":
+            yield self._condition.wait()
+
+        chunk = self._stream
+        self._stream = ""
+        raise tornado.gen.Return(chunk)
+
+    @tornado.gen.coroutine
+    def write(self, chunk):
```
abhinav: You can drop the `coroutine` for this one. You're not `yield`ing anything.

python/tchannel/io.py
```diff
@@ -19,8 +19,169 @@
...
+        if self._state == StreamState.completed:
+            raise StreamingException("Stream has been closed.")
+
+        self._stream += chunk
+        self._condition.notify()
+
+    @tornado.gen.coroutine
+    def close(self):
+        self._state = StreamState.completed
+        self._condition.notify()
```
abhinav: Consider `write -> read -> read -> close`. The first `read` will clear `_stream`, and the second one will block until notified. Then `close` will notify it and it will move on to returning an empty string as the result.

python/tchannel/io.py
```diff
@@ -19,8 +19,169 @@
...
+
+
+class PipeArgStream(BaseArgStream):
+
+    def __init__(self, rpipe, wpipe=None, auto=False):
+        """Pipe-based arg stream
+
+        NOTE: if you want to pass a file descriptor of a file, you have to
+        use os.open to get the file descriptor instead of python's open.
```
abhinav: Since this will accept any file descriptor and the problem exists only for `open`-based file descriptors, just mention that for reading from or writing to files, `os.open` must be used instead of `open`. Socket file descriptors and others will be just fine.

python/tchannel/io.py
```diff
@@ -19,8 +19,169 @@
...
+
+    @tornado.gen.coroutine
+    def close(self):
+        self._state = StreamState.completed
+        self._condition.notify()
+
+
+class PipeArgStream(BaseArgStream):
+
+    def __init__(self, rpipe, wpipe=None, auto=False):
```
abhinav: Are both optional or only one?

python/tchannel/io.py
```diff
@@ -19,8 +19,169 @@
...
+        chunk = ""
+        try:
+            chunk = yield self._rs.read_bytes(
+                common.MAX_PAYLOAD_SIZE, partial=True)
+
+        except StreamClosedError:
+            # reach the end of the pipe stream
+            self._state = StreamState.completed
+        finally:
+            raise tornado.gen.Return(chunk)
```
abhinav: Return the value from inside the try block. If you want an empty string returned for StreamClosedError, have a separate `raise Return` inside the `except`. This will return an empty string in case of other exceptions and the system will end up ignoring the actual exceptions.

python/tchannel/io.py
junchaowu: condition.wait needs coroutine

python/tchannel/io.py
```diff
@@ -19,8 +19,169 @@
...
+
+    @tornado.gen.coroutine
+    def write(self, chunk):
+        if self._wpipe is not None:
```
abhinav: Nit: I prefer not having too many unnecessary indentation levels.

If your method consists entirely of this,

    if x:
        do_stuff

Do this:

    if not x:
        return
    do_stuff

python/tchannel/io.py
junchaowu: won't happen. in our case, read() will block the reset read. only one read happen at once

python/tchannel/io.py
junchaowu: only write one is optional. I did this on purpose. from our point of view, we need rpipe to read.

python/tchannel/io.py
junchaowu: done

python/tchannel/io.py
abhinav: The implementation checks `if not None` on both. Assert the presence of the required one.

python/tchannel/dispatch.py
```diff
@@ -56,21 +56,32 @@ def _call_endpoint(self, endpoint, request, response):
...
-        endpoint = self.endpoints.get(request.endpoint, None)
+    @tornado.gen.coroutine
+    def handle_call(self, request, connection):
+        # read arg_1 so that handle_call is able to get the endpoint
+        # name and find the endpoint handler.
+        # the arg_1 value will be store in the request.endpoint field.
 
+        # NOTE: after here, the correct way to access value of arg_1 is through
+        # request.endpoint. The original argstream[0] is no longer valid. If
+        # user still tries read from it, it will return empty.
```
breerly: nit - this comment should probably be in a docblock

python/tchannel/dispatch.py
```diff
@@ -56,21 +56,32 @@ def _call_endpoint(self, endpoint, request, response):
         finally:
             response.finish()
 
-    def handle_call(self, message_id, call, connection):
-        request = Request(message_id, call, connection)
-        endpoint = self.endpoints.get(request.endpoint, None)
+    @tornado.gen.coroutine
```
breerly: code that is coupled to tornado should live in the `tchannel.tornado` package.

python/tchannel/dispatch.py
```diff
@@ -56,21 +56,32 @@ def _call_endpoint(self, endpoint, request, response):
...
+        # user still tries read from it, it will return empty.
+        chunk = yield request.argstreams[0].read()
+        while chunk:
+            request.endpoint += chunk
+            chunk = yield request.argstreams[0].read()
+
+        endpoint = self.endpoints.get(request.endpoint, None)
         if endpoint is None:
-            return connection.senderror(
+            raise tornado.gen.Return(connection.senderror(
```
breerly: This is basically a 404? Is `ErrorCode.bad_request` the right code for this?

python/tchannel/dispatch.py
```diff
@@ -105,58 +116,82 @@ def handler(request, response, opts):
...
 
-        self.endpoint = self.message.args[0]
-        self.header = self.message.args[1]
-        self.body = self.message.args[2]
+    # TODO decide which elements inside "message" object to expose to user.
+    def __init__(
+        self,
+        id=None,
+        flags=FlagsType.none,
+        ttl=10,
```
breerly: wondering if this should be a constant instead

python/tchannel/io.py
```diff
@@ -19,8 +19,154 @@
...
 
 try:
     from cStringIO import StringIO as BytesIO
 except ImportError:  # pragma: no cover
     from io import BytesIO  # noqa
+
+import tornado
+import tornado.gen
+import tornado.ioloop
+from tornado.iostream import PipeIOStream, StreamClosedError
```
breerly: looks like this module should be moved to the `tchannel.tornado` package.

python/tchannel/io.py
```diff
@@ -19,8 +19,154 @@
...
+
+    @tornado.gen.coroutine
+    def write(self, chunk):
+        if self._state == StreamState.completed:
+            raise StreamingException("Stream has been closed.")
+
+        self._stream += chunk
+        self._condition.notify()
+
+    @tornado.gen.coroutine
```
breerly: This method doesn't appear to yield anything.... does `self._confition.notify()` require unwrapping a coroutine or something?

python/tchannel/io.py
```diff
@@ -19,8 +19,154 @@
...
+
+try:
+    from tornado.locks import Condition
+except ImportError:  # pragma: no cover
+    from toro import Condition
+
+from .messages import common
+
+
+class BaseArgStream(object):
```
breerly: `Base*` classes make me really sad.

python/tchannel/messages/common.py
```diff
@@ -32,6 +32,18 @@
 PROTOCOL_VERSION = 0x02
 
 
+# 64KB Max frame size
+# 16B (size:2 | type:1 | reserved:1 | id:4 | reserved:8)
+# 1 2 Bytes can represent 0~2**16-1
+MAX_PAYLOAD_SIZE = 0xFFEF   # 64*1024 - 16 - 1
```
breerly: :+1: to the detailed comments here :)

python/tchannel/tcurl.py
```diff
@@ -31,6 +31,8 @@
 import time
 
 import tornado.ioloop
+from .io import InMemArgStream
+from .util import print_arg
 from .tornado import TChannel
```
breerly: we should move tcurl into the `tchannel.tornado` module as well

breerly:
I'd say move the tornado-specific stuff into `tchannel.tornado` and ship it.

python/tchannel/io.py
junchaowu: discussed offine. solved

python/tchannel/dispatch.py
junchaowu: this is protocol level error. There is a node and go task for this too.

python/tchannel/dispatch.py
junchaowu: for the request/response interface. we will change that in the future.

python/tchannel/io.py
junchaowu: yes.

python/tchannel/io.py
junchaowu: done

python/tchannel/io.py
junchaowu: changed name

python/tchannel/tcurl.py
junchaowu: keep tcurl at current level for now since tcurl is more like a wrapper not library

python/tchannel/tornado/util.py
```diff
@@ -0,0 +1,43 @@
...
+def get_arg(context, index):
+    if index < len(context.argstreams):
+        arg = ""
+        try:
+            chunk = yield context.argstreams[index].read()
+            while chunk:
+                arg += chunk
+                chunk = yield context.argstreams[index].read()
+        finally:
+            raise tornado.gen.Return(arg)
```
abhinav: When returning values in tornado, don't do it from finally blocks.

If any of the previous code threw an exception, this will cause that exception to be completely ignored and always return a response (because of how `raise` works)

python/tchannel/tornado/util.py
```diff
@@ -0,0 +1,43 @@
+from __future__ import absolute_import
+
+import tornado
+import tornado.gen
+from ..exceptions import TChannelException
+
+
+@tornado.gen.coroutine
+def get_arg(context, index):
```
abhinav: Document these methods.

python/tchannel/tornado/util.py
```diff
@@ -0,0 +1,43 @@
...
+        arg = ""
+        try:
+            chunk = yield context.argstreams[index].read()
+            while chunk:
+                arg += chunk
+                chunk = yield context.argstreams[index].read()
+        finally:
+            raise tornado.gen.Return(arg)
+    else:
+        raise TChannelException()
```
abhinav: This isn't a standard operating exception. `index < len(..)` must be an assertion.

python/tchannel/tornado/util.py
```diff
@@ -0,0 +1,43 @@
...
+def get_all_args(context):
+    arg1 = yield get_arg(context, 0)
+    arg2 = yield get_arg(context, 1)
+    arg3 = yield get_arg(context, 2)
+
+    raise tornado.gen.Return((arg1, arg2, arg3))
+
+
+@tornado.gen.coroutine
+def print_arg(request, index):
```
abhinav: Probably remove this if it's not being used. If it is being used, make it more generic -- something that writes the contents of an arg stream to a file-like object instead of just stdout.

python/tchannel/tornado/stream.py
```diff
@@ -0,0 +1,143 @@
...
+
+        except StreamClosedError:
+            # reach the end of the pipe stream
+            self._state = StreamState.completed
+        finally:
+            raise tornado.gen.Return(chunk)
+
+    @tornado.gen.coroutine
+    def write(self, chunk):
+        if self._wpipe is not None:
```
abhinav: This should be an assert instead of a silent failure

python/tchannel/handler.py
```diff
@@ -76,14 +77,33 @@ def handle(self, context, connection):
...
+        It passes all the messages into the message_factory to build the init
+        request object. Only when it get a CallRequestMessage and a completed
+        arg_1=argstream[0], the message_factory will return a request object.
+        Then it will trigger the async call_handle call.
+
+        :param message_id: message id
+        :param message: CallRequestMessage or CallRequestContinueMessage
+        :param connection: tornado connection
+        """
+        req = connection._request_message_factory.build(message_id, message)
```
abhinav: Accessing a protected attribute here.

Maybe this class should have its own message factory? Or perhaps assembly should happen somewhere else?

python/tchannel/tornado/connection.py
```diff
@@ -102,6 +98,9 @@ def __init__(self, connection):
         # handshake has been performed.
         self._loop_running = False
 
+        # response message callback
+        self.response_callback = response_callback
```
abhinav: Document what this is and how it being set affects the behavior of the class.

python/tchannel/tornado/connection.py
```diff
@@ -53,19 +53,15 @@
 
 class TornadoConnection(object):
     """Manages a bi-directional TChannel conversation between two machines.
-
     The following primary bi-directional operations are provided:
-
```
abhinav: Why are you removing all these lines in the docstrings? The general format used for docstrings is:

    One line summary.

    More detailed paragraph
    spanning multiple lines.

    Other paragraphs or sections in
    reStructuredText syntax, often
    separated by empty lines.

Docstrings use the reStructuredText markup syntax (see http://sphinx-doc.org/rest.html and http://docutils.sourceforge.net/docs/ref/rst/restructuredtext.html). Removing these lines breaks the formatting of the [definition lists](http://docutils.sourceforge.net/docs/ref/rst/restructuredtext.html#definition-lists).

python/tchannel/tornado/connection.py
```diff
@@ -405,18 +394,10 @@ def serve(self, handler):
         assert handler, "handler is required"
         assert self._loop_running, "Finish the handshake first"
 
+        connection = conn or self
```
abhinav: If you need to do this, might as well make this a class method and always pass in conn

python/tchannel/tornado/connection.py
```diff
@@ -445,7 +425,7 @@ def senderror(self, code, message, message_id):
 
         self._write(
             ErrorMessage(
-                code=ErrorCode.bad_request,
+                code=code or ErrorCode.bad_request,
```
abhinav: `code` doesn't have a default value. In fact, we explicitly check that it's in the list of accepted error codes 5 lines above. When do you expect it to be unset?

python/tchannel/tornado/dispatch.py
```diff
@@ -20,9 +20,101 @@
...
+            response.write('hello world')
+    """
+
+    def __init__(self):
+        super(RequestDispatcher, self).__init__()
+        self.endpoints = {}
+
+    def _call_endpoint(self, endpoint, request, response):
+        try:
+            endpoint.handler(request, response, endpoint.opts)
```
abhinav: Make this a coroutine and `yield maybe_future` on the response of the handler.

python/tchannel/tornado/connection.py
```diff
@@ -190,50 +190,49 @@ def _loop(self):
...
-                message = self._response_message_factory.build(
+                response = self.response_message_factory.build(
                     context.message_id, context.message
                 )
+                if (context.message.message_type == Types.ERROR or
+                        context.message.flags == FlagsType.none):
+                    future = self._outstanding.pop(context.message_id)
+                else:
+                    # still streaming, keep it for record
+                    future = self._outstanding.get(context.message_id)
```
abhinav: Cool. I think this if-else block needs a little more explanation though. It would be very easy to break this with minor changes.

python/tchannel/tornado/connection.py
```diff
@@ -190,50 +190,49 @@ def _loop(self):
...
                 self._messages.put(context)
                 continue
+
             elif context.message_id in self._outstanding:
-                message = self._response_message_factory.build(
+                response = self.response_message_factory.build(
                     context.message_id, context.message
                 )
+                if (context.message.message_type == Types.ERROR or
+                        context.message.flags == FlagsType.none):
```
abhinav: Instead of checking for just error messages, assume all messages are unfragmented except call req/res and their continuations.

python/tchannel/tornado/connection.py
```diff
@@ -190,20 +190,24 @@ def _loop(self):
...
                     context.message_id, context.message
                 )
 
-                if message is None:
-                    # Message fragment is incomplete. It'll probably be filled
-                    # by a future request.
-                    continue
+                # keep continue message in the list
+                # pop all other type messages including error message
+                if context.message.message_type != FlagsType.none:
```
abhinav: FlagsType and MessageType are different things.

python/tchannel/messages/call_continue.py
```diff
@@ -92,7 +91,6 @@ def fragment(self, space_left, fragment_msg):
         self.args = new_args
         if space_left >= 0 and len(fragment_msg.args) == 0:
             # don't need to fragment any more
-            self.flags = FlagsType.none
```
abhinav: Don't you need the flag set to None in this case?

junchaowu: yes

python/tchannel/tornado/connection.py
```diff
@@ -73,6 +73,7 @@ class TornadoConnection(object):
     """
 
     CALL_TYPES = frozenset([Types.CALL_REQ, Types.CALL_REQ_CONTINUE])
+    CALL_RES_TYPES = frozenset([Types.CALL_RES, Types.CALL_RES_CONTINUE])
```
abhinav: Rename CALL_TYPES to CALL_REQ_TYPES for consistency.

python/tchannel/tornado/connection.py
```diff
@@ -190,20 +191,24 @@ def _loop(self):
...
                     context.message_id, context.message
                 )
 
-                if message is None:
-                    # Message fragment is incomplete. It'll probably be filled
-                    # by a future request.
-                    continue
+                # keep continue message in the list
+                # pop all other type messages including error message
+                if context.message.message_type not in self.CALL_RES_TYPES:
```
abhinav: You need to check the flag here.

python/tchannel/tornado/connection.py
```diff
@@ -212,9 +217,7 @@ def _loop(self):
 
     def send(self, message, message_id=None):
         """Send the given message up the wire.
-
         Use this for messages which have a response message.
-
```
abhinav: Don't delete these lines

python/tchannel/tornado/connection.py
```diff
@@ -456,3 +450,87 @@ def ping(self):
 
     def pong(self):
         return self._write(messages.PingResponseMessage())
+
+
+class StreamConnection(TornadoConnection):
```
abhinav: Add a docstring to the class.

python/tchannel/tornado/connection.py
```diff
@@ -456,3 +450,87 @@ def ping(self):
...
+    def __init__(self, connection):
+        super(StreamConnection, self).__init__(connection)
```
abhinav: Don't need an init if all you're doing is calling super

python/tchannel/tornado/connection.py
```diff
@@ -456,3 +450,87 @@ def ping(self):
...
+        into args[] array, and continue to read next arg stream in order to
+        compose a larger message instead of sending multi small messages.
+
+        Note: the message built at this stage is not guaranteed the size is
+        less then 64KB.
+
+        Possible messages created sequence:
+
+        Take request as an example:
+        CallRequestMessage(flags=fragment)
```
abhinav: Use proper reStructuredText for this. Add a colon to "example:" to indicate that the paragraph that follows is a code block, add an empty line, and indent the example messages 4 spaces. Like so,

    Take request as an example::

        CallRequestMessage(flags=fragment)
        [..]

python/tchannel/tornado/connection.py
```diff
@@ -456,3 +450,87 @@ def ping(self):
...
+            --> CallRequestContinueMessage(flags=fragment)
+            ....
+            --> CallRequestContinueMessage(flags=fragment)
+                --> CallRequestMessage(flags=none)
+
+        :param context: Request or Response object
```
abhinav: Rename the parameter if it's not a Context object as defined in context.py.

python/tchannel/tornado/connection.py
```diff
@@ -456,3 +450,87 @@ def ping(self):
...
+                args = [chunk]
+                chunk = yield argstream.read()
+
+        # last piece of request/response.
+        message = (message_factory.
+                   build_raw_message(context, args, is_completed=True))
+        self.write(message, context.id)
+
+    @tornado.gen.coroutine
+    def post_response(self, response):
```
abhinav: docstring

python/tchannel/tornado/connection.py
```diff
@@ -456,3 +450,87 @@ def ping(self):
...
+        response.close_argstreams()
+        yield self._stream(response, self.response_message_factory)
+        response.close_argstreams(force=True)
```
abhinav: I don't understand why this pattern exists (`close; do_something; close(force=True)`).

Also, what happens if `_stream` throws an exception? Do you still want `close_argstreams(force=True)` to be called? Should it be in a `finally` block?

python/tchannel/tornado/connection.py
```diff
@@ -456,3 +450,87 @@ def ping(self):
...
+
+    @tornado.gen.coroutine
+    def post_request(self, request):
+        """send the given request and response is not required"""
```
abhinav: clarify

python/tchannel/tornado/connection.py
```diff
@@ -456,3 +450,87 @@ def ping(self):
...
+        request.close_argstreams()
+        yield self._stream(request, self.request_message_factory)
+        request.close_argstreams(force=True)
+
+    def send_request(self, request):
+        """Send the given request and response is required.
+        Use this for messages which have a response message.
```
abhinav: Add empty lines before and after "Use this for messages which have a response message"

python/tchannel/tornado/dispatch.py
```diff
@@ -20,9 +20,106 @@
...
+        # user still tries read from it, it will return empty.
+        chunk = yield request.argstreams[0].read()
+        while chunk:
+            request.endpoint += chunk
+            chunk = yield request.argstreams[0].read()
+
+        endpoint = self.endpoints.get(request.endpoint, None)
+        if endpoint is None:
+            raise tornado.gen.Return(connection.send_error(
+                ErrorCode.bad_request,
```
abhinav: You can't return futures from a coroutine like this. You have to yield it.

python/tchannel/tornado/dispatch.py
```diff
@@ -38,3 +135,84 @@ def _call_endpoint(self, endpoint, request, response):
...
+        # argstreams is a list of InMemArgStream/PipeArgStream objects
+        self.argstreams = argstreams
+        self.id = id
+        self.headers = headers
+        self.state = StreamState.init
+        self.endpoint = ""
+
+    @property
+    def arg_scheme(self):
+        return self.headers.get('as', None)
```
abhinav: The default value for headers is None. This will throw an exception. Check that it's non None.

python/tchannel/tornado/dispatch.py
```diff
@@ -38,3 +135,84 @@ def _call_endpoint(self, endpoint, request, response):
...
+
+    def close_argstreams(self, force=False):
+        for stream in self.argstreams:
```
abhinav: docstring

python/tchannel/tornado/message_factory.py
```diff
@@ -36,6 +41,156 @@ def __init__(self):
...
+        elif request.state == StreamState.streaming:
+            message = CallRequestContinueMessage(
+                flags=request.flags,
+                # checksum=request.checksum
+                args=args
+            )
+            request.state = (StreamState.completed if is_completed
+                             else StreamState.streaming)
+        else:
+            raise StreamingException("request state Error")
```
abhinav: This error message needs to be more helpful

python/tchannel/tornado/message_factory.py
```diff
@@ -36,6 +41,156 @@ def __init__(self):
         # value: incomplete streaming messages
         self.message_buffer = {}
 
+    def build_raw_request_message(self, request, args, is_completed=False):
```
abhinav: What does the `raw` here imply? What's the relationship between `build_raw_request_message` and `build_request`?

abhinav: Also, `self` is not used once. Maybe this should be a `classmethod`.

python/tchannel/tornado/message_factory.py
```diff
@@ -36,6 +41,156 @@ def __init__(self):
...
+                args=args
+            )
+            request.state = (StreamState.completed if is_completed
+                             else StreamState.streaming)
+        else:
+            raise StreamingException("request state Error")
+
+        return message
+
+    def build_raw_response_message(self, response, args, is_completed=False):
```
abhinav: Same as above.

python/tchannel/tornado/message_factory.py
```diff
@@ -36,6 +41,156 @@ def __init__(self):
...
+
+        return message
+
+    def build_raw_message(self, context, args, is_completed=False):
+        if isinstance(context, Request):
+            return self.build_raw_request_message(context, args, is_completed)
+        elif isinstance(context, Response):
+            return self.build_raw_response_message(context, args, is_completed)
+        else:
+            raise StreamingException("context object type error")
```
abhinav: Same as above. More helpful error message. Should probably be a class method.

python/tchannel/tornado/message_factory.py
```diff
@@ -36,6 +41,156 @@ def __init__(self):
...
+                args=args
+            )
+            response.state = (StreamState.completed if is_completed
+                              else StreamState.streaming)
+        else:
+            raise StreamingException("response state Error")
+
+        return message
+
+    def build_raw_message(self, context, args, is_completed=False):
```
abhinav: `context` implies it's a `Context` from context.py.

If not, use a different name.

python/tchannel/tornado/message_factory.py
```diff
@@ -36,6 +41,156 @@ def __init__(self):
...
+            # tracing=message.tracing,
+            service=message.service,
+            # headers=messaget.headers,
+            # checksum=message.checksum
+            argstreams=args,
+            id=message_id
+        )
+        return req
+
+    def build_response(self, message, message_id=None):
```
abhinav: Both of these should be `classmethod`s or `staticmethod`s.

python/tchannel/tornado/message_factory.py
```diff
@@ -36,6 +41,156 @@ def __init__(self):
...
+        created request may not contain whole three arguments.
+
+        :param message: CallResponseMessage
+        :param message_id: integer of message id
+        :return: response object
+        """
+
+        args = []
+        for arg in message.args:
+            args.append(InMemArgStream(arg, auto_close=False))
```
abhinav: What's the reason for `auto_close=False`

python/tchannel/tornado/message_factory.py
junchaowu: build_raw_request_message is building message,  build_request is building request. Should not be classmethod. Self is not used in stream here, but in tracing we will.

python/tchannel/tornado/message_factory.py
junchaowu: Because on one hand we are receiving fragments from socket, on the other hand user may start to consume. If we set auto_close to true, argsream may closed automatically. Therefore set it false here, we will close stream when we receive all fragments

python/tchannel/tornado/message_factory.py
```diff
@@ -47,44 +202,54 @@ def build(self, message_id, message):
...
                                       Types.CALL_RES_CONTINUE]:
-            call_msg = self.message_buffer.get(message_id)
-            if call_msg is None:
+            context = self.message_buffer.get(message_id)
+            if context is None:
                 # missing call msg before continue msg
                 raise StreamingException(
-                    "missing call message after receiving" +
-                    "continue message")
+                    "missing call message after receiving continue message")
```
abhinav: probably mention the message ID in the error

python/examples/handler.py
```diff
@@ -21,34 +21,53 @@
...
 
 from tchannel.tornado.dispatch import TornadoDispatcher
+from tchannel.tornado.stream import InMemArgStream
+from tchannel.tornado.util import print_arg, get_arg
 
 
+@tornado.gen.coroutine
 def say_hi(request, response, opts):
-    response.write(arg3="hi")
+    arg2 = yield get_arg(request, 1)
```
blampe: Why not `request.get_arg(1)`?

python/examples/handler.py
```diff
@@ -21,34 +21,53 @@
...
+    arg3 = yield get_arg(request, 2)
+    response.argstreams = [
+        InMemArgStream(request.endpoint),
+        InMemArgStream(arg2),
+        InMemArgStream(arg3)
```
blampe: "hi"

python/tchannel/tornado/stream.py
```diff
@@ -0,0 +1,143 @@
...
+try:
+    from tornado.locks import Condition
+except ImportError:  # pragma: no cover
+    from toro import Condition
+
+
+class ArgStream(object):
+
+    def __init__(self):
+        self._state = StreamState.init
```
abhinav: Move `_state` initialization into child classes (even if that leads to duplication). This class only defines the interface. It should not dictate internal state of implementations.

python/tchannel/tornado/stream.py
```diff
@@ -0,0 +1,143 @@
...
+        if it reaches the end of the stream, it will return empty(""), caller
+        will depend on the return value to tell if it reaches the end of the
+        stream.
+
+        It doesn't support seek functionality, which means it will only read
+        the stream data once in single direction.
+
+        NOTE: The data which has been read will be discarded in the stream.
+
+        :return:chunk of bytes read from stream
```
abhinav: It seems that the interface is that it returns a future that contains the next chunk. Mention that instead.

python/tchannel/tornado/stream.py
```diff
@@ -0,0 +1,143 @@
...
+        """
+        raise NotImplementedError()
+
+    def write(self, chunk):
+        """Async write to internal stream buffer
+
+        :return chunk of data in the stream
+
```
abhinav: Write does not return that

python/tchannel/tornado/stream.py
```diff
@@ -0,0 +1,143 @@
...
+        :raises StreamingException:
+            if stream has been closed, it will raise StreamingException
+        """
+        raise NotImplementedError()
+
+    def close(self):
+        raise NotImplementedError()
```
abhinav: docstring

python/tchannel/tornado/stream.py
```diff
@@ -0,0 +1,143 @@
...
+
+
+class InMemArgStream(ArgStream):
+
+    def __init__(self, buf=None, auto_close=True):
```
abhinav: Document the meaning of the parameters

python/examples/handler.py
```diff
@@ -21,34 +21,53 @@
...
+from tchannel.tornado.stream import InMemArgStream
+from tchannel.tornado.util import print_arg, get_arg
 
 
+@tornado.gen.coroutine
 def say_hi(request, response, opts):
-    response.write(arg3="hi")
+    arg2 = yield get_arg(request, 1)
+    arg3 = yield get_arg(request, 2)
+    response.argstreams = [
```
blampe: Importing `InMemArgStream` and assigning a 3-element list to `argstreams` is clunky and error-prone, we'll need to make some helper methods eventually.

python/tchannel/tornado/stream.py
```diff
@@ -0,0 +1,143 @@
...
+        self._state = StreamState.completed
+        self._condition.notify()
+
+
+class PipeArgStream(ArgStream):
+
+    def __init__(self, rpipe, wpipe=None, auto_close=False):
+        """Pipe-based arg stream
+
+        NOTE: reading from or writing to files, use os.open to get the file
```
abhinav: > Note: **When** reading from or writing to files, [...]

Also explain *why* this is the case. (`open` tries to close the already-closed file descriptor.)

python/examples/handler.py
```diff
@@ -58,10 +77,15 @@ def get_example_handler():
     dispatcher.register("ok", say_ok)
     dispatcher.register("echo", echo)
     dispatcher.register("slow", slow)
-    dispatcher.register("blocking", blocking)
```
blampe: Why'd you kill this one?

python/tchannel/tornado/stream.py
```diff
@@ -0,0 +1,143 @@
...
+        :param rpipe: an integer file descriptor which supports read ops
+        :param wpipe: an integer file descriptor which supports write ops
+        :param auto: flag to indicate to close the stream automatically or not
+        """
+        super(PipeArgStream, self).__init__()
+        self._rpipe = rpipe
+        self._wpipe = wpipe
+
+        self._rs = (PipeIOStream(self._rpipe) if
+                    self._rpipe is not None else None)
```
abhinav: Since `rpipe` is required, don't allow it to be None. Assert that it is not None.

python/tchannel/tornado/stream.py
```diff
@@ -0,0 +1,143 @@
...
+        NOTE: reading from or writing to files, use os.open to get the file
+        descriptor instead of python's open. Socket file descriptors and
+        others are fine.
+
+        when you use os.pipe to generate one write pipe and one read pipe, you
+        need to pass both of them into init method.
+
+        :param rpipe: an integer file descriptor which supports read ops
+        :param wpipe: an integer file descriptor which supports write ops
+        :param auto: flag to indicate to close the stream automatically or not
```
abhinav: We can fix this in a future diff but: `auto_close` is not actually used anywhere in this class. It's really a property of one of the consumers of this class.

python/tchannel/tornado/stream.py
abhinav: Comment still valid^

python/tchannel/tornado/stream.py
```diff
@@ -0,0 +1,143 @@
...
+    def __init__(self, buf=None, auto_close=True):
+        """In-Memory based arg stream
+
+        :param buf:
+        :return:
+        """
+        super(InMemArgStream, self).__init__()
+        self._stream = buf or ""
+        self._condition = Condition()
+        self.auto_close = auto_close
```
abhinav: It seems that this version of auto_close doesn't do anything for this class. Are you sure this shouldn't be a property of the parent class?

python/tchannel/tornado/stream.py
```diff
@@ -0,0 +1,143 @@
...
+        if self._wpipe is not None:
+            try:
+                yield self._ws.write(chunk)
+            except StreamClosedError:
+                self._state = StreamState.completed
+                raise StreamingException("Stream has been closed.")
+
+    def close(self):
+        self._state = StreamState.completed
+        if self._ws and self.auto_close:
```
abhinav: So this `auto_close` closes the underlying stream. Maybe it should be called something else. If disabled, it wouldn't close the underlying file descriptor and, assuming that `PipeIOStream` doesn't do that on GC, we would still be able to use `open` for file descriptors passed to this.

python/examples/handler.py
junchaowu: all are streaming now. when they read for args, they do yield arg.read()

python/tchannel/tornado/stream.py
```diff
@@ -0,0 +1,143 @@
...
+        :raises StreamingException:
+            if stream has been closed, it will raise StreamingException
+        """
+        raise NotImplementedError()
+
+    def close(self):
+        raise NotImplementedError()
+
+
+class InMemArgStream(ArgStream):
```
blampe: Naming nit: these implementations don't have anything to do with args, they're just reading/writing buffers.

python/examples/tchannel_server.py
```diff
@@ -21,9 +21,7 @@
 from __future__ import absolute_import
 
 import tornado.ioloop
-
```
blampe: Keep this. Whitespace between imports is used to distinguish between stdlib/third-party/local modules.

python/tchannel/tornado/tchannel.py
```diff
@@ -226,23 +226,17 @@ def __init__(self, hostport, service, tchannel):
     @tornado.gen.coroutine
     def send(self, arg_1, arg_2, arg_3):
         peer_connection = yield self.tchannel().get_peer(self.hostport)
-        self.message_id = message_id = peer_connection.next_message_id()
+        self.message_id = peer_connection.next_message_id()
```
abhinav: Actually, I don't think `self.message_id` is used anywhere. I would remove the attribute completely and just track the message ID in a local variable.

python/tchannel/tcurl.py
```diff
@@ -205,28 +207,25 @@ def multi_tcurl(
 
     all_requests = getattr(itertools, 'izip', zip)(hostports, headers, bodies)
 
-    with timing(profile=profile) as info:
```
blampe: :(

python/tchannel/tornado/util.py
junchaowu: Need these methods for tests and examples

python/tchannel/tornado/util.py
junchaowu: see above

python/tchannel/tornado/stream.py
junchaowu: done

python/tchannel/tornado/connection.py
junchaowu: keep context for now. call revisit in the future

python/tchannel/tornado/connection.py
junchaowu: add try..finally block

python/tchannel/tornado/connection.py
junchaowu: done

python/tchannel/tornado/dispatch.py
junchaowu: removed the raise

python/tchannel/tornado/message_factory.py
junchaowu: for all remaining naming issue. Won't fix in this patch

python/tchannel/tornado/dispatch.py
junchaowu: fixed

PR #382, opened 2015-04-13
trace: support manual trace propagation

rf:
Unless `enableAutoTracePropagate()` is called on the tracer, we don't install async listener, avoiding the performance hit. In this case spans must be manually propagated from the incoming request object to the outgoing request call.

@jcorbin @kriskowal @Raynos


node/test/trace/basic_server.js
```diff
@@ -71,7 +73,7 @@ test('basic tracing test', function (assert) {
         console.log("top level sending to subservice");
         setTimeout(function () {
             server
-                .request({host: '127.0.0.1:4042', service: 'subservice', trace: true})
+                .request({host: '127.0.0.1:4042', serviceName: 'subservice', trace: true})
```
jcorbin: really these shouldn't be one-liners anymore... (optional feedback)

node/trace/agent.js
```diff
@@ -30,13 +30,36 @@ function Agent () {
     }
     var self = this;
     self.refcnt = 0;
+    self.currentSpan = null;
+
+    self.autoTracePropagate = false;
+
+    self.logger = DebugLogtron('tchannelTrace');
```
jcorbin: why not a real logger, these are for tests

node/trace/agent.js
```diff
@@ -59,19 +82,10 @@ function Agent () {
...
 
-    self.listening = false;
-
-    self.logger = DebugLogtron('tchannelTrace');
-
-    // If this is set to true in a call to Agent#configure, all incoming
-    // requests will have their traceflags forced to 1. It's intended to be
-    // set on the 'top level service'.
-    self.forceTrace = false;
+    process.addAsyncListener(self.asyncListener);
```
jcorbin: we shouldn't add here right? wait for later refcnt trigger?

node/trace/agent.js
```diff
@@ -122,7 +136,17 @@ Agent.prototype.setupNewSpan = function setupNewSpan(options) {
...
     });
 
-    var parentSpan = self.getCurrentSpan();
+    var parentSpan;
+    if (self.asyncListener) {
+        parentSpan = self.getCurrentSpan();
+    } else {
+        parentSpan = options.parentSpan;
+        if (options.outgoing) {
+            self.logger.error("TChannel tracer: parent span not specified " +
```
jcorbin: there's no need for these line length gymnastics in this repo, we don't use a line-length limiting linter...

also this seems like a warn to me not an error

jcorbin:
Hardly any of your test code is wired up:
```
joshuacorbin@Joshuas-MacBook-Pro ~/Uber/tchannel/node$ grep require test/trace/index.js
require('./basic_server');
require('./server_2_requests');
joshuacorbin@Joshuas-MacBook-Pro ~/Uber/tchannel/node$ ls test/trace/
basic_server.js  basic_server_fixture.js  cross_instance.js  index.js  server_2_requests.js  server_2_requests_fixture.js
```

So no one but you has been running your tests (least of all Travis CI); please fix it in this PR before merge.

rf:
Wired tests in with something that forks per test. Since some tests load in async-listener and some don't it makes more sense to run them individually.

node/test/trace/index.js
```diff
@@ -18,6 +18,26 @@
...
+    return function (done) {
+        var proc = fork(__dirname + '/' + file);
+        proc.on('exit', done);
+    };
+}
+
+function runSeries(items) {
+    if (items.length === 0) return;
+    items[items.length - 1](function () {
+        runSeries(items.slice(0, -1));
```
jcorbin: blowing out the stack and recursing is cool and all, but let's just use run-series please

rf: didn't want to add the dep for such a trivial case

PR #383, opened 2015-04-13
Add request option defaults

jcorbin:


Raynos:
Do we want to keep this global default to timeouts ?

As a user of tchannel in autobahn and ncar-client I actually preferred to force myself to set timeouts for all my requests instead of the default 2s

PR #385, opened 2015-04-13
Carry remoteAddr on reqs, shift away from host option

jcorbin:


node/outgoing_request.js
```diff
@@ -86,8 +87,8 @@ function TChannelOutgoingRequest(id, options) {
             traceid: null,
             parentid: null,
             flags: options.trace? 1 : 0,
-            hostPort: options.host,
-            serviceName: options.serviceName,
+            hostPort: self.remoteAddr,
+            serviceName: self.service,
```
jcorbin: cc @rf on this change, since tests are fully wired up, I don't know if this breaks any of your assumptions; should the reported span service differ from the service which the request is being made to here?

rf: This looks fine to me

PR #386, opened 2015-04-13
remove legacy register() support

Raynos:
We want to re-purpose the top level `register()` to work
in the clean new v2 world.

This breaks the old legacy interface that was used in
various integration tests in ringpop + autobahn as well
as in ringpop itself.

This is going to be a painful breaking change + refactor.

We will want to do this; but the question is when.

As a secondary question; do we want to repurpose `.send()`
or just completely remove it ? 

I think it should be removed as it goes against the 
`as/json` and `as/thrift` implementations.

r: @kriskowal @jcorbin

### DIFF
--- a/node/channel.js
+++ b/node/channel.js
@@ -299,7 +299,6 @@ TChannel.prototype.listen = function listen(port, host, callback) {
     self.getServer().listen(port, host, callback);
 };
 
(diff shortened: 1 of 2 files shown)


kriskowal:
:ok:

PR #389, opened 2015-04-14
Retries

jcorbin:
r @Raynos @kriskowal @rf

### DIFF
--- a/node/request.js
+++ b/node/request.js
@@ -20,6 +20,8 @@
 
 'use strict';
 
+var DEFAULT_RETRY_LIMIT = 5;
+
 function TChannelRequest(channel, options) {
     options = options || {};
     var self = this;
@@ -30,6 +32,7 @@ function TChannelRequest(channel, options) {
     self.options = options;
     self.outReqs = [];
     self.timeout = self.options.timeout;
+    self.limit = self.options.retryLimit || DEFAULT_RETRY_LIMIT;
     self.start = 0;
     self.end = 0;
     self.elapsed = 0;
@@ -93,6 +96,10 @@ TChannelRequest.prototype.onReqDone = function onReqDone(err, res, arg2, arg3) {
 
 TChannelRequest.prototype.shouldRetry = function shouldRetry(err, res, arg2, arg3) {
     var self = this;
+
+    if (self.outReqs.length >= self.retryLimit) {
+        return false;
+    }
 
     if (err) {
         switch (err.type) {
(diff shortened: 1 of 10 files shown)


node/request.js
```diff
@@ -0,0 +1,133 @@
...
+};
+
+TChannelRequest.prototype.shouldRetry = function shouldRetry(err, res, arg2, arg3) {
+    var self = this;
+
+    if (self.outReqs.length >= self.retryLimit) {
+        return false;
+    }
+
+    if (err) {
```
Raynos: can we check the `err.isErrorFrame` boolean ?

Raynos: If we do not check that boolean we will retry on all network errors; i.e. `ECONNRESET` and pals.

kriskowal:
This looks solid to me. There are pending Travis errors, so I’m going to let this sit. We can cut a tchannel tag in the morning. I won’t have a PR based on this ready today regardless.

Raynos:
Other then improving `shouldRetry()` with better error detection (having a default case where we always retry seems a bit loose); this looks good to me.

jcorbin:
As confirmed with @mranney this morning, the entire point of the design here is to default into retrying with only blacklisted "why not".

This PR doesn't add any opt-out path, which should / will come as needed.  However we're intentionally going with the design of defaulted-in to retries.

Especially so for socket errors.

node/request.js
```diff
@@ -0,0 +1,133 @@
...
+TChannelRequest.prototype.onReqDone = function onReqDone(err, res, arg2, arg3) {
+    var self = this;
+    var now = self.channel.timers.now();
+    self.elapsed = now - self.start;
+    if (self.elapsed < self.timeout &&
+        self.shouldRetry(err, res, arg2, arg3)) {
+        self._lastErr = err;
+        self._lastArg2 = arg2;
+        self._lastArg3 = arg3;
+        process.nextTick(deferResend);
```
Raynos: In the near future we will want some kind of backoff here.

Raynos:
Are we sure all socket errors are remote errors ? Does the node `net` library emit `net` errors that are basically a `BadRequest`.

To be honest I don't mind either way as long as we have stats and information on what the hell is going through our processes.

node/request.js
jcorbin: no we don't, backoff is a per-peer concern; there's no sane reason to blow your timeout deadline sleeping.

Raynos: true; makes sense.

PR #392, opened 2015-04-14
remove as legacy layer

Raynos:
This is no longer used by anyone. This has been removed from
both autobahn and autobahn-node-client.

r: @kriskowal @jcorbin


kriskowal:
:100:

Issue #396, opened 2015-04-14
Add CRC32C support in golang implementation


jcorbin:
cc @mmihic 

Apropos of #390

PR #397, opened 2015-04-14
Switch to using the Castagnoli poly for crc32c

mmihic:
R: @jcorbin

### DIFF
--- a/golang/checksum.go
+++ b/golang/checksum.go
@@ -25,6 +25,10 @@ import (
 	"hash/crc32"
 )
 
+var (
+	crc32CastagnoliTable = crc32.MakeTable(crc32.Castagnoli)
+)
+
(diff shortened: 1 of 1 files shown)


golang/checksum.go
```diff
@@ -59,7 +63,7 @@ func (t ChecksumType) New() Checksum {
 	case ChecksumTypeNone:
 		return nullChecksum{}
 	case ChecksumTypeCrc32:
```
jcorbin: sadly this needs to maintain the old adler-32 setting, there's a new constat (0x03) for castagnoli

jcorbin:
lgtm

mmihic:
Resolves https://github.com/uber/tchannel/issues/396

PR #398, opened 2015-04-14
trace: removing async-listener

rf:
I don't think this was a good design decision in the first place. This makes everything simpler and faster and removes a lot of unnecessary complexity.


jcorbin:
LGTM... less code less problems...

Raynos:
@rf what will the diff for tracing propagation in autobahn look like without async-listener ?

Raynos:
lgtm; yolo shipping it.

PR #400, opened 2015-04-15
make as/json consistent with as/thrift

Raynos:
This makes the `register()` method of json and thrift
consistent.

The json one had the arguments in the wrong way.

r: @kriskowal @jcorbin


kriskowal:
:+1:

PR #404, opened 2015-04-15
Add error classification

jcorbin:


kriskowal:
:ok: Perhaps we should consider making the tchannel class of an error a property of each individual error.

Raynos:
lgtm.

PR #405, opened 2015-04-15
Cleanup out req / res state paths a bit

jcorbin:


node/outgoing_request.js
```diff
@@ -104,14 +104,28 @@ function TChannelOutgoingRequest(id, options) {
 
     self.err = null;
     self.res = null;
-    self.start = self.timers.now();
```
Raynos: what happened to start?

node/self_peer.js
```diff
@@ -38,7 +38,7 @@ TChannelSelfPeer.prototype.connect = function connect() {
     var self = this;
     while (self.connections[0] &&
            self.connections[0].closing) {
-        self.connections[0].shift();
+        self.connections.shift();
```
Raynos: nice fix XD

PR #412, opened 2015-04-15
Define new "network error" error frame code, implement in node

jcorbin:
Need felt when implementing the forwarding layer of *bahn, the core of which is the `RelayHandler` now in tchannel.

Basically the forwarder had no way of differentiating network errors to its caller.  This is chiefly a visibility concern at present, network errors aren't notable (sentry and/or logging), but an unexpected error is.

### DIFF
--- a/docs/protocol.md
+++ b/docs/protocol.md
@@ -464,6 +464,7 @@ code   | name                 | description
 `0x04` | declined             | Node declined request for reasons other than load, the request is safe to retry elsewhere if desired, do not change peer preferencing for future requests.
 `0x05` | unexpected error     | Request resulted in an unexpected error. The request may have been completed before the error, retry only if the request is idempotent or duplicate execution can be handled.
 `0x06` | bad request          | Request args do not match expectations, request will never be satisfied, do not retry
+`0x07` | network error        | A network error (e.g. socket error) occurred.
 `0xFF` | fatal protocol error | Connection will close after this frame. message ID of this frame should be `0xFFFFFFFF`.
 
 ### id:4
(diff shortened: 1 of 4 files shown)


Raynos:
lgtm; +1

kriskowal:
It would be good to note what a network error implies for future attempts to retry. My guess: "this is probably an intermittent failure, possibly a restarting server. Do retry, favor other peers for a while", semantically similar to "declined", possibly means "declined after accept" so retry might be more dangerous if the endpoint is not idempotent.

cc @mranney

mmihic:
Declined means "I have explicitly not acted on your request" while network failure could occur after the request has already been fully or partially processed (e.g. server crash / connection loss / network partition after receiving and processing initial fragment but before generating response).  Semantically network error is effectively the same as "unexpected failure" - you don't know if the request was processed, but can retry if the request is idempotent or duplicate execution can be handled.

jcorbin:
Indeed, I'm not intending this to be acted on differently, but mean it
primarily for not log/sentry spamming about network blips.

mmihic:
Yep, I'm +1 on this change.

mranney:
I like that sometimes we know for sure that the "unexpected" error is a network error, but in a world with 0x07, when would we ever use 0x05?

Raynos:
In systems that have "continue on exceptions" if an application developers handler function throws an unexpected error we would send an error frame saying "this incoming call request failed because we ran the application handler and it blew up in an unexpected way".

This is a useful piece of information; at least for logging purposes, maybe not so for retry or automated logic.

jcorbin:
Also in tchannel, if there was an programming error or unexpected condition, it's really useful to send some an unexpected error frame.

PR #416, opened 2015-04-16
TChannel Router

jcorbin:


Raynos:
This is old; let's abandon this PR.

jcorbin:
indeed

PR #417, opened 2015-04-16
Implement retry flags per protocol doc

jcorbin:


node/request.js
```diff
@@ -195,20 +204,28 @@ TChannelRequest.prototype.shouldRetry = function shouldRetry(err, res, arg2, arg
...
             case 'tchannel.canceled':
                 return false;
 
-            case 'tchannel.socket':
-            case 'tchannel.timeout':
             case 'tchannel.busy':
             case 'tchannel.declined':
-            case 'tchannel.unexpected':
-            case 'tchannel.protocol':
                 return true;
```
jcorbin: cc @mranney @mmihic does this seem like a missed flag? or intentionally left out?

node/request.js
```diff
@@ -195,20 +204,28 @@ TChannelRequest.prototype.shouldRetry = function shouldRetry(err, res, arg2, arg
...
 
+            case 'tchannel.timeout':
+                return !!self.options.retryFlags.onTimeout;
+
+            case 'tchannel.socket':
+            case 'tchannel.protocol':
+            case 'tchannel.unexpected':
```
jcorbin: cc @mranney @mmihic it's correct to treat all 3 the same here right?

node/request.js
```diff
@@ -195,20 +204,28 @@ TChannelRequest.prototype.shouldRetry = function shouldRetry(err, res, arg2, arg
...
     if (err) {
         switch (err.type) {
             case 'tchannel.bad-request':
             case 'tchannel.canceled':
                 return false;
 
-            case 'tchannel.socket':
-            case 'tchannel.timeout':
             case 'tchannel.busy':
             case 'tchannel.declined':
```
jcorbin: cc @mranney @mmihic does this seem like a missed flag? or intentionally left out?

node/request.js
mmihic: yes

node/request.js
mmihic: Sorry, which flag?

jcorbin: The busy/declined path, it's hardwired to true here.

Seems a bit of a minor miss to me:
- would be better explicate a flag, say `"d"` for declined
- change the default retryflags then to be `"cd"`

We may not feel the lack of it much, but implementing the code seemed to point it out to me as an oversight.

Raynos:
lgtm.

PR #420, opened 2015-04-18
Set start time for non-streamed outgoing requests

jwolski:
Start time used for timeout checks was never set on (non-streamed) outgoing requests.

@jcorbin @Raynos @kriskowal @rf


Raynos:
:+1:

jcorbin:
Great catch, merging this so that I can roll it into other related work.

PR #421, opened 2015-04-18
Refactor TChannelRequest handling

jcorbin:


node/request.js
```diff
@@ -43,7 +44,7 @@ function TChannelRequest(channel, options) {
     self.options = options;
     self.triedRemoteAddrs = {};
     self.outReqs = [];
-    self.timeout = self.options.timeout;
+    self.timeout = self.options.timeout || DEFAULT_TIMEOUT;
```
Raynos: there is another default timeout somewhere else of 5s. We should make the default timeout consistent across the board.

node/request.js
```diff
@@ -188,6 +190,29 @@ TChannelRequest.prototype.resend = function resend() {
...
 };
 
+TChannelRequest.prototype.checkTimeout = function checkTimeout(err, res) {
+    var self = this;
+    var now = self.timers.now();
+    self.elapsed = now - self.start;
+    if (self.elapsed < self.timeout) return false;
+    if (err) {
+        if (!self.err) {
+            self.emit('error', err);
```
Raynos: show we `self.err = err;` here ?

node/request.js
```diff
@@ -188,6 +190,29 @@ TChannelRequest.prototype.resend = function resend() {
...
+        }
+    } else if (res) {
+        if (!self.err && !self.res) {
+            self.emit('response', res);
```
Raynos: should we `self.res = res;` here ?

node/request.js
```diff
@@ -188,6 +190,29 @@ TChannelRequest.prototype.resend = function resend() {
...
+        }
+    } else if (!self.err) {
+        self.emit('error', errors.TimeoutError({
```
Raynos: This looks like a dead branch; maybe leave a comment here ?

node/request.js
jcorbin: No, that's done by `TChannelRequest#onError` which gets hooked up by constructor so that we don't have to do that in every emission site.

node/request.js
jcorbin: No, that's done by `TChannelRequest#onResponse` which gets hooked up by constructor so that we don't have to do that in every emission site.

node/request.js
jcorbin: it's not dead, it's used by the two new call sites circa L134 and L148 as added by https://github.com/uber/tchannel/commit/57e60e81f64e1866fb4737af9923578288284aa1

node/request.js
jcorbin: sure, this is a dead branch anyhow at present, just added for defense in depth, will update

Raynos:
lgtm

PR #422, opened 2015-04-18
Make in/out req/res more cross-consistent

jcorbin:
- span annotate at start of outgoing
- remoteAddr on both incoming
- start for everyone
- end for outgoing

### DIFF
--- a/node/outgoing_response.js
+++ b/node/outgoing_response.js
@@ -46,6 +46,8 @@ function TChannelOutgoingResponse(id, options) {
     self.random = options.random;
     self.timers = options.timers;
 
+    self.start = 0;
+    self.end = 0;
     self.state = States.Initial;
     self.id = id || 0;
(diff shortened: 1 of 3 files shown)


node/outgoing_response.js
```diff
@@ -111,13 +117,12 @@ TChannelOutgoingResponse.prototype.sendCallResponseFrame = function sendCallResp
...
         case States.Initial:
+            self.start = self.timers.now();
             self.sendFrame.callResponse(args, isLast);
-            if (isLast) {
-                if (self.span) {
-                    self.span.annotate('ss');
-                }
-                self.state = States.Done;
+            if (self.span) {
+                self.span.annotate('ss');
```
jcorbin: cc @rf since this is a semantic change, but one for the better and one that doesn't make much difference right now anyhow

Raynos:
lgtm.

PR #423, opened 2015-04-18
Explicate RelayRequest

jcorbin:
- reifies the prior closure `RelayHandler#handleRequest` closure
- made more obvious to handle pathological cases
- allows to add more visibility work for forwarding going forward


jcorbin:
Review #427 first, will re-open once rebased on its land

PR #424, opened 2015-04-18
Correct request TTL

jcorbin:
- trim any known over head on first request
- correctly decremented for retries


Raynos:
decrement TTL. This is good feature.

Raynos:
lgtm.

PR #425, opened 2015-04-19
Add multicast messaging library

jcorbin:
Based on my earlier multicast demo, and later TRouter demo


kriskowal:
This seems more like a BufRW example than a TChannel example although I’m sure you intend to subsume some of the functionality here.

jcorbin:
sure... if it were an "example", I had started using it in the trouter branch towards production... we don't need this yet, so I'll likely just let this branch/PR sit until we do

Raynos:
Let's repurpose this and add it to examples? Or abandon branch?

PR #434, opened 2015-04-20
Initial hyperbahn implementation in tchannel

Raynos:
r: @jcorbin

### DIFF
--- a/node/endpoint-handler.js
+++ b/node/endpoint-handler.js
@@ -84,12 +84,12 @@ TChannelEndpointHandler.prototype.handleArg1 = function handleArg1(req, buildRes
             'no such endpoint service=%j endpoint=%j',
             req.service, name));
     } else if (handler.canStream) {
-        handler(req, buildResponse);
+        handler.call(self, req, buildResponse);
     } else if (req.streamed) {
         self.bufferArg23(req, buildResponse, handler);
     } else {
         res = buildResponse({streamed: false});
-        handler(req, res, req.arg2, req.arg3);
+        handler.call(self, req, res, req.arg2, req.arg3);
     }
 };
 
--- a/node/hyperbahn/handler.js
+++ b/node/hyperbahn/handler.js
@@ -44,7 +44,6 @@ function HyperbahnHandler(options) {
     // TODO support blackList
 
     self.channel = options.channel;
-    self.type = 'hyperbahn.advertisement-handler';
     self.egressNodes = options.egressNodes;
 
     self.tchannelJSON = TChannelJSON({
@@ -58,6 +57,8 @@ function HyperbahnHandler(options) {
         self.handleRelayAdvertise);
 }
 util.inherits(HyperbahnHandler, TChannelEndpointHandler);
+
+HyperbahnHandler.prototype.type = 'hyperbahn.advertisement-handler';
 
 /*  req: {
         services: Array<{
--- a/node/hyperbahn/handler.js
+++ b/node/hyperbahn/handler.js
@@ -26,7 +26,6 @@ var assert = require('assert');
 var util = require('util');
 
 var TChannelJSON = require('../as/json');
-var EgressNodes = require('./egress-nodes');
 var TChannelEndpointHandler = require('../endpoint-handler');
 
 module.exports = HyperbahnHandler;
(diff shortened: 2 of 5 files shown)


node/hyperbahn/handler.js
```diff
@@ -0,0 +1,229 @@
...
+    TChannelEndpointHandler.call(self, 'hyperbahn');
+
+    assert(options && options.channel, 'channel required');
+    assert(options && options.ringpop, 'ringpop required');
+    assert(options && options.egressNodes, 'egressNodes required');
+
+    // TODO support blackList
+
+    self.channel = options.channel;
+    self.type = 'hyperbahn.advertisement-handler';
```
jcorbin: proto

node/hyperbahn/handler.js
```diff
@@ -0,0 +1,229 @@
...
+            hostPort, exitNodeServices, onFinish
+        );
+    }
+
+    onFinish();
+
+    // TODO remove blocking on fanout finish. Requires fixing
+    // autobahn tests upstream
+    function onFinish() {
+        if (--counter === 0) {
```
jcorbin: <=

node/hyperbahn/handler.js
```diff
@@ -0,0 +1,229 @@
...
+
+    if (typeof options === 'function') {
+        callback = options;
+        options = {};
+    }
+
+    var channel = options.channel || self.channel;
+
+    self.tchannelJSON.send(channel.request({
+        timeout: options.timeout,
```
jcorbin: all opt

node/hyperbahn/handler.js
```diff
@@ -0,0 +1,229 @@
...
+            hostPort: String,
+            cost: Number
+        }>
+    }
+
+    res: {}
+*/
+HyperbahnHandler.prototype.handleRelayAdvertise =
+function handleRelayAdvertise(self, req, arg2, arg3, cb) {
+    /*eslint max-params: [2, 5]*/
```
jcorbin: missing var self

node/hyperbahn/handler.js
```diff
@@ -0,0 +1,229 @@
...
+    var logger = self.channel.logger;
+
+    for (var i = 0; i < services.length; i++) {
+        var serviceObj = services[i];
+
+        var exitNodes = self.egressNodes
+            .exitsFor(serviceObj.serviceName);
+        var exitHosts = Object.keys(exitNodes);
+
+        var myHost = self.ringpop.whoami();
```
jcorbin: chan hostport

node/hyperbahn/handler.js
```diff
@@ -0,0 +1,229 @@
...
+    });
+};
+
+HyperbahnHandler.prototype.sendRelayAdvertise =
+function sendRelayAdvertise(hostPort, services, callback) {
+    var self = this;
+
+    self.tchannelJSON.send(self.channel.request({
+        host: hostPort,
+        timeout: 500,
```
jcorbin: fun hardcoded timeout

node/test/trace/basic_server.js
```diff
@@ -1,4 +1,24 @@
 // Copyright (c) 2015 Uber Technologies, Inc.
+//
```
jcorbin: derp

node/endpoint-handler.js
```diff
@@ -84,12 +84,12 @@ TChannelEndpointHandler.prototype.handleArg1 = function handleArg1(req, buildRes
             'no such endpoint service=%j endpoint=%j',
             req.service, name));
     } else if (handler.canStream) {
-        handler(req, buildResponse);
+        handler.call(self, req, buildResponse);
```
kriskowal: I’m going to push to uniformly accept handleRequest functions or handler.handleRequest objects in all these places that take handler. This has implications for how that would work. Might be better to pass the endpoint handler as an argument.

See unmerged: https://github.com/uber/tchannel/blob/circuit-breaker/node/request-handler.js

node/hyperbahn/handler.js
```diff
@@ -0,0 +1,223 @@
...
+
+var TChannelJSON = require('../as/json');
+var TChannelEndpointHandler = require('../endpoint-handler');
+
+module.exports = HyperbahnHandler;
+
+function HyperbahnHandler(options) {
+    if (!(this instanceof HyperbahnHandler)) {
+        return new HyperbahnHandler(options);
+    }
```
kriskowal: We should decide whether we’re going to kill this pattern off.

Raynos: This pattern is useful under two conditions:

 - The constructor is not called often. (per server, not per request)
 - The constructor is part of the public interface.

Under those two conditions it's useful to hide the fact that `new` is required. This makes it easier for developers to use the interfaces exposed by your package without having to remember whether its a constructor or not.

node/hyperbahn/handler.js
kriskowal: I would stand by ===. Callback should not be called more than once. We should also verify that "send" guarantees at-most-once callback by discarding tardy extra results, otherwise, onFinish may count down earlier than intended.

kriskowal:
:ok:

PR #437, opened 2015-04-20
add "network error" error code

junchaowu:
@blampe @breerly @abhinav


abhinav:
:shipit: 

We can make another change later to actually throw this error when we run into socket errors.

PR #441, opened 2015-04-20
Emit operational errors rather than throwing

jcorbin:


kriskowal:
:ok:

Yeah, these should not crash.

Raynos:
lgtm.

PR #442, opened 2015-04-20
Golang implementation of thrift-over-tchannel

weikai77:


golang/thrift/outbound.go
```diff
@@ -0,0 +1,323 @@
...
+import "bytes"
+import "errors"
+import "github.com/apache/thrift/lib/go/thrift"
+import tchannel "github.com/uber/tchannel/golang"
+import "golang.org/x/net/context"
+import "time"
+
+// NewTChannelOutboundProtocol creates a TChannelOutboundProtocol
+func NewTChannelOutboundProtocol(remoteHostPort, remoteServiceName, remoteProcessorName string) (*TChannelOutboundProtocol, error) {
+	tchannel, err := tchannel.NewChannel("0.0.0.0:0", nil)
```
mmihic: We should pass the actual TChannel object into this function, so that we can re-use connections across calls.

golang/thrift/outbound.go
```diff
@@ -0,0 +1,323 @@
...
+// it as arg3 to the underlying tchannel.
+func (p *TChannelOutboundProtocol) Flush() error {
+	// flush to write buffer
+	if err := p.writer.Flush(); err != nil {
+		return err
+	}
+	payload := p.writeBuffer.Bytes()
+
+	// begin the outbound call
+	ctx, _ := context.WithTimeout(context.Background(), time.Second*time.Duration(1))
```
mmihic: The timeout needs to be controllable by the calling code.  In fact the entire context needs to be passed from the caller, or else you won't be able to do zipkin trace flags propagation, request deadline propagation, etc.  I'd recommend passing the context into NewTChannelOutboundProtocol.

golang/thrift/server.go
```diff
@@ -0,0 +1,57 @@
...
+	return s.tchannel.Stop()
+}
+
+func (s *Server) HostPort() string {
+	return s.tchannel.HostPort()
+}
+
+// NewServer creates a service that speaks thrift over tchannel.
+func NewServer(bindAddr, serviceName string) (*Server, error) {
+	ch, err := tchannel.NewChannel(bindAddr, &tchannel.ChannelOptions{})
```
mmihic: need to allow control over the ChannelOptions; this determines things like process name etc which are needed for peering.

golang/thrift/inbound.go
```diff
@@ -0,0 +1,297 @@
...
+}
+
+// WriteMessageEnd delegates to the TBinaryProtocol writer
+func (p *TChannelInboundProtocol) WriteMessageEnd() error {
+	return p.writer.WriteMessageEnd()
+}
+
+// WriteStructBegin delegates to the TBinaryProtocol writer
+func (p *TChannelInboundProtocol) WriteStructBegin(name string) error {
+	err := p.writer.WriteStructBegin(name)
```
mmihic: nit: return err directly

golang/thrift/outbound.go
```diff
@@ -0,0 +1,323 @@
...
+}
+
+// WriteMessageEnd delegates to the TBinaryProtocol writer
+func (p *TChannelOutboundProtocol) WriteMessageEnd() error {
+	return p.writer.WriteMessageEnd()
+}
+
+// WriteStructBegin delegates to the TBinaryProtocol writer
+func (p *TChannelOutboundProtocol) WriteStructBegin(name string) error {
+	err := p.writer.WriteStructBegin(name)
```
mmihic: nit: return err directly

golang/thrift/inbound.go
```diff
@@ -0,0 +1,297 @@
...
+// and tchannel (tchannel.InboundCall).
+//
+// Incoming and outgoing data is buffered in memory buffers and the
+// actual parsing of the data is delegated to thrift.TBinaryProtocol.
+//
+type TChannelInboundProtocol struct {
+	call *tchannel.InboundCall
+
+	writer      *thrift.TBinaryProtocol
+	writeBuffer *MemoryBufferTransport
```
mmihic: This double buffering is a shame.  We should probable extend InboundCall / InboundCallResponse with BeginArg3 methods that return an io.Reader/io.Writer, so that the thrift.TBinaryProtocol to stream directly to them vs needing to go through an additional intermediate in-memory buffer.

golang/thrift/outbound.go
```diff
@@ -0,0 +1,323 @@
...
+	tchannel            *tchannel.Channel
+	remoteHostPort      string
+	remoteServiceName   string
+	remoteProcessorName string
+
+	// state per call
+	remoteOperationName string
+	call                *tchannel.OutboundCall
+	writer              *thrift.TBinaryProtocol
+	writeBuffer         *MemoryBufferTransport
```
mmihic: Same comments as above re: double buffering.

golang/Makefile
```diff
@@ -27,6 +27,7 @@ clean:
 	echo Cleaning build artifacts...
 	go clean
 	rm -rf $(BUILD)
+	rm -rf examples/thrift/gen-go
```
mmihic: nit: generally preferable to put generated sources in a different directory from hand-written code.  failing that, please add a .gitignore line to ensure that generated sources aren't accidentally committed to the repo.

golang/connection_test.go
```diff
@@ -168,5 +168,8 @@ func withTestChannel(t *testing.T, f func(ch *Channel)) {
 
 	go ch.ListenAndHandle()
 
-	f(ch)
+	func() {
```
mmihic: You can get rid of the wrapping func - since you are calling it immediately, it isn't doing anything for you.

golang/thrift/outbound.go
weikai77: @mmihic Thank you for the quick feedback!

The connections are being re-used across calls from the same client. But I'll add a different version of the NewTChannelOutboundProtocol function to take a TChannel object so connections can be re-used across *clients*.

mmihic: Is the TChannelOutboundProtocol object shared across multiple calls from the client?  This seems bad; the protocol as currently written is specific to a single request (it has references to request-specific data) so if there are multiple concurrent requests you'll end up with corruption.  

What is the scope of a client?

golang/thrift/server.go
weikai77: Will add another form of NewServer that takes a tchannel object. That should serve the same purpose.

golang/thrift/inbound.go
weikai77: Will do.

golang/thrift/outbound.go
weikai77: Will do.

golang/thrift/inbound.go
```diff
@@ -0,0 +1,297 @@
...
+	reader     *thrift.TBinaryProtocol
+	readBuffer *MemoryBufferTransport
+}
+
+// WriteMessageBegin creates the writer buffer and a TBinaryProtocol
+// writer and delegates the write to the writer
+func (p *TChannelInboundProtocol) WriteMessageBegin(name string, typeId thrift.TMessageType, seqId int32) error {
+	p.writeBuffer = NewMemoryBufferTransport()
+	p.writer = thrift.NewTBinaryProtocol(p.writeBuffer, false, false)
+	return p.writer.WriteMessageBegin(name, typeId, seqId)
```
abhinav: This doesn't match what Node and Python are doing. This information is not written to the payload.

golang/thrift/inbound.go
```diff
@@ -0,0 +1,297 @@
...
+	// parse the thrift method from arg1
+	if arg1 := string(p.call.Operation()); strings.Index(arg1, "::") < 0 {
+		err = errors.New(fmt.Sprintf("Malformed arg1: %s", arg1))
+		return
+	} else {
+		name = strings.Split(arg1, "::")[1]
+	}
+
+	// read from the read buffer
+	_, typeId, seqId, err = p.reader.ReadMessageBegin()
```
abhinav: Same. Discrepancy between languages.

golang/Makefile
weikai77: Tried to put it in a top level `generated` folder, but that messes up the package naming. Will add to .gitignore.

golang/thrift/memory_buffer_transport.go
```diff
@@ -0,0 +1,43 @@
...
+import (
+	"bytes"
+)
+
+// MemoryBufferTransport is a memory buffer-based implementation of the
+// thrift.TTransport interface. It's similar to thrift.TMemoryBuffer but
+// this implementation can be initialized with a given byte buffer.
+type MemoryBufferTransport struct {
+	*bytes.Buffer
+	size int
```
abhinav: This isn't used anywhere?

golang/Makefile
mmihic: You can put it into a generated folder if you update the Makefile GOPATH line to look in that folder.  This is how we handle the Godeps/_workspace for example.

golang/connection_test.go
weikai77: Need the wrapping func for the defer call--unless I'm missing something?

golang/thrift/outbound.go
weikai77: Yes. TChannelOutboundProtocol is not thread-safe, but can be re-used across multiple calls as long as these calls are from the same goroutine.

This does create a problem for the context stuff though. We may have to require a new TChannelOutboundProtocol for each call to get the context thing working.

golang/thrift/inbound.go
weikai77: Ah, missed that. Will fix.

golang/connection_test.go
mmihic: You don't need the wrapping func for the defer - you can just defer and then call f(ch).  The defer won't kick in until you exit the withTestChannel scope, which won't happen until f(ch) is complete.

golang/thrift/outbound.go
mmihic: The context also typically represents a goroutine level object, so I think it's fine to pass it into the TChannelOutboundProtocol and share it between calls.  In server applications, the context is created when an incoming request is received, and then is shared by all calls made during the processing of that request.  In pure clients (e.g. command line utilities), you usually create the context once at the start of main() and then use it for every subsequent invocation.

golang/thrift/inbound.go
weikai77: Yeah I agree. Will look into this approach a bit later.

prashantv:
Thrift integration (partially based on this PR) added in #667

PR #444, opened 2015-04-20
v3: Proposal for advertisement.

Raynos:
@jcorbin put out an advertisement demo last week. 

We want to add advertisement to tchannel as a first class citizen
to remove the registration logic from autobahn.

I put out a seperate PR to move registration out of autobahn
into tchannel as a seperate serviceName with well defined
call requests and call responses ( #434 ).

Ideally we want the advertisement logic to grow up and become
first class frames and part of the protocol.

Here is an initial draft of what that might look like.

r: @jcorbin @kriskowal @mranney

### DIFF
--- a/docs/advertisement.md
+++ b/docs/advertisement.md
@@ -1,0 +1,89 @@
+# TChannel advertisement
+
+TChannel advertisement is an addendum to the TChannel protocol.
+The advertisement document defines a couple of new frame types
+used to advertise the availability of service names.
+
+## Message flow
+
+There are two workflows for advertisement; one is for advertising
+to a hyperbahn routing service and the other is for general
+advertisement between peer to peer tchannel servers.
+
+### Advertising to a hyperbahn routing service
+
+A service X sends an `advertise` frame to a hyperbahn routing
+service. It tells it what services it has available at what
+costs.
+
+The hyperbahn routing service will internally assign
+a set of egress affinity routing instances that are responsible
+for the advertised services. It will send a `relay-advertise`
+frame to each egress affinity routing instance.
+
+The hyperbahn routing instance that got the `advertise`
+message will respond with a `forward-advertise` message indicating
+that it will forward traffic to service X. It will inform service
+X how many connections for forwarding it should expect.
+
+### Advertising to a peer
+
+A tchannel server X can send an `advertise` frame to a tchannel
+server Y. It will advertise what services it has available
+at what costs.
+
+## Frames
+
+### Payload: advertise (0x20)
+
+Schema:
+```
+tracing:25 num:2 ( service~1 cost:1 ){num}
+```
+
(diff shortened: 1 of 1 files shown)


docs/advertisement.md
```diff
@@ -0,0 +1,89 @@
...
+server Y. It will advertise what services it has available
+at what costs.
+
+## Frames
+
+### Payload: advertise (0x20)
+
+Schema:
+```
+tracing:25 num:2 ( service~1 cost:1 ){num}
```
jcorbin: don't need tracing

docs/advertisement.md
```diff
@@ -0,0 +1,89 @@
...
+  - send a `forward-advertise` message back to the server that
+    advertised.
+
+It is valid to `advertise` multiple times down the same connection
+
+### Payload: relay-advertise (0x21)
+
+Schema:
+```
+tracing:25 num:2 ( service~1 hostport~1 cost:1 ){num}
```
jcorbin: don't need tracing

docs/advertisement.md
```diff
@@ -0,0 +1,89 @@
...
+This means that if the tchannel server that receives the 
+`relay-advertise` message were to connect directly to the
+hostport it would be able to send call requests for the service
+and know that it gets delivered with the cost.
+
+### Payload: forward-advertise (0x22)
+
+Schema:
+```
+tracing:25 conncount:2
```
jcorbin: don't need tracing

docs/advertisement.md
```diff
@@ -0,0 +1,89 @@
...
+```
+
+A `forward-advertise` message is a response to an `advertise`
+message. It means that the remote intents to forward messages
```
jcorbin: s/intents/intends/

jcorbin:
At least the frame specs should go into the protocol document.

Until we start breaking out longer feature-deep-dives into separate documents (such as fragmentation/streaming) we should probably just add an advertisement section to it.

mranney:
I like advertise going into the protocol, but I'm worried about breaking existing implementations. Is there some way to do this with v2, or does it need to wait until v3?

Also, I think we should have the client regularly re-advertise on a server-specified interval. This should be something like 10 or 60 seconds. Upon receipt of an advertise message, a receiving router will fan this out to the other nodes in the egress affinity set. That way, a transient failure of a router node will be easy to recover form.

jcorbin:
At this point we'll almost certainly punt any advertisement frame(s) until
v3.

I'd especially like to have a discussion about the beacon interval
semantics.  We're for sure doing them in v2, but I strongly suspect that we
simply don't need them, and moving to low level frames could be an
opportunity to get away from fixed beacon intervals.

Raynos:
This will be completely punted until v3

PR #446, opened 2015-04-20
Revert "v2/tracing: reduce buffer reads/writes 3:1"

jcorbin:
This turned out to the opposite of good as it meant doing 3+1 slices, rather than just 3.


Raynos:
lgtm.

PR #447, opened 2015-04-20
Document json as handler

Raynos:
r: @jcorbin @kriskowal


jcorbin:
:+1: let's go forward

PR #448, opened 2015-04-20
OutgoingResponse: note doubled app error

jcorbin:


node/outgoing_response.js
```diff
@@ -194,8 +194,15 @@ TChannelOutgoingResponse.prototype.sendOk = function sendOk(res1, res2) {
 
 TChannelOutgoingResponse.prototype.sendNotOk = function sendNotOk(res1, res2) {
     var self = this;
-    self.setOk(false);
-    self.send(res1, res2);
+    if (self.state === States.Error) {
+        self.logger.error('cannot send application error, already sent error frame', {
```
Raynos: can we print `self.arg2` and `self.arg3` as well to see what the first Error was ?

Raynos:
lgtm.

PR #452, opened 2015-04-21
Handle socket closed errors correctly

jcorbin:


node/connection_base.js
```diff
@@ -157,7 +157,7 @@ TChannelConnectionBase.prototype.resetAll = function resetAll(err) {
         err = new Error('unknown connection reset'); // TODO typed error
     }
 
-    var isError = err.type !== 'tchannel.socket-closed';
+    var isError = err.type.indexOf('tchannel.socket') !== 0;
     self.logger[isError ? 'warn' : 'info']('resetting connection', {
```
kriskowal: I bet that looking up a method by name is a speed bump, and maybe even switches the internal representation of logger from hidden class to hash table. Looking into it.

kriskowal: My latter suspicion is invalid. The logger does not get turned into a hash table. `%HasFastProperties(logger)` with `--allow-natives-syntax` after this pattern is still `true`.

PR #453, opened 2015-04-21
change stream by using deque instead of string

junchaowu:
to improve the performance

### DIFF
--- a/python/tchannel/tornado/stream.py
+++ b/python/tchannel/tornado/stream.py
@@ -11,6 +11,8 @@ try:
 except ImportError:  # pragma: no cover
     from toro import Condition
 
+from collections import deque
+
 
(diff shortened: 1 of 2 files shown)


junchaowu:
@blampe @abhinav @breerly

python/tchannel/tornado/stream.py
```diff
@@ -71,7 +77,7 @@ def write(self, chunk):
         if self._state == StreamState.completed:
             raise StreamingException("Stream has been closed.")
 
-        self._stream += chunk
+        self._stream.append(chunk)
```
abhinav: In both appends, you probably want to make sure that you append only if the string is non-empty.

abhinav:
:ship:

PR #454, opened 2015-04-21
Fix test hang

jcorbin:
This was an interesting one, at least affecting `test/send.js`:
- .request contained host option
- so TChannelRequest#choosePeer was always choosing the same (self) peer
- so even past channel / peers / peer / connection / socket closes, it kept trying to next-tick retry


jcorbin:
tests pass locally, travis seems swapmed, not waiting

kriskowal:
Nice catch.

PR #456, opened 2015-04-21
Implement unhealthy state and transitions

kriskowal:
Unit test peer state transitions

### DIFF
--- a/node/peer_states.js
+++ b/node/peer_states.js
@@ -22,13 +22,65 @@
 
 var inherits = require('util').inherits;
 
+var errors = require('./errors');
 var TChannelPeerState = require('./peer_state');
 
 module.exports.TChannelPeerHealthyState = TChannelPeerHealthyState;
+module.exports.TChannelPeerUnhealthyState = TChannelPeerUnhealthyState;
 
+/*
+ * Collectively, the health states receive additional options through the peer
+ * options:
+ * - maxErrorRate (error rate to go from healthy to unhealthy)
+ * - minResponseCount (response count to go from unhealthy to healthy)
+ * - TODO
+ *
+ * They also inherit:
+ * - channel.timers
+ * - channel.random
+ */
+
+var symptoms = {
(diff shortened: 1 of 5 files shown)


kriskowal:
r @jcorbin

node/peer_states.js
```diff
@@ -22,13 +22,65 @@
...
+ * - TODO
+ *
+ * They also inherit:
+ * - channel.timers
+ * - channel.random
+ */
+
+var symptoms = {
+    'BadRequest': false, // not an indicator of bad health
+    'Cancelled': false, // not an indicator of bad health
```
breerly: Are there not constants that can be used in place of these strings?

kriskowal: There are, but as it happens there is nothing to be gained by a level of indirection through that lookup table. We are getting the code as a string from the error classifier.

breerly: Aaah, makes sense ;)

kriskowal:
Looks like some commits got redrawn in the rebase. Fixing.

node/test/send.js
```diff
@@ -203,57 +206,64 @@ allocCluster.test('request().send() to a pool of servers', 4, function t(cluster
             res.sendOk(arg2, arg3 + ' served by ' + chanNum);
         });
         channel.peers.add(chan.hostPort);
+        var peer = channel.peers.get(chan.hostPort);
+        var conn = peer.connect(chan.hostPort);
+        console.log(chan.hostPort);
```
jcorbin: spam

node/test/send.js
```diff
@@ -203,57 +206,64 @@ allocCluster.test('request().send() to a pool of servers', 4, function t(cluster
...
+        conn.on('identified', ready.signal);
```
jcorbin: this whole pattern might be worthy of getting hauled back into allocCluster or an ancillary utility function around it.

jcorbin:
lgtm, spammy test is spammy

PR #457, opened 2015-04-22
Drafting Circuit Breaking design.

breerly:
The following PR attempts to:

* Introduce the problem area and define Circuit Breaking.
* Choose a simple vocabulary we can use when talking about the problem.
* Outlines a implementation that prevents cascading failure.
* Details failure scenarios and what their outcome would be under this design.
* Highlight challenges we will no doubt face when attempting to implement/rollout.

Sorry for the typo's and grammatical errors, I didn't get this out till late :sleepy: 

View the rendered markdown here:
https://github.com/uber/tchannel/blob/docs_circuit_breaking/docs/circuit-breaking.md


leizha:
this design looks good to me.
please note that it doesn't have an OPEN state which completely marks down the service for a fixed period of time. Instead, it keeps sending probe calls to the service 1 per second. I think it gives us more insight during the the time service being unhealthy.

breerly:
That's true, the `Unhealthy` state detailed in the draft is more analogous to `HALF OPEN` state. This is an intentional design decision that favors using actual traffic for health requests instead of replaying traffic as a probe.

abhinav:
lgtm

jwolski:
questions:
1) was there a deliberate decision made about passive probe checks vs. explicit health checks upon failure (akin to lbpool)? there's some obvious scalability challenges with explicit health checking, but are there any non-obvious reasons?
2) since individual service instances are registering, have you thought about circuit breaking at the service instance level instead of the service level?

jwolski:
3) seems like service response times can also be used as criteria to trip the circuit. have you considered this for next phases (beyond v1)?

breerly:
Good questions @jwolski 

1. It was deliberate - the thought was that passive healthchecks (using actual traffic) was the best way to state change circuits on the very specific `SOURCE->SERVICE::ENDPOINT` criteria.
2. Initially I actually did start by thinking that a service would have a circuit that looked like this `SOURCE->NODE::SERVICE::ENDPOINT` where `NODE` was the hostname of the machine that the service lived on but then I realized that node failure (hardware failure) is already handled by Hyperbahn - if a node dies then all the services on it lose its connection to Hyperbahn causing that node to be "deregistered".
3. Totally. Timeouts act as failures and contribute to circuit health. Average timeouts could totally be factored in if we see a use for it.

I created this PR mostly to layout the problem area and so we could collaborate on how to attack it. I imagine @kriskowal will start with something reasonably simple and gradually enhance it to cover all the scenarios we think of.

jwolski:
I feel satisfied with these answers.

kriskowal:
Work in progress is here: https://github.com/uber/tchannel/pull/456

First pass is just a neighboring peer health monitor using the state machine described here. Still searching for the socket to plug `source->service:endpoint` health state machines into.

Nit: particularly with regard to `source->service`, we should use words that make pairs, source to target, client to server, requester to responder, or caller to callee.

breerly:
@kriskowal that's really good feedback, give me a bit to update the terminology :)

junchaowu:
:ship:

mranney:
This looks great to me.

I think we should consider adding a fully open state as @leizha suggested, but we can do this in the future.

docs/circuit-breaking.md
```diff
@@ -0,0 +1,119 @@
...
+
+The following diagram illustrates Hyperbahn's Circuit Breaking implementation:
+
+![Circuit Breaking](images/circuit_breaking.png)
+[Click Here For Full Size Diagram](images/circuit_breaking.png)
+
+
+* Each Service that is registered on the Hyperbahn maintains a set of Circuits that are responsible for denying traffic when the Circuit is Unhealthy.
+* The Circuits live on the Service's Egress nodes so that traffic can be denied before reaching the Service. The state of the Circuits can be optionally gossiped between that Service's Egress nodes to enhance the Circuit's sensitivity.
+* The Service maintains a Circuit for every Caller it has, times every endpoint it exposes, with a name format of `SOURCE->SERVICE::ENDPOINT`. In our diagram, this means that even though the `CatsForCharity->Petshop::listCats` Circuit is Unhealthy and denying traffic, the `PetBreeders->Petshop::listCats` Circuit is Healthy and continues to allow traffic to flow.
```
blampe: Nit: In the protocol we don't give any semantic meaning to arg1, but here (and in other places) we're assuming it's an endpoint, triggering behavior like circuit breaking and request dispatching based on that. I wonder if it's worth clarifying vocabulary in the protocol.

blampe:
This looks great.

docs/circuit-breaking.md
breerly: That's an interesting point.

@Raynos @jcorbin has `arg1` in practice always always always been endpoint?

PR #460, opened 2015-04-22
Allow client only channels

mmihic:
Also makes channel api symmetric with net/http

R: @abhinav / @weikai77

### DIFF
--- a/golang/channel.go
+++ b/golang/channel.go
@@ -21,12 +21,23 @@ package tchannel
 // THE SOFTWARE.
 
 import (
+	"errors"
 	"fmt"
-	"golang.org/x/net/context"
 	"net"
 	"os"
 	"path/filepath"
+	"sync"
 	"time"
+
+	"golang.org/x/net/context"
+)
+
+var (
+	errAlreadyListening = errors.New("channel already listening")
+)
+
+const (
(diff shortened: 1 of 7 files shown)


golang/channel.go
```diff
@@ -21,12 +21,23 @@ package tchannel
...
+
+	"golang.org/x/net/context"
+)
+
+var (
+	errAlreadyListening = errors.New("channel already listening")
+)
+
+const (
+	clientOnlyHostPort = "0.0.0.0:0"
```
abhinav: IMO, ephemeral might be a better term for this than client-only.

golang/channel.go
```diff
@@ -21,12 +21,23 @@ package tchannel
...
 	"os"
 	"path/filepath"
+	"sync"
 	"time"
+
+	"golang.org/x/net/context"
+)
+
+var (
+	errAlreadyListening = errors.New("channel already listening")
```
abhinav: Don't you want this to be a publicly accessible value so that users can match on it?

abhinav:
Looks good besides minor issues mentioned above.

golang/channel.go
mmihic: This is a programming error, so not something you'd branch on.  Usually logic errors are kept unexported.

PR #463, opened 2015-04-22
Adding Caller Name as Transport Header to Protocol

breerly:
This will be useful for:

* Stats - Answering questions like "How much traffic was being sent to Service A from Service B?".
* Circuit Breaking - We'll need this in order to create `SOURCE->SERVICE::ENDPOINT` Circuits.
* Tracing - We should be able to use this with Zipkin.

### DIFF
--- a/docs/protocol.md
+++ b/docs/protocol.md
@@ -558,6 +558,7 @@ name  | req | res | description
 `as`  | Y   | Y   | the Arg Scheme
 `cas` | Y   | N   | Claim At Start
 `caf` | Y   | N   | Claim At Finish
+`cn`  | Y   | N   | Caller Name
 `re`  | Y   | N   | Retry Flags
 `se`  | Y   | N   | Speculative Execution
 `fd`  | Y   | Y   | Failure Domain
@@ -589,6 +590,10 @@ Send a claim message when work is started.
 Value is string "host:port".
 
 Send claim message to `host:port` when response is being sent.
+
+### Transport Header `cn` -- Caller Name
+
+Value is name of the service making the call.
 
 ### Transport Header `re` -- Retry Flags
 


kriskowal:
This looks good to me.

jcorbin:
:+1:

breerly:
@Raynos @rf @mranney - Since this is a protocol change could I get your guys' stamp as well?

Raynos:
lgtm.

docs/protocol.md
```diff
@@ -590,6 +591,10 @@ Value is string "host:port".
 
 Send claim message to `host:port` when response is being sent.
 
+### Transport Header `cn` -- Caller Name
+
+Value is name of the service making the call.
```
mmihic: other information of this sort (such as process name and host_port) is sent in the init req/res, not as part of every call.  can we not do the same with the caller name?

breerly: This should correlate with the calls to register that the service is making. You think that's enough to lookup all desired info? My thought is that the service name is the key into this type of data.

mmihic: Having the service name makes sense, my question is whether it needs to be sent with every request vs sent once during initialization handshake.

breerly: Aaah, interesting. Not sure which is best, my thought was that this was more practical. @jcorbin @Raynos what do you guys think?

Raynos: No you cannot put this in init. For autobahn, various call reqs will have cn set to ringpop; others will have cn set to hyperbahn; or autobahn;

breerly: Thanks for the clarification.

jcorbin: Yeah, there may be multiple services registering or calling on a single connection.

Certainly this will be the case for internal hyperbahn hopes (Entry -> Exit, Exit -> CalledService).

PR #467, opened 2015-04-23
Explicate node streaming mode

jcorbin:
Break out Streaming{Out,In}{Request,Response} subclasses

r @kriskowal @Raynos

### DIFF
--- a/node/streaming_out_request.js
+++ b/node/streaming_out_request.js
@@ -1,0 +1,68 @@
+// Copyright (c) 2015 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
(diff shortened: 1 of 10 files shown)


node/streaming_out_request.js
```diff
@@ -0,0 +1,68 @@
...
+    function onFrame(parts, isLast) {
+        self.sendParts(parts, isLast);
+    }
+
+    function onFinish() {
+        self.emit('finish');
+    }
+}
+
+inherits(StreamingOutRequest, TChannelOutRequest);
```
Raynos: Can we set `type` to 'tchannel.outgoing-response.streaming'; ?

Raynos:
lgtm.

Consider setting the types for the 4 streaming classes.

PR #468, opened 2015-04-23
Pivot frame send injection

jcorbin:
Use explicit {v2,self}Out{Request,Response} classes rather than injecting frame senders with a closure.

r: @Raynos @kriskowal

### DIFF
--- a/node/out_response.js
+++ b/node/out_response.js
@@ -58,6 +58,30 @@ inherits(TChannelOutResponse, EventEmitter);
 
 TChannelOutResponse.prototype.type = 'tchannel.outgoing-response';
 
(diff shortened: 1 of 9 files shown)


node/out_response.js
```diff
@@ -126,7 +123,7 @@ TChannelOutResponse.prototype.sendCallResponseContFrame = function sendCallRespo
             }));
             break;
         case States.Streaming:
-            self.sendFrame.callResponseCont(args, isLast);
+            self._sendCallResponseCont(args, isLast);
```
Raynos: should we define these methods on the base class and make their body throw not implemented errors ?

Raynos:
Looks reasonable.

Raynos:
lgtm.

PR #470, opened 2015-04-23
add event hook into tchannel

junchaowu:
@breerly @abhinav @blampe


python/tchannel/event.py
```diff
@@ -0,0 +1,70 @@
...
+    def receive_request(self, context):
+        """event hook for receiving request"""
+        pass
+
+    def receive_response(self, context):
+        """event hook for receiving response"""
+        pass
+
+
+class EventManager(object):
```
breerly: Please don't ever end a class with `Manager`. Suggest renaming to something specific, like `EventEmitter`.

breerly:
Tests?

python/tchannel/event.py
```diff
@@ -0,0 +1,70 @@
+from __future__ import absolute_import
+from enum import IntEnum
+import logging
+from .exceptions import InternalEventException
+
+log = logging.getLogger('tchannel')
+
+
+class EventType(IntEnum):
```
breerly: Can you detail what these variable actually are and how they are used?

python/tchannel/event.py
```diff
@@ -0,0 +1,70 @@
...
+    def receive_request(self, context):
+        """event hook for receiving request"""
+        pass
+
+    def receive_response(self, context):
+        """event hook for receiving response"""
+        pass
+
+
+class EventEmitter(object):
```
breerly: Quick docblock here would be nice.

breerly:
Without seeing an actual real-life usage of this, it's a bit difficult to judge whether we reached the optimal API, but I'm fine with shipping it for now. Peez add tests.

:shipit:

python/tchannel/event.py
```diff
@@ -0,0 +1,70 @@
...
+
+
+class EventHook(object):
+    """provide all event hook interfaces
+
+    Customized Hook should should inherit from EventHook class and implement
+    the events' hooks that it wants to listen.
+
+    Example::
+        TraceHook(EventHook):
```
abhinav: Add empty line between code block and the line ending with `::`

python/tchannel/event.py
```diff
@@ -0,0 +1,70 @@
...
+            def send_request(self, context):
+                ....
+
+    """
+
+    def __init__(self, priority=0):
+        self.priority = priority
+
+    def __cmp__(self, other):
+        return cmp(self.priority, other.priority)
```
abhinav: You don't need cmp on something that's supposed to represent a computation of some kind.

python/tchannel/event.py
```diff
@@ -0,0 +1,70 @@
...
+    def __init__(self):
+        self.hooks = []
+
+    def register_hooks(self, hook):
+        assert isinstance(hook, EventHook)
+        self.hooks.append(hook)
+
+    def fire(self, event, *args, **kwargs):
+        if not isinstance(event, EventType):
+            raise InternalEventException()
```
abhinav: Add useful error messages when raising exceptions

python/tchannel/tornado/connection.py
```diff
@@ -75,7 +76,7 @@ class TornadoConnection(object):
     CALL_REQ_TYPES = frozenset([Types.CALL_REQ, Types.CALL_REQ_CONTINUE])
     CALL_RES_TYPES = frozenset([Types.CALL_RES, Types.CALL_RES_CONTINUE])
 
-    def __init__(self, connection):
+    def __init__(self, connection, tchannel=None):
```
abhinav: You access the variable only as as `tchannel.event_emitter`. That suggests you should be injecting the event emitter here instead.

python/tchannel/event.py
```diff
@@ -0,0 +1,70 @@
...
+        """event hook for receiving response"""
+        pass
+
+
+class EventEmitter(object):
+    def __init__(self):
+        self.hooks = []
+
+    def register_hooks(self, hook):
+        assert isinstance(hook, EventHook)
```
abhinav: You're registering a single hook. The method should be called register_Hook.

junchaowu:
ignore the style issue in ttypes.py . This is thrift generated code.

python/tchannel/tornado/tchannel.py
```diff
@@ -161,13 +173,46 @@ def handle_stream(self, stream, address):
...
+    def __init__(self, tchannel, parent_tracing=None):
+        self._tchannel = tchannel
+        self.parent_tracing = parent_tracing
+
+    @property
+    def closed(self):
+        return self._tchannel.closed
+
+    @property
+    def hostport(self):
```
breerly: +1 law of demeter

python/tchannel/tornado/tchannel.py
```diff
@@ -161,13 +173,46 @@ def handle_stream(self, stream, address):
...
+
+class TChannelProxy(object):
+    __slots__ = ('_tchannel', 'parent_tracing')
+
+    def __init__(self, tchannel, parent_tracing=None):
+        self._tchannel = tchannel
+        self.parent_tracing = parent_tracing
+
+    @property
+    def closed(self):
```
breerly: +1 law of demeter

python/tchannel/event.py
```diff
@@ -0,0 +1,72 @@
+from __future__ import absolute_import
```
blampe: license header

newline after `__future__` import

python/tchannel/event.py
junchaowu: update comments

python/tchannel/tornado/connection.py
junchaowu: It will work for current needs. If we consider future extension, we may need pass more things from tchannel. That's why I pass tchannel

python/tchannel/event.py
junchaowu: didn't setup git properly. since next commit, it will automatically add license

python/tchannel/event.py
```diff
@@ -0,0 +1,92 @@
...
+
+    Events:
+
+    send_request: before client sends request
+
+    send_response: before server sends response
+
+    receive_request: after server receive request
+
+    receive_response: after server receive response
```
abhinav: There's a lack of consistency here. One of them is for a client operation and all others are for server operations.

python/tchannel/event.py
```diff
@@ -0,0 +1,92 @@
...
+
+    Example::
+        TraceHook(EventHook):
+            def send_request(self, context):
+                ....
+
+    """
+
+    def send_request(self, context):
+        """event hook for sending request"""
```
abhinav: The hooks are not for sending/receiving anything. They're called before/after certain operations happen. The naming and the docs should reflect that.

python/tchannel/event.py
```diff
@@ -0,0 +1,92 @@
...
+        pass
+
+    def send_response(self, context):
+        """event hook for sending response"""
```
abhinav: Also, the docs need to contain information on what argument you'll receive. As mentioned before, "context" doesn't mean anything here. Is it the Context as defined in tchannel.context? Is it some arbitrary object. Adding new parameters is cheap. Consider specifying actual separate objects instead of arbitrary parameter objects.

python/tchannel/event.py
junchaowu: sorry. typo. receive_response should be client.

python/tchannel/event.py
```diff
@@ -0,0 +1,92 @@
...
+        pass
+
+    def receive_request(self, context):
+        """event hook for receiving request"""
+        pass
```
abhinav: Also, docstrings are full sentences. They should start with a capital letter and the first line must be a one-line summary ending with a period.

python/tchannel/event.py
```diff
@@ -0,0 +1,92 @@
...
+        self.hooks.append(hook)
+
+    def fire(self, event, *args, **kwargs):
+        # TODO find proper hook name
+        event_hook_name = event.name
+        for hook in self.hooks:
+            try:
+                getattr(hook, event_hook_name)(*args, **kwargs)
+            except Exception as e:
+                log.error(e.message)
```
abhinav: This is janky. Instead of this, you could make EventEmitter conform to the EventHook interface. `send_request` would call `send_request` on all registered hooks.

python/tchannel/tornado/connection.py
abhinav: When that happens, we can pass that information down. This just seems like a bad violation of concerns.

python/tchannel/tornado/connection.py
```diff
@@ -75,13 +76,18 @@ class TornadoConnection(object):
...
         assert connection, "connection is required"
 
         self.closed = False
         self.connection = connection
+        try:
+            (self.remote_host,
+             self.remote_host_port) = connection.socket.getsockname()
+        except:
+            (self.remote_host, self.remote_host_port) = ("0.0.0.0", 0)
```
abhinav: 1. Don't do catch-all like that.
2. If getsockname actually ever throws an exception, that's an actual problem and should be addressed. Not silently suppressed.

python/tchannel/tornado/connection.py
```diff
@@ -518,6 +528,11 @@ def _stream(self, context, message_factory):
 
     @tornado.gen.coroutine
     def post_response(self, response):
+
+        # event: send_response
+        if self.tchannel:
+            self.tchannel.event_emitter.fire(EventType.send_response, response)
```
abhinav: So this is a good place to expand on that idea of EventEmitter conforming to the EventHook interface:
If EventEmitter conformed to the EventHook interface and acted as a composite EventHook of sorts, we could remove the knowledge of EventEmitter from TornadoConnection completely. It would only depend on a single EventHook-like object. In this place where you fire the event, it would just call "self.hook.send_response(..)".

When we injected an EventEmitter here, this send_response call would cause the EventEmitter to loop through all EventHooks and call send_response on them.

That'll would also remove the need of having this EventType enum which doesn't appear to have much value.

python/tchannel/tornado/connection.py
```diff
@@ -554,4 +574,19 @@ def send_request(self, request):
         future = tornado.gen.Future()
         self._outstanding[request.id] = future
         self.post_request(request)
+
+        def adapt_tracing(f):
+            # fetch the request tracing for response
+            f.result().tracing = request.tracing
```
abhinav: No. You can't be mutating values like that. Block on that future's completion, create a new result, and produce that inside a future.

abhinav: Basically, your hooks shouldn't be modifying values like that. If you have a use case where you need to modify a value, that's not a hook. That's part of the main flow of the program.

python/tchannel/tornado/dispatch.py
```diff
@@ -88,17 +93,25 @@ def handle_call(self, request, connection):
...
         else:
-            response = Response(id=request.id, connection=connection)
-            yield self._call_endpoint(endpoint, request, response)
-
-    def route(self, rule, **opts):
+            response = Response(id=request.id,
+                                tracing=request.tracing,
+                                connection=connection)
+
+            from .tchannel import TChannelProxy
```
abhinav: Yeah, if you need to do this, the class shouldn't be defined in the same class as TChannel.

python/tchannel/tornado/dispatch.py
```diff
@@ -88,17 +93,25 @@ def handle_call(self, request, connection):
...
+    def route(self, rule):
         def decorator(handler):
-            self.register(rule, handler, **opts)
+            self.register(rule, handler)
             return handler
 
         return decorator
 
-    def register(self, rule, handler, **opts):
+    def register(self, rule, handler):
```
abhinav: You need to update docstrings to reflect this interface change

python/tchannel/tornado/dispatch.py
```diff
@@ -54,10 +52,10 @@ def __init__(self):
...
     @tornado.gen.coroutine
-    def _call_endpoint(self, endpoint, request, response):
+    def _call_endpoint(self, endpoint, request, response, proxy):
         try:
             yield tornado.gen.maybe_future(
-                endpoint.handler(request, response, endpoint.opts))
+                endpoint.handler(request, response, proxy))
         except:
             # TODO send application error
             pass
```
abhinav: You've deleted the Endpoint class. What's this `.handler` on?

Also, remove that catch-all  `except:`.

python/tchannel/tornado/connection.py
junchaowu: I don't see any difference between this and create a new result. Also I am appending tracing result which is None.

python/tchannel/tornado/dispatch.py
```diff
@@ -78,6 +76,13 @@ def handle_call(self, request, connection):
             request.endpoint += chunk
             chunk = yield request.argstreams[0].read()
 
+        # event: receive_request
+        if connection.tchannel:
+            request.tracing.name = request.endpoint
+            connection.tchannel.event_emitter.fire(
+                EventType.receive_request, request
```
abhinav: You shouldn't be storing a `TChannel` in `TornadoConnection`. Nor should you be accessing that `TChannel` inside the request handler through the `TornadoConnection`. Inject the necessary information (`EventEmitter` or `EventHook`) through other means instead of playing the 5-level-of-attribute-access game.

python/tchannel/tornado/connection.py
junchaowu: I thought about this before. I didn't do this way because it requires the hook to have each event methods. In side the require methods, they basically have the same logic. 

I take the pattern that you write in the BaseRequestHandler method, doing "handle_"+name style.  I think this pattern is quite concise and reuse the code.

I agree both will work.

python/tchannel/tornado/message_factory.py
```diff
@@ -60,7 +63,11 @@ def build_raw_request_message(self, request, args, is_completed=False):
             message = CallRequestMessage(
                 flags=request.flags,
                 # ttl=request.ttl,
-                # tracing=request.tracing,
+                tracing=Tracing(request.tracing.span_id,
+                                request.tracing.parent_span_id,
+                                request.tracing.trace_id,
+                                request.tracing.traceflags
+                                ),
```
abhinav: Move `)` to previous line. Usually, `)` goes on a separate line only if the `(` was on a separate line too. Like so,

    Tracing(
        arg1,
        arg2,
    )

python/tchannel/tornado/dispatch.py
junchaowu: I know this is not good. But based on what's TChannelProxy doing, they should be in the same file.

Although it is not good.

python/tchannel/tornado/message_factory.py
```diff
@@ -154,11 +165,21 @@ def build_request(self, message, message_id=None):
 
         args = self.prepare_args(message)
 
+        tracing = Trace(
+            trace_id=message.tracing.trace_id,
+            span_id=message.tracing.span_id,
+            parent_span_id=message.tracing.parent_id,
+            endpoint=Endpoint(self.connection.remote_host,
+                              self.connection.remote_host_port,
```
abhinav: Again, given that you need just those two attributes, pass only those in instead of consuming `connection`.

python/tchannel/tornado/connection.py
abhinav: The difference is that you cannot guarantee ordering. If I have two consumers of the future, there's no guarantee which one will be called first. If one of them mutates the future, the second one may or may not get the mutated objects.

You don't modify objects out of band like that. You're opening yourself up to all kinds of race conditions.

python/tchannel/tornado/dispatch.py
junchaowu: why do we need to review catch-all

python/tchannel/tornado/peer.py
```diff
@@ -409,12 +410,13 @@ def __init__(self, peer, service):
 
         self.peer = peer
         self.service = service
+        self.parent_tracing = kwargs.get('parent_tracing', None)
```
abhinav: Just use `parent_tracing=None` in the parameter list.

python/tchannel/tornado/peer.py
```diff
@@ -434,14 +436,37 @@ def send(self, arg1, arg2, arg3):
             maybe_stream(arg1), maybe_stream(arg2), maybe_stream(arg3)
         )
 
+        # hack to get endpoint from arg_1 for trace name
+        arg1.close()
+        chunk = yield arg1.read()
+        endpoint = ''
+        while chunk:
+            endpoint += chunk
+            chunk = yield arg1.read()
```
abhinav: Use `read_full` defined in `tchannel.tornado.stream`

python/tchannel/tornado/peer.py
```diff
@@ -434,14 +436,37 @@ def send(self, arg1, arg2, arg3):
...
+
+        parent_span_id = (self.parent_tracing.span_id if
+                          self.parent_tracing else None)
+
+        trace_id = (self.parent_tracing.trace_id if
+                    self.parent_tracing else None)
```
abhinav: Use `if self.parent_tracing: [...]` instead of this

python/tchannel/tornado/tchannel.py
```diff
@@ -151,7 +163,8 @@ def __init__(self, tchannel):
     @tornado.gen.coroutine
     def handle_stream(self, stream, address):
         log.debug("New incoming connection from %s:%d" % address)
-        conn = StreamConnection(connection=stream)
+        # TODO peer how do
```
abhinav: You can remove this TODO. It's taken care of. I probably forgot to remove ti.

python/tchannel/tornado/connection.py
```diff
@@ -75,13 +76,18 @@ class TornadoConnection(object):
...
 
-    def __init__(self, connection):
+    def __init__(self, connection, tchannel=None):
         assert connection, "connection is required"
 
         self.closed = False
         self.connection = connection
+        try:
+            (self.remote_host,
+             self.remote_host_port) = connection.socket.getsockname()
```
abhinav: The remote_host is set correctly when we receive a handshake. We expect the information in the handshake. It includes both, the address and the port. So, for a tornado connection, remote_host is the address (with port) on which you can reach that service.

python/tchannel/tornado/tchannel.py
```diff
@@ -159,13 +172,57 @@ def handle_stream(self, stream, address):
...
+            "Successfully completed handshake with %s:%s (%s)",
             conn.remote_host,
+            conn.remote_host_port,
             conn.remote_process_name,
         )
 
-        self.tchannel.peers.get(conn.remote_host).register_incoming(conn)
+        self.tchannel.peers.get(
+            "%s:%s" % (conn.remote_host,
+                       conn.remote_host_port)
```
abhinav: When the handshake is received, the remote_host is set to whatever was received in the handshake -- which is the address at which the peer can be reached. You don't need to manually concatenate this. This will probably result in `$ipAddress:$port:$port`.

python/tchannel/tornado/tchannel.py
```diff
@@ -159,13 +172,57 @@ def handle_stream(self, stream, address):
...
+class TChannelProxy(object):
+    """TChannel Proxy with additional runtime info
+
+    TChannelProxy contains parent_tracing information which is created by
+    received request.
+
+    TChannelProxy will be used as one parameter for the request handler.
+
+    Example::
+        def handler(request, response, proxy):
```
abhinav: Newline between code block and previous paragraph

python/tchannel/tornado/dispatch.py
abhinav: I looked at the implementation. There's no reason it couldn't be moved to this file. Especially because this is the only place it is consumed. 

There's no hard and fast rule dictating that classes need to be in a file named similar to their class name.

python/tchannel/tornado/tchannel.py
```diff
@@ -159,13 +172,57 @@ def handle_stream(self, stream, address):
...
+        return self._tchannel.hostport
+
+    def host(self, handler):
+        return self._tchannel.host(handler)
+
+    def listen(self):
+        return self._tchannel.listen()
+
+    def request(self, hostport=None, service=None, **kwargs):
+        kwargs['parent_tracing'] = self.parent_tracing
```
abhinav: Consider using [`kwargs.setdefault`](https://docs.python.org/2/library/stdtypes.html#dict.setdefault) instead. That way, the caller can specify `parent_tracing` manually.

python/tchannel/zipkin/zipkin_trace.py
```diff
@@ -0,0 +1,70 @@
...
+from tchannel.zipkin import annotation
+from tchannel.zipkin.tracers import DebugTracer, TChannelZipkinTracer
+
+
+class ZipkinTraceHook(EventHook):
+    """generate zipkin-style span for tracing"""
+
+    def __init__(self, tchannel=None, dst=None):
+        if tchannel:
+            # TChannelZipkinTracer generates 64 based encoded span
```
abhinav: base 64 encoded

python/tests/integration/conftest.py
```diff
@@ -26,8 +26,10 @@
 
 
 @pytest.yield_fixture
-def tchannel_server(random_open_port):
-    with TChannelServerManager(random_open_port) as manager:
+def tchannel_server(random_open_port, dispatcher=None):
+    with TChannelServerManager(random_open_port,
+                               dispatcher
+                               ) as manager:
```
abhinav: Again, `)` on separate line only if the `(` was on a separate line.

python/tchannel/tornado/dispatch.py
abhinav: The catch all hides away all errors, including critical errors that should not be caught. In this specific case, the widest we should go is `Exception`, and we should log details even if we're currently ignoring the actual exception.

Also see: http://stackoverflow.com/questions/18982610/difference-between-except-and-except-exception-as-e-in-python

python/tchannel/tornado/tchannel.py
junchaowu: remote_host is only the ip. 
remote_host_port is for port. I changed it in the connection.py

python/tchannel/tornado/connection.py
junchaowu: First if someone uses multi consumers for one future, I don't think they are using future in right way. Second let's step back. If one of the consumers mutates the future, then the second one may not get the same object regardless my change for the tracing.

python/tchannel/tornado/dispatch.py
junchaowu: You can fire a task. It is not ideal put in this patch.

python/tchannel/tornado/dispatch.py
junchaowu: when we initialize TChannelProxy, we need to pass tchannel. I understand your concern, currently I haven't found a better to do that. We can add a TODO here to track this issue.

python/tchannel/tornado/connection.py
abhinav: No. You're not using a future wrong if you have multiple consumers waiting on it. It's part of the design to have multiple consumers waiting on the result of an asynchronous operation. We do it all the time. You can set up a callback on a future to perform some cleanup in the calling context based on the result of the future and still return that same future up the stack where it's consumed as part of the main flow. 

And what I'm asking here is to not mutate the object at all. You shouldn't be mutating an object out of band in a callback that has no guarantee of ordering.

python/tchannel/tornado/dispatch.py
abhinav: You're touching some code and a line below it there's a known bad practice. All you need to do to fix it is add the word `Exception`. Why is making the fix not ideal for this patch.

python/tchannel/tornado/connection.py
junchaowu: Got you. I will create another future. Set that future inside this.

junchaowu: :)

python/tchannel/tornado/dispatch.py
junchaowu: all I change is to make the argument consistent. Actually this piece of code is not executed and override by the subclass. I plan to delete in next small patch. But I really don't want to do anything beyond the purpose of this patch. Don't be stuck at code review by adding more works

abhinav: If you know that it's not being used, delete it now. If it is being used, fix the catch-all. It doesn't take a lot of effort to do either thing in this CR.

python/tchannel/tornado/peer.py
breerly: +1 here

python/tchannel/event.py
```diff
@@ -0,0 +1,109 @@
...
+
+
+class EventType(IntEnum):
+    """Types to represent system events
+
+    Events:
+
+    send_request: before client sends request
+
+    send_response: before server sends response
```
blampe: Why is this happening *before* the response is sent? It seems like *after* would be more useful, although eventually we'll probably want both.

python/tchannel/tornado/tchannel.py
```diff
@@ -78,6 +80,16 @@ def __init__(self, hostport=None, process_name=None):
         # RequestHandler to handle incoming calls.
         self._handler = None
 
+        # register event hooks
+        self.event_emitter = EventEmitter()
```
blampe: Why put `event_emitter` on the `tchannel` instead of the `dispatcher`?

junchaowu: when client sends request, it doesn't goes into dispatcher

junchaowu: when client receives response, it doesn't goes into dispatcher either

python/tchannel/event.py
junchaowu: I moved it to after send response. If we need more events, we can add more event types for them.

PR #471, opened 2015-04-23
Make `serviceName` not go into requestDefaults

Raynos:
This removes the "feature" from subChannels where we put
the serviceName of a sub channel into the requestDefaults

In this PR I also wrote a test for the peers feature.

r: @kriskowal @jcorbin

### DIFF
--- a/node/channel.js
+++ b/node/channel.js
@@ -83,7 +83,6 @@ function TChannel(options) {
     self.serviceName = '';
     if (self.options.serviceName) {
         self.serviceName = self.options.serviceName;
-        self.requestDefaults.serviceName = self.serviceName;
         delete self.options.serviceName;
     }
 
--- a/node/test/relay.js
+++ b/node/test/relay.js
@@ -24,7 +24,7 @@ var allocCluster = require('./lib/alloc-cluster');
 var TChannel = require('../channel');
 var RelayHandler = require('../relay_handler');
 
-allocCluster.test('request retries', {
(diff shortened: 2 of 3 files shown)


node/peers.js
```diff
@@ -136,12 +136,6 @@ TChannelPeers.prototype.delete = function del(hostPort) {
     var self = this;
     var peer = self._map[hostPort];
     delete self._map[hostPort];
-    if (self.subChannels) {
```
jcorbin: This should be corrected not removed.  The invariant we want is that if a peer is in a sub channel, then it's on the top channel too.  So when you remove a peer from a top channel, it must be removed from all sub channels.

The clear method is even more/differently wrong in that it doesn't have any affordance.

Raynos: This does not match the `add()` semantics;

If you `add()` a peer to subchannel A; we add it to top channel.

What it sounds like we want for sub channel is;

 - if you `remove()` a peer from a sub channel AND no other sub channel has that peer; remove it from the top channel.

Raynos:
@jcorbin all the peers stuff is a seperate PR; I think this one is ready to go in now;

jcorbin:
Updated.  Not only is this clearer / nicer, but it has the potential to go faster: we can do an anon-object to explicit prototyped object conversion once when we create the channel, then the "extension" per-request gets really cheap.

#detailsmatter #uxmatters

jcorbin:
forgot to add: lgtm, merge at will

PR #472, opened 2015-04-24
Fix peers.delete() and write tests.

Raynos:
There was a bug in peers.delete(); I wrote a bunch
of tests including a failing test.

r: @jcorbin @kriskowal


node/test/peers.js
```diff
@@ -0,0 +1,221 @@
+'use strict';
```
jcorbin: license missing

jcorbin:
lgtm, merging for so can do a similar thing

PR #473, opened 2015-04-24
Fix peers clear

jcorbin:
Follow on to #472 

r @Raynos

### DIFF
--- a/node/test/peers.js
+++ b/node/test/peers.js
@@ -24,8 +24,6 @@ var parallel = require('run-parallel');
 
 var allocCluster = require('./lib/alloc-cluster');
 
-// clear()
-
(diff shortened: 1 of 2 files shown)


node/test/peers.js
```diff
@@ -211,8 +209,76 @@ allocCluster.test('delete peer() on top channel', {
...
+
+        assert.equal(bob1.peers.keys().length, 0,
+            'bob1 has no peers');
+        assert.equal(bob2.peers.keys().length, 0,
+            'bob2 has no peers');
+
+        assert.end();
+    }
+
+    function thunkSend(channel, reqOpts, arg1, arg2, arg3) {
```
Raynos: nitpick: `thunkSend` is not needed in this test

PR #476, opened 2015-04-24
as/json: eliminate last explicit throw

jcorbin:
- assert instead for programmer error
- convert strictMode into bossMode

r @Raynos


Raynos:
Should we have a `logger.warn()` or `logger.error()` if your in bossMode ?

PR #477, opened 2015-04-24
[py] Peer management

abhinav:
- Move peer management responsibilities from `TChannel` to `PeerManager`. We'll be able to use `tchannel.request(service='foo').send(...)` in addition to `tchannel.request('localhost:4040', 'foo').send(...)`. If the `hostport` is not specified, the system will automatically pick one of the existing known peers.
- API change: Use `tchannel.host(handler).listen(port)` instead of `tchannel.listen(handler)` because the port to listen on is already known, or is 0, in which case it will be OS-assigned.

CC: @blampe @breerly @junchaowu

### DIFF
--- a/python/examples/thrift/server.py
+++ b/python/examples/thrift/server.py
@@ -35,7 +35,7 @@ class HelloServiceHandler(object):
 
 def run():
     dispatcher = TornadoDispatcher()
-    TChannel().host(dispatcher).listen(4040)
+    TChannel('localhost:4040').listen(dispatcher)
     thrift.register(dispatcher, HelloService, HelloServiceHandler())
     ioloop.IOLoop.current().start()
 
--- a/python/examples/thrift/server.py
+++ b/python/examples/thrift/server.py
@@ -35,7 +35,7 @@ class HelloServiceHandler(object):
 
 def run():
     dispatcher = TornadoDispatcher()
-    TChannel('localhost:4040').listen(dispatcher)
+    TChannel('localhost:4040').host(dispatcher).listen()
     thrift.register(dispatcher, HelloService, HelloServiceHandler())
     ioloop.IOLoop.current().start()
 
--- a/python/tchannel/tornado/peer.py
+++ b/python/tchannel/tornado/peer.py
@@ -43,14 +43,18 @@ from ..handler import CallableRequestHandler
 log = logging.getLogger('tchannel')
 
 
-class PeerManager(object):
-    """PeerManager is responsible for managing peer connections."""
+class PeerGroup(object):
+    """A PeerGroup represents a collection of Peers.
+
+    Requests routed through a PeerGroup can be sent to either a specific peer
+    or a peer chosen at random.
+    """
 
     def __init__(self, tchannel, score_threshold=None):
-        """Initializes a new PeerManager.
+        """Initializes a new PeerGroup.
 
         :param tchannel:
-            The TChannel for which this manager will manage peers.
+            TChannel used for communication by this PeerGroup
         :param score_threshold:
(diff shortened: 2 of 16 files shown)


python/tchannel/tornado/peer.py
```diff
@@ -0,0 +1,448 @@
...
+from .dispatch import Request
+from .stream import InMemStream, Stream
+from .connection import StreamConnection
+from ..handler import CallableRequestHandler
+
+
+log = logging.getLogger('tchannel')
+
+
+class PeerManager(object):
```
breerly: Please don't end classes in `Manager` - in my experience this always leads to bucket classes that grow far beyond their original intention (`UserManager` gives me nightmares, for example). Instead, lets choose a name that is more specific. 

Suggest renaming to `PeerGroup`.

junchaowu:
Do we still have anyway to run our example code since we may need interface to request with the ip.

abhinav:
Tested examples/tchannel_server.py with tcurl. Seems to work just fine.

junchaowu:
One use case is in side the handler, it makes another request to another server. How does it set ip ?

abhinav:
Clarified description. Basically,

    tchannel.request(hostport='123.45.67.89:4040', service='foo').send(...)

Says "call service `foo` hosted at `123.45.67.89:4040`."

    tchannel.request(service='foo').send(...)

Says "call service `foo` through any of the known peers."

python/examples/thrift/server.py
```diff
@@ -35,7 +35,7 @@ def hello(self, name):
 
 def run():
     dispatcher = TornadoDispatcher()
-    TChannel().host(dispatcher).listen(4040)
+    TChannel('localhost:4040').listen(dispatcher)
```
junchaowu: listen dispatcher looks weird.
What's wrong with previous one.

python/tchannel/tornado/peer.py
abhinav: Renamed to PeerGroup

python/examples/thrift/server.py
abhinav: See PR description. Port is already known from TChannel construction time. You shouldn't need to specify it again.

abhinav: Although, `host(dispatcher)` or `serve(dispatcher)` may make more sense.

abhinav: Renaming to serve

breerly: +1 to `serve`.

breerly:
Does this mean we can finally delete trouter.py?

abhinav:
Not yet but it's a step in that direction.

breerly:
thank god :+1:

python/examples/thrift/server.py
abhinav: We need to go back to `.host(..).listen()` (but without the port argument). Without the separation of "set a `RequestHandler`" and "listen for connections", you can't serve requests received via an outgoing connection.

python/tchannel/tornado/peer.py
```diff
@@ -0,0 +1,459 @@
...
+    Requests routed through a PeerGroup can be sent to either a specific peer
+    or a peer chosen at random.
+    """
+
+    def __init__(self, tchannel, score_threshold=None):
+        """Initializes a new PeerGroup.
+
+        :param tchannel:
+            TChannel used for communication by this PeerGroup
+        :param score_threshold:
```
blampe: What's the range for this param?

abhinav: Added range

PR #478, opened 2015-04-24
as/thrift: simplify and improve error handling

jcorbin:
- use AResult from thriftify alpha13
- don't throw errors in the send case, pass to callback
- make remaining programmer error an assert
- add bossMode for handlers similar to as/json in #476

r @kriskowal @Raynos 
cc @leizha


Raynos:
lgtm.

PR #479, opened 2015-04-25
trace: using null logtron instead of debug

rf:
debug is no longer a dep so current beta won't start with trace enabled

@Raynos @jcorbin


Raynos:
lgtm.

PR #482, opened 2015-04-27
[py] Prefer connected peers rather than creating new connections

abhinav:
@breerly @blampe @junchaowu


abhinav:
Don't worry about the test failure. Something funny going on with the Node test.

blampe:
lgtm

PR #481, opened 2015-04-27
Add and convert to a faster (custom) EventEmitter

jcorbin:
r @Raynos @kriskowal @mranney

PR #484, opened 2015-04-28
Drop more closure surface area

jcorbin:


kriskowal:
Nice. How does the benchmark respond?

jcorbin:
Benchmark is inconclusive (some ups, some downs), this is primarily done:
- as requested by jake on #481 towards less closure spew
- it actually helped clarity a bit in a few cases (primarily looking at the ConnectionBase handling path)
- more consistent overall

The travis fail appears to be because node 0.12, something we should address separately.

Merging this, we can dial back any part if "inconclusive" proves to be an actual problem on a flame graph, but better to start in this direction with tiny binding closures.

PR #485, opened 2015-04-28
Max pending

kriskowal:
Adds maxPending and maxPendingForService options (maxPendingPerService a better name?) and a trackPending option for requests (or requestDefaults for a subchannel). Between these features, we will be able to track all pending requests per service at egress nodes of Hyperbahn.

r @jcorbin @Raynos 
cc @ShanniLi

### DIFF
--- a/node/channel.js
+++ b/node/channel.js
@@ -42,6 +42,7 @@ var errors = require('./errors');
 
 var TChannelConnection = require('./connection');
 var TChannelPeers = require('./peers');
+var TChannelServices = require('./services');
 
 var TracingAgent = require('./trace/agent');
 
@@ -110,6 +111,15 @@ function TChannel(options) {
     // - manually api (.peers.add etc)
     // - incoming connections on any listening socket
     self.peers = TChannelPeers(self, self.options);
+
+    // For tracking the number of pending requests to any service
+    self.services = new TChannelServices();
+    if (self.options.maxPending !== undefined) {
+        self.services.maxPending = self.options.maxPending;
+    }
+    if (self.options.maxPendingForService !== undefined) {
+        self.services.maxPendingForService = self.options.maxPendingForService;
+    }
 
     // TChannel advances through the following states.
     self.listened = false;
--- a/node/request.js
+++ b/node/request.js
@@ -35,6 +35,7 @@ function TChannelRequest(channel, options) {
     self.responseEvent = self.defineEvent('response');
 
     self.channel = channel;
+    self.services = self.channel.services;
     self.logger = self.channel.logger;
     self.random = self.channel.random;
     self.timers = self.channel.timers;
@@ -137,11 +138,14 @@ TChannelRequest.prototype.send = function send(arg1, arg2, arg3, callback) {
     }
     self.start = self.timers.now();
     self.resendSanity = self.limit + 1;
+
     self.resend();
 };
 
 TChannelRequest.prototype.resend = function resend() {
     var self = this;
+
+    if (self.checkPending()) return;
 
     if (self.checkTimeout()) return;
 
@@ -177,8 +181,11 @@ TChannelRequest.prototype.resend = function resend() {
     outReq.errorEvent.on(onError);
     outReq.send(self.arg1, self.arg2, self.arg3);
 
+    self.services.onRequest(self);
+
     function onError(err) {
         self.onSubreqError(err);
+        self.services.onRequestError(self);
     }
 
     function onResponse(res) {
@@ -218,6 +225,16 @@ TChannelRequest.prototype.deferResend = function deferResend() {
     function doResend() {
         self.resend();
     }
+};
+
+TChannelRequest.prototype.checkPending = function checkPending() {
+    var self = this;
+    var err = self.services.errorIfExceedsMaxPending(self);
+    if (err) {
+        self.errorEvent.emit(self, err);
+        return true;
+    }
+    return false;
 };
 
 TChannelRequest.prototype.checkTimeout = function checkTimeout(err, res) {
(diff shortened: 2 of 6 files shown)


node/test/max_pending.js
```diff
@@ -0,0 +1,85 @@
...
+            serviceName: 'battle'
+        });
+        var dumChannel = tweedleDum.makeSubChannel({
+            serviceName: 'battle'
+        });
+
+        dumChannel.peers.add(tweedleDee.hostPort);
+
+        deeChannel.register('start', function start() {
+            // Let's just say we did
```
kriskowal: Or,

```js
// A black crow swoops down
```

node/services.js
```diff
@@ -0,0 +1,129 @@
...
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+'use strict';
+
+var errors = require('./errors');
+
+function TChannelServices() {
+    var self = this;
+    self.services = Object.create(null);
```
jcorbin: we suspect this to be abundantly slow, just use an anon-object?

node/services.js
```diff
@@ -0,0 +1,129 @@
...
+}
+
+TChannelServices.prototype.considerDecline = function considerDecline(req) {
+    var self = this;
+    if (self.pending >= self.maxPending) {
+        return errors.MaxPendingError({
+            pending: self.pending
+        });
+    }
+    if (!req.options.serviceName) {
```
jcorbin: just use req.service, reaching into a foreign options isn't stable.

node/services.js
```diff
@@ -0,0 +1,129 @@
...
+
+TChannelServices.prototype.onRequestResponse = function onRequestResponse(req) {
+    var self = this;
+    self.pending--;
+    if (!req.options.serviceName) {
+        return;
+    }
+    var key = mangle(req.options.serviceName);
+    var service = self.services[key];
+    service.onRequestResponse();
```
jcorbin: if (service)

node/services.js
```diff
@@ -0,0 +1,129 @@
...
+
+TChannelServices.prototype.onRequestError = function onRequestError(req) {
+    var self = this;
+    self.pending--;
+    if (!req.options.serviceName) {
+        return;
+    }
+    var key = mangle(req.options.serviceName);
+    var service = self.services[key];
+    service.onRequestError();
```
jcorbin: if (service)

node/test/max_pending.js
```diff
@@ -0,0 +1,85 @@
...
+        });
+
+        parallel([challengeSender(), challengeSender()], function (err, results) {
```
jcorbin: function name

node/test/max_pending.js
```diff
@@ -0,0 +1,85 @@
...
+            return function sendChallenge(cb) {
+                dumChannel.request({
+                    serviceName: 'battle',
+                    timeout: 1000
+                }).send('start', '', '', regardless(cb));
+            };
+        }
+
+        function regardless(cb) {
+            return function (err, value) {
```
jcorbin: name

node/request.js
```diff
@@ -217,6 +225,16 @@ TChannelRequest.prototype.deferResend = function deferResend() {
     }
 };
 
+TChannelRequest.prototype.checkPending = function checkPending() {
+    var self = this;
+    var err = self.services.considerDecline(self);
+    if (err) {
+        self.emit('error', err);
```
jcorbin: this is now self.errorEvent.emit =D

node/services.js
```diff
@@ -0,0 +1,129 @@
...
+};
+
+TChannelService.prototype.onRequestError = function onRequestError() {
+    var self = this;
+    self.pending -= 1;
+};
+
+// Mangling service names ensures that no joker can confuse us with
+// '__proto__'.
+function mangle(serviceName) {
```
jcorbin: suggest name trunderdunder.... dang this is ugly, but I guess the lastIndexOf trick is really a clever way of keeping the overhead minimal, so let the flame graph will out.

jcorbin:
looks sensible, but this review pass has been primarily mechanical, will consider further once have more brains.

node/services.js
kriskowal: is req.service the public api going forward or do we intend to converge on serviceName?

node/services.js
kriskowal: yeah, i'll do the defense in depth, just in case we forget to onRequest() first.

node/services.js
jcorbin: good qn, should probably shift to `req.serviceName`

node/services.js
kriskowal: Will do. I’m going to use '$' prefix in all cases to avoid prototype collisions. I could alternately sieve on typeof service !== 'function' && prefix iff `__` prefix, but this is definitely simpler and probably faster. Might need to review other uses of object as hashtable with user provided keys to head off stupid outages.

kriskowal:
@jcrobin  @Raynos I’ve followed up with a request option to track, which we will be able to add to requestDefaults on egress affinity subchannels, and added some test cases. I think this is ready to land.

Raynos:
Left feedback. The biggest thing I would check is do we decrement the counters in all cases; will we leak maxPending slowly for certain edge cases ?

kriskowal:
Addressed. Further fixes in follow-up. Merging.

node/channel.js
```diff
@@ -111,6 +112,15 @@ function TChannel(options) {
     // - incoming connections on any listening socket
     self.peers = TChannelPeers(self, self.options);
 
+    // For tracking the number of pending requests to any service
+    self.services = new TChannelServices();
```
jcorbin: wait a second, I thought I just reviewed a similar line in another PR???

PR #491, opened 2015-04-29
[py] Implement synchronous client on top of Tornado.

abhinav:
@breerly @blampe @junchaowu

### DIFF
--- a/python/setup.py
+++ b/python/setup.py
@@ -4,13 +4,27 @@ from setuptools import find_packages, setup
 setup(
     name='tchannel',
     version='0.4.1',
-    author='Aiden Scandella, Abhinav Gupta, Bryce Lampe, Junchao Wu, Grayson Koonce',
+    author=", ".join([
+        'Aiden Scandella',
+        'Abhinav Gupta',
+        'Bryce Lampe',
+        'Junchao Wu',
+        'Grayson Koonce',
+    ]),
     author_email='dev@uber.com',
     description='Network multiplexing and framing protocol for RPC',
     license='MIT',
     url='https://github.com/uber/tchannel',
     packages=find_packages(),
-    install_requires=['contextlib2', 'enum34', 'futures'],
+    install_requires=[
+        'contextlib2',
+        'crcmod'
+        'enum34',
+        'futures',
+        # 'pyfarmhash', TODO not yet used
+        'tornado>=4.0,<5.0',
+        'toro>=0.8,<0.9',
+    ],
     entry_points={
         'console_scripts': [
             'tcurl.py = tchannel.tcurl:start_ioloop'
--- a/python/tchannel/glossary.py
+++ b/python/tchannel/glossary.py
@@ -22,3 +22,7 @@
 #
 # Message ID 0xffffffff is reserved
 MAX_MESSAGE_ID = 0xfffffffe
+
+# Hostport used by TChannel peers that are ephemeral and don't expect incoming
+# connections.
+EPHEMERAL_HOSTPORT = '0.0.0.0:0'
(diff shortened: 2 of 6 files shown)


abhinav:
Tests pending

python/tchannel/glossary.py
```diff
@@ -22,3 +22,7 @@
 #
 # Message ID 0xffffffff is reserved
 MAX_MESSAGE_ID = 0xfffffffe
+
+# Hostport used by TChannel peers that are ephemeral and don't expect incoming
+# connections.
+EPHEMERAL_HOSTPORT = '0.0.0.0:0'
```
breerly: yay, glossary for holding config constants :+1:

breerly:
Have you driven this bad boy by hand yet?

python/setup.py
```diff
@@ -4,13 +4,27 @@
...
     license='MIT',
     url='https://github.com/uber/tchannel',
     packages=find_packages(),
-    install_requires=['contextlib2', 'enum34', 'futures'],
+    install_requires=[
+        'contextlib2',
+        'crcmod'
+        'enum34',
+        'futures',
+        # 'pyfarmhash', TODO not yet used
```
breerly: Are we still running sans-hashing?

python/tchannel/sync/client.py
```diff
@@ -0,0 +1,130 @@
...
+        arg1 = arg1 or ''
+        arg2 = arg2 or ''
+        arg3 = arg3 or ''
+
+        @gen.coroutine
+        def go():
+            # This is the response that contains argstreams. We need to read
+            # everything in memory.
+            resp = yield self.ops.send(arg1, arg2, arg3)
+
```
junchaowu: check resp error here.

python/setup.py
junchaowu: currently we have them, but not used.

python/setup.py
```diff
@@ -4,13 +4,27 @@
 setup(
     name='tchannel',
     version='0.4.1',
-    author='Aiden Scandella, Abhinav Gupta, Bryce Lampe, Junchao Wu, Grayson Koonce',
+    author=", ".join([
```
breerly: fancy pants

python/tchannel/sync/client.py
abhinav: It will throw an exception if there's an error. The future will then contain an exception which will be re-raised by the run_sync call.

python/setup.py
abhinav: >80 characters. Unacceptable.

abhinav:
Yes, I ran it manually against examples/tchannel_server.py. Seemed to work.

abhinav:
In [14]: client.request('localhost:8888').send('hi', None, '1')
    Out[14]: Response(arg1='hi', arg2='', arg3='1')
    
    In [15]: client.request('localhost:8888').send('slow', None, '1')
    Out[15]: Response(arg1='', arg2='', arg3='done')
    
    In [17]: client.request('localhost:8888').send('echo', None, 'foo bar baz qux')
    Out[17]: Response(arg1='echo', arg2='', arg3='foo bar baz qux')

breerly:
NOW that's what I call music. Tests and :ship: :it:

breerly:
Can you run it with `time` in iPython?

python/tchannel/sync/client.py
junchaowu: you should not call yield resp.arg1 if it is an error.

blampe:
No fanout?

breerly:
Me like fanout.

abhinav:
T
his isn't a priority for now. We can revisit this later.

blampe:
Let's prioritize this next.

PR #493, opened 2015-04-29
Converge on serviceName over service for requests

kriskowal:
Old and busted:
 - service meant service name
 - serviceObj meant service instance

New hawtness:
 - serviceName means service name
 - service means service instance

### DIFF
--- a/node/README.md
+++ b/node/README.md
@@ -225,7 +225,7 @@ This is used to avoid race conditions in the network.
 ```jsig
 type TChannelInRequest : {
     id: Number,
-    service: String,
+    serviceName: String,
 
     arg1: Buffer,
     arg2: Buffer,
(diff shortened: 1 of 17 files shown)


kriskowal:
r @jcorbin @Raynos

Raynos:
Rename of `serviceObj` to `service` does not help.

`service` is ambigious enough as it is; calling it `serviceObj` or something `serviceTuple` or whatever helps in this case.

jcorbin:
lgtm

-1 on @Raynos 's serviceObj feedback, better to have `service` just be the danged "service" thing at hand

Raynos:
Do we want to rename `v2.CallRequest.service` to `serviceName` as well ? at what point are we talking about the `service` field on the call request vs the `serviceName` ?

Should we go the other way and rename everything to `service` to have consistency with `v2.CallRequest`

Raynos:
lgtm.

Issue #488, opened 2015-04-28
node: Don't (always) buffer arg2/3 for application retry logic


jcorbin:
Currently `TChannelRequest#resend` always buffers arg2/3 to call '#onSubreqRespons` with fully realized arg2/3, to pass to `#shouldRetry`, in case `options.shouldApplicationRetry` was set.

This should change:
- [x] simply pass err/res through `#resend` -> `#onSubreqResponse` -> `#shouldRetry`
- [x] make `#shouldRetry :: (err, res, callback) -> void` where `callback :: (?err, Boolean)`
- [x] make `options.shouldApplicationRetry :: (err, res, callback) -> void` also

jcorbin:
done.

PR #496, opened 2015-04-29
Factor out {req,res}.withArg23

jcorbin:
Towards #488, this makes it easier to shift around who / when buffering occurs.


kriskowal:
:ok:

Raynos:
lgtm.

PR #497, opened 2015-04-29
Refactor should retry logic to defer arg2 / arg3 buffering

jcorbin:
Implements all points outlined in #488 

r @kriskowal @Raynos

### DIFF
--- a/node/request.js
+++ b/node/request.js
@@ -199,6 +199,7 @@ TChannelRequest.prototype.onSubreqError = function onSubreqError(err) {
 
 TChannelRequest.prototype.onSubreqResponse = function onSubreqResponse(res) {
     var self = this;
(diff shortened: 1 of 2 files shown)


kriskowal:
To be clear, the way I read this story, whether a request should be retried is an application level concern. Previously, applications had to decide synchronously based and had to wait for arguments to accumulate before asking. With this change, shouldRetry becomes asynchronous and has the option of deciding whether a request is retry-able based on partial information.

:ok:

jcorbin:
@kriskowal correct, for certain application, deciding on arg2 may be sufficient, or on a first chunk of arg3

jcorbin:
but chances are, they'll just use res.withArg23, which is now easily available to all

Issue #122, opened 2015-03-05
Cut node 2.0.0


jcorbin:
Current blockers (2015-04-28):
- [x] finalize MIGRATION guide
- [ ] ~~circuit breaking model~~
- [x] complete, if minimal, protocol implementation
  - [x] ping
  - [x] claim
  - [x] cancel
  - [x] init handshake blocks call req (no init+call pipelining)
- [x] protocol errata
  - [x] arg1 shouldn't be fragmented or streamed
  - [x] limit allowed total overhead of a call req frame, currently it's nearly 2x max frame size, leaving no room for arg 1 (current max size is 0x1ff23 bytes, aka (128Ki - 221)B

Raynos:
Possible breaking changes up for grabs:

 - [ ] make `opts.span` required in `tchannel.request()`
 - [ ] make `opts.inreq` required in `tchannel.request()`
 - [ ] change `nameAsThrift` from field on err to field on ThriftResponse
 - [ ] consider making `.request()` and `.register()` fail on the top channel; there has been confusion recently around whether you should use the topChannel or a subChannel ( we always want to use sub channels ).

jcorbin:
re: client libraries and multiple copies, can we not just inject a `tchannelAsThrift` instance?

Raynos:
We have consensus that the "cn" header is required.

We have consensus that the "as" header is required.

Raynos:
We should encourage libraries to never depend on `tchannel` and to take a `TChannelAsThrift` constructor.

Raynos:
All the features for 2.0.0 are in;

There are only four **BLOCKERS** left

 - [ ] update the README with documentation.
 - [ ] write a quick node guide
 - [ ] update the MIGRATION guide
 - [ ] get all the examples operational again.

Raynos:
That's it. 2.0.0 we are done.

PR #499, opened 2015-04-29
add frame/handler support for ping req, ping res, claim, and cancel.

ShanniLi:
r @jcorbin @kriskowal 

support operation claim
claim frame parser
internal handler support

support operation cancel
cancel frame parser
internal handler support

ping req/res frame parsers
internal handler support

### DIFF
--- a/node/v2/claim.js
+++ b/node/v2/claim.js
@@ -30,7 +30,7 @@ function Claim(tracing) {
     self.tracing = tracing || Tracing.emptyTracing;
 }
 
-Claim.TypeCode = 0xC1;
+Claim.TypeCode = 0xc1;
 
 Claim.RW = bufrw.Struct(Claim, [
     {name: 'tracing', rw: Tracing.RW}  // tracing:25
--- a/node/v2/claim.js
+++ b/node/v2/claim.js
@@ -20,15 +20,20 @@
 
 'use strict';
 
-require('./frame.js');
-require('./init.js');
-require('./checksum.js');
-require('./header.js');
-require('./tracing.js');
-require('./call.js');
-require('./cancel.js');
-require('./cont.js');
-require('./claim.js');
-require('./ping.js');
-require('./error_response.js');
-require('./args.js');
+var bufrw = require('bufrw');
+var Tracing = require('./tracing');
+
+// tracing:25
+function Claim(tracing) {
+    var self = this;
+    self.type = Claim.TypeCode;
+    self.tracing = tracing || Tracing.emptyTracing;
+}
+
+Claim.TypeCode = 0xC1;
+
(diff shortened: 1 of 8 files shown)


kriskowal:
This looks good to me. When this lands, check off relevant parts on https://github.com/uber/tchannel/issues/122

kriskowal:
And wait for @jcorbin to weigh in.

node/v2/handler.js
```diff
@@ -50,8 +50,12 @@ function TChannelV2Handler(options) {
     self.callIncomingErrorEvent = self.defineEvent('callIncomingError');
     self.callIncomingRequestEvent = self.defineEvent('callIncomingRequest');
     self.callIncomingResponseEvent = self.defineEvent('callIncomingResponse');
+    self.cancelEvent = self.defineEvent('cancel');
     self.initRequestEvent = self.defineEvent('initRequest');
     self.initResponseEvent = self.defineEvent('initResponse');
+    self.claimEvent = self.defineEvent('claim');
```
jcorbin: let's move this up to b/w L53/54 to maintain alpahabetic ordering

node/v2/claim.js
```diff
@@ -0,0 +1,39 @@
...
+var Tracing = require('./tracing');
+
+// tracing:25
+function Claim(tracing) {
+    var self = this;
+    self.type = Claim.TypeCode;
+    self.tracing = tracing || Tracing.emptyTracing;
+}
+
+Claim.TypeCode = 0xC1;
```
jcorbin: The convention in the node implementation so far is to lowercase hex literals, would be nice to maintain that here and in Cancel.TypeCode above

node/v2/handler.js
```diff
@@ -208,6 +220,12 @@ TChannelV2Handler.prototype.callResponseFrameHandled = function callResponseFram
     callback();
 };
 
+TChannelV2Handler.prototype.handleCancel = function handleCancel(frame, callback) {
```
jcorbin: Let's at add a TODO noting that we should implement clearing of `self.streaming{Req,Res}`.

- should clear out of self.streaming{Req,Res}
- but they're currently keyed by id not tracing
- however we probably can't just shift the primary key of these collections
- rather we'll probably need to add a secondary index to protect against pathological "zero tracing" being sent... unless of course we make that a hard error

Similar concerns apply at the `TChannelConnection.requests.{in,out} layer`.

jcorbin:
:+1: optional feedback, this is good enough to merge and go forward

node/v2/claim.js
ShanniLi: fixed.

node/v2/handler.js
ShanniLi: TODO added.

PR #501, opened 2015-04-30
refactor the interface of response and request

junchaowu:


python/examples/handler.py
```diff
@@ -45,10 +39,7 @@ def say_ok(request, response, proxy):
     yield print_arg(request, 1)
     yield print_arg(request, 2)
 
-    response.argstreams = [
-        InMemStream(),
-        InMemStream(),
-        InMemStream("world")]
+    response.set_body(InMemStream("world"))
```
breerly: You guys think we should get away from the `InMemStream` instantiation? @blampe @abhinav

python/examples/handler.py
```diff
@@ -56,20 +47,15 @@ def echo(request, response, proxy):
     print "echo"
     yield tornado.gen.sleep(1)
     # stream args right back to request side
-    response.argstreams = [
-        InMemStream(request.endpoint),
-        request.argstreams[1],
-        request.argstreams[2]
-    ]
+    response.set_header(request.get_header_s())
```
breerly: I think this should probably be:

`response.set_headers()` - takes a list of headers to set

and

`response.add_header()` - add a single header

python/examples/handler.py
```diff
@@ -56,20 +47,15 @@ def echo(request, response, proxy):
...
+    response.set_body(request.get_body_s())
```
breerly: This feels right. +1

python/examples/handler.py
breerly: `request.get_header_s()` feels akward. 

What about:

`request.headers` - a dict of all headers
`request.headers.get('X-Uber-Source')` - get the header with the `X-Uber-Source` key.

python/examples/handler.py
breerly: `request.get_body_s()` feels akward.

What about:

`request.body`

python/tchannel/tornado/dispatch.py
```diff
@@ -239,50 +229,138 @@ def __init__(
...
+        """
+
+        if self.flushed:
+            raise TChannelException("write operation invalid after flush call")
+
+        if (self.argstreams[0].state != StreamState.completed and
+                self.argstreams[0].auto_close):
+            self.argstreams[0].close()
+
+        yield self.argstreams[1].write(chunk)
```
blampe: Don't need to make these functions coroutines, you can just return the future here instead of yielding to it.

python/examples/handler.py
abhinav: We should get these methods to accept strings/bytes and automatically convert them to in-memory streams.

Consider using `maybe_stream` in `tchannel.tornado.peer` to do that. (Probably move it to `tchannel.tornado.stream` while you're at it.)

python/examples/handler.py
```diff
@@ -31,13 +31,7 @@
...
 @tornado.gen.coroutine
 def say_hi(request, response, proxy):
-    arg2 = yield request.arg2()
-    arg3 = yield request.arg3()
-    response.argstreams = [
-        InMemStream(request.endpoint),
-        InMemStream(arg2),
-        InMemStream(arg3)
-    ]
+    yield response.write_body("hi")
```
blampe: Much nicer right? :)

breerly:
Without seeing how you are planning to add in JSON and Thrift types, it's very difficult to comment on how to best structure the raw request/responses.

python/examples/handler.py
blampe: Definitely. In most cases I would expect the user just wants to write strings, in which case we can wrap the string with this at a lower level.

breerly:
Suggest renaming these methods to be more like:

```
response.set_raw_header
response.get_raw_header
response.set_raw_body
response.get_raw_body
```

Also, I wonder if it makes sense to keep the arg2/arg3 naming semantics at this level, and only switch to header/body naming semantics once we add in json/thrift.

python/tchannel/tornado/dispatch.py
```diff
@@ -88,13 +79,13 @@ def handle_call(self, request, connection):
                 ErrorCode.bad_request,
                 "Endpoint '%s' for service '%s' is not defined" % (
                     request.endpoint, request.service),
-                request.id,
-            )
+                request.id)
```
abhinav: Styling nits:

If the call is in the form

    foo(
        a,
        b,
        c,
    )

It's okay to leave the `)` on a separate line. If the call is in the form,

    foo(a,
        b,
        c)

Then, you want the `)` on the same line as the last parameter.

python/tchannel/thrift/transport/tornado_base.py
```diff
@@ -89,8 +89,8 @@ def _flush_internal(self, endpoint, response, seqid):
                 Thrift.TMessageType.REPLY,
                 seqid,
             )
-            arg3 = yield response.arg3()
-            buff.write(arg3)
+            body = yield response.get_body()
```
breerly: Suggest renaming to `response.get_raw_body()`.

python/examples/handler.py
```diff
@@ -56,20 +47,15 @@ def echo(request, response, proxy):
...
 
 @tornado.gen.coroutine
 def slow(request, response, proxy):
     yield tornado.gen.sleep(random.random())
-    response.argstreams = [
-        InMemStream(),
-        InMemStream(),
-        InMemStream("done")]
+    yield response.write_body("done")
+    response.flush()
```
blampe: Suggest doing this in the handler if it hasn't been called manually (e.g., `if not response.flushed: flush()`; and flushing twice would raise an error).

abhinav:
@breerly I would call them `set_header_stream`/etc. instead of "raw"

python/examples/handler.py
breerly: Suggest renaming to:

```
response.set_raw_header(request.get_raw_header())
```

python/examples/handler.py
junchaowu: that's the method called write_body not set_body.
set_body is used to give them a chance to provide the customized stream

python/examples/handler.py
breerly: Suggest renaming to:

```
response.set_raw_body(request.get_raw_body())
```

python/examples/handler.py
breerly: Good idea.

python/examples/handler.py
blampe: Oh you already did that, nice!

python/tchannel/tornado/dispatch.py
```diff
@@ -184,34 +175,33 @@ def close_argstreams(self, force=False):
             if stream.auto_close or force:
                 stream.close()
 
-    def arg1(self):
-        """get value for arg1
+    def get_header(self):
+        """Get the header value from the request.
 
-        :return: return the future object contains the value for arg1
+        :return: the value of header
```
abhinav: Actually these methods return futures. In the `:return:` section, mention that it returns "a future producing the value of [..]"

python/examples/handler.py
junchaowu: flush twice , I think it's okay. Basically nothing happened. But we can raise error too.

blampe: no-op'ing is also fine.

python/examples/handler.py
abhinav: Yeah, but `set_body` also implies that you're proving the full content. That's different from passing a string to `write_body`.

I mean, `write_body` says you're adding the next chunk of the body. `set_body` says you're providing the full body content (which may be in form of a stream). In fact, if you look at the behavior of `maybe_stream` it closes the stream after building it because you're not going to be writing anything more to the stream.

blampe: `set_body` and `write_body` are too easy to confuse. As a user I'd rather have one method that can take a string *or* a mem stream, that way I don't need to remember specifics like that.

junchaowu: we also have pipestream. stream a file. we need to provide the ability to support that. We can find a better name

junchaowu:
We will keep using the name "get_body" "get_header" "get_body_s" "set_body_s" for now.

blampe:
Great first steps, lgtm!

python/tests/test_stream.py
```diff
@@ -50,8 +51,8 @@ def test_InMemStream():
 def test_PipeStream():
     r, w = os.pipe()
     stream = PipeStream(r, w, auto_close=True)
-    yield stream.write("1")
-    yield stream.write("2")
+    stream.write("1")
+    stream.write("2")
```
abhinav: Doesn't write return a future?

junchaowu: let me change it back.

PR #504, opened 2015-04-30
Add options to endpoint registration

jcorbin:
Encapsulates streaming handler interface

r @kriskowal @Raynos

### DIFF
--- a/node/endpoint-handler.js
+++ b/node/endpoint-handler.js
@@ -50,6 +50,7 @@ TChannelEndpointHandler.prototype.register = function register(name, options, ha
     if (typeof handler !== 'function') {
         throw errors.InvalidHandlerError();
     }
+    if (options.streamed) handler.canStream = true;
     self.endpoints[name] = handler;
     return handler;
 };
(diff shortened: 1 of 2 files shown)


node/endpoint-handler.js
```diff
@@ -41,11 +41,16 @@ inherits(TChannelEndpointHandler, EventEmitter);
...
+TChannelEndpointHandler.prototype.register = function register(name, options, handler) {
     var self = this;
+    if (typeof options === 'function') {
+        handler = options;
+        options = {};
+    }
     if (typeof handler !== 'function') {
         throw errors.InvalidHandlerError();
     }
+    if (options.streamed) handler.canStream = true;
```
kriskowal: It seems unclear to me whether the handler is accepting a streamed request or providing a streamed response.

kriskowal:
:ok:

node/endpoint-handler.js
jcorbin: It means you can handle streaming incoming.  Whether you send a streamed response is down to how you call `buildResponse(options)`.

Raynos:
lgtm.

PR #506, opened 2015-04-30
[node] Coerce request handlers

kriskowal:
Thus allowing a request handler to be implemented either as a closure or
as an object implementing handleRequest.

### DIFF
--- a/node/endpoint-handler.js
+++ b/node/endpoint-handler.js
@@ -93,15 +93,8 @@ TChannelEndpointHandler.prototype.handleArg1 = function handleArg1(req, buildRes
         res.sendError('BadRequest', util.format(
             'no such endpoint service=%j endpoint=%j',
             req.serviceName, name));
-    } else if (handler.canStream) {
-        handler.handleRequest(req, buildResponse);
-    } else if (req.streamed) {
-        // And implicitly, !handler.canStream, meaning fragmented but
-        // reaccumulated:
-        self.withArg23(req, buildResponse, handler);
     } else {
-        res = buildResponse({streamed: false});
-        handler.handleRequest(req, res, req.arg2, req.arg3);
+        handler.handleRequest(req, buildResponse);
     }
 };
 
--- a/node/endpoint-handler.js
+++ b/node/endpoint-handler.js
@@ -24,6 +24,7 @@ var EventEmitter = require('./lib/event_emitter');
 var inherits = require('util').inherits;
 var util = require('util');
 var errors = require('./errors');
+var coerceRequestHandler = require('./request-handler');
 
 function TChannelEndpointHandler(serviceName) {
     if (!(this instanceof TChannelEndpointHandler)) {
@@ -50,7 +51,7 @@ TChannelEndpointHandler.prototype.register = function register(name, options, ha
         throw errors.InvalidHandlerError();
     }
     if (options.streamed) handler.canStream = true;
-    self.endpoints[name] = handler;
+    self.endpoints[name] = coerceRequestHandler(handler, self, options);
     return handler;
 };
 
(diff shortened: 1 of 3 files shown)


kriskowal:
r @jcorbin @Raynos

node/endpoint-handler.js
```diff
@@ -77,7 +78,7 @@ TChannelEndpointHandler.prototype.waitForArg1 = function waitForArg1(req, buildR
 TChannelEndpointHandler.prototype.handleArg1 = function handleArg1(req, buildResponse, arg1) {
     var self = this;
     var name = String(arg1);
-    var handler = self.endpoints[name];
+    var handler = coerceRequestHandler(self.endpoints[name], self);
```
jcorbin: coerces on every request rather than once at registration time, please make it once

node/endpoint-handler.js
```diff
@@ -89,16 +90,18 @@ TChannelEndpointHandler.prototype.handleArg1 = function handleArg1(req, buildRes
...
+        self.withArg23(req, buildResponse, handler);
     } else {
         res = buildResponse({streamed: false});
-        handler.call(self, req, res, req.arg2, req.arg3);
+        handler.handleRequest(req, res, req.arg2, req.arg3);
     }
 };
 
-TChannelEndpointHandler.prototype.bufferArg23 = function bufferArg23(req, buildResponse, handler) {
+TChannelEndpointHandler.prototype.withArg23 = function withArg23(req, buildResponse, handler) {
```
jcorbin: if you're going to go for "consistency" with req.withArg23... actually use req.withArg23

node/request-handler.js
```diff
@@ -0,0 +1,49 @@
...
+'use strict';
+
+var assert = require('assert');
+
+// This function allows EndpointHandler to accept either a callback or a
+// RequestHandler object that implements handleRequest.
+
+module.exports = coerceRequestHandler;
+
+function coerceRequestHandler(handler, thisp) {
```
jcorbin: nit: `thisp` seems weird, can we call it `that` instead?

node/request-handler.js
```diff
@@ -0,0 +1,49 @@
...
+        return new RequestCallbackHandler(handler, thisp);
+    } else {
+        assert(typeof handler.handleRequest === 'function', 'handler must have handleRequest method');
+        return handler;
+    }
+}
+
+function RequestCallbackHandler(callback, thisp) {
+    var self = this;
+    self.canStream = !!callback.canStream;
```
jcorbin: you do realize that now we have the potential:
- implement two callback wrapper handlers in this module
- instantiate one or the other based on the streamed option (masquerading as callback.canStream here, which go away...)
- then the canStream buffering convenience is resolved once at registration time, rather than branched on every request

jcorbin:
Works, modulo a rebase.

If you don't follow through on the canStream idea, I probably will try soon ;-)

Raynos:
lgtm.

node/endpoint-handler.js
kriskowal: Are you recommending that we delete bufferArg23?

node/request-handler.js
kriskowal: Consulting precedent:

- thisArg https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach
- receiver https://chromium.googlesource.com/v8/v8.git/+/master/src/array.js
- context https://github.com/es-shims/es6-shim/blob/master/es6-shim.js#L2204
- thisp https://github.com/es-shims/es5-shim/blob/master/es5-shim.js#L439

node/endpoint-handler.js
jcorbin: nearly, you'll need a shim of some sort, since:
```
r.withArg23 :: (callback) -> void
callback :: (err, arg2, arg3) -> void
```
In other words, it relies on the fact that its caller had access already to `r` (whether it was a req or res).

So most of the `argsDone` callback function remains, just with a slightly different call signature.  Basically:
- stick with the rename
- use req.withArg23 rather than parallel

As a minor note, this introduces zalgo wrt the type of req (non-streaming calls immed, streaming calls eventually), however I'm 90% sure that all zalgo concerns for handlers v channel have been solved by `TChannelConnectionBase#runHandler` having been deferred circa `connection_base.js:258`.

jcorbin:
:+1: good to go, nit pick optional feedback about tightening a var decl

PR #508, opened 2015-05-01
add tchannel json arg schema

junchaowu:
cc @breerly @abhinav @blampe 

test_json_trace is example how to use it in both client and server sides.

### DIFF
--- a/python/tchannel/messages/common.py
+++ b/python/tchannel/messages/common.py
@@ -41,7 +41,8 @@ MAX_PAYLOAD_SIZE = 0xFFEF   # 64*1024 - 16 - 1
 class StreamState(IntEnum):
     init = 0x00,
     streaming = 0x01,
-    completed = 0x02
+    completed = 0x02,
+    none = 0x03
 
 
 class FlagsType(IntEnum):
--- a/python/tchannel/tornado/broker.py
+++ b/python/tchannel/tornado/broker.py
@@ -1,0 +1,85 @@
+# Copyright (c) 2015 Uber Technologies, Inc.
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+
+from __future__ import absolute_import
+
+import tornado
+
+from .stream import Stream
+from ..exceptions import TChannelException, ProtocolException
+
+
+class ArgSchemeBroker(object):
+    """Use serializer to broker request/response."""
+
+    def __init__(self, arg_scheme):
+        self.endpoint = {}
+        self.arg_scheme = arg_scheme
+
+    def register(self, rule, handler):
+        """Register handler.
+
+        :param rule: endpoint
+        :param handler: endpoint handler
+        """
+        self.endpoint[rule] = handler
+
+    def handle_call(self, req, resp, proxy):
+        if not req.headers.get('as', None) == self.arg_scheme.type():
+            raise TChannelException("Invalid arg scheme in request header")
+
+        req.scheme = self.arg_scheme
+        resp.scheme = self.arg_scheme
+
+        handler = self.endpoint.get(req.endpoint, None)
+        if handler is None:
+            raise ProtocolException(
+                "Endpoint '%s' for service '%s' is not defined" % (
+                    req.endpoint, req.service))
+
+        return handler(req, resp, proxy)
+
+    @tornado.gen.coroutine
+    def send(self, client, endpoint, header, body, traceflag=False):
+        try:
+            if not isinstance(header, Stream):
+                raw_header = self.arg_scheme.serialize(header)
+            else:
+                raw_header = header
+
+            if not isinstance(body, Stream):
+                raw_body = self.arg_scheme.serialize(body)
+            else:
+                raw_body = body
+
+        except Exception as e:
+            raise TChannelException(e.message)
+
+        resp = yield client.send(
+            arg1=endpoint,
+            arg2=raw_header,
+            arg3=raw_body,
+            traceflag=traceflag,
+            headers={'as': self.arg_scheme.type()},
+        )
+
+        resp.scheme = self.arg_scheme
+
+        raise tornado.gen.Return(resp)
--- a/python/tchannel/tornado/connection.py
+++ b/python/tchannel/tornado/connection.py
@@ -24,7 +24,6 @@ import logging
 import os
 import socket
 import sys
-from tchannel.event import EventType
 
 import tornado.gen
 import tornado.iostream
@@ -34,20 +33,23 @@ try:
 except ImportError:
     import toro as queues
 
+from .message_factory import MessageFactory
 from .. import glossary
 from .. import frame
 from .. import messages
 from .. import exceptions
 from ..io import BytesIO
 from ..context import Context
-from ..exceptions import ConnectionClosedException, InvalidErrorCodeException
+from ..exceptions import ConnectionClosedException
+from ..exceptions import InvalidErrorCodeException
 from ..messages.types import Types
-from ..messages.common import (
-    PROTOCOL_VERSION, generate_checksum, FlagsType, verify_checksum
-)
-from ..messages.error import ErrorMessage, ErrorCode
-from .message_factory import MessageFactory
-
+from ..messages.common import PROTOCOL_VERSION
+from ..messages.common import generate_checksum
+from ..messages.common import FlagsType
+from ..messages.common import verify_checksum
+from ..messages.error import ErrorCode
+from ..messages.error import ErrorMessage
+from ..event import EventType
 
 log = logging.getLogger('tchannel')
 
--- a/python/tchannel/tornado/data.py
+++ b/python/tchannel/tornado/data.py
@@ -1,0 +1,300 @@
+# Copyright (c) 2015 Uber Technologies, Inc.
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+
+from __future__ import absolute_import
+
+import tornado
+import tornado.gen
+
+from .stream import InMemStream
+from .util import get_arg
+from ..exceptions import TChannelException
+from ..messages.common import FlagsType
+from ..messages.common import StreamState
+from ..zipkin.trace import Trace
+
+
+class Request(object):
+    """Represents an incoming request to an endpoint.
+
+    Request class is used to represent the CallRequestMessage at User's level.
+    This is going to hide the protocol level message information.
+    """
+
+    # TODO decide which elements inside "message" object to expose to user.
+    def __init__(
+            self,
+            id=None,
+            flags=FlagsType.none,
+            ttl=10,
+            tracing=None,
+            service=None,
+            headers=None,
+            checksum=None,
+            argstreams=None,
+            scheme=None,
+    ):
+        self.flags = flags
+        self.ttl = ttl
+        self.service = service
+        self.tracing = tracing or Trace()
+        # argstreams is a list of InMemStream/PipeStream objects
+        self.argstreams = argstreams
+        self.checksum = checksum
+        self.id = id
+        self.headers = headers or {}
+        self.state = StreamState.init
+        self.endpoint = ""
+        self.header = None
+        self.body = None
+        self.scheme = scheme
+
+    @property
+    def arg_scheme(self):
+        return self.headers.get('as', None)
+
+    def close_argstreams(self, force=False):
+        for stream in self.argstreams:
+            if stream.auto_close or force:
+                stream.close()
+
+    @tornado.gen.coroutine
+    def get_header(self):
+        """Get the header value from the request.
+
+        :return: a future contains the deserialized value of header
+        """
+        raw_header = yield get_arg(self, 1)
+        if not self.scheme:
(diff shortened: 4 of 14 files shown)


python/tchannel/tornado/connection.py
```diff
@@ -225,7 +225,7 @@ def _loop(self):
                 if response and future.running():
                     future.set_result(response)
                 continue
-
+            print context.message
```
abhinav: Remove

python/tchannel/tornado/data.py
```diff
@@ -0,0 +1,421 @@
...
+    def __init__(self, req=None, serializer=None, deserializer=None):
+        if isinstance(req, Request):
+            super(JsonRequest, self).__init__(
+                id=req.id,
+                flags=req.flags,
+                ttl=req.ttl,
+                tracing=req.tracing,
+                service=req.service,
+                headers=req.headers,
+                checksum=req.checksum,
```
abhinav: Indentation

python/tchannel/messages/common.py
```diff
@@ -41,7 +41,7 @@
 class StreamState(IntEnum):
     init = 0x00,
     streaming = 0x01,
-    completed = 0x02
+    completed = 0x02,
```
blampe: This isn't doing what you think it is :)

Within parenthesis, `('foo', 'bar')` is equivalent to `('foo', 'bar',)` because they represent the same tuple. BUT if you're not already in a tuple, then a dangling comma will implicitly create a single-element tuple. So what you have here means:

```
> StreamState.streaming == 0x01
False
> StreamState.streaming == (0x01,)
True
```

python/tchannel/tornado/connection.py
blampe: Remove.

python/tchannel/tornado/data.py
```diff
@@ -0,0 +1,421 @@
...
+
+
+import tornado
+import tornado.gen
+
+from ..messages.common import StreamState, FlagsType
+from .stream import InMemStream
+from .util import get_arg, serialize_json, deserialize_json
+from ..zipkin.trace import Trace
+from ..exceptions import TChannelException
```
blampe: `!sort`

python/tchannel/tornado/data.py
```diff
@@ -0,0 +1,421 @@
...
+
+from __future__ import absolute_import
+
+
+import tornado
+import tornado.gen
+
+from ..messages.common import StreamState, FlagsType
+from .stream import InMemStream
+from .util import get_arg, serialize_json, deserialize_json
```
blampe: One import per line.

python/tchannel/tornado/data.py
```diff
@@ -0,0 +1,421 @@
...
+
+        :return: the argstream of body
+        """
+        return self.argstreams[2]
+
+
+class JsonRequest(Request):
+    def __init__(self, req=None, serializer=None, deserializer=None):
+        if isinstance(req, Request):
+            super(JsonRequest, self).__init__(
```
abhinav: what? `isinstance` and conditional `super` initialization? You probably need to refactor something.

python/tchannel/tornado/data.py
```diff
@@ -0,0 +1,421 @@
...
+        for stream in self.argstreams:
+            if stream.auto_close or force:
+                stream.close()
+
+
+class JsonResponse(Response):
+    def __init__(self, resp=None, serializer=None, deserializer=None):
+        if isinstance(resp, Response):
+            super(JsonResponse, self).__init__(
+                resp.connection,
```
abhinav: Same as above

python/tchannel/tornado/data.py
```diff
@@ -0,0 +1,421 @@
...
+        :except TChannelException:
+            Raise TChannelException if the body is not a valid json object.
+        """
+        raw_header = yield super(JsonRequest, self).get_header()
+        jheader = {}
+        try:
+            jheader = deserialize_json(raw_header)
+        except Exception as e:
+            # TODO define error handling for this case
+            raise tornado.gen.Return(TChannelException(e.message))
```
abhinav: This is not how you raise exceptions in Tornado

python/tchannel/tornado/dispatch.py
```diff
@@ -91,14 +87,14 @@ def handle_call(self, request, connection):
                                           connection.tchannel,
                                           request.tracing))
 
-    def route(self, rule):
+    def route(self, rule, format=None):
```
abhinav: Document the parameters

python/tchannel/tornado/dispatch.py
```diff
@@ -115,10 +111,20 @@ def handler(request, response, proxy):
         :param handler:
             A function that gets called with ``Request``, ``Response``, and
             the ``proxy''.
+        :param protocol:
+            Register handler for specific type of format
+            protocol = json/http/thrift/None
+            protocol=None means it registers as raw handle. It deals with raw
+            buffer in the request/response.
```
abhinav: How will the `thrift` protocol work.

Also, not sure you want to call it `protocol`. `scheme` or `arg_scheme`?

python/tchannel/tornado/dispatch.py
```diff
@@ -115,10 +111,20 @@ def handler(request, response, proxy):
...
         """
         assert rule, "rule must not be None"
         assert handler, "handler must not be None"
-        self.endpoints[rule] = handler
+
+        if protocol:
+            protocol.register(rule, handler)
+            self.endpoints[rule] = protocol.handle_call
```
abhinav: The docs suggest `protocol` is a string. Clearly it's something else.

python/tchannel/tornado/format/__init__.py
```diff
@@ -0,0 +1,19 @@
...
+#
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
```
abhinav: Don't add to __init__.py. Use a recent version of the `uber-licence` package. It doesn't add copyright notices to empty files.

python/tchannel/tornado/format/tchannel_json.py
```diff
@@ -0,0 +1,77 @@
...
+            raise TChannelException("Invalid arg schema in request header")
+
+        handler = self.endpoint.get(req.endpoint, None)
+        return handler(
+            JsonRequest(req, self.serializer, self.deserializer),
+            JsonResponse(resp, self.serializer, self.deserializer),
+            proxy)
+
+    @tornado.gen.coroutine
+    def send(self, client, endpoint, jheader, jbody, traceflag=False):
```
abhinav: Don't need to prepend `j` to every parameter name.

python/tchannel/tornado/format/tchannel_json.py
```diff
@@ -0,0 +1,77 @@
...
+
+            if not isinstance(jbody, Stream):
+                body = self.serializer(jbody)
+        except Exception as e:
+            raise TChannelException(e.message)
+
+        resp = yield client.send(
+            arg1=endpoint,
+            arg2=header,
+            arg3=body,
```
abhinav: Indentation

python/tchannel/tornado/util.py
```diff
@@ -52,3 +53,15 @@ def print_arg(request, index):
                 print chunk
     except Exception:
         pass
+
+
+def deserialize_json(content):
+    if content is None:
+        return ""
```
abhinav: the result of deserializing None is not an empty string. It's an error. 

assert that content is not None

python/tchannel/tornado/util.py
```diff
@@ -52,3 +53,15 @@ def print_arg(request, index):
...
+    return json.loads(content)
+
+
+def serialize_json(jobject):
+    if jobject is None:
+        return {}
```
abhinav: The result of serializing `None` is not an empty dictionary. `serialize_json` must always return a string.

python/tests/integration/json/__init__.py
```diff
@@ -0,0 +1,102 @@
...
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+
+import random
+import math
+
+
+class Generator(object):
```
abhinav: You can't use this

python/tchannel/tornado/data.py
breerly: I don't think inheritance is the right choice here. Why not have an object that is responsible for creating a `Request` from json? For example, why not have a `JsonRequestBuilder.build()` that returns an instantiated `Request`?

breerly: Or perhaps a `RequestBuilder.from_json()` - in other-words, there is no benefit to this inheritance tree, it just makes the program brittle.

Raynos:
No markdown file :(

python/tchannel/tornado/connection.py
```diff
@@ -34,20 +33,23 @@
...
-from ..messages.error import ErrorMessage, ErrorCode
-from .message_factory import MessageFactory
-
+from ..messages.common import PROTOCOL_VERSION
+from ..messages.common import generate_checksum
+from ..messages.common import FlagsType
+from ..messages.common import verify_checksum
+from ..messages.error import ErrorCode
+from ..messages.error import ErrorMessage
+from tchannel.event import EventType
```
blampe: relative import

python/tchannel/tornado/dispatch.py
```diff
@@ -91,14 +87,14 @@ def handle_call(self, request, connection):
...
+    def route(self, rule, format=None):
         def decorator(handler):
-            self.register(rule, handler)
+            self.register(rule, handler, format)
             return handler
 
         return decorator
 
-    def register(self, rule, handler):
+    def register(self, rule, handler, protocol=None):
```
blampe: format/protocol are inconstant, let's call it argscheme.

blampe:
Agree that injecting a serializer to the request should simplify things.

python/tchannel/tornado/helper.py
```diff
@@ -0,0 +1,91 @@
...
+
+from .util import deserialize_default
+from .util import serialize_default
+from .stream import Stream
+from ..exceptions import TChannelException, ProtocolException
+from ..messages.common import ArgSchema
+from ..messages.common import ARG_SCHEMA_NAME
+
+
+class TChannelHelper(object):
```
breerly: Oh man, this class name has to change. Never end a class in `Helper` or `Manager` or `Util` or any other "bucket" functionality name. I'll think of something.

python/tchannel/tornado/data.py
```diff
@@ -0,0 +1,294 @@
...
+            id=None,
+            flags=FlagsType.none,
+            ttl=10,
+            tracing=None,
+            service=None,
+            headers=None,
+            checksum=None,
+            argstreams=None,
+            serializer=None,
+            deserializer=None,
```
blampe: Just use one `serializer` object that defines `serialize` and `deserialize` methods.

breerly:
It's going in the right direction, we still have to get the serializers right.

python/tchannel/tornado/data.py
```diff
@@ -0,0 +1,294 @@
...
+        self.argstreams = argstreams
+        self.checksum = checksum
+        self.id = id
+        self.headers = headers or {}
+        self.state = StreamState.init
+        self.endpoint = ""
+        self.header = None
+        self.body = None
+
+        self.serializer = serializer or serialize_default
```
blampe: No need for `serialize_default`; if `serializer` is `None` just don't do anything.

python/tchannel/tornado/data.py
abhinav: To expand on that: what we discussed earlier was that you'll have some interface `Scheme` (or a different name) that has `serialize` and `deserialize` methods that know how to serialize/deserialize types. It can also have another method or property (if you're using `abc` with `abstractproperty`) that knows what its `as` header value is.

python/tchannel/messages/common.py
```diff
@@ -41,7 +41,22 @@
...
+    none = 0x03
+
+
+class ArgSchema(IntEnum):
```
abhinav: This enum is a bad idea. See other comment about `Scheme` interface.

python/tchannel/messages/common.py
```diff
@@ -41,7 +41,22 @@
...
+    http = 0x01,
+    thrift = 0x02,
+    test = 0x03
+
+ARG_SCHEMA_NAME = {
+    ArgSchema.json: "json",
+    ArgSchema.http: "http",
+    ArgSchema.thrift: "thrift",
+    ArgSchema.test: "test",
+}
```
abhinav: Same as above. Having a fixed mapping like that removes all extensibility.

python/tchannel/tornado/data.py
junchaowu: The default serializer will just return what we pass.
In that case, I don't need if-else in the code.

python/tchannel/tornado/helper.py
abhinav: How about `TChannelAgent`?

python/tchannel/tornado/helper.py
```diff
@@ -0,0 +1,91 @@
...
+    def register(self, rule, handler):
+        """Register json handler
+
+        :param rule: endpoint
+        :param handler: endpoint handler
+        """
+        self.endpoint[rule] = handler
+
+    def handle_call(self, req, resp, proxy):
+        if not req.headers.get('as', None) == self.arg_schema_name:
```
abhinav: `!=`

python/tchannel/tornado/helper.py
```diff
@@ -0,0 +1,91 @@
...
+from .util import deserialize_default
+from .util import serialize_default
+from .stream import Stream
+from ..exceptions import TChannelException, ProtocolException
+from ..messages.common import ArgSchema
+from ..messages.common import ARG_SCHEMA_NAME
+
+
+class TChannelHelper(object):
+    """Handle Json format data."""
```
abhinav: Well, this doesn't really care about JSON vs whatever.

python/tchannel/tornado/helper.py
```diff
@@ -0,0 +1,91 @@
...
+
+    def __init__(self, arg_schema=None, serializer=None, deserializer=None):
+        self.endpoint = {}
+        self.arg_schema = arg_schema or ArgSchema.test
+        self.arg_schema_name = ARG_SCHEMA_NAME[self.arg_schema]
+        self.serializer = serializer or serialize_default
+        self.deserializer = deserializer or deserialize_default
+
+    def register(self, rule, handler):
+        """Register json handler
```
abhinav: Not exclusive to JSON

python/tchannel/tornado/util.py
```diff
@@ -52,3 +53,21 @@ def print_arg(request, index):
...
+
+
+def deserialize_json(obj):
+    assert obj
+    return json.loads(obj)
+
+
+def serialize_json(obj):
+    assert obj
+    return json.dumps(obj)
```
abhinav: Actually, `json.dumps(None)` is the string `"null"`.

Only deserialize doesn't know how to handle empty strings or None.

python/tchannel/tornado/helper.py
breerly: Do this. Explain the exact, single responsibility of the class. When I hear `Agent` it sounds like a `Manager` class - it's not obvious what the responsibility is and therefor it's a poor name.

python/tchannel/tornado/util.py
```diff
@@ -52,3 +53,21 @@ def print_arg(request, index):
                 print chunk
     except Exception:
         pass
+
```
breerly: Dude. Rename this file to `json.py` so we don't end up dumping a bunch of un-related functions in it.

blampe: I'm pretty opposed to this whole file. The `print_arg` and `get_arg` methods seem unnecessary now that we have methods on the request, and these things could all be defined inline.

python/tchannel/tornado/helper.py
breerly: Suggest renaming to `FormatBroker`.

python/tchannel/tornado/broker.py
```diff
@@ -0,0 +1,85 @@
...
+
+from __future__ import absolute_import
+
+import tornado
+
+from .stream import Stream
+from ..exceptions import TChannelException, ProtocolException
+
+
+class TChannelBroker(object):
```
breerly: If this isn't part of the public API, which it isn't, then you shouldn't start the object with `TChannel`.

Suggest renaming to `ReqResBroker`.

breerly: After you do this, lgtm.

python/tchannel/tornado/broker.py
```diff
@@ -0,0 +1,85 @@
...
+
+        except Exception as e:
+            raise TChannelException(e.message)
+
+        resp = yield client.send(
+            arg1=endpoint,
+            arg2=raw_header,
+            arg3=raw_body,
+            traceflag=traceflag,
+            headers={'as': self.arg_schema.type},
```
abhinav: `.type()` is a function. You're not calling it here.

abhinav:
looks good besides that and the scheme-vs-schema thing

PR #509, opened 2015-05-01
GoLang README and examples

dacamp:
Replaced getopt with STDLIB flag and removed dependency from Getdeps
  Added default values to examples
 
Move HelloWorld examples into hello directory

Created working PingPong example from README

### DIFF
--- a/golang/examples/hello/client/main.go
+++ b/golang/examples/hello/client/main.go
@@ -21,27 +21,28 @@ package main
 // THE SOFTWARE.
 
 import (
-	"code.google.com/p/getopt"
 	"github.com/uber/tchannel/golang"
 	"golang.org/x/net/context"
+	"flag"
 	"os"
 	"time"
 )
 
 var log = tchannel.SimpleLogger
 
-var peerAddr = getopt.StringLong("peer", 'p', "", "Host and port of remote peer")
-var serviceName = getopt.StringLong("service", 's', "", "Name of service to invoke")
-var operationName = getopt.StringLong("operation", 'o', "", "Name of operation to invoke")
-var arg2 = getopt.StringLong("arg2", '2', "", "Input for arg2.  Curl-style, use @foo.txt to read from foo.txt")
-var arg3 = getopt.StringLong("arg3", '3', "", "Input for arg3.  Curl-style, use @foo.txt to read from foo.txt")
-var timeout = getopt.IntLong("timeout", 't', 30, "Timeout (in seconds)")
+var peerAddr = flag.String("peer", "localhost:10500", "Host and port of remote peer")
+var serviceName = flag.String("service", "TestService", "Name of service to invoke")
+var operationName = flag.String("operation", "echo", "Name of operation to invoke")
+var arg2 = flag.String("arg2", "hello", "Input for arg2.  Curl-style, use @foo.txt to read from foo.txt")
+var arg3 = flag.String("arg3", "world", "Input for arg3.  Curl-style, use @foo.txt to read from foo.txt")
+var timeout = flag.Int("timeout", 30, "Timeout (in seconds)")
 
 func asArgument(arg string) tchannel.Output {
 	if arg[0] == '@' {
 		f, err := os.Open(arg[1:])
 		if err != nil {
-			panic(err)
+			log.Errorf(err.Error())
+			os.Exit(-1)
 		}
 
 		return tchannel.NewStreamingOutput(f)
@@ -52,18 +53,13 @@ func asArgument(arg string) tchannel.Output {
 }
 
 func main() {
-	getopt.Parse()
-	if *peerAddr == "" || *serviceName == "" || *operationName == "" ||
-		*arg2 == "" || *arg3 == "" {
-		getopt.Usage()
-		os.Exit(-1)
-	}
+	flag.Parse()
 
 	ch, err := tchannel.NewChannel(&tchannel.ChannelOptions{
 		Logger: log,
 	})
 	if err != nil {
-		log.Errorf("Could not create client channel: %v", err)
+		log.Errorf("Could not create client channel: %v\n", err)
 		os.Exit(-1)
 	}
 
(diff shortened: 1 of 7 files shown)


golang/examples/hello/client/main.go
```diff
@@ -21,27 +21,28 @@ package main
...
+var arg3 = flag.String("arg3", "world", "Input for arg3.  Curl-style, use @foo.txt to read from foo.txt")
+var timeout = flag.Int("timeout", 30, "Timeout (in seconds)")
 
 func asArgument(arg string) tchannel.Output {
 	if arg[0] == '@' {
 		f, err := os.Open(arg[1:])
 		if err != nil {
-			panic(err)
+			log.Errorf(err.Error())
+			os.Exit(-1)
```
abhinav: Why no `log.Fatal(err)` instead?

dacamp: I tried that first, but tchannel.SimpleLogger doesn't mimic stdlib log (yet).  There's a few different problems within the  SimpleLogger. It's on my radar to take a look at.

abhinav: Ah, didn't realize this was using our logger

golang/examples/hello/server/main.go
```diff
@@ -100,7 +100,7 @@ func main() {
 	ch.Register(tchannel.HandlerFunc(timeout), "TestService", "timeout")
 
 	if err := ch.ListenAndServe(*bindAddr); err != nil {
-		log.Errorf("Could not listen on %s: %v", *bindAddr, err)
+		// log.Errorf("Could not listen on %s: %v HERE!!!", *bindAddr, err)
```
mmihic: cleanup

golang/examples/ping/main.go
```diff
@@ -0,0 +1,96 @@
...
+	// ch.ListenAndServe("127.0.0.1:10500")
+	go listenAndHandle(ch, "127.0.0.1:10500")
+
+	// Create a new TChannel for sending requests.
+	client, err := tchannel.NewChannel(nil)
+	if err != nil {
+		fmt.Println(err)
+	}
+
+	// go listenAndHandle(client, "localhost:10500")
```
mmihic: Cleanup

golang/examples/ping/main.go
```diff
@@ -0,0 +1,96 @@
...
+	if err != nil {
+		fmt.Println(err)
+		return
+	}
+
+	// Register a handler for the ping message on the PingService
+	ch.Register(tchannel.HandlerFunc(pingHandler), "PingService", "ping")
+
+	// Listen for incoming requests
+	// ch.ListenAndServe("127.0.0.1:10500")
```
mmihic: Cleanup

golang/examples/ping/main.go
```diff
@@ -0,0 +1,96 @@
+package main
+
+import (
```
mmihic: needs standard copyright

golang/examples/hello/client/main.go
```diff
@@ -72,22 +68,24 @@ func main() {
...
 	}
 
-	if err := call.WriteArg2(asArgument(*arg2)); err != nil {
+	if err = call.WriteArg2(asArgument(*arg2)); err != nil {
 		log.Errorf("Could not write arg2: %v", err)
 		os.Exit(-1)
 	}
 
-	if err := call.WriteArg3(asArgument(*arg3)); err != nil {
+	if err = call.WriteArg3(asArgument(*arg3)); err != nil {
```
mmihic: even though it's declared above, it's usually still preferential to limit the scope of the specific err

golang/examples/hello/client/main.go
mmihic: fwiw I intentionally didn't use stdlib log directly so that we could report structured metadata through the loggers, ala logrus / logtron.

golang/examples/ping/main.go
```diff
@@ -0,0 +1,96 @@
...
+	// go listenAndHandle(client, "localhost:10500")
+	// Make a call to ourselves, with a timeout of 10s
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second * 10)
+	defer cancel()
+
+	var responseHeaders Headers
+	ping := Ping{"Hello, World!"}
+	var pong Pong
+
+	// RoundTrip returns bool, error.  The bool here is false.  Not sure why.
```
mmihic: It's false because the success case is returning the value of Response.ApplicationError, which of course is false on success.

golang/examples/hello/client/main.go
dacamp: I'll change it back, but variables should only be initialized once.  I'm actually surprised the compiler doesn't bitch here, it should.

golang/examples/ping/main.go
dacamp: Would it make more sense to return `Response.ApplicationError`as an error then?  You could then return a single variable back from `RoundTrip`, error or nil.

mmihic: Yes, that works for me.

dacamp:
Made updates from feedback.

golang/examples/hello/client/main.go
mmihic: They aren't the same variable. The ``err`` in ``if err :=`` is limited to the block inside.

golang/examples/ping/main.go
```diff
@@ -0,0 +1,114 @@
...
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+import (
+	"time"
+	"github.com/uber/tchannel/golang"
+	"golang.org/x/net/context"
+	"fmt"
+	"os"
+	// log "github.com/Sirupsen/logrus"
```
mmihic: eliminate

golang/examples/ping/main.go
```diff
@@ -0,0 +1,114 @@
...
+		fmt.Printf("Could not read body from client: %v", err)
+		return
+	}
+
+	if err := call.Response().WriteArg2(tchannel.NewJSONOutput(headers)); err != nil {
+		fmt.Printf("Could not echo response headers to client: %v", err)
+		return
+	}
+
+	pong := Pong{Message: fmt.Sprintf("ping %s", inArg3)}
```
mmihic: nit: shouldn't this be "pong %s"?

mmihic:
minor comments.  lgtm

golang/examples/ping/main.go
dacamp: I dunno, you wrote it! :)

dacamp:
Removed commented import, rebased, pushed... After a successful build, I'll merge this.

golang/examples/ping/main.go
mmihic: ha, ok

golang/examples/ping/main.go
mmihic: When you do this, we should also change InboundCallResponse so that it only
has a single Err method which either returns the system error or
ErrApplication.  Right now I believe there are seperate methods for
getting system errors vs checking the application error flag.

mmihic: I take that last comment back; I want to preserve this distinction in the
lower level interface.

Just so you know, the reason we draw a hard distinction between these two
is that they represent very different conditions.  If a system error
occurs, the response is complete - you can't do anything more with it.  But
if an application error occurs, the response is still valid - you can still
read args, which might even be streamed.  We can sortof achieve the same
effect with a distinct error, but it's not as good IMO as being explicit in
the api.

It's the same reason that the net/http package doesn't use errors to return
http status information, and reserves errors only for protocol level
problems.

dacamp: Good call, I agree we should explicitly draw the line between a protocol error and an application error.
 
If there's an application error, `RoundTrip` should return nil (the transaction *was* successful). Then, inside the handler callback, `*tchannel.InboundCall` should have a struct field (or method) that returns the application error.

This allow users to break out logical error handling between the protocol and the application levels.

Sound reasonable?

dacamp: I'm going to merge this.  I'll copy this thread over to an issue.

PR #510, opened 2015-05-01
Upgrade debug-logtron & cleanup logs

Raynos:
- DebugLogtron() has enabled to to true so writes
    info, warn, error & fatal.
 - Fix log spew for socket closes
 - Fix warns around self response race condition
 - Add allocCluster.test.only() for sanity.

r: @jcorbin @kriskowal

### DIFF
--- a/node/connection_base.js
+++ b/node/connection_base.js
@@ -314,6 +314,7 @@ TChannelConnectionBase.prototype.onReqDone = function onReqDone(req) {
     if (self.requests.in[req.id] !== req) {
         self.logger.warn('mismatched onReqDone callback', {
             hostPort: self.channel.hostPort,
+            hasInReq: self.requests.in[req.id] !== undefined,
             id: req.id
         });
     } else {
--- a/node/self_out_response.js
+++ b/node/self_out_response.js
@@ -85,8 +85,12 @@ function passError(codeString, message) {
         originalId: self.id,
         message: message
     });
-    self.inreq.outreq.errorEvent.emit(self, err);
     if (!self.closing) self.conn.lastTimeoutTime = 0;
+    process.nextTick(emitError);
+
+    function emitError() {
+        self.inreq.outreq.errorEvent.emit(self, err);
+    }
 };
 
 module.exports.OutResponse = SelfOutResponse;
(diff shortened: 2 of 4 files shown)


node/test/lib/alloc-cluster.js
```diff
@@ -158,6 +145,26 @@ allocCluster.test = function testCluster(desc, opts, t) {
...
+    test.only(desc, function t2(assert) {
+        allocCluster(opts).ready(function clusterReady(cluster) {
+            assert.once('end', function testEnded() {
+                cluster.destroy();
+            });
+            t(cluster, assert);
+        });
+    });
+};
+
```
kriskowal: Consider a less copypasta solution. Perhaps thread test.only as an arg into test or a shared impl.

kriskowal:
:ok:

node/test/lib/alloc-cluster.js
Raynos: Will create shared impl in the future.

node/connection_base.js
```diff
@@ -307,8 +309,11 @@ TChannelConnectionBase.prototype.buildResponse = function buildResponse(req, opt
 TChannelConnectionBase.prototype.onReqDone = function onReqDone(req) {
     var self = this;
     if (self.requests.in[req.id] !== req) {
+        var inreq = self.requests.in[req.id];
+
         self.logger.warn('mismatched onReqDone callback', {
             hostPort: self.channel.hostPort,
+            inreq: !!inreq,
```
jcorbin: how about `hasInReq: self.requests.in[req.id] !== undefined`

node/self_out_response.js
```diff
@@ -85,8 +85,12 @@ function passError(codeString, message) {
         originalId: self.id,
         message: message
     });
-    self.inreq.outreq.errorEvent.emit(self, err);
     if (!self.closing) self.conn.lastTimeoutTime = 0;
+    process.nextTick(emitError);
+
+    function emitError() {
```
jcorbin: why defer

Raynos: necessary to avoid a race condition; can explain in person.

Raynos: defer should not happen here; its the wrong place; let me defer in the correct place.

jcorbin:
Funny, I had taken a run at the info/warn log branch during xlang testing, I propose #512

node/self_out_response.js
Raynos: no wait; this is the correct location.

jcorbin:
lgtm, after fix inreq feedback

Raynos:
Fixed the inreq.

PR #512, opened 2015-05-01
ConnectionBase: cleanup reset all logging

jcorbin:
- don't log stack traces for info logs
- don't log socket close errors


Raynos:
lgtm.

PR #515, opened 2015-05-02
GoLang: Rewrite Logger by embedding Logrus

dacamp:
Takes some of the changes from #185 and cleans up logger.go

Maybe irrelevant due to comments from #509.  I'm not sure if you even want a Logger inside of the library or you want clients to set the logger instead.

### DIFF
--- a/golang/Godeps/Godeps.json
+++ b/golang/Godeps/Godeps.json
@@ -21,6 +21,10 @@
 		{
 			"ImportPath": "golang.org/x/net/context",
 			"Rev": "b6fdb7d8a4ccefede406f8fe0f017fb58265054c"
+		},
+	        {
+			"ImportPath": "github.com/Sirupsen/logrus",
+			"Rev": "aaf92c95712104318fc35409745f1533aa5ff327"
 		}
 	]
 }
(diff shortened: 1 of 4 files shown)


dacamp:
Gotta update godeps.

mmihic:
imo this is not what we should do.  There is no real standardization of log approaches in the golang community yet.  Logger is intentionally an interface that users of the library can provide, so that applications can plugin whatever logging framework they use without the library mandating the use of a particular framework.

dacamp:
I absolutely agree, allowing plugins is a much better approach.  The current logger as well as calling `var log = tchannel.SimpleLogger` in the examples lead me to believe you were going for more of an inside out approach.

BTW, travis failed for the same reason I was having yesterday. It's why I pulled getopt from the examples.
> godep: https://code.google.com/p/getopt/source/checkout?repo=: 502 Bad Gateway

mmihic:
Ah yeah, google closed down google code.  pfft.  I don't know where getopt has moved, but it's probably better to remove anyway.  We only used it because it has better support for repeated flags, but that's not needed for these examples and the fewer external dependencies the better.

PR #518, opened 2015-05-04
seperate serialize/deserialize for body and header in argscheme

junchaowu:
@blampe @abhinav @breerly

### DIFF
--- a/python/tchannel/scheme.py
+++ b/python/tchannel/scheme.py
@@ -24,13 +24,34 @@ import json
 
 
 class ArgScheme(object):
+    """ArgScheme defines interface on how to serialize/deserialize
+    header and body.
+
+    Customized ArgScheme subclass must implement all methods::
+
+        def type()
+
+        def serialize_header(obj)
+
+        def deserialize_header(obj)
+
+        def serialize_body(obj)
+
+        def deserialize_body(obj)
+    """
     def type(self):
         raise NotImplementedError()
 
-    def serialize(self, obj):
+    def serialize_header(self, obj):
         raise NotImplementedError()
 
-    def deserialize(self, obj):
+    def deserialize_header(self, obj):
+        raise NotImplementedError()
+
+    def serialize_body(self, obj):
+        raise NotImplementedError()
+
+    def deserialize_body(self, obj):
         raise NotImplementedError()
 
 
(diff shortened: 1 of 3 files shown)


python/tchannel/scheme.py
```diff
@@ -41,10 +62,18 @@ def __init__(self):
     def type(self):
         return self._type
 
-    def deserialize(self, obj):
+    def deserialize_header(self, obj):
+        assert obj
```
blampe: Don't use `assert` outside of tests. If it's an error then raise an exception, or handle it. In this case I think it would be more appropriate for the JSON library to raise an exception that we can catch.

abhinav: I think it's okay to use `assert`s for what are essentially 'programmer errors'. Things like invalid state, for example.

breerly:
I'm ok with merging this, but I want to point out the OOP smell:

There is serialization for header and body mixed into the argscheme. Have you ever heard about the term `encapsulate the concept that varies`? That principle points to this situation - you have objects that have the same needs, but different implementation. The concept that varies is the fact that its a header or body. To encapsulate that, I probably would do something like:

```
SerializableInterface(abc):
   seriliaze
   deserialize

HeaderSerializer(SerializableInterface):
   pass

FooterSerializer(SerializableInterface):
   pass
```

Then the argscheme could use those objects, effectively delegating the work and encapsulating the concept that varies.

python/tchannel/scheme.py
abhinav: Although in this specific case, we should try to handle as many cases as we can. @junchaowu consider returning `None` if `obj` is `None` or an empty string.

junchaowu:
Good point. Will keep in mind next time.

python/tchannel/scheme.py
abhinav: Also, nit: `obj` is a poor argument name for this case because we know it's a string. `s` or `string` would be preferable.

python/tchannel/scheme.py
```diff
@@ -41,10 +62,18 @@ def __init__(self):
...
+        return json.loads(obj)
+
+    def serialize_header(self, obj):
+        assert obj
+        return json.dumps(obj)
```
abhinav: It's totally possible to serialize None. No reason to assert here.  `json.dumps` knows how to handle `None`.

python/tchannel/scheme.py
```diff
@@ -41,10 +62,18 @@ def __init__(self):
...
+
+    def deserialize_body(self, obj):
         assert obj
         return json.loads(obj)
 
-    def serialize(self, obj):
+    def serialize_body(self, obj):
         assert obj
         return json.dumps(obj)
```
abhinav: Same as above.

python/tchannel/scheme.py
junchaowu: For the edge cases like None or empty, I am planning to include them in the error handling tasks. At that time, I'll list all the possible cases in tchannel with corresponding responses of tchannel.

abhinav: I'm not suggesting you treat it as an error. Treat it as valid input. Make `deserialize(None) == deserialize('') == None`

abhinav:
:shipit: 

+1 @breerly's comment. We should move the actual serialize-deserialize concerns somewhere else in the future.

PR #519, opened 2015-05-04
Refactor streaming_bisect test => ClusterSearch

jcorbin:
This sets the stage for:
- other cluster strategies than inproc (allocCluster)
- easy use by other future tests

r @kriskowal @Raynos


kriskowal:
:ok:

PR #521, opened 2015-05-05
peer_state: add more whitespace to peer string

Raynos:
I noticed that the toString() of the peer state was
missing some whitespace so I added it.

r: @jcorbin @kriskowal

### DIFF
--- a/node/peer_states.js
+++ b/node/peer_states.js
@@ -90,7 +90,7 @@ TChannelPeerHealthyState.prototype.type = 'tchannel.healthy';
 
 TChannelPeerHealthyState.prototype.toString = function healthyToString() {
     var self = this;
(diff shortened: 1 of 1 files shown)


node/peer_states.js
```diff
@@ -90,7 +90,7 @@ TChannelPeerHealthyState.prototype.type = 'tchannel.healthy';
 
 TChannelPeerHealthyState.prototype.toString = function healthyToString() {
     var self = this;
-    return '[HealthyPeer ' + self.okCount + 'ok ' + self.notOkCount + 'err]';
+    return '[HealthyPeer ' + self.okCount + ' ok ' + self.notOkCount + ' err]';
```
jcorbin: Mistakes like this are easy to make when you don't use something like util.format...

jcorbin:
:+1:

kriskowal:
:ok:

PR #522, opened 2015-05-05
implement the ping API for tchannel(node)

ShanniLi:
r @jcorbin @kriskowal @Raynos 

1. Implement the ping API for tchannel(node).
2. add a ping example in example/ping.js
3. add a ping test in test/ping.js


node/connection.js
```diff
@@ -151,6 +153,14 @@ TChannelConnection.prototype.setupHandler = function setupHandler() {
         self.onCallResponse(res);
     }
 
+    function onPingRequest(req) {
+        self.onPingRequest(req);
+    }
+
+    function onPingResponse(req) {
```
jcorbin: s/req/res

node/connection.js
```diff
@@ -213,6 +223,26 @@ TChannelConnection.prototype.onCallResponse = function onCallResponse(res) {
     req.responseEvent.emit(req, res);
 };
 
+TChannelConnection.prototype.onPingRequest = function onPingRequest(req) {
+    var self = this;
+    req.remoteAddr = self.remoteName;
```
jcorbin: you shouldn't use an event emitter to mutate the thing being observed, this needs reconsidered.

node/connection.js
```diff
@@ -213,6 +223,26 @@ TChannelConnection.prototype.onCallResponse = function onCallResponse(res) {
...
+};
+
+TChannelConnection.prototype.onPingResponse = function onPingResponse(res) {
+    var self = this;
+    var req = self.popOutReq(res.id);
```
jcorbin: Did I miss where you're putting ping reqs into `self.requests.out`?  I thought we agreed to not do any frame id correlation like we do for call req/res since there's no need for correlated callbacks, and we don't intend to make it an external api.

node/connection.js
ShanniLi: fixed.

node/examples/ping.js
```diff
@@ -0,0 +1,57 @@
...
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+var TChannel = require('../channel.js');
```
jcorbin: missing 'use strict';

node/examples/ping.js
```diff
@@ -0,0 +1,57 @@
+// Copyright (c) 2015 Uber Technologies, Inc.
```
jcorbin: drop this example, ping is not to be an external api at present.

node/out_request.js
```diff
@@ -104,6 +104,14 @@ TChannelOutRequest.prototype._sendCallRequestCont = function _sendCallRequestCon
     });
 };
 
+TChannelOutRequest.prototype._sendPingRequest = function _sendPingRequest() {
```
jcorbin: this should not be on the req res objects, it's a concern external to call req/res.

node/out_request.js
```diff
@@ -185,6 +193,12 @@ TChannelOutRequest.prototype.sendCallRequestContFrame = function sendCallRequest
     }
 };
 
+TChannelOutRequest.prototype.sendPingRequestFrame = function sendPingRequestFrame() {
```
jcorbin: ditto, wrong place

node/out_request.js
```diff
@@ -197,6 +211,16 @@ TChannelOutRequest.prototype.send = function send(arg1, arg2, arg3, callback) {
     return self;
 };
 
+TChannelOutRequest.prototype.ping = function ping(callback) {
```
jcorbin: public api, not what we want.

node/connection.js
jcorbin: this can just be `handler.sendPingRequest()`, which has access to `self.hostPort`, which is the same as `self.remoteName` on conn.

jcorbin:
What we need:
- [ ] `TChannelV2Handler#sendPing{Request,Respanse}` as an analog of `TChannelV2Handler#sendInit{Request,Response}`
- [ ] an api for internal use only, so `conn.ping()` would suffice for now (you'll need to do something so that self connections are ping able too)

kriskowal:
@jcorbin Likewise, I’ve made a note to self to implement a no-op connect() on self peer.

jcorbin:
erm, self-peer does have a .connect method, it returns a self connection or makes one.

jcorbin:
needs a rebase on latest master, waiting on that before review

ShanniLi:
done rebasing master. :)

kriskowal:
Looks good to me.

node/connection_base.js
```diff
@@ -266,6 +267,12 @@ TChannelConnectionBase.prototype.handleCallRequest = function handleCallRequest(
     }
 };
 
+TChannelConnectionBase.prototype.handlePingResponse =
+function handlePingResponse(res) {
```
jcorbin: this line breaking is locally inconsistent (within class, within module)

node/connection_base.js
```diff
@@ -266,6 +267,12 @@ TChannelConnectionBase.prototype.handleCallRequest = function handleCallRequest(
...
+    var self = this;
+    self.pingResponseEvent.emit(self, res);
```
jcorbin: Exposes version-specific frame shapes out of a connection-external api, connection exists to abstract over specifcs... the fact that you had to introduce a v2 require within `self_connection` should've been the clue.

node/self_connection.js
```diff
@@ -24,6 +24,7 @@ var OutRequest = require('./self_out_request').OutRequest;
 var OutResponse = require('./self_out_response').OutResponse;
 var StreamingOutRequest = require('./self_out_request').StreamingOutRequest;
 var StreamingOutResponse = require('./self_out_response').StreamingOutResponse;
+var v2 = require('./v2/index');
```
jcorbin: this, adding this line should've been a red flag

node/connection_base.js
```diff
@@ -266,6 +267,12 @@ TChannelConnectionBase.prototype.handleCallRequest = function handleCallRequest(
     }
 };
 
+TChannelConnectionBase.prototype.handlePingResponse =
```
jcorbin: also this method should be on TChannelConnection, not this base class... handler entry points like this are to be out of base class

jcorbin: also once you get it there, it will be clear that the arg should be spelled `resFrame` to make it clearer

node/self_connection.js
```diff
@@ -76,4 +77,12 @@ TChannelSelfConnection.prototype.buildOutResponse = function buildOutResponse(in
     }
 };
 
+TChannelSelfConnection.prototype.ping = function ping() {
+    var self = this;
+    var id = self.idCount++;
+    var body = new v2.PingResponse();
+    var res = new v2.Frame(id, body);
+    self.handlePingResponse(res);
```
jcorbin: we can just get away with an anonymous object in both cases for now, with something like `// TODO: explicit type`:

So here we just have: 
```
self.pingResponseEvent.emit({self, 
  id: id
});
```

Above, in connection, we have:
```
self.pingResponseEvent.emit({self, 
  id: resFrame.id
});
```

node/test/ping.js
```diff
@@ -0,0 +1,66 @@
...
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+/* jshint maxparams:5 */
+
+'use strict';
+
+var tape = require('tape');
```
jcorbin: please call this `var test = require('tape');` like all our other tests do

node/test/ping.js
```diff
@@ -0,0 +1,66 @@
...
+var TChannel = require('../channel.js');
+var Ping = require('../v2/ping.js');
+
+tape('ping with a remote connection', function (assert) {
+
+    var client = new TChannel();
+    var server = new TChannel();
+
+    server.listen(0, '127.0.0.1', function onListen() {
```
jcorbin: why are you doing all of this yourself rather than using `allocCluster` like all our other tests do?

Raynos:
Let's get this one shipped after the init req/res blocking

node/connection.js
```diff
@@ -115,6 +115,7 @@ TChannelConnection.prototype.setupHandler = function setupHandler() {
     self.handler.errorEvent.on(onHandlerError);
     self.handler.callIncomingRequestEvent.on(onCallRequest);
     self.handler.callIncomingResponseEvent.on(onCallResponse);
+    self.handler.pingIncomingResponseEvent.on(onPingResponse);
```
Raynos: what about `pingIncomingRequest` ?

Raynos: oh pingIncomignRequset was already hooked up.

Raynos:
lgtm. Matching id's is left as an exercise to the user.

node/self_connection.js
```diff
@@ -76,4 +76,11 @@ TChannelSelfConnection.prototype.buildOutResponse = function buildOutResponse(in
     }
 };
 
+TChannelSelfConnection.prototype.ping = function ping() {
+    var self = this;
+    var id = self.idCount++;
+    // TODO: explicit type
+    self.pingResponseEvent.emit(self, {id: id});
```
Raynos: we should return the id here.

node/connection.js
```diff
@@ -247,6 +258,11 @@ TChannelConnection.prototype.onCallResponse = function onCallResponse(res) {
     }
 };
 
+TChannelConnection.prototype.ping = function ping() {
+    var self = this;
+    self.handler.sendPingRequest();
```
Raynos: we should return the id here

PR #523, opened 2015-05-05
add thrfit arg scheme

junchaowu:
@blampe @breerly @abhinav 
The python generated code gives us so many limitations to integrated with thrift. Probably this is what we can do so far.

### DIFF
--- a/python/tchannel/scheme.py
+++ b/python/tchannel/scheme.py
@@ -77,3 +77,25 @@ class JsonArgScheme(ArgScheme):
 
     def serialize_body(self, obj):
         return json.dumps(obj)
+
+
+class ThriftArgScheme(ArgScheme):
+    def __init__(self):
+        self._type = "thrift"
+
+    def type(self):
+        return self._type
+
+    def deserialize_header(self, obj):
+        # TODO implement after we figure out how to handle thrift header
+        return obj
+
+    def serialize_header(self, obj):
+        # TODO implement after we figure out how to handle thrift header
+        return obj
+
+    def deserialize_body(self, obj):
+        return obj
+
+    def serialize_body(self, obj):
+        return obj
--- a/python/tchannel/thrift/server.py
+++ b/python/tchannel/thrift/server.py
@@ -21,10 +21,13 @@
 from __future__ import absolute_import
 
 import inspect
+import tornado
 
 from thrift.transport import TTransport
 from thrift.protocol import TBinaryProtocol
-from tchannel.tornado.stream import InMemStream
+from ..scheme import ThriftArgScheme
+from ..tornado.stream import InMemStream
+from ..tornado.broker import ArgSchemeBroker
 
 
 def register(dispatcher, service_module, handler, service_name=None):
@@ -74,7 +77,9 @@ def register(dispatcher, service_module, handler, service_name=None):
     for method in methods:
         endpoint = "%s::%s" % (service_name, method)
         dispatcher.register(
-            endpoint, build_handler(service_module, method, handler)
+            endpoint,
+            build_handler(service_module, method, handler),
+            ArgSchemeBroker(ThriftArgScheme())
         )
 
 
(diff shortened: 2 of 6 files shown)


python/tchannel/scheme.py
```diff
@@ -77,3 +77,25 @@ def deserialize_body(self, obj):
...
     def serialize_body(self, obj):
         return json.dumps(obj)
+
+
+class ThriftArgScheme(ArgScheme):
+    def __init__(self):
+        self._type = "thrift"
+
+    def type(self):
+        return self._type
```
abhinav: Why assign to this attribute at all? This will never change. Just make it a class attribute with a constant value and return that.

python/tchannel/thrift/server.py
```diff
@@ -82,23 +87,28 @@ def build_handler(service_module, method_name, handler):
...
             TBinaryProtocol.TBinaryProtocolAccelerated(
-                TTransport.TMemoryBuffer(request.body)
+                TTransport.TMemoryBuffer(body)
             )
         )
 
         result = result_type()
         try:
+            thrift_args = [
+                getattr(args, spec[2]) for spec in args.thrift_spec[1:]]
```
abhinav: If the `[` is not on the same line as the expression, the `]` must be on the next line too.

    thrift_args = [
        getattr(...) ...
    ]

python/tchannel/thrift/server.py
```diff
@@ -82,23 +87,28 @@ def build_handler(service_module, method_name, handler):
...
+            thrift_args.append(proxy)
```
abhinav: No. This is a bad way to inject the proxy. An arbitrary "last argument is proxy" is a bad interface and really paints us in a corner if we want to make any change to the scheme. It breaks all existing Thrift handlers (which currently just conform to the interface specified for the service in the IDL).

A possible better way would be to make it an instance attribute and construct a new Handler for each request with the Proxy for that request set to this value. But we wouldn't want to enforce a constructor shape either.

We need to have an offline discussion about the Thrift interface overall before we make a change like this.

python/tchannel/thrift/server.py
```diff
@@ -82,23 +87,28 @@ def build_handler(service_module, method_name, handler):
...
             result.success = getattr(handler, method_name)(
-                *[getattr(args, spec[2]) for spec in args.thrift_spec[1:]]
+                *thrift_args
```
abhinav: If the expression is short enough, move it to the previous line.

python/tchannel/thrift/server.py
```diff
@@ -21,10 +21,13 @@
 from __future__ import absolute_import
 
 import inspect
+import tornado
 
 from thrift.transport import TTransport
 from thrift.protocol import TBinaryProtocol
```
blampe: Import order: stdlib, packages, local.

```
import inspect # stdlib

from thrift ...
import tornado

from . ...
```

python/tchannel/thrift/server.py
blampe: This looks like it will open us up to versioning issues if the spec is expanded.

python/tchannel/thrift/transport/tornado.py
```diff
@@ -20,9 +20,13 @@
 
 from __future__ import absolute_import
 
+
 from tornado import gen
 from tchannel.io import BytesIO
 from tchannel.tornado.stream import InMemStream
+from tchannel.scheme import ThriftArgScheme
+from tchannel.tornado.broker import ArgSchemeBroker
```
blampe: These should be relative imports.

junchaowu: Relative imports are possible only inside a package.
Here two levels. Use absolute import

python/tchannel/scheme.py
breerly: +1

abhinav:
this was handled in the other CR

PR #524, opened 2015-05-05
Presenting: cross language testing -- streaming_bisect

jcorbin:
- implemented minimal python and node test server entry points
- add a runner script which manages spinning up a language's test server, and then runs a test against it 
- support running against remote process in `node/test/streaming_bisect.js`


node/test/lib/cluster_search.js
```diff
@@ -280,6 +285,59 @@ function inprocClusterCreator(options) {
     }
 }
 
+function clusterClientCreator(options) {
+    var TChannel = require('../../channel');
```
Raynos: nitpick: require at top of file

Raynos:
lgtm.

PR #527, opened 2015-05-06
node: restore coverage tooling

jcorbin:
r @Raynos @kriskowal


node/.istanbul.yml
```diff
@@ -0,0 +1,17 @@
...
+        - benchmarks/*
+        - examples/*
+        - v1/*
+check:
+    global:
+        statements: 80
+        lines: 80
+        branches: 60
+        functions: 80
+    each:
```
jcorbin: Once we get any hyperbahn tests, we can have non-zero per-file expectations.

node/.istanbul.yml
```diff
@@ -0,0 +1,17 @@
...
+    include-all-sources: true
+    excludes:
+        - benchmarks/*
+        - examples/*
+        - v1/*
+check:
+    global:
+        statements: 80
+        lines: 80
+        branches: 60
```
Raynos: #getdat80

Raynos:
lgtm

PR #528, opened 2015-05-06
Cull back unneeded dependencies

Raynos:
This is a general cleanup PR to reduce the volume of dependencies

We currently have 44 dependencies and I would love to get that
to around ~20-25.

This PR brings it down to 36 and blocks on the bufrw PR

r: @jcorbin @kriskowal

### DIFF
--- a/node/package.json
+++ b/node/package.json
@@ -25,7 +25,6 @@
     "farmhash": "^0.2.0",
     "is-error": "2.0.0",
     "json-stringify-safe": "5.0.0",
-    "null-logtron": "2.2.0",
     "readable-stream": "1.0.33",
     "ready-signal": "^1.1.1",
     "run-parallel": "^1.1.0",
--- a/node/package.json
+++ b/node/package.json
@@ -23,7 +23,7 @@
     "crc": "^3.2.1",
     "error": "^6.2.0",
     "farmhash": "^0.2.0",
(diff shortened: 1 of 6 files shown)


kriskowal:
:ok:

node/package.json
```diff
@@ -19,16 +19,15 @@
...
+    "error": "^6.2.0",
     "farmhash": "^0.2.0",
-    "is-error": "2.0.0",
-    "json-stringify-safe": "5.0.0",
-    "null-logtron": "2.2.0",
+    "json-stringify-safe": "^5.0.0",
     "readable-stream": "1.0.33",
     "ready-signal": "^1.1.1",
-    "run-parallel": "^1.1.0",
+    "rezult": "^1.0.0",
```
jcorbin: so wait, that's strictly speaking a new dependency... we already had this in bufrw?

jcorbin:
:-1:  there's way too much going on in this PR, please break it up
- dropping the logger and forcing all the use sites to branch is a big change
- maybe there's some other okay changes in here, but it's hard to see because of the logger change pervading
- shifting bufrw to a git link seems hasty, let's fix whatever the issue is that led you to that, heck let's even talk about wth the issue even was?

Raynos:
Closing in favor of 4 other PRs.

PR #529, opened 2015-05-06
fix chain checksum bug

junchaowu:
@blampe @breerly @abhinav

### DIFF
--- a/python/tchannel/handler.py
+++ b/python/tchannel/handler.py
@@ -19,8 +19,9 @@
 # THE SOFTWARE.
 from __future__ import absolute_import
 
-from .messages import Types
+from .messages import Types, ErrorCode
 from .messages import PingResponseMessage
+from .exceptions import InvalidChecksumException
 
 
 class RequestHandler(object):
@@ -92,8 +93,20 @@ class BaseRequestHandler(RequestHandler):
         :param message_id: message id
         :param message: CallRequestMessage or CallRequestContinueMessage
         :param connection: tornado connection
+
+        :except InvalidChecksumException:
+            Raise InvalidChecksumException when the checksum in the message
+            is wrong.
         """
-        req = connection.request_message_factory.build(message_id, message)
+        try:
+            req = connection.request_message_factory.build(message_id, message)
+        except InvalidChecksumException as e:
+            connection.send_error(
+                ErrorCode.bad_request,
+                e.message,
+                message_id,
+            )
+
         # call handler only for the call request message not continue message
         if req:
             self.handle_call(req, connection)
--- a/python/tchannel/handler.py
+++ b/python/tchannel/handler.py
@@ -19,9 +19,10 @@
 # THE SOFTWARE.
 from __future__ import absolute_import
 
-from .messages import Types, ErrorCode
-from .messages import PingResponseMessage
 from .exceptions import InvalidChecksumException
+from .messages import ErrorCode
+from .messages import PingResponseMessage
+from .messages import Types
 
 
 class RequestHandler(object):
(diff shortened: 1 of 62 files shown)


breerly:
screen shot 2015-05-06 at 11 12 31 am

breerly:
This is great, let's write a test on our end to prevent regression.

junchaowu:
Test will cover the chain checksum.

python/tchannel/messages/common.py
```diff
@@ -112,30 +112,39 @@ def compute_checksum(checksum_type, args, csum=0):
     return csum
 
 
-def generate_checksum(message):
+def generate_checksum(message, pre_csum=0):
```
blampe: `previous_checksum`

python/tchannel/messages/common.py
```diff
@@ -112,30 +112,39 @@ def compute_checksum(checksum_type, args, csum=0):
...
+            message.args,
+            pre_csum,
+        )
 
         message.checksum = (message.checksum[0], csum)
 
 
-def verify_checksum(message):
-    """
+def verify_checksum(message, pre_csum=0):
```
blampe: `previous_checksum`

python/tchannel/handler.py
```diff
@@ -19,8 +19,9 @@
 # THE SOFTWARE.
 from __future__ import absolute_import
 
-from .messages import Types
+from .messages import Types, ErrorCode
```
blampe: One import per line.

python/tchannel/handler.py
```diff
@@ -92,8 +94,20 @@ def handle_pre_call(self, message_id, message, connection):
         :param message_id: message id
         :param message: CallRequestMessage or CallRequestContinueMessage
         :param connection: tornado connection
+
+        :except InvalidChecksumException:
+            Raise InvalidChecksumException when the checksum in the message
+            is wrong.
```
abhinav: The docstring is wrong. This method never raises that exception. It actually catches it.

python/tchannel/tornado/message_factory.py
```diff
@@ -45,6 +53,9 @@ def __init__(self, remote_host=None, remote_host_port=None):
         self.remote_host = remote_host
         self.remote_host_port = remote_host_port
 
+        self.in_checksum = {}
+        self.out_checksum = {}
```
abhinav: Consider adding a comment (not a docstring) about what these dictonaries will hold.

python/tchannel/tornado/message_factory.py
```diff
@@ -315,10 +329,40 @@ def fragment(message):
...
+        ):
+            self.in_checksum[message_id] = message.checksum[1]
+
+            if message.flags == FlagsType.none:
+                self.in_checksum.pop(message_id)
+        else:
+            try:
+                self.in_checksum.pop(message_id)
+            except KeyError:
+                pass
```
abhinav: Please don't use exceptions for control flow. There's a significant performance hit to that. Check if `message_id in self.in_checksum` manually.

python/tchannel/tornado/message_factory.py
```diff
@@ -315,10 +329,38 @@ def fragment(message):
...
+                self.in_checksum.get(message_id, 0),
+        ):
+            self.in_checksum[message_id] = message.checksum[1]
+
+            if message.flags == FlagsType.none:
+                self.in_checksum.pop(message_id)
+        else:
+            if message_id in self.in_checksum:
+                self.in_checksum.pop(message_id)
+            raise InvalidChecksumException("Checksum does not match!")
```
abhinav: This is fine for now but a code-style nit for future changes:

Generally, instead of

    if foo:
        # do stuff
    else:
        raise Something

It's cleaner to do:

    if not foo:
        raise Something
    # do stuff

PR #530, opened 2015-05-06
package: inline is-error

Raynos:
One of many PRs to reduce the number of dependencies in tchannel

r: @jcorbin @kriskowal


jcorbin:
:+1:

PR #532, opened 2015-05-06
Cleanup ranges in package.json

Raynos:
General cleanup; we want to favor ranges instead of strict
versions to give application developers and upstream more
control over what version they want

r: @jcorbin @kriskowal


jcorbin:
:+1:  once you rebase

PR #533, opened 2015-05-06
package: remove null-logtron

Raynos:
We have a `null-logtron` dependency purely for as/json and
for the tracer.

Ideally I would like to remove it; this can be done in two
ways.

 - use `if (logger)` statements everywhere
 - allow the tracer and as layers to re-use `channel.logger`

This PR currently does the former; the `if` statements.

r: @jcorbin @kriskowal

### DIFF
--- a/node/as/json.js
+++ b/node/as/json.js
@@ -38,7 +38,8 @@ function TChannelJSON(options) {
 
     var self = this;
 
-    self.logger = options && options.logger || null;
+    // lazily populated from tchannel
+    self.logger = null;
 
     var bossMode = options && options.bossMode;
     self.bossMode = typeof bossMode === 'boolean' ? bossMode : false;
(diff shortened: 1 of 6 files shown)


jcorbin:
please use the null-logger module that we inlined, or shift them all to use null-logtron instead and keep the dep

jcorbin:
but don't introduce if (logger) branching for part

Raynos:
Ready for second review: @jcorbin @kriskowal

node/as/json.js
```diff
@@ -39,7 +38,8 @@ function TChannelJSON(options) {
 
     var self = this;
 
-    self.logger = options && options.logger || NullLogtron();
+    // lazily populated from tchannel
+    self.logger = null;
```
jcorbin: can you please require `../null-logger` and default to it? then drop the `if (self.logger)` guarding.

Raynos:
@jcorbin fixed.

jcorbin:
🎁

PR #537, opened 2015-05-07
As http

jcorbin:
Minimal working example of an HTTP "bridge" over TChannel:
1) start an http service:
```
$ python -m SimpleHTTPServer # listens on port 8000
```
2) start an ingress proxy which will convert TChannel calls to http requests:
```
$ node examples/http_ingress.js --tchannel-port 4040 example 127.0.0.1:8000
```

3) start an egress proxy which will convert HTTP requests to TChannel calls: 
```
$ node examples/http_egress.js --http-port 8080 --peers 127.0.0.1:4040
```

4) use it through the egress node:
```
$ curl http://localhost:8080 # or open in browser to taste
```

### DIFF
--- a/node/as/http.js
+++ b/node/as/http.js
@@ -40,7 +40,7 @@ function HTTPReqArg2(method, url, headerPairs) {
 
 HTTPReqArg2.RW = bufrw.Struct(HTTPReqArg2, {
     method: bufrw.str1,   // method~1
-    url: bufrw.str2,      // url~2
+    url: bufrw.strn,      // url~N
     headerPairs: headerRW // numHeaders:2 (headerName~2 headerValue~2){numHeaders}
 });
 
@@ -53,7 +53,7 @@ function HTTPResArg2(statusCode, message, headerPairs) {
 
 HTTPResArg2.RW = bufrw.Struct(HTTPResArg2, {
     statusCode: bufrw.UInt16BE, // statusCode:2
-    message: bufrw.str2,        // message~2
+    message: bufrw.strn,        // message~N
     headerPairs: headerRW       // numHeaders:2 (headerName~2 headerValue~2){numHeaders}
 });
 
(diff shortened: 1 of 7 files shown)


node/as/http.js
```diff
@@ -0,0 +1,257 @@
...
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+'use strict';
+
+var assert = require('assert');
+var bufrw = require('bufrw');
+var PassThrough = new require('readable-stream').PassThrough;
```
kriskowal: `new` does not appear to be intended in this case.

jcorbin:
ready for review @kriskowal @Raynos

node/as/http.js
```diff
@@ -0,0 +1,340 @@
...
+
+TChannelHTTP.prototype.sendRequest = function send(treq, hreq, options, callback) {
+    assert(treq.streamed, 'as http must have a streamed tchannel request');
+
+    if (typeof options === 'function') {
+        callback = options;
+        options = null;
+    }
+
+    var i = hreq.url.indexOf('?');
```
Raynos: unused variable?

Raynos:
Looks good enough.

Let's add a `docs/http.md` before we shipping it.

jcorbin:
Added documentation, and shifted url and method to be ~N (new bufrw varint support).

blampe:
Very cool! Interesting `arg2` semantics. I'm a little skeptical of circuit-breaking on `arg1`, because that would require clients to all agree on a name for the path they're hitting, and there's nothing to enforce that name. This is just an edge case that we probably don't care about, though. In the worst case this would still provide per-client circuits, which is better than nothing.

PR #538, opened 2015-05-07
add default 'as': 'raw' in the headers

junchaowu:
@blampe @abhinav @breerly 
create default rawargscheme

### DIFF
--- a/python/tchannel/scheme.py
+++ b/python/tchannel/scheme.py
@@ -56,8 +56,7 @@ class ArgScheme(object):
 
 
 class JsonArgScheme(ArgScheme):
-    def __init__(self):
-        self._type = "json"
+    _type = 'json'
 
     def type(self):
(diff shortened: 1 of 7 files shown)


breerly:
Sure. :ship:

python/tchannel/scheme.py
```diff
@@ -77,3 +76,22 @@ def deserialize_body(self, obj):
 
     def serialize_body(self, obj):
         return json.dumps(obj)
+
+
+class RawArgScheme(ArgScheme):
+    _type = "raw"
+
+    def type(self):
+        return self._type
```
blampe: Not necessary.

junchaowu: I want to make it as a required field. If user creates customerized the argscheme subclass, make sure they have type in it.

PR #539, opened 2015-05-07
move hyperbahn client into tchannel

blampe:
WIP. Can successfully talk to hyperbahn but need to update this to use peering logic.

### DIFF
--- a/python/tchannel/tornado/hyperbahn.py
+++ b/python/tchannel/tornado/hyperbahn.py
@@ -36,7 +36,7 @@ class HyperbahnClient(object):
     def __init__(
         self,
         service,
-        hyperbahn_routers,
+        routers,
         tchannel=None
     ):
         """
@@ -44,7 +44,7 @@ class HyperbahnClient(object):
             Address at which this service can be reached. For example,
             "127.0.0.1:2499".
 
-        :param hyperbahn_routers: list of hyperbahn addresses, e.g.,
+        :param routers: list of hyperbahn addresses, e.g.,
             ``["127.0.0.1:21300", "127.0.0.1:21301"]``.
 
         :param tchannel: ``TChannel`` instance to make Hyperbahn requests with.
@@ -52,7 +52,7 @@ class HyperbahnClient(object):
         self.service = service
         self.tchannel = tchannel or TChannel()
 
-        for known_peer in hyperbahn_routers:
+        for known_peer in routers:
             self.tchannel.peers.add(known_peer)
 
     @tornado.gen.coroutine
--- a/python/tchannel/tornado/hyperbahn.py
+++ b/python/tchannel/tornado/hyperbahn.py
@@ -21,34 +21,77 @@
 from __future__ import absolute_import
 
 import json
-import random
 import tornado.gen
 
-from .messages.error import ErrorCode
-from .tornado.tchannel import TChannel
+from ..messages.error import ErrorCode
+from .tchannel import TChannel
 
 
 DEFAULT_TTL = 60  # seconds
 
 
-class HyperbahnRegistrant(object):
-    """Register a service with a Hyperbahn routing mesh."""
+class HyperbahnClient(object):
+    """Client for talking with the Hyperbahn."""
 
-    def __init__(self, hyperbahn_client, service):
+    def __init__(
+        self,
+        service,
+        hyperbahn_routers,
+        tchannel=None
+    ):
         """
-        :param hyperbahn_client: ``HyperbahnClient`` to make ``register``
-            requests with.
+        :param hostport:
+            Address at which this service can be reached. For example,
+            "127.0.0.1:2499".
 
-        :param service: Name of service to register with Hyperbahn, eg "maps".
+        :param hyperbahn_routers: list of hyperbahn addresses, e.g.,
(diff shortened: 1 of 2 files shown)


python/tchannel/tornado/hyperbahn.py
```diff
@@ -0,0 +1,142 @@
...
+                    {
+                        "serviceName": self.service,
+                        "cost": 0,
+                    },
+                ],
+            }),
+            protocol_headers={
+                "as": "json",
+            },
+        )
```
junchaowu: why not use yield ArgSchemeBroker(JsonArgScheme()).send(
        client,
        endpoint,
        header,
        body,
    )

python/tchannel/tornado/hyperbahn.py
```diff
@@ -0,0 +1,142 @@
...
+
+        # TODO: it would be nice to have message_type on the Response
+        response = yield self.hyperbahn_client.request(**forward_request)
+
+        if response.code not in ErrorCode:
+            # re-register every ``DEFAULT_TTL`` seconds
+            ioloop = tornado.ioloop.IOLoop.current()
+            ioloop.call_later(
+                delay=DEFAULT_TTL,
```
junchaowu: shouldn't we read the ttl from the response and then reset the ttl

python/tchannel/tornado/hyperbahn.py
Raynos: we use json in our node client as well.

python/tchannel/tornado/hyperbahn.py
Raynos: there is no ttl in the response.

python/tchannel/tornado/hyperbahn.py
blampe: > why not use yield ArgSchemeBroker

Because two imports to do this is very awkward and feels like it's leaking an implementation detail. Something like `hyperbahn_client.request(..., argscheme="json")` would be a lot cleaner.

python/tchannel/tornado/hyperbahn.py
```diff
@@ -0,0 +1,122 @@
...
+DEFAULT_TTL = 60  # seconds
+
+
+class HyperbahnClient(object):
+    """Client for talking with the Hyperbahn."""
+
+    def __init__(
+        self,
+        service,
+        hyperbahn_routers,
```
breerly: Suggest rename to just `routers`.

python/tchannel/tornado/hyperbahn.py
```diff
@@ -0,0 +1,122 @@
...
+from __future__ import absolute_import
+
+import json
+import tornado.gen
+
+from ..messages.error import ErrorCode
+from .tchannel import TChannel
+
+
+DEFAULT_TTL = 60  # seconds
```
breerly: @abhinav do you think we should move this to the `glossary`?

python/tchannel/tornado/hyperbahn.py
breerly: In this case we don't need to use the broker because we know it's `json` ahead of time.

python/tchannel/tornado/hyperbahn.py
breerly: @blampe what about @Raynos comment here?

python/tests/tornado/test_hyperbahn.py
```diff
@@ -0,0 +1,62 @@
...
+
+import mock
+import pytest
+
+from tchannel.exceptions import ConnectionClosedException
+from tchannel.tornado.hyperbahn import HyperbahnClient
+
+
+def test_new_client_establishes_peers():
+    tchannel = mock.MagicMock()
```
breerly: nit: would rather see `doubles` being used here.

breerly:
Mostly nit's. :shipit:

python/tchannel/tornado/hyperbahn.py
junchaowu: or message common. It's part of message

abhinav: glosarry sounds preferable IMO. messages.common seems to contain all kinds of unrelated information.

python/tchannel/tornado/hyperbahn.py
junchaowu: this is fine.

python/tests/tornado/test_hyperbahn.py
```diff
@@ -0,0 +1,62 @@
...
+    routers = [mock.sentinel.foo, mock.sentinel.bar]
+
+    HyperbahnClient(
+        service='baz',
+        hyperbahn_routers=routers,
+        tchannel=tchannel,
+    )
+
+    for router in routers:
+        assert tchannel.peers.add.called_with(router)
```
abhinav: Instead of testing the behavior by asserting this call, can you instead assert that peers passed into the constructor exist? Something like,

    assert tchannel.peers.lookup('localhost:4040')

For all peers

python/tchannel/tornado/hyperbahn.py
blampe: agree

PR #540, opened 2015-05-07
refactor the exception control for endpoint

junchaowu:
@blampe @abhinav @breerly

### DIFF
--- a/python/tchannel/messages/common.py
+++ b/python/tchannel/messages/common.py
@@ -47,11 +47,6 @@ class StreamState(IntEnum):
     none = 0x03
 
 
-class StatusCode(IntEnum):
-    ok = 0x00,
-    error = 0x01
-
-
 class FlagsType(IntEnum):
     none = 0x00,
     fragment = 0x01
--- a/python/tchannel/messages/common.py
+++ b/python/tchannel/messages/common.py
@@ -47,6 +47,11 @@ class StreamState(IntEnum):
     none = 0x03
 
 
+class StatusCode(IntEnum):
+    ok = 0x00,
+    error = 0x01
+
+
 class FlagsType(IntEnum):
     none = 0x00,
     fragment = 0x01
--- a/python/tchannel/tornado/data.py
+++ b/python/tchannel/tornado/data.py
@@ -124,6 +124,11 @@ class Response(object):
     level. This is going to hide the protocol level message information.
     """
 
+    OK = 0x00
+    ERROR = 0x01
+
+    resp_status = [OK, ERROR]
+
     # TODO decide which elements inside "message" object to expose to user.
     def __init__(
             self,
@@ -138,7 +143,7 @@ class Response(object):
             scheme=None,
     ):
 
-        self.flags = flags
+        self.flags = flags or self.OK
         self.code = code
         self.tracing = tracing
         self.checksum = checksum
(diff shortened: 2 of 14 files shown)


python/tchannel/tornado/dispatch.py
```diff
@@ -92,15 +91,27 @@ def handle_call(self, request, connection):
...
+                    )
+                )
+                response.flush()
+            except Exception as e:
+                # refine the exception in the following patches
+                if response.flushed:
+                    # this is bad if user called flush before exception happens
+                    return
+
+                response.write_body(e.message)
```
abhinav: For unexpected exceptions, you want to send an internal error back, not an application error.

abhinav:
Here's what I think we should go for:

For non-Thrift arg schemes, the definition of an application error is not up to us. We can't do anything special for it. If the handler throws an exception, it should be treated as an "unexpected internal failure" and we should send an actual TChannel-level Error frame back with code 0x05 ("unexpected error"). From the user's point of view, they're expected to catch all application errors at the top-level in their request handler and decide how they want to propagate the error. The only thing we need to provide there is a way for them to say "send this body in the call response and set the error flag to 1".

For Thrift arg scheme, we actually know exactly what application errors can occur. We want to catch all top-level exceptions and if they're application errors (defined in the IDL), send back a call response with the exception information and the error flag set to 1. If they're not defined in the IDL, we want to treat them as "unexpected internal failures" and send back a TChannel-level Error frame.

abhinav:
So basically what you want to do at this level is treat all exceptions as an internal error.

Then at the layer above this where we call the Thrift handler, we'll add logic to catch all exceptions, check if they are application-level (defined in the IDL), and propagate that as a call response. Otherwise, re-raise the exception and let this layer convert it into an "internal failure" error frame. Some of this logic already exists in the Thrift server code.

breerly:
Abhinav hit it on the head. +1

python/tchannel/tornado/data.py
```diff
@@ -124,6 +124,11 @@ class Response(object):
     level. This is going to hide the protocol level message information.
     """
 
+    OK = 0x00
+    ERROR = 0x01
+
+    resp_status = [OK, ERROR]
```
abhinav: Use an Enum or use `glossary` to store constants.

python/tchannel/tornado/data.py
```diff
@@ -154,6 +159,17 @@ def __init__(
...
         self.scheme = scheme
 
+    @property
+    def status(self):
+        return self.flags
+
+    @status.setter
+    def status(self, status):
+        if status not in self.resp_status:
+            raise TChannelException("Not valid status!")
```
abhinav: "Invalid status"

python/tchannel/tornado/data.py
```diff
@@ -154,6 +159,17 @@ def __init__(
...
+
+        self.flags = status
```
abhinav: Once using an enum, map that to a flag manually. That will allow decoupling protocol level flags from response status codes.

python/tchannel/tornado/dispatch.py
```diff
@@ -92,15 +91,25 @@ def handle_call(self, request, connection):
...
+                            connection.tchannel,
+                            request.tracing,
+                        ),
+                    )
+                )
+                response.flush()
+            except Exception as e:
+                # refine the exception in the following patches
+                if response.flushed:
+                    # this is bad if user called flush before exception happens
```
abhinav: Explain? What's happening here? Why is it bad? How should that be handled?

python/tchannel/messages/common.py
```diff
@@ -47,6 +47,11 @@ class StreamState(IntEnum):
     none = 0x03
 
 
+class StatusCode(IntEnum):
+    ok = 0x00,
+    error = 0x01
```
abhinav: This is not a message-level concern. It doesn't belong in messages.common.

abhinav:
:+1:

PR #541, opened 2015-05-07
[py] Make it possibly to specify known peers upon instantiation

abhinav:
@blampe @breerly @junchaowu 

This will make the autobahn stuff simpler

### DIFF
--- a/python/tchannel/tornado/peer.py
+++ b/python/tchannel/tornado/peer.py
@@ -36,12 +36,12 @@ from .stream import InMemStream
 from .stream import Stream
 from .stream import read_full
 
+
 try:
     # included in Tornado 4.2
     from tornado.locks import Condition
 except ImportError:  # pragma: no cover
     from toro import Condition
-
 
 log = logging.getLogger('tchannel')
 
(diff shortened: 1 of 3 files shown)


python/tchannel/tornado/peer.py
```diff
@@ -26,7 +26,6 @@
 from random import random
 
 from tornado import gen
-
```
blampe: wut

abhinav: PyCharm auto format. Fixing.

python/tchannel/tornado/tchannel.py
```diff
@@ -86,6 +87,10 @@ def __init__(self, hostport=None, process_name=None):
         self.event_emitter = EventEmitter()
         self.hooks = EventRegistrar(self.event_emitter)
 
+        if known_peers:
```
blampe: If I have an existing `TChannel` do I need to this myself?

abhinav: Yes

abhinav: .peers is part of the TChannel interface. Accessing it and adding/getting peers is completely okay.

abhinav:
Added a quick test

blampe:
lgtm

PR #542, opened 2015-05-07
max length limit for arg1

ShanniLi:
r @jcorbin @kriskowal  

Two checks:
1. TChannelV2Handler.prototype.handleCallRequest
2. V2OutRequest.prototype._sendCallRequest

### DIFF
--- a/node/errors.js
+++ b/node/errors.js
@@ -37,6 +37,12 @@ module.exports.ArgChunkOutOfOrderError = TypedError({
     got: null
 });
 
+module.exports.Arg1OverLengthLimit = TypedError({
+    type: 'tchannel.arg1-over-length-limit',
+    message: 'arg 1 length {length} is larger than the limit 0x4000',
+    length: null
+});
+
 module.exports.BodyParserError = WrappedError({
     type: 'tchannel-handler.parse-error.body-failed',
     message: 'Could not parse body (arg3) argument.\n' +
--- a/node/errors.js
+++ b/node/errors.js
@@ -39,8 +39,9 @@ module.exports.ArgChunkOutOfOrderError = TypedError({
 
 module.exports.Arg1OverLengthLimit = TypedError({
     type: 'tchannel.arg1-over-length-limit',
-    message: 'arg 1 length {length} is larger than the limit 0x4000',
-    length: null
+    message: 'arg 1 length {length} is larger than the limit {limit}',
+    length: null,
+    limit: null
 });
 
 module.exports.BodyParserError = WrappedError({
--- a/node/v2/frame.js
+++ b/node/v2/frame.js
@@ -44,7 +44,6 @@ Frame.MaxSize = 0xffff;
 Frame.MaxBodySize = Frame.MaxSize - Frame.Overhead;
 Frame.MaxId = 0xfffffffe;
 Frame.NullId = 0xffffffff;
-Frame.MaxArg1Size = 0x4000;
 
 // size:2: type:1 reserved:1 id:4 reserved:8 ...
 Frame.RW = bufrw.Base(frameLength, readFrameFrom, writeFrameInto);
--- a/node/v2/frame.js
+++ b/node/v2/frame.js
@@ -44,6 +44,7 @@ Frame.MaxSize = 0xffff;
 Frame.MaxBodySize = Frame.MaxSize - Frame.Overhead;
 Frame.MaxId = 0xfffffffe;
 Frame.NullId = 0xffffffff;
+Frame.MaxArg1Size = 0x4000;
 
 // size:2: type:1 reserved:1 id:4 reserved:8 ...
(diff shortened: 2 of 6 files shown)


Raynos:
Consider limiting arg1 in call responses as well for symmetry.

node/errors.js
```diff
@@ -37,6 +37,12 @@ module.exports.ArgChunkOutOfOrderError = TypedError({
     got: null
 });
 
+module.exports.Arg1OverLengthLimit = TypedError({
+    type: 'tchannel.arg1-over-length-limit',
+    message: 'arg 1 length {length} is larger than the limit 0x4000',
```
jcorbin: please don't hardcode the limit, pass it in as another field

node/v2/frame.js
```diff
@@ -44,6 +44,7 @@ Frame.MaxSize = 0xffff;
 Frame.MaxBodySize = Frame.MaxSize - Frame.Overhead;
 Frame.MaxId = 0xfffffffe;
 Frame.NullId = 0xffffffff;
+Frame.MaxArg1Size = 0x4000;
```
jcorbin: not on frame, on CallRequest / CallResponse

node/v2/handler.js
```diff
@@ -181,6 +181,13 @@ TChannelV2Handler.prototype.handleCallRequest = function handleCallRequest(reqFr
         return callback(new Error('call request before init request')); // TODO typed error
     }
     var req = self.buildInRequest(reqFrame);
+    if (reqFrame.body.args && reqFrame.body.args[0] &&
+        reqFrame.body.args[0].length > v2.Frame.MaxArg1Size) {
+        req.res = self.buildOutResponse(req);
+        self.sendErrorFrame(req.res, 'BadRequest',
+            'arg1 exceeds the max size of 0x4000');
```
jcorbin: hardcode, use the value

Raynos:
Should we send back a protocol error error frame when we get an INCOMING call response with arg1 that is too big.

The only thing we can do is send back a `fatal protocol error`; it might be nice to do this so we dont have to buffer arg1 anywhere in the node implementation.

cc: @jcorbin

node/v2/out_request.js
```diff
@@ -49,6 +51,12 @@ function _sendCallRequest(args, isLast) {
     var self = this;
     var flags = 0;
     if (!isLast) flags |= CallFlags.Fragment;
+    if (args && args[0] && args[0].length > v2.CallRequest.MaxArg1Size) {
+        self.errorEvent.emit(self, errors.Arg1OverLengthLimit({
+                length: '0x' + args[0].length.toString(16),
+                limit: '0x' + v2.CallRequest.MaxArg1Size.toString(16)
+        }));
```
Raynos: missing `return;`

node/v2/out_response.js
```diff
@@ -48,6 +50,12 @@ V2StreamingOutResponse.prototype._sendCallResponse =
 function _sendCallResponse(args, isLast) {
     var self = this;
     var flags = 0;
+    if (args && args[0] && args[0].length > v2.CallResponse.MaxArg1Size) {
+        self.errorEvent.emit(self, errors.Arg1OverLengthLimit({
+                length: '0x' + args[0].length.toString(16),
+                limit: '0x' + v2.CallResponse.MaxArg1Size.toString(16)
+        }));
```
Raynos: `return;`

node/v2/out_request.js
ShanniLi: fixed.

node/v2/out_response.js
ShanniLi: fixed.

node/v2/handler.js
```diff
@@ -203,6 +211,13 @@ TChannelV2Handler.prototype.handleCallResponse = function handleCallResponse(res
         return callback(new Error('call response before init response')); // TODO typed error
     }
     var res = self.buildInResponse(resFrame);
+    if (resFrame.body.args && resFrame.body.args[0] &&
+        resFrame.body.args[0].length > v2.CallResponse.MaxArg1Size) {
+        return callback(errors.Arg1OverLengthLimit({
+                length: '0x' + resFrame.body.args[0].length.toString(16),
+                limit: '0x' + v2.CallResponse.MaxArg1Size.toString(16)
```
kriskowal: please remove extra indent

kriskowal: @jcorbin mentioned that calling this callback with an error constitutes a protocol error which would cause this connection to be torn down. This seems appropriate to me since the stream is not longer in a consistent state.

On a side note, if we go with UDP in v2+, connection scoped identifiers would become irrelevant, protocol errors would not break the connection, and we would need to use tracing to identify a req/res pair.

jcorbin:
This looks good enough to land, however it made me realize #550/#551

PR #543, opened 2015-05-08
add protocol error object

junchaowu:
@blampe @breerly @abhinav

### DIFF
--- a/python/tchannel/event.py
+++ b/python/tchannel/event.py
@@ -32,19 +32,8 @@ log = logging.getLogger('tchannel')
 
 @unique
 class EventType(IntEnum):
-    """Types to represent system events
+    """Types to represent system events"""
 
-    Events:
-
-    send_request: before client sends request
-
-    send_response: after server sends response
-
-    receive_request: after server receive request
-
-    receive_response: after client receive response
-
-    """
     before_send_request = 0x00
     after_send_request = 0x01
 
@@ -56,6 +45,9 @@ class EventType(IntEnum):
 
     before_receive_response = 0x30
     after_receive_response = 0x31
+
+    after_receive_protocol_error = 0x40
+    after_send_protocol_error = 0x41
 
 
 class EventHook(object):
@@ -101,6 +93,14 @@ class EventHook(object):
 
     def after_receive_response(self, response):
         """Called after a ``CALL_RESP`` message is read."""
+        pass
+
+    def after_receive_protocol_error(self, error):
+        """Called after a ''error'' message is read."""
+        pass
+
+    def after_send_protocol_error(self, error):
+        """Called after a ''error'' message is sent."""
         pass
 
 
--- a/python/tchannel/event.py
+++ b/python/tchannel/event.py
@@ -46,8 +46,8 @@ class EventType(IntEnum):
     before_receive_response = 0x30
     after_receive_response = 0x31
 
-    after_receive_protocol_error = 0x40
-    after_send_protocol_error = 0x41
+    after_receive_error = 0x40
+    after_send_error = 0x41
 
 
 class EventHook(object):
@@ -95,11 +95,11 @@ class EventHook(object):
         """Called after a ``CALL_RESP`` message is read."""
         pass
 
-    def after_receive_protocol_error(self, error):
+    def after_receive_error(self, err):
         """Called after a ``error`` message is read."""
         pass
 
-    def after_send_protocol_error(self, error):
+    def after_send_error(self, err):
         """Called after a ``error`` message is sent."""
         pass
 
--- a/python/tchannel/event.py
+++ b/python/tchannel/event.py
@@ -96,11 +96,11 @@ class EventHook(object):
         pass
 
     def after_receive_protocol_error(self, error):
-        """Called after a ''error'' message is read."""
+        """Called after a ``error`` message is read."""
         pass
 
     def after_send_protocol_error(self, error):
-        """Called after a ''error'' message is sent."""
+        """Called after a ``error`` message is sent."""
         pass
 
 
--- a/python/tchannel/tornado/connection.py
+++ b/python/tchannel/tornado/connection.py
@@ -42,7 +42,6 @@ from ..messages.common import PROTOCOL_VERSION
 from ..messages.common import FlagsType
 from ..messages.error import ErrorMessage
 from ..messages.types import Types
-from .data import ProtocolError
 from .message_factory import MessageFactory
 
 try:
@@ -211,12 +210,15 @@ class TornadoConnection(object):
                 continue
 
             elif context.message_id in self._outstanding:
+
                 response = self.response_message_factory.build(
                     context.message_id, context.message
                 )
 
                 # keep continue message in the list
                 # pop all other type messages including error message
+
+                result = {"type": context.message.message_type}
                 if (context.message.message_type in self.CALL_RES_TYPES and
                         context.message.flags == FlagsType.fragment):
                     # still streaming, keep it for record
@@ -224,18 +226,20 @@ class TornadoConnection(object):
                 else:
                     future = self._outstanding.pop(context.message_id)
 
-                    if context.message.message_type == Types.ERROR:
-                        protocol_error = (
-                            self.response_message_factory.build_protocol_error(
-                                context.message,
-                                context.message_id,
-                            ))
-
-                        future.set_result(protocol_error)
-                        continue
+                if context.message.message_type == Types.ERROR:
+                    protocol_error = (
+                        self.response_message_factory.build_protocol_error(
+                            context.message,
+                            context.message_id,
+                        )
+                    )
+                    result['value'] = protocol_error
+                    future.set_result(result)
+                    continue
 
                 if response and future.running():
-                    future.set_result(response)
+                    result['value'] = response
+                    future.set_result(result)
                 continue
 
             log.warn('Unconsumed message %s', context)
@@ -456,13 +460,13 @@ class TornadoConnection(object):
                 # TODO Send error frame back
                 logging.exception("Failed to process %s", repr(context))
 
-    def send_error(self, code, message, message_id):
+    def send_error(self, code, description, message_id):
         """Convenience method for writing Error frames up the wire.
 
         :param code:
             Error code
-        :param message:
-            Error message
+        :param description:
+            Error description
         :param message_id:
             Message in response to which this error is being sent
         """
@@ -472,7 +476,7 @@ class TornadoConnection(object):
         return self._write(
             ErrorMessage(
                 code=code,
-                message=message
+                description=description
             ),
             message_id
         )
@@ -552,11 +556,10 @@ class StreamConnection(TornadoConnection):
             yield self._stream(response, self.response_message_factory)
 
             # event: send_response
-            if self.tchannel:
-                self.tchannel.event_emitter.fire(
-                    EventType.after_send_response,
-                    response,
-                )
+            self.tchannel.event_emitter.fire(
+                EventType.after_send_response,
+                response,
+            )
         finally:
             response.close_argstreams(force=True)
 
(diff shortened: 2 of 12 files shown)


python/tchannel/event.py
```diff
@@ -103,6 +95,14 @@ def after_receive_response(self, response):
         """Called after a ``CALL_RESP`` message is read."""
         pass
 
+    def after_receive_protocol_error(self, error):
+        """Called after a ''error'' message is read."""
```
blampe: backticks

python/tchannel/tornado/connection.py
```diff
@@ -222,6 +223,17 @@ def _loop(self):
...
                     future = self._outstanding.pop(context.message_id)
+
+                    if context.message.message_type == Types.ERROR:
+                        protocol_error = (
+                            self.response_message_factory.build_protocol_error(
+                                context.message,
+                                context.message_id,
+                            )
+                        )
+                        future.set_result(protocol_error)
```
abhinav: Why not `set_exception` here?

python/tchannel/tornado/connection.py
```diff
@@ -222,6 +223,17 @@ def _loop(self):
                     future = self._outstanding.get(context.message_id)
                 else:
                     future = self._outstanding.pop(context.message_id)
+
+                    if context.message.message_type == Types.ERROR:
+                        protocol_error = (
+                            self.response_message_factory.build_protocol_error(
+                                context.message,
+                                context.message_id,
```
blampe: Unrelated, what's to stop us from putting `message_id` on the `message`? That seems to be the primary use for `context` right now.

python/tchannel/tornado/connection.py
```diff
@@ -593,19 +605,39 @@ def send_request(self, request):
...
             # fetch the request tracing for response
             f.result().tracing = request.tracing
-            response_future.set_result(f.result())
-            # event: receive_response
-            if self.tchannel:
-                self.tchannel.event_emitter.fire(
-                    EventType.after_receive_response,
-                    f.result(),
+
+            if isinstance(f.result(), ProtocolError):
```
abhinav: See if you can refactor this to avoid the `isinstance`

python/tchannel/tornado/connection.py
```diff
@@ -222,6 +223,17 @@ def _loop(self):
...
+
+                    if context.message.message_type == Types.ERROR:
+                        protocol_error = (
+                            self.response_message_factory.build_protocol_error(
+                                context.message,
+                                context.message_id,
+                            )
+                        )
+                        future.set_result(protocol_error)
+                        continue
```
blampe: Why keep going?

python/tchannel/tornado/connection.py
junchaowu: because it needs to get corresponding request's tracing information

python/tchannel/tornado/data.py
```diff
@@ -316,3 +316,22 @@ def close_argstreams(self, force=False):
         for stream in self.argstreams:
             if stream.auto_close or force:
                 stream.close()
+
+
+class ProtocolError(object):
+    """Object to represent protocol error message
+
+    Note: Internal use only.
```
abhinav: If this is only used in connection.py and is for internal use only there, declare it in that file.

python/tchannel/tornado/connection.py
junchaowu: Agreed. Context is the legacy code.

python/tchannel/tornado/connection.py
```diff
@@ -593,19 +605,39 @@ def send_request(self, request):
         # TODO: fire before_receive_response
 
         def adapt_tracing(f):
```
blampe: This method is doing way more than tracing, and it's complicated enough that it shouldn't be defined inline.

python/tchannel/tornado/connection.py
blampe: If we need to use `isinstance` that's a sign that we need a better interface.

python/tchannel/tornado/message_factory.py
```diff
@@ -56,6 +58,30 @@ def __init__(self, remote_host=None, remote_host_port=None):
         self.in_checksum = {}
         self.out_checksum = {}
 
+    def build_raw_error_message(self, err):
+        """build protocol level error message based on Error object"""
+        message = ErrorMessage(
+            code=err.code,
+            tracing=err.Tracing(err.tracing.span_id,
```
abhinav: Did you mean Tracing? What's err.Tracing?

python/tchannel/tornado/connection.py
```diff
@@ -593,19 +605,39 @@ def send_request(self, request):
...
-                    EventType.after_receive_response,
-                    f.result(),
+
+            if isinstance(f.result(), ProtocolError):
+                protocol_error = f.result()
+                response_future.set_exception(
+                    TChannelException(protocol_error.message)
                 )
+                # event: after_receive_protocol_error
+                if self.tchannel:
```
blampe: We should kill all of these `if self.tchannel:` checks if they're only for some legacy test code -- we can just update the test code.

python/tchannel/tornado/data.py
blampe: What does this mean?

junchaowu: No. This task is in progress. There are other places will use this object

python/tchannel/tornado/data.py
```diff
@@ -316,3 +316,22 @@ def close_argstreams(self, force=False):
         for stream in self.argstreams:
             if stream.auto_close or force:
                 stream.close()
+
```
blampe: Can we think of a better name for this file than `data`?

python/tchannel/tornado/data.py
```diff
@@ -316,3 +316,22 @@ def close_argstreams(self, force=False):
...
+class ProtocolError(object):
+    """Object to represent protocol error message
+
+    Note: Internal use only.
+    """
+
+    def __init__(
+        self,
+        code,
+        message,
```
blampe: Is this `message` an instance or a string?

python/tchannel/tornado/message_factory.py
```diff
@@ -56,6 +58,30 @@ def __init__(self, remote_host=None, remote_host_port=None):
...
+    def build_protocol_error(self, message, message_id=None):
+        """build protocol level error message based on Error object"""
+
+        error = ProtocolError(
+            code=message.code,
+            message=message.message,
+            id=message_id,
+        )
+
+        return error
```
abhinav: Why is this a method in this class at all? The functionality is needed in connection.py only and doesn't depend on anything in instances of this class.

python/tchannel/tornado/message_factory.py
```diff
@@ -56,6 +58,30 @@ def __init__(self, remote_host=None, remote_host_port=None):
         self.in_checksum = {}
         self.out_checksum = {}
 
+    def build_raw_error_message(self, err):
```
blampe: What is `err`? An `ErrorMessage`? A `ProtocolError `? We have a lot of name collisions happening.

python/tchannel/tornado/message_factory.py
```diff
@@ -56,6 +58,30 @@ def __init__(self, remote_host=None, remote_host_port=None):
...
+        """build protocol level error message based on Error object"""
+        message = ErrorMessage(
+            code=err.code,
+            tracing=err.Tracing(err.tracing.span_id,
+                                err.tracing.parent_span_id,
+                                err.tracing.trace_id,
+                                err.tracing.traceflags),
+            message=err.message,
+        )
+
```
abhinav: Same as below. Why is this defined in this class?

python/tchannel/zipkin/zipkin_trace.py
```diff
@@ -39,34 +39,50 @@ def __init__(self, tchannel=None, dst=None):
             # to dst. By default it writes to stdout
             self.tracer = DebugTracer(dst)
 
-    def before_send_request(self, context):
-        if not context.tracing.traceflags:
+    def before_send_request(self, request):
+        if not request.tracing.traceflags:
```
blampe: Thanks :)

python/tchannel/tornado/message_factory.py
junchaowu: error message also has tracing span defined in protocol

python/tchannel/tornado/connection.py
junchaowu: I do that

python/tchannel/tornado/connection.py
junchaowu: This is because we will mutate the future object for tracing staff after we receive it but we can't guarantee the execution order of this add_done_callback, and in order to send same future object to  whoever yield this future, we create two futures.

python/tchannel/tornado/message_factory.py
```diff
@@ -56,6 +58,31 @@ def __init__(self, remote_host=None, remote_host_port=None):
         self.in_checksum = {}
         self.out_checksum = {}
 
+    def build_raw_error_message(self, protocol_error):
+        """build protocol level error message based on Error object"""
+        message = ErrorMessage(
+            code=protocol_error.code,
+            tracing=protocol_error.Tracing(
+                protocol_error.tracing.span_id,
```
abhinav: Again, did you mean 'Tracing'? What's 'protocol_error.Tracing'?

python/tchannel/tornado/message_factory.py
```diff
@@ -56,6 +58,31 @@ def __init__(self, remote_host=None, remote_host_port=None):
...
+                protocol_error.tracing.parent_span_id,
+                protocol_error.tracing.trace_id,
+                protocol_error.tracing.traceflags),
+            description=protocol_error.description,
+        )
+
+        return message
+
+    def build_protocol_error(self, message, message_id=None):
```
abhinav: As mentioned before: I don't think these methods belong in this class. This class context is providing no value to the instantiation of these objects.

python/tchannel/tornado/data.py
```diff
@@ -316,3 +316,19 @@ def close_argstreams(self, force=False):
...
+        self,
+        code,
+        description,
+        id=None,
+        tracing=None,
+    ):
+        self.code = code
+        self.tracing = tracing
+        self.id = id
+        self.description = description
```
abhinav: You need to call the super constructor with the message too

python/tchannel/tornado/data.py
```diff
@@ -316,3 +316,19 @@ def close_argstreams(self, force=False):
         for stream in self.argstreams:
             if stream.auto_close or force:
                 stream.close()
+
+
+class ProtocolError(Exception):
+    """Object to represent protocol error message"""
+
```
abhinav: I would add `__slots__` to this

python/tchannel/tornado/data.py
```diff
@@ -316,3 +316,19 @@ def close_argstreams(self, force=False):
         for stream in self.argstreams:
             if stream.auto_close or force:
                 stream.close()
+
+
+class ProtocolError(Exception):
+    """Object to represent protocol error message"""
```
abhinav: This should inherit TChannelException

python/tchannel/tornado/connection.py
```diff
@@ -592,20 +601,30 @@ def send_request(self, request):
...
         )
-
         return response_future
+
+    def adapt_result(self, f, request, response_future):
+        if f.exception():
+            protocol_error = f.exception()
+            protocol_error.tracing = request.tracing
+            response_future.set_exception(
+                TChannelException(protocol_error.description)
```
abhinav: The user loses information here. Just give them the protocol_error.

junchaowu: yes, as we discussed

python/tchannel/tornado/message_factory.py
```diff
@@ -34,13 +34,42 @@
...
 from ..zipkin.annotation import Endpoint
 from ..zipkin.trace import Trace
+from .data import ProtocolError
 from .data import Request
 from .data import Response
 from .stream import InMemStream
 
 
+def build_raw_error_message(protocol_error):
+    """build protocol level error message based on Error object"""
```
abhinav: Docstrings are full sentences. Start with a capital letter. End with a period.

abhinav:
:+1:

PR #544, opened 2015-05-08
TChannel: factor out #requestOptions

jcorbin:
r @Raynos


Raynos:
lgtm.

kriskowal:
Great change. I notice that this isolates the for/in doptimization.

PR #545, opened 2015-05-08
Improve piplineStreams

jcorbin:
- add debug feature
- add optional callback

r @kriskowal


Raynos:
lgtm.

Should the debug feature be at a lower level then pipeline streams ? Maybe in argstream itself either for the reading or writing side?

kriskowal:
:ok:

PR #549, opened 2015-05-08
add server error handling

junchaowu:
@blampe @abhinav @breerly 
Still a dependency PR.


python/tchannel/handler.py
```diff
@@ -19,7 +19,7 @@
 # THE SOFTWARE.
 from __future__ import absolute_import
 
-from .exceptions import InvalidChecksumException
+from .exceptions import InvalidChecksumException, StreamingException
```
blampe: One per line.

python/tchannel/handler.py
```diff
@@ -97,12 +97,18 @@ def handle_pre_call(self, message_id, message, connection):
...
         try:
             req = connection.request_message_factory.build(message_id, message)
-        except InvalidChecksumException as e:
+        except (InvalidChecksumException, StreamingException) as e:
             connection.send_error(
                 ErrorCode.bad_request,
                 e.message,
                 message_id,
             )
+        except Exception:
```
blampe: Is this just defensive? Is there a higher level where we could do this?

python/tchannel/tornado/connection.py
```diff
@@ -538,21 +539,29 @@ def _stream(self, context, message_factory):
...
-            chunk = yield argstream.read()
-            args.append(chunk)
-            chunk = yield argstream.read()
-            while chunk:
-                message = (message_factory.
-                           build_raw_message(context, args))
-                yield self.write(message, context.id)
-                args = [chunk]
+        try:
+            for i, argstream in enumerate(context.argstreams):
```
blampe: I don't think you use `i`?

python/tchannel/tornado/data.py
```diff
@@ -172,6 +177,13 @@ def status_code(self, status):
 
         self.flags = status.value
 
+    @property
+    def ok(self):
+        if self.flags == StatusCode.ok.value:
```
blampe: `return self.flags == StatusCode.ok.value`

When I think of flags I usually think of a bitmask, so it's strange that you're using equality here. Don't we also have a `code` attribute?

python/tchannel/tornado/stream.py
```diff
@@ -161,8 +161,13 @@ def __init__(self, rpipe, wpipe=None, auto_close=False):
         self.auto_close = auto_close
         self.state = StreamState.init
 
+        self.exception = None
+
     @tornado.gen.coroutine
     def read(self):
+        if self.exception:
+            raise self.exception
```
blampe: What does this do to the stack trace?

python/tchannel/tornado/data.py
junchaowu: flags is the field defined in protocol for request and response message

python/tchannel/tornado/stream.py
junchaowu: for async exception, what the stack trace we want to keep. We need discuss more about it

python/tchannel/tornado/data.py
blampe: Right, so it seems like this isn't right -- `flags` controls fragmentation, what we want is `code`: https://github.com/uber/tchannel/blob/master/docs/protocol.md#code1

PR #550, opened 2015-05-08
TODO: mishandling of streaming responses (and especially their erros) between handler and connnection

jcorbin:
This bug will be especially insidious if we ever have a streamed call req/res that lasts for longer than it takes to cycle frame id space (~4 billion calls on a connection).  However (un)likely this is for any single service, such a probability is compounded for a forwarding entity.

I discovered this bug while reviewing #542 while considering the response arg1 too large error handling path.

cc @ShanniLi @kriskowal


Raynos:
lgtm.

Raynos:
lgtm; please ship.

jcorbin:
oh, you want me to land the TODO comments? I suppose..

PR #552, opened 2015-05-08
Make {req,res}.send/.sendStreams consistent

jcorbin:
- return `self` for chaining
- take callbacks

r @kriskowal


kriskowal:
:ok:

PR #553, opened 2015-05-08
Decouple health states from peer

kriskowal:
This is preparatory work for circuit breaking. @ShanniLi is going to be doing overlapping work on init hand shake, so getting this out now to avoid churn.


kriskowal:
r @jcorbin @Raynos

node/peer.js
```diff
@@ -45,11 +45,22 @@ function TChannelPeer(channel, hostPort, options) {
...
     self.connections = [];
     self.random = self.channel.random;
+
+    self.stateOptions = {
+        timers: self.channel.timers,
+        random: self.channel.random,
+        period: self.options.period,
+        maxErrorRate: self.options.maxErrorRate,
+        probation: self.options.probation,
+        stateMachine: self,
```
Raynos: Why call the peer the state machine and the next handler ?

Why not just call it the peer.

Raynos:
Lgtm.

node/peer_states.js
```diff
@@ -170,31 +149,30 @@ TChannelPeerUnhealthyState.prototype.shouldRequest = function shouldRequest(req,
...
+UnhealthyState.prototype.onRequestError = function onRequestError(err) {
     var self = this;
     var codeString = errors.classify(err);
     if (symptoms[codeString]) {
         self.successCount = 0;
     }
 };
+
+UnhealthyState.prototype.close = function close(callback) {
+    callback();
```
jcorbin: pass explicit null err please

jcorbin:
lgtm 🏮

node/peer.js
kriskowal: Intent is to reuse for circuit. In the circuit case, the stateMachine is the circuit, and the nextHandler.shouldRequest just returns 1.

kriskowal:
I’ve slipped two more commits in, one to rename peer_states to states. The other factors the state machine setState method into a separate module, so I can use it in Circuit.

PR #560, opened 2015-05-08
Error on incoming call req/res frame before init res frame is sent

ShanniLi:
r @jcorbin @Raynos 

1. handle the issue by moving self.sendInitResponse up. 
2. did a little refinement adding the typed error: TChannelInitProtocolError

### DIFF
--- a/node/request.js
+++ b/node/request.js
@@ -303,7 +303,7 @@ TChannelRequest.prototype.onIdentified = function onIdentified(peer, opts, perAt
             'retry-count': self.outReqs.length - 1
         });
     }
-}
+};
 
 TChannelRequest.prototype.onSubreqError = function onSubreqError(err) {
     var self = this;
--- a/node/request.js
+++ b/node/request.js
@@ -70,9 +70,6 @@ function TChannelRequest(channel, options) {
 
     self.err = null;
     self.res = null;
-
-    self.errorEvent.on(self.onError);
-    self.responseEvent.on(self.onResponse);
 }
 
 inherits(TChannelRequest, EventEmitter);
@@ -83,17 +80,20 @@ TChannelRequest.defaultTimeout = 5000;
 
 TChannelRequest.prototype.type = 'tchannel.request';
 
-TChannelRequest.prototype.onError = function onError(err, self) {
+TChannelRequest.prototype.emitError = function emitError(err) {
+    var self = this;
     if (!self.end) self.end = self.timers.now();
     self.err = err;
     self.services.onRequestError(self);
-
+    self.errorEvent.emit(self, err);
 };
 
-TChannelRequest.prototype.onResponse = function onResponse(res, self) {
+TChannelRequest.prototype.emitResponse = function emitResponse(res) {
+    var self = this;
     if (!self.end) self.end = self.timers.now();
     self.res = res;
     self.services.onRequestResponse(self);
+    self.responseEvent.emit(self, res);
 };
 
 TChannelRequest.prototype.hookupStreamCallback = function hookupCallback(callback) {
@@ -213,12 +213,12 @@ TChannelRequest.prototype.resend = function resend() {
         if (self.outReqs.length) {
             var lastReq = self.outReqs[self.outReqs.length - 1];
             if (lastReq.err) {
-                self.errorEvent.emit(self, lastReq.err);
+                self.emitError(lastReq.err);
             } else {
-                self.responseEvent.emit(self, lastReq.res);
+                self.emitResponse(lastReq.res);
             }
         } else {
-            self.errorEvent.emit(self, errors.NoPeerAvailable());
+            self.emitError(errors.NoPeerAvailable());
         }
         return;
     }
@@ -313,7 +313,7 @@ TChannelRequest.prototype.onSubreqError = function onSubreqError(err) {
     if (self.shouldRetryError(err)) {
         self.deferResend();
     } else {
-        self.errorEvent.emit(self, err);
+        self.emitError(err);
     }
 };
 
@@ -321,18 +321,18 @@ TChannelRequest.prototype.onSubreqResponse = function onSubreqResponse(res) {
     var self = this;
     if (self.checkTimeout(null, res)) return;
     if (res.ok) {
-        self.responseEvent.emit(self, res);
+        self.emitResponse(res);
     } else if (self.options.shouldApplicationRetry) {
         self.maybeAppRetry(res);
     } else {
-        self.responseEvent.emit(self, res);
+        self.emitResponse(res);
     }
 };
 
 TChannelRequest.prototype.deferResend = function deferResend() {
     var self = this;
     if (--self.resendSanity <= 0) {
-        self.errorEvent.emit(self, new Error('TChannelRequest out of resend sanity'));
+        self.emitError(new Error('TChannelRequest out of resend sanity'));
     } else {
         process.nextTick(doResend);
     }
@@ -345,7 +345,7 @@ TChannelRequest.prototype.checkPending = function checkPending() {
     var self = this;
     var err = self.services.errorIfExceedsMaxPending(self);
     if (err) {
-        self.errorEvent.emit(self, err);
+        self.emitError(err);
         return true;
     }
     return false;
@@ -358,14 +358,14 @@ TChannelRequest.prototype.checkTimeout = function checkTimeout(err, res) {
     if (self.elapsed < self.timeout) return false;
     if (err) {
         if (!self.err) {
-            self.errorEvent.emit(self, err);
+            self.emitError(err);
         }
     } else if (res) {
         if (!self.err && !self.res) {
-            self.responseEvent.emit(self, res);
+            self.emitResponse(res);
         }
     } else if (!self.err) {
-        self.errorEvent.emit(self, errors.TimeoutError({
+        self.emitError(errors.TimeoutError({
             start: self.start,
             elapsed: self.elapsed,
             timeout: self.timeout
@@ -427,9 +427,9 @@ TChannelRequest.prototype.maybeAppRetry = function maybeAppRetry(res) {
 
     function done(err) {
         if (err) {
-            self.errorEvent.emit(self, err);
+            self.emitError(err);
         } else {
-            self.responseEvent.emit(self, res);
+            self.emitResponse(res);
         }
     }
 };
--- a/node/request.js
+++ b/node/request.js
@@ -232,46 +232,51 @@ TChannelRequest.prototype.resend = function resend() {
     self.services.onRequest(self);
     var conn = peer.connect();
     if (!peer.isConnected() && conn.handler) {
-        var timer = self.timers.setTimeout(function connectTimeOut() {
-            self.checkTimeout();
-        },
-        opts.timeout);
-        if (timer.unref) timer.unref();
-        conn.on('identified', function retry() {
-            self.timers.clearTimeout(timer);
-            onConnected();
-        });
+        conn.on('identified', onIdentified);
+        var timer = self.timers.setTimeout(onIdentifyTimeout, opts.timeout);
         return;
     } else {
-        onConnected();
+        onIdentified();
     }
 
-    function onConnected() {
-        var outReq = peer.request(opts);
-        self.outReqs.push(outReq);
+    function onIdentifyTimeout() {
+        self.checkTimeout();
+    }
 
-        if (self.outReqs.length === 1) {
-            self.channel.outboundCallsSentStat.increment(1, {
-                'target-service': outReq.serviceName,
-                'service': outReq.headers.cn,
-                // TODO should always be buffer
-                'target-endpoint': String(self.arg1)
-            });
-        } else {
-            self.channel.outboundCallsRetriesStat.increment(1, {
-                'target-service': outReq.serviceName,
-                'service': outReq.headers.cn,
-                // TODO should always be buffer
-                'target-endpoint': String(self.arg1),
-                'retry-count': self.outReqs.length - 1
-            });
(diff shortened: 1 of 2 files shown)


Raynos:
Ideally we should have one typed error per case rather then a single one with a reason string.

Raynos:
lgtm. We still need to have erroring if we get incoming call reqs before the init response.

jcorbin:
This breaks tests and can't land first, but good progress

node/v2/handler.js
```diff
@@ -143,7 +143,8 @@ TChannelV2Handler.prototype.handleFrame = function handleFrame(frame, callback)
 TChannelV2Handler.prototype.handleInitRequest = function handleInitRequest(reqFrame, callback) {
     var self = this;
     if (self.remoteHostPort !== null) {
-        return callback(new Error('duplicate init request')); // TODO typed error
+        return callback(errors.TChannelInitProtocolError(
```
jcorbin: awesome, glad you implemented these TODOs in passing.  If you wanted to, this error conversion could separate out first since it won't break tests.

node/channel.js
```diff
@@ -370,7 +370,8 @@ TChannel.prototype.request = function channelRequest(options) {
     }
 
     var req = null;
-    if (opts.host || // retries are only between hosts
```
Raynos: why? don't do this.

node/request.js
```diff
@@ -174,16 +176,34 @@ TChannelRequest.prototype.resend = function resend() {
         opts[key] = self.options[key];
     }
     opts.timeout = self.timeout - self.elapsed;
-    var outReq = peer.request(opts);
+    onInitResponse();
```
Raynos: why have a func if its called directly.

Raynos: I see.

node/request.js
```diff
@@ -174,16 +176,34 @@ TChannelRequest.prototype.resend = function resend() {
...
+    function onInitResponse() {
+        if (self.checkTimeout() ||
```
Raynos: please dont break lines so agressively.

node/request.js
```diff
@@ -174,16 +176,34 @@ TChannelRequest.prototype.resend = function resend() {
...
+            self.channel.destroyed) {
+            return;
+        }
+
+        self.services.onRequest(self);
```
Raynos: what does this line do?

node/request.js
ShanniLi: fixed

node/channel.js
ShanniLi: Since self.peers.request doesn't have resend for call req/res. 

@jcorbin, should I add a blocking and retry logic in v2/out_request.js?

node/request.js
```diff
@@ -174,16 +176,34 @@ TChannelRequest.prototype.resend = function resend() {
...
+        var conn = peer.connect();
+        if (!peer.isConnected('out', true) &&
+            conn.handler) {
+            require('timers').setImmediate(onInitResponse);
```
Raynos: you can't do this; this is going to be a crazy async loop.

node/request.js
Raynos: I understand now.

node/request.js
```diff
@@ -272,10 +292,15 @@ TChannelRequest.prototype.shouldRetryError = function shouldRetryError(err) {
         return false;
     }
 
+    if (self.options.host) {
```
Raynos: let's not make this change unless needed.

node/streaming_out_request.js
```diff
@@ -45,7 +47,17 @@ function StreamingOutRequest(id, options) {
...
-        self.sendParts(parts, isLast);
+
+        onInitResponse();
+        function onInitResponse() {
+            if (self.handler.remoteHostPort !== null) {
+                self.sendParts(parts, isLast);
+            } else if (self.checkTimeout()) {
+                return;
+            } else {
+                require('timers').setImmediate(onInitResponse);
```
Raynos: you cannot loop like this; let's just making streaming be thrown exceptions when init handshake is not finished.

Raynos:
These PR has gotten too big; let's break it out into multiple ones.

Easy to land:
 
 - https://github.com/uber/tchannel/commit/984705abdff9292430bc647938cf15dba95aaf28
 - https://github.com/uber/tchannel/commit/4b63efa69c3d3042e321adbaa893295870a8afc3
 
The rest is harder to land.

node/channel.js
Raynos: if you specify `host` and its not ready just throw.

jcorbin:
agreed with @Raynos , declining to review until fragmented / the rest of feedback is addressed.

Raynos:
Let's not implement delay/buffering/retry for streaming on init req/res.

Raynos:
Let's re-implement retries to be waiting on `peer.onInitialized`

Raynos:
We have to refactor tests to not be `request({ host: host })` but be `request({ serviceName: service })` so that we don't change the `opts.host` semantics in `channel.js`

Raynos:
We should update the streaming tests to wait on initialized manually.

node/request.js
```diff
@@ -226,32 +228,50 @@ TChannelRequest.prototype.resend = function resend() {
...
-            'target-service': outReq.serviceName,
-            'service': outReq.headers.cn,
-            // TODO should always be buffer
-            'target-endpoint': String(self.arg1)
+    self.services.onRequest(self);
+    var conn = peer.connect();
+    if (!peer.isConnected() && conn.handler) {
+        var timer = self.timers.setTimeout(function connectTimeOut() {
+            self.checkTimeout();
+        },
```
Raynos: please dont break line here.

node/request.js
```diff
@@ -226,32 +228,50 @@ TChannelRequest.prototype.resend = function resend() {
...
-    if (self.outReqs.length === 1) {
-        self.channel.outboundCallsSentStat.increment(1, {
-            'target-service': outReq.serviceName,
-            'service': outReq.headers.cn,
-            // TODO should always be buffer
-            'target-endpoint': String(self.arg1)
+    self.services.onRequest(self);
+    var conn = peer.connect();
+    if (!peer.isConnected() && conn.handler) {
+        var timer = self.timers.setTimeout(function connectTimeOut() {
```
jcorbin: please hoist this function down to above onConnected

node/request.js
```diff
@@ -226,32 +228,50 @@ TChannelRequest.prototype.resend = function resend() {
...
+            self.checkTimeout();
+        },
+        opts.timeout);
+        if (timer.unref) timer.unref();
+        conn.on('identified', function retry() {
```
jcorbin: please call it `function onIdentified` and hoist it down

node/request.js
```diff
@@ -198,6 +198,8 @@ TChannelRequest.prototype.send = function send(arg1, arg2, arg3, callback) {
 TChannelRequest.prototype.resend = function resend() {
     var self = this;
 
+    if (self.channel.destroyed) return;
```
Raynos: do we need this check? Should someone else be doing this ?

node/request.js
jcorbin: Really we need to get this timeout checking unified with the channel's interval tick for same on outgoing requests, but this will suffice for now.

node/request.js
```diff
@@ -226,32 +228,50 @@ TChannelRequest.prototype.resend = function resend() {
...
 
-    if (self.outReqs.length === 1) {
-        self.channel.outboundCallsSentStat.increment(1, {
-            'target-service': outReq.serviceName,
-            'service': outReq.headers.cn,
-            // TODO should always be buffer
-            'target-endpoint': String(self.arg1)
+    self.services.onRequest(self);
+    var conn = peer.connect();
+    if (!peer.isConnected() && conn.handler) {
```
Raynos: What is `conn.handler`  ?

node/request.js
```diff
@@ -226,32 +228,50 @@ TChannelRequest.prototype.resend = function resend() {
...
+        var timer = self.timers.setTimeout(function connectTimeOut() {
+            self.checkTimeout();
+        },
+        opts.timeout);
+        if (timer.unref) timer.unref();
```
Raynos: do not unref timers.

node/request.js
Raynos: there should be no need to time this out.

Raynos:
Good progress; just a few comments.

node/request.js
```diff
@@ -226,32 +228,50 @@ TChannelRequest.prototype.resend = function resend() {
         opts[key] = self.options[key];
     }
     opts.timeout = self.timeout - self.elapsed;
```
jcorbin: this is now wrong whenever the onConnected starts to send it, since non-trivial time has elapsed; please:
- move the `opts` building into `onConnected`
- inline the RHS expression here in the setTImeout use inside the not connected timer set

node/request.js
jcorbin: having to chechk `timer.unfer` seems wrong to me, we're not running in the browser

node/request.js
```diff
@@ -226,32 +228,50 @@ TChannelRequest.prototype.resend = function resend() {
...
-    var outReq = peer.request(opts);
-    self.outReqs.push(outReq);
 
-    if (self.outReqs.length === 1) {
-        self.channel.outboundCallsSentStat.increment(1, {
-            'target-service': outReq.serviceName,
-            'service': outReq.headers.cn,
-            // TODO should always be buffer
-            'target-endpoint': String(self.arg1)
+    self.services.onRequest(self);
```
jcorbin: is it correct to call this observer before actually sending the request? question mainly for @kriskowal

node/request.js
ShanniLi: unref removed.

node/request.js
ShanniLi: removed.

node/request.js
ShanniLi: timeout removed.

node/request.js
kriskowal: For the purposes of back-pressure, services.onRequest should probably be called as soon as the request is queued, and only once for each corresponding error or response case. Having this in resend might actually be too late since resend gets called once for each retry. To maintain the invariant that there’s one error or response for each request, we must either onError before scheduling a retry, or onRequest once per send, not resend. I’m favoring the latter since a request that is scheduled for retry should count toward the max pending requests limit.

node/request.js
ShanniLi: Actually, we cannot remove it. The channel may be already destroyed due to the retry ....

node/request.js
ShanniLi: fixed.

node/request.js
ShanniLi: fixed.

node/request.js
ShanniLi: opts will become out of date during retry.

node/request.js
ShanniLi: there needs a timeout; otherwise, it may wait for ever for 'identified'.

ShanniLi: In order to keep this PR small, I will do a different PR to move the timeout check to TChannelConnectionBase.prototype.onTimeoutCheck.

node/request.js
ShanniLi: Talked to @Raynos, I will update TChannelConnectionBase.prototype.resetAll to handle this scenario. I will do it in a separate PR.

Raynos:
tests pass. This is good enough to ship.

We still need find that commit that makes incoming call reqs/res before init req/res and turns them into errors.

node/request.js
```diff
@@ -83,14 +80,20 @@ TChannelRequest.defaultTimeout = 5000;
 
 TChannelRequest.prototype.type = 'tchannel.request';
 
-TChannelRequest.prototype.onError = function onError(err, self) {
+TChannelRequest.prototype.emitError = function emitError(err) {
+    var self = this;
```
jcorbin: there was no need for this change

node/request.js
jcorbin: should that be done firts, then this PR rebased on it and this addition removed?

jcorbin:
lgtm

node/test/max_pending.js
```diff
@@ -28,7 +28,7 @@ var allocCluster = require('./lib/alloc-cluster');
 testBattle('cap maximum pending requests', {
     numPeers: 2,
     channelOptions: {
-        maxPending: 1,
+        maxPending: 2,
```
kriskowal: This change makes me wonder whether the max pending feature is broken.

node/request.js
```diff
@@ -83,14 +80,20 @@ TChannelRequest.defaultTimeout = 5000;
...
+    self.errorEvent.emit(self, err);
 };
 
-TChannelRequest.prototype.onResponse = function onResponse(res, self) {
+TChannelRequest.prototype.emitResponse = function emitResponse(res) {
+    var self = this;
     if (!self.end) self.end = self.timers.now();
     self.res = res;
+    self.services.onRequestResponse(self);
+    self.responseEvent.emit(self, res);
```
kriskowal: This is exactly the change I was looking for. Thanks. (@jcorbin, the motivating principle is that we should avoid using event listeners to communicate within an instance, since we can just use a method, avoid a closure, try to avoid more than one listener on an event)

node/test/max_pending.js
ShanniLi: Actually, it was working on the edge. After we lift the max pending to TChannelRequest.send(), max pending checks have a higher priority than timeout checks. So, we need to update the test.

node/request.js
ShanniLi: @jcorbin fixing it now.

PR #561, opened 2015-05-08
Emit stats for outbound call requests

Raynos:
This is a first pass at emitting outbound call request
stats. It adds an `emitStat()` helper to the channel
and wires up all the stats.

 - Implement all outbound stats except for request sizes
    and response sizes.
 - Added non-standard operational error stat

I plan to add connection stats in a seperate PR.

r: @jcorbin @kriskowal @mmihic

### DIFF
--- a/node/request.js
+++ b/node/request.js
@@ -141,6 +141,29 @@ TChannelRequest.prototype.hookupCallback = function hookupCallback(callback) {
         if (called) return;
         called = true;
         res.withArg23(function gotArg23(err, arg2, arg3) {
+            if (res.ok) {
+                self.channel.emitStat(new Stat.Counter(
+                    'outbound.calls.success', 1, {
+                        'target-service': self.serviceName,
+                        'service': self.headers.cn,
+                        // TODO should always be buffer
+                        'target-endpoint': String(self.arg1)
+                    }
+                ));
+            } else {
+                self.channel.emitStat(new Stat.Counter(
+                    'outbound.calls.app-errors', 1, {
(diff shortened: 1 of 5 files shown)


node/request.js
```diff
@@ -110,6 +111,31 @@ TChannelRequest.prototype.hookupCallback = function hookupCallback(callback) {
...
         called = true;
+
+        if (err.isErrorFrame) {
+            self.channel.emitStat(new Stat.Counter(
+                'outbound.calls.systems-errors', 1, {
+                    'target-service': self.serviceName,
+                    'service': self.headers.cn,
+                    // TODO should always be buffer
+                    'target-endpoint': String(self.arg1),
+                    'error-type': err.codeName
```
kriskowal: Recommending "caller", "callee", "method".

This would look like the following with StatEmitter:

```
// Channel
self.outboundCallsSystemErrors = self.defineCounter('outbound.calls.system-errors')
// Request
self.channel.outboundCallsSystemErrors.increment(1, {…})
```

mmihic: If you are going to drift from metrics doc, please go back and update the metrics doc so we can keep all of the language implementations in sync.

kriskowal: Ah, I see that this tracks closely. I’ll step away from the bike shed.

mmihic: :)  I have no strong preference on the names (I kindof like yours better tbh) - just want to make sure that we record whatever changes we make.

Raynos:
Abandoning in favor of #565

PR #562, opened 2015-05-08
Error on incoming call req/res frame before init res frame is sent

ShanniLi:
cc: @jcorbin @Raynos 

Separate the none breaking logic and do a merge first.
1. handle the issue by moving self.sendInitResponse up.
2. did a little refinement adding the typed error: TChannelInitProtocolError

### DIFF
--- a/node/errors.js
+++ b/node/errors.js
@@ -235,6 +235,12 @@ module.exports.SocketClosedError = TypedError({
     reason: null
 });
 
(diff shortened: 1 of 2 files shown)


kriskowal:
Be sure to follow up with individual TypedError for each of these cases. There should be one type for each reason and the reason should not be a parameter to the TypedError.

PR #564, opened 2015-05-09
Include full Thrift payload in Thrift calls

abhinav:
Currently, we skip out on the method name, message type, and sequence ID to avoid duplicate information in the payload. That avoidance of duplication has the result of complicating the Go and Python implementations a lot because they rely on the code generated by the Apache Thrift compiler.

By including this information in the payload like a standard Thrift request, we get to simplify the client implementations significantly without affecting end users in any way.

CC: @breerly @kriskowal @leizha @weikai77

### DIFF
--- a/docs/thrift.md
+++ b/docs/thrift.md
@@ -51,16 +51,18 @@ name).
 
 `arg3` must contain a Thrift struct encoded using `TBinaryProtocol`.
 
-For `call req` messages, it is a struct containing JUST the parameters of the
-method.
+For `call req` messages, it is the full encoded Thrift payload. This includes
+the method name, the Thrift message type, and sequence ID, and the parameters
+of the call.
 
-For `call res` messages,
+Similarly, for `call res` messages, it is the full encoded Thrift payload: the
+method name, message type, sequence ID, and the result.
 
--   In case of success, the response contains a struct with a single field with
-    identifier `0` that contains the return value of the method. For methods
-    with a `void` return type, the struct must be empty.
+-   In case of success, the result is a struct with a single field with
+    identifier `0` that contains the return value. For methods with a `void`
+    return type, the struct must be empty.
 
--   In case of failure, the response contains a struct with a single exception
+-   In case of failure, the result is a struct with a single exception
     field identifier with the exception struct as the value.
 
 For example,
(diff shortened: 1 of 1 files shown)


leizha:
I don't think mixing thrift seqid within tchannel payload is a good idea. do you plan to maintain a map of seqid to outstanding requests above tchannel? how does tchannel retries or proxy work with thrift seqid?
it also breaks all existing thrift-over-tchannel protocol.

Raynos:
:-1: having two sources of truth on the wire complicates matters.

These two pieces of information can get out of sync.

Can we define what the sequence id is? Is it the id for the call req? If it is we have to rewrite arg3 in the hyperbahn relay.

If it's not the call request id then how do you avoid duplicates? A service can take incoming call req from arbitary services; we could always set it to zero but will that break thrift implementations?

Putting an id in arg3 seems like a bad idea. We should avoid adding cruft to our protocol just to support legacy Apache code.

junchaowu:
In standard thrift, the static generated code will generate seqid for it's purpose. We don't actually use it.
We understand the extra data will be sent but compared with the dirty code or difficulty to work with static generated code on Python and Go, it is worth to keep it.

Raynos:
In the long run we want a lean and performant system. Adding a short term hack to the binary encoding format to make it "convenient" to bolt apache code into your codebase seems like a step back.

We should do the simplest thing possible on the wire.

Not to mention the pain of changing the binary interface right now when we are this close to cutting a v2.

abhinav:
For the record, we all discussed this yesterday in-person and the consensus we arrived at is this: Cross-language interoperability takes precedence. The current Thrift-over-TChannel protocol makes it so that all other languages we want to support will need custom Thrift compilers (this is definitely true for Go and Python). This is a major blocker in the short term, especially if we want to see this go to production soon. The path forward that we decided on was that we'll treat the Thrift binary protocol as a black box and go ahead with these changes.

In the future, when we add RPC streaming, we will most likely need some sort of custom compiler anyway. At that point, we could introduce a new arg scheme (name TBD) which would represent a more "native" Thrift-based RPC system over TChannel.

abhinav:
We decided to not do this after all. We can revisit this later.

PR #566, opened 2015-05-09
out_response: log the currentState for already done error

Raynos:
Add more information to typed error;

r: @jcorbin @kriskowal


jcorbin:
:+1:

PR #569, opened 2015-05-11
Locked states

kriskowal:
Wherein we generalize states and state machines, then add support for locked states. We can use LockedHealthyState if peer-scoped health state-machines have undesirable effects.

These are leading to having Circuit state machines using the same logic.


node/states.js
```diff
@@ -50,6 +53,24 @@ var symptoms = {
     'FatalProtocolError': true
 };
 
+// ## abstract State
+
+function State() {
+}
+
+State.prototype.onRequest = function onRequest(/* req */) {
```
Raynos: should we throw not implemented exceptions ?

node/states.js
```diff
@@ -173,6 +191,40 @@ UnhealthyState.prototype.onRequestError = function onRequestError(err) {
...
+inherits(LockedHealthyState, State);
+
+LockedHealthyState.prototype.type = 'tchannel.healthy-locked';
+
+LockedHealthyState.prototype.toString = function lockedHealthyToString() {
+    return '[Healthy state (locked)]';
+};
+
+LockedHealthyState.prototype.shouldRequest = function shouldRequest(req, options) {
+    return self.nextHandler.shouldRequest(req, options);
```
Raynos: what is `nextHandler` ? Can we rename it to `peer` ?

node/state_machine.js
```diff
@@ -0,0 +1,27 @@
+'use strict';
+
+// Peer and Circuit are state machines.
+
+module.exports = StateMachine;
+
+function StateMachine() {
+    var self = this;
+    self.state = null;
```
Raynos: can we allocate `self.stateOptions` in this constructor instead of in peer ?

Raynos:
lgtm.

node/states.js
kriskowal: No, no-op is a default behavior depended upon by many of the states.

node/states.js
kriskowal: We have to call it nextHandler because it will not be peer when it is a circuit state.

node/state_machine.js
kriskowal: stateOptions will vary between Circuit and Peer.

kriskowal:
Thanks. Waiting for CI to run.

kriskowal:
cc @jcorbin

Raynos:
lgtm; #shippingit

kriskowal:
Looks like coverage dropped. Will have to revisit.

node/states.js
```diff
@@ -173,6 +191,41 @@ UnhealthyState.prototype.onRequestError = function onRequestError(err) {
     }
 };
 
-UnhealthyState.prototype.close = function close(callback) {
-    callback(null);
+// ## LockedHealthyState
```
jcorbin: can we make this `// ## -- "` just so that we have all language's single-line commenting approaches in one?

Right, so please just make it a normal comment.

PR #570, opened 2015-05-11
Server error handling

junchaowu:
@blampe @abhinav @breerly

### DIFF
--- a/python/tchannel/exceptions.py
+++ b/python/tchannel/exceptions.py
@@ -49,12 +49,12 @@ class ProtocolException(TChannelException):
         self.description = description
 
 
-class InvalidMessageException(ProtocolException):
+class InvalidMessageException(TChannelException):
     """Represent an invalid message."""
     pass
 
 
-class InvalidEndpointException(ProtocolException):
+class InvalidEndpointException(TChannelException):
     """Represent an message containing invalid endpoint."""
     pass
 
--- a/python/tchannel/exceptions.py
+++ b/python/tchannel/exceptions.py
@@ -54,6 +54,11 @@ class InvalidMessageException(ProtocolException):
     pass
 
 
+class InvalidEndpointException(ProtocolException):
+    """Represent an message containing invalid endpoint."""
+    pass
+
+
 class TimeoutException(TChannelException):
     pass
 
--- a/python/tchannel/handler.py
+++ b/python/tchannel/handler.py
@@ -98,6 +98,11 @@ class BaseRequestHandler(RequestHandler):
         """
         try:
             req = connection.request_message_factory.build(message_id, message)
+            # call handler only for the call request message
+            # not continue message
+            if req:
+                self.handle_call(req, connection)
+
         except (InvalidChecksumException, StreamingException) as e:
             connection.send_error(
                 ErrorCode.bad_request,
@@ -110,10 +115,6 @@ class BaseRequestHandler(RequestHandler):
                 "An unexpected error has occurred!",
                 message_id,
             )
-
-        # call handler only for the call request message not continue message
-        if req:
-            self.handle_call(req, connection)
 
     def handle_ping(self, message_id, ping, connection):
         return connection.write(PingResponseMessage(), message_id)
--- a/python/tchannel/handler.py
+++ b/python/tchannel/handler.py
@@ -19,7 +19,8 @@
 # THE SOFTWARE.
 from __future__ import absolute_import
 
-from .exceptions import InvalidChecksumException, StreamingException
+from .exceptions import InvalidChecksumException
+from .exceptions import StreamingException
 from .messages import ErrorCode
 from .messages import PingResponseMessage
 from .messages import Types
--- a/python/tchannel/handler.py
+++ b/python/tchannel/handler.py
@@ -98,8 +98,9 @@ class BaseRequestHandler(RequestHandler):
         """
         try:
             req = connection.request_message_factory.build(message_id, message)
-            # call handler only for the call request message
-            # not continue message
+            # message_factory will create Request only when it receives
+            # CallRequestMessage. It will return None, if it receives
+            # CallRequestContinueMessage.
             if req:
                 self.handle_call(req, connection)
 
--- a/python/tchannel/handler.py
+++ b/python/tchannel/handler.py
@@ -19,7 +19,7 @@
 # THE SOFTWARE.
 from __future__ import absolute_import
 
-from .exceptions import InvalidChecksumException
+from .exceptions import InvalidChecksumException, StreamingException
 from .messages import ErrorCode
 from .messages import PingResponseMessage
 from .messages import Types
@@ -97,10 +97,16 @@ class BaseRequestHandler(RequestHandler):
         """
         try:
             req = connection.request_message_factory.build(message_id, message)
-        except InvalidChecksumException as e:
+        except (InvalidChecksumException, StreamingException) as e:
             connection.send_error(
                 ErrorCode.bad_request,
                 e.message,
+                message_id,
+            )
+        except Exception:
+            connection.send_error(
+                ErrorCode.unexpected,
+                "An unexpected error has occurred!",
                 message_id,
             )
 
--- a/python/tchannel/tornado/connection.py
+++ b/python/tchannel/tornado/connection.py
@@ -36,14 +36,15 @@ from ..context import Context
 from ..event import EventType
 from ..exceptions import ConnectionClosedException
 from ..exceptions import InvalidErrorCodeException
+from ..exceptions import TChannelException
 from ..io import BytesIO
 from ..messages.common import PROTOCOL_VERSION
-from ..messages.common import StreamState
 from ..messages.common import FlagsType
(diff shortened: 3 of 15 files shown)


python/tchannel/handler.py
```diff
@@ -97,16 +98,23 @@ def handle_pre_call(self, message_id, message, connection):
         """
         try:
             req = connection.request_message_factory.build(message_id, message)
-        except InvalidChecksumException as e:
+            # call handler only for the call request message
+            # not continue message
+            if req:
```
blampe: When would this be `None`?

python/tests/integration/test_error_handling.py
```diff
@@ -0,0 +1,151 @@
...
+            'a',
+        ],
+    )
+
+    with pytest.raises(ProtocolException) as e:
+        resp_future = connection.send(callrequest, 1)
+        for _ in xrange(10):
+            yield connection.write(callreqcontinue, 1)
+
+        # by pass the default checksum calculation
```
blampe: bypass

python/tchannel/tornado/connection.py
```diff
@@ -527,21 +528,29 @@ def _stream(self, context, message_factory):
...
+            # last piece of request/response.
+            message = (message_factory.
+                       build_raw_message(context, args, is_completed=True))
+            yield self.write(message, context.id)
+            context.state = StreamState.completed
+        # Stop streamming immediately if exception occurs on the handler side
+        except TChannelException:
+            # raise by tchannel intentionally
+            pass
+        except Exception:
```
blampe: The `except Exception / raise` is a no-op and can be removed.

python/tchannel/handler.py
junchaowu: it will be none, when we receive callreqcontinue message. We only build request when we receive callrequest message,

python/tests/integration/test_error_handling.py
```diff
@@ -0,0 +1,151 @@
...
+    )
+
+    callreqcontinue = CallRequestContinueMessage(
+        flags=FlagsType.fragment,
+        args=[
+            'a',
+        ],
+    )
+
+    with pytest.raises(ProtocolException) as e:
```
blampe: This is wrapping a pretty large block of code, just put this around the statement you expect to raise the error.

python/tests/test_stream.py
```diff
@@ -80,3 +80,23 @@ def test_response_exception():
...
+    stream = InMemStream()
+
+    @pytest.mark.gen_test
+    def set_exception():
+        try:
+            1 / 0
+        except Exception as e:
+            stream.set_exception(e)
+
+    io_loop.call_later(
```
blampe: `ioloop.add_callback(set_exception)`

python/tests/test_stream.py
```diff
@@ -80,3 +80,23 @@ def test_response_exception():
     resp.flush()
     with pytest.raises(TChannelException):
         yield resp.write_body("aaaa")
+
+
+@pytest.mark.gen_test
+def test_error_during_stream(io_loop):
+    stream = InMemStream()
+
+    @pytest.mark.gen_test
```
blampe: This shouldn't be another `gen_test`. It would be a `coroutine`, but if you `return stream.set_exception(e)` (which should be a future) then you don't need a decorator at all.

python/tchannel/exceptions.py
```diff
@@ -54,6 +54,11 @@ class InvalidMessageException(ProtocolException):
     pass
 
 
+class InvalidEndpointException(ProtocolException):
+    """Represent an message containing invalid endpoint."""
+    pass
```
abhinav: InvalidExceptionEndpoint shouldn't be a ProtocolException.

ProtocolException represents an error caused by protocol error frames.

python/tchannel/handler.py
blampe: The comment could be clearer.

python/tchannel/tornado/connection.py
```diff
@@ -527,21 +528,29 @@ def _stream(self, context, message_factory):
...
+
+            # last piece of request/response.
+            message = (message_factory.
+                       build_raw_message(context, args, is_completed=True))
+            yield self.write(message, context.id)
+            context.state = StreamState.completed
+        # Stop streamming immediately if exception occurs on the handler side
+        except TChannelException:
+            # raise by tchannel intentionally
+            pass
```
abhinav: What do you mean? Isn't this the exception we use for all custom exceptions defined in this project?

Also, regardless, usually you don't want to silently blackhole exceptions.

breerly: At the very least, you'll want logging.

python/tchannel/tornado/data.py
```diff
@@ -163,14 +168,18 @@ def __init__(
...
     def status_code(self, status):
         if status not in StatusCode:
             raise TChannelException("Invalid status code!")
 
-        self.flags = status.value
+        self.code = status.value
+
+    @property
+    def ok(self):
+        return self.cdoe == StatusCode.ok.value
```
abhinav: typo

python/tchannel/tornado/dispatch.py
```diff
@@ -102,13 +105,22 @@ def handle_call(self, request, connection):
...
+            except (InvalidMessageException, InvalidEndpointException) as e:
+                response.set_exception(e)
+                connection.request_message_factory.remove_buffer(response.id)
+                connection.send_error(
+                    ErrorCode.bad_request,
+                    e.message,
+                    response.id,
+                )
+            except Exception as e:
+                response.set_exception(TChannelException())
```
abhinav: What? Why are you abandoning `e`?

python/tchannel/tornado/data.py
```diff
@@ -312,6 +321,11 @@ def flush(self):
         self.flushed = True
         self.close_argstreams()
 
+    def set_exception(self, exception):
+        for stream in self.argstreams:
+            stream.set_exception(exception)
+            stream.close()
```
abhinav: so these will cause the stream to fail on `.write`s?

python/tchannel/tornado/stream.py
```diff
@@ -98,11 +98,17 @@ def __init__(self, buf=None, auto_close=True):
         self._condition = Condition()
         self.auto_close = auto_close
 
+        self.err_future = tornado.gen.Future()
```
abhinav: Why are you using a future to track this. You can just track the value inside an instance variable and raise that.

python/tchannel/tornado/stream.py
```diff
@@ -118,6 +124,12 @@ def write(self, chunk):
             self._condition.notify()
```
abhinav: Why doesn't `write` fail with an exception if the stream has failed with an exception

python/tests/test_stream.py
```diff
@@ -80,3 +80,23 @@ def test_response_exception():
...
+    def set_exception():
+        try:
+            1 / 0
+        except Exception as e:
+            stream.set_exception(e)
```
abhinav: Why does this need to be done in a callback at all. It's an asynchronous exception. You can set it right away and check that it's thrown when you `yield` the `read()`.

python/tchannel/tornado/data.py
junchaowu: it will. Stream will be closed

python/tchannel/tornado/dispatch.py
junchaowu: added e.message info

python/tchannel/tornado/data.py
junchaowu: fixed

python/tchannel/tornado/connection.py
junchaowu: log the exception info

python/tchannel/tornado/stream.py
junchaowu: Good point

python/tchannel/exceptions.py
junchaowu: That's still a protocolexception. It doesn't reaches to handler yet.

abhinav: From your original implementation, ProtocolException is only for error frame exceptions. ProtocolExceptions have a code, description, id, and tracing. 

Also, you're instantiating `InvalidEndpointException(..)` with just the message. The constructor for `ProtocolException` accepts at least 2 arguments. This won't work anyway. Most likely same with `InvalidMessageException`.

junchaowu: Right.

breerly:
One thing I've noticed that we seem to be doing wrong. When libraries expose exceptions, they should end in `Error`. This seems to be the case in the standard library as well.

What do you guys think? (of course we can fix in another commit).

abhinav:
@breerly Sounds good.

breerly:
:shipit:

python/tchannel/tornado/stream.py
abhinav: Ping?

PR #574, opened 2015-05-12
Test sorted errors

jcorbin:
Render unto computers what humans are bad at...

Turns out many of the ones I added were also out of order...

r @kriskowal @Raynos @ShanniLi


kriskowal:
:ok:

ShanniLi:
nice! :)

PR #575, opened 2015-05-12
Fix out req popping for incoming call response

Raynos:
We had a bunch of false positive around logging
`error received for unknown or lost operation`
because we were not tracking the call req vs incoming call res
data correctly.

This PR cleans up the logic to only pop an out req when an
incoming call response is truly finished

r: @jcorbin @kriskowal

### DIFF
--- a/node/connection.js
+++ b/node/connection.js
@@ -235,12 +235,21 @@ TChannelConnection.prototype.onCallResponse = function onCallResponse(res) {
 
 TChannelConnection.prototype.onCallError = function onCallError(err) {
     var self = this;
-    var req = self.popOutReq(err.originalId);
-    if (!req) {
-        self.logger.info('error received for unknown or lost operation', err);
-        return;
+
+    var req = self.requests.out[err.originalId];
+
+    if (req && req.res) {
+        req.res.errorEvent.emit(req.res, err);
+    } else {
+        // Only popOutReq if there is no call response object yet
+        req = self.popOutReq(err.originalId);
+        if (!req) {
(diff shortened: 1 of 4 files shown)


node/test/streaming-resp-err.js
```diff
@@ -0,0 +1,68 @@
+'use strict';
```
jcorbin: copyright blurb

jcorbin:
lgtm, but blurb it

node/connection.js
```diff
@@ -214,16 +210,46 @@ TChannelConnection.prototype.onCallResponse = function onCallResponse(res) {
...
     var self = this;
-    var req = self.popOutReq(err.originalId);
-    if (!req) {
-        self.logger.info('error received for unknown or lost operation', err);
-        return;
+
+    var req = self.requests.out[err.originalId];
+
+    if (req && req.res) {
+        req.res.errorEvent.emit(req.res, err);
```
kriskowal: woot

node/connection.js
```diff
@@ -190,20 +191,15 @@ TChannelConnection.prototype.handleReadFrame = function handleReadFrame(frame) {
...
-            code: res.code,
-            arg1: Buffer.isBuffer(res.arg1) ?
-                String(res.arg1) : "streamed-arg1",
-            remoteAddr: self.remoteAddr,
-            direction: self.direction,
-        });
-        return;
+
+    var req = self.requests.out[res.id];
+    var called = false;
```
kriskowal: Perhaps call this “popped”

kriskowal:
:ok: Thanks

PR #576, opened 2015-05-12
Make tracing tests silent by default

Raynos:
Our tests should be silent by default and only output
TAP to work nicely with tap workflows.

For example: using `grep` to parse out TAP output

```
$ node test/index.js | grep -v '^ok \|^#'
TAP version 13
TCHANNEL WARN: localhost:37892 destroying socket from timeouts ~ null
TCHANNEL WARN: mismatched onReqDone callback ~ { hostPort: 'localhost:42207', hasInReq: false, id: 1 }
TCHANNEL WARN: mismatched onReqDone callback ~ { hostPort: 'localhost:38391', hasInReq: false, id: 2 }

1..2194
```

I will fix the warnings in seperate PRs

r: @rf @kriskowal


rf:
lgtm

PR #577, opened 2015-05-12
Remove unnecessary opReqDone warnings

Raynos:
There are a bunch of false positives for this warning that
I'm removing with these changes.

This makes actual production issues easier to debug.

r: @jcorbin @kriskowal

### DIFF
--- a/node/self_out_response.js
+++ b/node/self_out_response.js
@@ -71,9 +71,13 @@ function passResponse(args, isLast ) {
         self.inres.code = self.code;
         self.inres.ok = self.ok;
         self.first = false;
-        self.inreq.responseEvent.emit(self, self.inres);
+        process.nextTick(emitResponse);
     }
     if (!self.closing) self.conn.lastTimeoutTime = 0;
+
+    function emitResponse() {
+        self.inreq.responseEvent.emit(self, self.inres);
+    }
 };
 
 SelfOutResponse.prototype._sendError =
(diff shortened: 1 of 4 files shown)


node/self_out_response.js
```diff
@@ -71,9 +71,13 @@ function passResponse(args, isLast ) {
         self.inres.code = self.code;
         self.inres.ok = self.ok;
         self.first = false;
-        self.inreq.responseEvent.emit(self, self.inres);
+        process.nextTick(emitResponse);
```
jcorbin: why the defer?

node/test/timeouts.js
```diff
@@ -61,12 +61,26 @@ allocCluster.test('requests will timeout', {
...
+
+        var adv = Math.max(1001, (
+            twoConn.options.timeoutCheckInterval +
+            (twoConn.options.timeoutFuzz / 2) + 1
+        ));
+
+        // timers module has weird semantics
+        // TODO: less magic
+        var newTime = timers.now() + adv;
+        timers.now = function now() {
```
jcorbin: shared mutation... wat?

node/self_out_response.js
Raynos: necessary to avoid race condition; similar to nextTick in error path.

otherwise we emit `finish` on the outres before the handler completes

```js
function (req, res, arg2, arg3) {
  res.sendOk();
  console.log('please first');
  console.log('why am I second :(');
}

.request().send(..., function onResponse(err, res) {
  console.log('out res is actually first lol; zalgo lol);
})
```

node/test/timeouts.js
Raynos: i dont have time to fix the `time-mock` library so I punched it in the face.

jcorbin:
makes sense, I can cope with it

PR #579, opened 2015-05-13
move request from data.py to request.py

junchaowu:
move response from data.py to response.py
@blampe @breerly @abhinav


breerly:
:+1: request.py makes alot of sense. Kickass

blampe:
nice

PR #581, opened 2015-05-13
refactor tests to be request({ serviceName: service })

ShanniLi:
r @jcorbin @Raynos
cc @kriskowal 

1. refactor tests to be request({ serviceName: service })
2. refactor streaming tests to block on init res
3. three tests left that I haven't figured out a good way to update:
'./streaming_bisect.js'
'./as-http.js'
'./trace/'


kriskowal:
Looks good to me

Raynos:
lgtm!

jcorbin:
:heart_eyes:

PR #582, opened 2015-05-13
TChannelRequest: no retries when connection is closing

ShanniLi:
r @jcorbin @Raynos 
cc @kriskowal 

When a connection is closing, TChannelRequest should not do retries.

### DIFF
--- a/node/connection_base.js
+++ b/node/connection_base.js
@@ -198,9 +198,7 @@ TChannelConnectionBase.prototype.resetAll = function resetAll(err) {
         var req = self.requests.out[id];
         delete self.requests.out[id];
         
-        err = errors.TChannelConnectionResetError(err, {
-            connection: self
-        });
+        err = errors.TChannelConnectionResetError(err);
 
         req.errorEvent.emit(req, err);
     });
--- a/node/connection_base.js
+++ b/node/connection_base.js
@@ -197,9 +197,7 @@ TChannelConnectionBase.prototype.resetAll = function resetAll(err) {
     outOpKeys.forEach(function eachOutOp(id) {
         var req = self.requests.out[id];
         delete self.requests.out[id];
-        
         err = errors.TChannelConnectionResetError(err);
-
         req.errorEvent.emit(req, err);
     });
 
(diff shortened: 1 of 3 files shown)


node/connection_base.js
```diff
@@ -197,7 +197,8 @@ TChannelConnectionBase.prototype.resetAll = function resetAll(err) {
     outOpKeys.forEach(function eachOutOp(id) {
         var req = self.requests.out[id];
         delete self.requests.out[id];
-        // TODO: shared mutable object... use Object.create(err)?
+        err = Object.create(err);
+        err.connectionClosing = true;
```
kriskowal: Will need to ensure that we use the good is-error implementation to handle this case. Might be better to clone properties and use a fresh Error instance for each. TypedError.clone(error)?

Raynos: We have `WrappedError` for wrapping errors and adding context; I recommend using that.

node/connection_base.js
```diff
@@ -197,7 +197,11 @@ TChannelConnectionBase.prototype.resetAll = function resetAll(err) {
     outOpKeys.forEach(function eachOutOp(id) {
         var req = self.requests.out[id];
         delete self.requests.out[id];
-        // TODO: shared mutable object... use Object.create(err)?
+        
+        err = errors.TChannelConnectionResetError(err, {
+            connection: self
```
Raynos: you do not want to put non trivial objects on errors. All of these fields are enumerable by default and if someone logs this err it will write a rather large object.

I recommend putting simpler meta data on this object.

ShanniLi: Make sense. I will remove connection field, since the error is very much self explanatory.

Raynos:
lgtm!

node/errors.js
```diff
@@ -259,6 +259,11 @@ module.exports.TChannelCallResContBeforeInitResError = TypedError({
     message: 'call response cont before init response'
 });
 
+module.exports.TChannelConnectionResetError = WrappedError({
+    type: 'tchannel.connection.reset',
+    message: 'tchannel: {origMessage}'
```
kriskowal: Recommend cause and causeMessage properties. I contributed a change to the errors module to make these property names more concise.

jcorbin:
:heart:

PR #583, opened 2015-05-13
fix the last four tests

ShanniLi:
r @jcorbin @kriskowal @Raynos

The rest would be adding the errors back to v2/handler.js. :)

### DIFF
--- a/node/as/http.js
+++ b/node/as/http.js
@@ -169,7 +169,6 @@ TChannelHTTP.prototype.forwardToTChannel = function forwardToTChannel(tchannel, 
     var options = tchannel.requestOptions({
         streamed: true
     });
-
     var peer = tchannel.peers.choosePeer(null, options);
     if (!peer) {
         hres.writeHead(503, 'Service Unavailable: no tchannel peer');
--- a/node/as/http.js
+++ b/node/as/http.js
@@ -169,6 +169,7 @@ TChannelHTTP.prototype.forwardToTChannel = function forwardToTChannel(tchannel, 
     var options = tchannel.requestOptions({
         streamed: true
     });
+
     var peer = tchannel.peers.choosePeer(null, options);
     if (!peer) {
         hres.writeHead(503, 'Service Unavailable: no tchannel peer');
@@ -177,7 +178,14 @@ TChannelHTTP.prototype.forwardToTChannel = function forwardToTChannel(tchannel, 
     } else {
         // TODO: observable
         var treq = peer.request(options);
-        self.sendRequest(treq, hreq, forwarded);
+        if (!peer.isConnected()) {
+            peer.connect().on('identified', function onIdentified() {
+                self.sendRequest(treq, hreq, forwarded);
+            });
+        } else {
+            self.sendRequest(treq, hreq, forwarded);
+        }
+
         return treq;
     }
 
--- a/node/test/as-http.js
+++ b/node/test/as-http.js
@@ -189,7 +189,14 @@ function allocHTTPBridge(opts) {
     }
 
     function onEgressRequest(hreq, hres) {
-        cluster.asHTTP.forwardToTChannel(cluster.egressChan, hreq, hres);
+        var options = cluster.egressChan.requestOptions({
+            streamed: true
+        });
+        var peer = cluster.egressChan.peers.choosePeer(null, options);
+        var conn = peer.connect();
+        conn.on('identified', function onId() {
+            cluster.asHTTP.forwardToTChannel(cluster.egressChan, hreq, hres);
+        });
     }
 
     function destroy(callback) {
--- a/node/test/as-http.js
+++ b/node/test/as-http.js
@@ -189,14 +189,7 @@ function allocHTTPBridge(opts) {
     }
 
     function onEgressRequest(hreq, hres) {
-        var options = cluster.egressChan.requestOptions({
-            streamed: true
-        });
-        var peer = cluster.egressChan.peers.choosePeer(null, options);
-        var conn = peer.connect();
-        conn.on('identified', function onId() {
-            cluster.asHTTP.forwardToTChannel(cluster.egressChan, hreq, hres);
-        });
+        cluster.asHTTP.forwardToTChannel(cluster.egressChan, hreq, hres);
     }
 
     function destroy(callback) {
--- a/node/test/streaming_bisect.js
+++ b/node/test/streaming_bisect.js
@@ -55,6 +55,7 @@
  */
 
 var async = require('async');
+var CountedReadySignal = require('ready-signal/counted');
 var extend = require('xtend');
 var minimist = require('minimist');
 var util = require('util');
@@ -234,16 +235,12 @@ var search = clusterSearch.ClusterIsolateSearch(extend({
                 serviceName: 'test_as_raw',
             }
         });
-        var options = cluster.testRawClient.requestOptions({
-            serviceName: 'test_as_raw',
-            headers: {
-                as: 'raw'
-            },
-            streamed: true
+        var peers = cluster.testRawClient.peers.values();
+        var ready = new CountedReadySignal(peers.length);
+        peers.forEach(function each(peer) {
+            peer.connect().on('identified', ready.signal);
         });
-        var peer = cluster.testRawClient.peers.choosePeer(null, options);
-        var conn = peer.connect();
-        conn.on('identified', function onId() {
+        ready(function onReady() {
             callback(null);
         });
     },
--- a/node/test/streaming_bisect.js
+++ b/node/test/streaming_bisect.js
@@ -234,7 +234,18 @@ var search = clusterSearch.ClusterIsolateSearch(extend({
                 serviceName: 'test_as_raw',
             }
         });
-        callback(null);
+        var options = cluster.testRawClient.requestOptions({
+            serviceName: 'test_as_raw',
+            headers: {
+                as: 'raw'
+            },
+            streamed: true
+        });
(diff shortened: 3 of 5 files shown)


node/test/as-http.js
```diff
@@ -189,7 +189,14 @@ function allocHTTPBridge(opts) {
     }
 
     function onEgressRequest(hreq, hres) {
-        cluster.asHTTP.forwardToTChannel(cluster.egressChan, hreq, hres);
+        var options = cluster.egressChan.requestOptions({
```
jcorbin: no, this logic should be put into `asHTTP.forwardToTChannel` itself since this logic will also be needed for real.

node/test/streaming_bisect.js
```diff
@@ -234,7 +234,18 @@ var search = clusterSearch.ClusterIsolateSearch(extend({
                 serviceName: 'test_as_raw',
             }
         });
-        callback(null);
+        var options = cluster.testRawClient.requestOptions({
```
jcorbin: You shouldn't need any of these options, in fact you really don't even need to use choosePeer, rather:
```
var peers = cluster.testRawClient.peers.values();
var ready = new CountedReadySignal(peers.length);
peers.forEach(function each(peer) {
  peer.connect().on('identified', ready.signal);
});
ready(callback);
```

node/test/trace/basic_server.js
```diff
@@ -70,17 +70,22 @@ test('basic tracing test', function (assert) {
     server.handler.register('/top_level_endpoint', function (req, res) {
         logger.info("top level sending to subservice");
         setTimeout(function () {
-            server
-                .request({
-                    host: '127.0.0.1:4042', 
-                    serviceName: 'subservice', 
+            var options = server.requestOptions({
```
jcorbin: same remark as above, use countedreadysignal, and just connect to all peers

node/test/trace/basic_server.js
```diff
@@ -92,14 +97,17 @@ test('basic tracing test', function (assert) {
             throw err;
         }
 
-        logger.info("client making req");
-        client
-            .request({host: '127.0.0.1:4040', serviceName: 'server', trace: true})
-            .send('/top_level_endpoint', "arg 1", "arg 2", function (err, res) {
+        logger.info('client making req');
+        var options = client.requestOptions({host: '127.0.0.1:4040', serviceName: 'server', trace: true});
```
jcorbin: likewise

node/test/trace/server_2_requests.js
```diff
@@ -92,18 +92,23 @@ test('basic tracing test', function (assert) {
         }, 40);
 
         process.nextTick(function () {
-            server
-                .request({
-                    host: '127.0.0.1:4042', 
-                    serviceName: 'subservice', 
+            var options = server.requestOptions({
```
jcorbin: likewise

node/test/trace/server_2_requests.js
```diff
@@ -120,14 +125,17 @@ test('basic tracing test', function (assert) {
             throw err;
         }
 
-        logger.info("client making req");
-        client
-            .request({host: '127.0.0.1:4040', serviceName: 'server', trace: true})
-            .send('/top_level_endpoint', "arg 1", "arg 2", function (err, res) {
+        logger.info('client making req');
+        var options = client.requestOptions({host: '127.0.0.1:4040', serviceName: 'server', trace: true});
```
jcorbin: likewise

node/test/as-http.js
ShanniLi: fixed.

node/test/streaming_bisect.js
ShanniLi: updated.

node/test/trace/basic_server.js
ShanniLi: updated

node/test/trace/basic_server.js
```diff
@@ -70,17 +70,28 @@ test('basic tracing test', function (assert) {
...
-                    logger.info("top level recv from subservice");
-                    if (err) return res.sendOk('error', err);
-                    res.sendOk('result', 'success: ' + subRes);
-                });
+            var servReq = server.request({
+                        host: '127.0.0.1:4042',
+                        serviceName: 'subservice',
+                        parentSpan: req.span,
+                        trace: true
+                    });
```
kriskowal: Hanging indent undesirable in this case. No method chain.

node/as/http.js
```diff
@@ -177,7 +177,14 @@ TChannelHTTP.prototype.forwardToTChannel = function forwardToTChannel(tchannel,
...
         var treq = peer.request(options);
-        self.sendRequest(treq, hreq, forwarded);
+        if (!peer.isConnected()) {
+            peer.connect().on('identified', function onIdentified() {
+                self.sendRequest(treq, hreq, forwarded);
+            });
+        } else {
+            self.sendRequest(treq, hreq, forwarded);
+        }
+
```
kriskowal: If this is necessary, this smells like a sendRequest bug, or we should be using something else for this test.

node/test/trace/basic_server.js
```diff
@@ -70,17 +70,28 @@ test('basic tracing test', function (assert) {
...
+            peers.forEach(function each(peer) {
+                if (peer.isConnected()) {
+                    ready.signal();
+                } else {
+                    peer.connect().on('identified', ready.signal);
+                }
+            });
+            ready(function send() {
+                servReq.send('/foobar', 'arg1', 'arg2', function (err, subRes) {
+                        logger.info('top level recv from subservice');
```
Raynos: double indent

node/test/trace/basic_server.js
```diff
@@ -70,17 +70,28 @@ test('basic tracing test', function (assert) {
...
-                    serviceName: 'subservice', 
-                    parentSpan: req.span,
-                    trace: true
-                }).send('/foobar', 'arg1', 'arg2', function (err, subRes) {
-                    logger.info("top level recv from subservice");
-                    if (err) return res.sendOk('error', err);
-                    res.sendOk('result', 'success: ' + subRes);
-                });
+            var servReq = server.request({
+                        host: '127.0.0.1:4042',
```
Raynos: mega epic indent

node/test/trace/basic_server.js
```diff
@@ -70,17 +70,28 @@ test('basic tracing test', function (assert) {
...
+            var ready = new CountedReadySignal(peers.length);
+            peers.forEach(function each(peer) {
+                if (peer.isConnected()) {
+                    ready.signal();
+                } else {
+                    peer.connect().on('identified', ready.signal);
+                }
+            });
+            ready(function send() {
+                servReq.send('/foobar', 'arg1', 'arg2', function (err, subRes) {
```
Raynos: function name lol

Raynos:
lgtm; can use some whitespace cleanup

node/test/trace/basic_server.js
ShanniLi: fixed

node/as/http.js
ShanniLi: Since it uses stream, we cannot add blocking/retry for init req/res now. Maybe for v3 to fix it? :)

node/test/trace/basic_server.js
ShanniLi: fixed

node/test/trace/basic_server.js
ShanniLi: fixed

node/test/trace/basic_server.js
ShanniLi: fixed

node/as/http.js
kriskowal: That sounds like the right call. Perhaps we can simplify this by adding a withConnection(callback) utility to DRY this up (land first).

node/test/trace/basic_server.js
```diff
@@ -70,17 +70,27 @@ test('basic tracing test', function (assert) {
...
-                }).send('/foobar', 'arg1', 'arg2', function (err, subRes) {
-                    logger.info("top level recv from subservice");
+            var servReq = server.request({
+                host: '127.0.0.1:4042',
+                serviceName: 'subservice',
+                parentSpan: req.span,
+                trace: true});
+            var peers = server.peers.values();
+            var ready = new CountedReadySignal(peers.length);
+            peers.forEach(function each(peer) {
```
kriskowal: This would be another candidate for a peer.withConnection(ready.signal) utility to DRY it up.

node/test/trace/basic_server.js
```diff
@@ -70,17 +70,27 @@ test('basic tracing test', function (assert) {
...
+                if (peer.isConnected()) {
+                    ready.signal();
+                } else {
+                    peer.connect().on('identified', ready.signal);
+                }
```
kriskowal: And here’s another situation where it looks like we should use signals instead of events. `peer.connectedSignal.on(…)`, to simplify monitoring the connected state, whether it happened before or will happen eventually, whether it needs to be initiated (peer.connect()) or whether it has already been initiated (peer.isConnected()), and perhaps even capture and replay errors to new observers introduced while connecting via the `peer.connect(cb(err))` interface. cc @jcorbin @Raynos

kriskowal:
:ok:

node/as/http.js
jcorbin: :+1: to @kriskowal's idea , but maybe for another PR:
```
peer.withConnection(function connected() {
  ... proceed to send
});
```

jcorbin:
:ok: enough to merge, the peer method can come later

PR #585, opened 2015-05-14
Fix request error on connection reset

kriskowal:
Revealed by testing against Hyperbahn.

r @jcorbin @ShanniLi 
cc @Raynos

### DIFF
--- a/node/connection_base.js
+++ b/node/connection_base.js
@@ -39,6 +39,10 @@ function TChannelConnectionBase(channel, direction, remoteAddr) {
     self.timedOutEvent = self.defineEvent('timedOut');
     self.spanEvent = self.defineEvent('span');
 
+    self.closing = false;
+    self.closeError = null;
+    self.closeEvent = self.defineEvent('close');
+
     self.channel = channel;
     self.options = self.channel.options;
     self.logger = channel.logger;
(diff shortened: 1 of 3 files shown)


node/connection_base.js
```diff
@@ -175,7 +180,7 @@ TChannelConnectionBase.prototype.resetAll = function resetAll(err) {
         outPending: self.pending.out
     };
 
-    if (err.type.indexOf('tchannel.socket') < 0) {
+    if (err.type.indexOf('tchannel.socket', 0) < 0) {
```
Raynos: why add the `0`; it should default to `0`.

node/request.js
```diff
@@ -224,19 +224,36 @@ TChannelRequest.prototype.resend = function resend() {
...
+            onIdentified();
+        }
+    }
+
+    function onConnectionError(err) {
+        conn.closeEvent.removeListener(onConnectionError);
+        if (timer) {
+            self.timers.clearTimeout(timer);
+        }
+        self.emitError(err);
```
Raynos: `err || new Error('connection reset')`

We dont know if `err` will always exist or not.

node/connection_base.js
kriskowal: Oh, half done. Should be `lastIndexOf` to check for `/^tchannel.socket/`

kriskowal:
@raynos Addressed feedback

Raynos:
lgtm

ShanniLi:
lgtm

jcorbin:
:+1:

PR #586, opened 2015-05-14
Implement TChannel configuration using `with_options`.

abhinav:
This will allow us to build `as_json`, `as_thrift`, on top of it.

The `TChannel.call` method will replace the `.request.send` chain but for now
it's just calling `.request.send`.

After I add the `as_*` methods, client-side arg-scheme story will change from:

    ArgSchemeBroker(JsonArgSchema()).send(
         tchannel.request(service='foo', hostport='localhost:4040'), arg1, arg2, arg3
     )

To any of:

    tchannel.as_json(service='foo', hostport='localhost:4040').call(args=(arg1, arg2, arg3))
    tchannel.as_json().call(args=(arg1, arg2, arg3), service='foo', hostport='localhost:4040')
    tchannel.as_json(service='foo').call(args=(arg1, arg2, arg3), hostport='localhost:4040')

Where the object returned by `as_json` is re-usable. Note that `as_json(**foo)` will just be shorthand for `with_options(scheme=JsonArgScheme(), **foo)`.

These proxy objects will track Zipkin information and be re-used for
server-side code (that currently uses its own custom proxy type).

Note that a caveat of this approach is that `.call` requires keyword parameters. We can make that go away if we specialize `.call` to have only the `arg{1,2,3}` arguments and require everything else to use `.with_options` or the corresponding `.as_*` function.

----

Tested manually against example server.

Right now, this PR is mostly to gather opinions from @blampe, @breerly, and @junchaowu. I'll update it with tests, the zipkin stuff, and the server-side stuff once I have some feedback.

### DIFF
--- a/python/tchannel/tornado/tchannel.py
+++ b/python/tchannel/tornado/tchannel.py
@@ -32,11 +32,13 @@ from tornado.netutil import bind_sockets
 
 from enum import IntEnum
 
+from ..errors import TChannelError
 from ..event import EventEmitter
 from ..event import EventRegistrar
 from ..handler import CallableRequestHandler
 from ..net import local_ip
 from .connection import StreamConnection
+from .stream import Stream
 from .peer import PeerGroup
 
 log = logging.getLogger('tchannel')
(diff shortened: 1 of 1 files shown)


breerly:
Don't feel super strongly, but I think Python might in general be anti-chaining, pro-named-parameters.

cc @blampe does that resonate with you?

See http://stackoverflow.com/questions/3883907/designing-an-python-api-fluent-interface-or-arguments

abhinav:
Used the wrong words. You wouldn't usually chain too many calls of this. This is wrong:

    tchannel.with_options(service='foo').with_options(hostport='localhost:4040')

blampe:
I'm not opposed to chaining at all. As an API consumer, I really like that this hides the broker/schema implementation detail from me. I also like your first example:

`tchannel.as_json(service='foo', hostport='localhost:4040').call(args=(arg1, arg2, arg3))`

Just spitballing, but a chain like this "reads" nicer to me:

`tchannel.call(service='foo', hostport='...').with(arg1, arg2=None, arg3=None).as_json()

In this case I think a `call` would return a `Peer`, `with` would return a `Request`, and the `as_*` would basically be the `send`.

abhinav:
Let's hold off on this change for now.

PR #587, opened 2015-05-14
Service proxy

kriskowal:
Note that I will rebase out the original before merging, but see commit 2 for changes from Hyperbahn’s original content.


jcorbin:
works, those warn logs about exit <-> forward change should probably just be infos

kriskowal:
Reduced warnings to info. Squashed the version from Hyperbahn out.

PR #588, opened 2015-05-14
hyperbahn/handler: retry the send relay ad endpoint

Raynos:
We noticed that we are getting socket errors during
restarts of our hyperbahn cluster.

This adds two retries with upto 4s of latency to the
advertisement logic

This allows us to smooth over socket errors that
occur because of restarts as part of deploys.

r: @jcorbin @kriskowal

### DIFF
--- a/node/hyperbahn/handler.js
+++ b/node/hyperbahn/handler.js
@@ -28,6 +28,9 @@ var util = require('util');
 var TChannelJSON = require('../as/json');
 var TChannelEndpointHandler = require('../endpoint-handler');
 
+var MAX_RELAY_AD_ATTEMPTS = 2;
+var RELAY_AD_RETRY_TIME = 2 * 1000;
+
 module.exports = HyperbahnHandler;
 
 function HyperbahnHandler(options) {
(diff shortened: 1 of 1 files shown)


kriskowal:
:ok:

node/hyperbahn/handler.js
```diff
@@ -28,6 +28,9 @@ var util = require('util');
 var TChannelJSON = require('../as/json');
 var TChannelEndpointHandler = require('../endpoint-handler');
 
+var MAX_RELAY_AD_ATTEMPTS = 2;
```
jcorbin: even if we define defaults here, can we then use them as an option / self property later on?

node/hyperbahn/handler.js
```diff
@@ -193,27 +196,41 @@ HyperbahnHandler.prototype.sendRelayAdvertise =
 function sendRelayAdvertise(hostPort, services, callback) {
     var self = this;
 
-    self.tchannelJSON.send(self.channel.request({
-        host: hostPort,
-        serviceName: 'hyperbahn'
-    }), 'relay-ad', null, {
-        services: services
-    }, onRelayAdvertise);
+    var attempts = 0;
```
jcorbin: prefer this to be passed as function arg... could even then hoist it to a method, so we have:
```
HH#sendRelayAd(h, s, cb) {
  self._tryRelayAd(0, h, s, cb);
};

HH#_tryRelayAd(attempt, h, s, cb) {
  ...
  if (maybe) self._tryRelayAd(attemp + 1, h, s, cb);
  else cb(err, res);
};
```

jcorbin:
:ok:

node/hyperbahn/handler.js
Raynos: :+1:

Raynos:
This code works; I'll clean it up later.

PR #591, opened 2015-05-14
add retry logic

junchaowu:
@blampe @breerly @abhinav 
The updated retry logic:
user can specify number of retries, timeout for each request, and no more total timeout.

### DIFF
--- a/python/tchannel/errors.py
+++ b/python/tchannel/errors.py
@@ -83,7 +83,7 @@ class StreamingError(TChannelError):
 
 
 class NoAvailablePeerError(TChannelError):
-    """Represent Error that TChannel can't find available peer for request"""
+    """Represents a failure to find any peers for a request."""
     pass
 
 
--- a/python/tchannel/errors.py
+++ b/python/tchannel/errors.py
@@ -83,7 +83,7 @@ class StreamingError(TChannelError):
 
 
 class NoAvailablePeerError(TChannelError):
-    """Represent error from maintaining the peer"""
+    """Represent Error that TChannel can't find available peer for request"""
     pass
 
 
--- a/python/tchannel/errors.py
+++ b/python/tchannel/errors.py
@@ -26,7 +26,7 @@ class TChannelError(Exception):
     pass
 
 
-class ProtocolException(TChannelError):
+class ProtocolError(TChannelError):
     """Represent a protocol-level exception"""
     __slots__ = (
         'code',
@@ -79,6 +79,11 @@ class InvalidChecksumError(TChannelError):
 
 class StreamingError(TChannelError):
     """Represent Streaming Message Exception"""
+    pass
+
+
+class NoAvailablePeerError(TChannelError):
+    """Represent error from maintaining the peer"""
     pass
 
 
--- a/python/tchannel/glossary.py
+++ b/python/tchannel/glossary.py
@@ -25,3 +25,5 @@ MAX_MESSAGE_ID = 0xfffffffe
 
 # CallRequestMessage uses it as the default TTL value for the message.
 DEFAULT_TTL = 1000  # ms
+MAX_ATTEMPT_TIMES = 3
+RETRY_DELAY = 0.3  # 300 ms
--- a/python/tchannel/theader.py
+++ b/python/tchannel/theader.py
@@ -1,0 +1,39 @@
+# Copyright (c) 2015 Uber Technologies, Inc.
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+
+
+class RetryType(object):
+    NEVER = 'n'
+    CONNECTION_ERROR = 'c'
+    TIMEOUT = 't'
+    CONNECTION_ERROR_AND_TIMEOUT = 'ct'
+    TIMEOUT_AND_CONNECTION_ERROR = 'tc'
+    DEFAULT = 'c'
+
+
+class ArgSchemeType(object):
+    RAW = 'raw'
+    JSON = 'json'
+    HTTP = 'http'
+    THRIFT = 'thrift'
+    DEFAULT = 'raw'
+
+
+transport_headers = ['as', 're']
--- a/python/tchannel/tornado/peer.py
+++ b/python/tchannel/tornado/peer.py
@@ -31,8 +31,6 @@ from ..errors import NoAvailablePeerError
 from ..errors import ProtocolError
 from ..errors import TimeoutError
 from ..handler import CallableRequestHandler
-from ..messages import ErrorCode
-from ..messages.common import StreamState
 from ..transport_header import ArgSchemeType
 from ..transport_header import RetryType
 from ..zipkin.annotation import Endpoint
@@ -533,7 +531,7 @@ class PeerClientOperation(object):
         # black list to record all used peers, so they aren't chosen again.
         blacklist = set()
         # only retry on non-stream request
-        if is_streaming_request(original_request):
+        if original_request.is_streaming_request():
             connection = yield self.peer.connect()
             message_id = connection.next_message_id()
             original_request.id = message_id
@@ -542,7 +540,7 @@ class PeerClientOperation(object):
             req = original_request
             # mac number of times to retry 5.
             peer = self.peer
-            for i in range(retry_times):
+            for num_of_retry in range(retry_times):
                 try:
                     req = original_request.clone()
                     response = yield self._retry_send(
@@ -550,10 +548,10 @@ class PeerClientOperation(object):
                     break
                 except (ProtocolError, TimeoutError) as e:
                     req.set_exception(e)
-                    if not should_retry_on_error(req, e) or self._hostport:
+                    if not req.should_retry_on_error(e) or self._hostport:
                         raise
 
-                    if i != retry_times - 1:
+                    if num_of_retry != retry_times - 1:
                         # delay further retry
                         yield gen.sleep(retry_delay)
                     else:
@@ -582,37 +580,6 @@ class PeerClientOperation(object):
             response = yield response_future
 
         raise gen.Return(response)
-
-
-def should_retry_on_error(request, error):
-    """rules for retry"""
-
-    retry_flag = request.headers.get('re', RetryType.DEFAULT)
-
-    if retry_flag == RetryType.NEVER:
-        return False
-
-    if error.code in [ErrorCode.bad_request, ErrorCode.cancelled]:
-        return False
-    elif error.code in [ErrorCode.busy, ErrorCode.declined]:
-        return True
-    elif error.code is ErrorCode.timeout:
-        return retry_flag is not RetryType.CONNECTION_ERROR
-    elif error.code in [ErrorCode.network_error, ErrorCode.unexpected]:
-        return retry_flag is not RetryType.TIMEOUT
-    else:
-        log.error("unknown error type in request retry!")
-        return False
-
-
-def is_streaming_request(request):
-    """check request is stream request or not"""
-    arg2 = request.argstreams[1]
-    arg3 = request.argstreams[2]
-    return not ((isinstance(arg2, InMemStream) and
-                isinstance(arg3, InMemStream) and
-                arg2.state == StreamState.completed and
-                arg3.state == StreamState.completed))
 
 
 def maybe_stream(s):
--- a/python/tchannel/tornado/peer.py
+++ b/python/tchannel/tornado/peer.py
@@ -27,6 +27,10 @@ from random import random
 
 from tornado import gen
 
+from tchannel.glossary import DEFAULT_TTL
+from tchannel.glossary import MAX_ATTEMPT_TIMES
+from tchannel.glossary import RETRY_DELAY
+
 from ..errors import NoAvailablePeerError
 from ..errors import ProtocolError
 from ..errors import TimeoutError
@@ -462,9 +466,9 @@ class PeerClientOperation(object):
     def send(self, arg1, arg2, arg3,
              traceflag=False,
              headers=None,
-             retry_times=Request.MAX_RETRY_TIMES,
-             timeout_per_req=Request.TIMEOUT,
-             retry_delay=Request.RETRY_DELAY):
+             attempt_times=MAX_ATTEMPT_TIMES,
+             ttl=DEFAULT_TTL,
+             retry_delay=RETRY_DELAY):
         """Make a request to the Peer.
 
         :param arg1:
@@ -480,12 +484,12 @@ class PeerClientOperation(object):
             Flag is for tracing.
         :param headers:
             Headers will be put int he message as protocol header.
-        :param retry_times:
-            Number of times to retry.
-        :param timeout_per_req:
-            Timeout for each request.
+        :param attempt_times:
+            Number of times to attempts.
+        :param ttl:
+            Timeout for each request (ms).
         :param retry_delay:
-            Delay between each retry.
+            Delay between each retry (s).
         :return:
             Future that contains the response from the peer. If None, an empty
             stream is used.
@@ -519,6 +523,7 @@ class PeerClientOperation(object):
             argstreams=[InMemStream(endpoint), arg2, arg3],
             id=connection.next_message_id(),
             headers=headers,
+            ttl=ttl,
             tracing=Trace(
                 name=endpoint,
                 trace_id=trace_id,
@@ -534,22 +539,22 @@ class PeerClientOperation(object):
         # black list to record all used peers, so they aren't chosen again.
         blacklist = set()
         # only retry on non-stream request
-        if request.is_streaming_request:
-            retry_times = 1
-            timeout_per_req = 0
+        if request.is_streaming_request or self._hostport:
+            attempt_times = 1
+            request.ttl = 0
 
         # mac number of times to retry 3.
-        for num_of_retry in range(retry_times):
+        for num_of_retry in range(attempt_times):
             try:
-                response = yield self._retry_send(
-                    connection, request, timeout_per_req)
+                response = yield self._send(
+                    connection, request)
                 break
             except (ProtocolError, TimeoutError) as e:
                 request.set_exception(e)
-                if not request.should_retry_on_error(e) or self._hostport:
+                if not request.should_retry_on_error(e):
                     raise
 
-                if num_of_retry != retry_times - 1:
+                if num_of_retry != attempt_times - 1:
                     # delay further retry
                     yield gen.sleep(retry_delay)
                 else:
@@ -572,9 +577,9 @@ class PeerClientOperation(object):
         raise gen.Return(response)
 
     @gen.coroutine
-    def _retry_send(self, connection, req, timeout_per_req):
+    def _send(self, connection, req):
         response_future = connection.send_request(req)
-        with timeout(response_future, timeout_per_req):
+        with timeout(response_future, req.ttl):
             response = yield response_future
 
         raise gen.Return(response)
--- a/python/tchannel/tornado/peer.py
+++ b/python/tchannel/tornado/peer.py
@@ -546,7 +546,7 @@ class PeerClientOperation(object):
             request.ttl = 0
 
         # mac number of times to retry 3.
-        for num_of_retry in range(attempt_times):
+        for num_of_attempt in range(attempt_times):
             try:
                 response = yield self._send(
                     connection, request)
@@ -556,7 +556,7 @@ class PeerClientOperation(object):
                 if not request.should_retry_on_error(e):
                     raise
 
-                if num_of_retry != attempt_times - 1:
+                if num_of_attempt != attempt_times - 1:
                     # delay further retry
                     yield gen.sleep(retry_delay / 1000)
                 else:
--- a/python/tchannel/tornado/peer.py
+++ b/python/tchannel/tornado/peer.py
@@ -558,7 +558,7 @@ class PeerClientOperation(object):
 
                 if num_of_attempt != attempt_times - 1:
                     # delay further retry
-                    yield gen.sleep(retry_delay / 1000)
+                    yield gen.sleep(retry_delay / 1000.0)
                 else:
                     raise
 
--- a/python/tchannel/tornado/peer.py
+++ b/python/tchannel/tornado/peer.py
@@ -241,7 +241,6 @@ class Peer(object):
         'state',
         'host',
         'port',
-        'peer_group',
 
         '_out_conns',
         '_in_conns',
@@ -424,7 +423,10 @@ class PeerClientOperation(object):
     """Encapsulates client operations that can be performed against a peer."""
 
     def __init__(self, peer, service, arg_scheme=None,
-                 retry=None, parent_tracing=None, **kwargs):
+                 retry=None,
+                 parent_tracing=None,
+                 hostport=None,
+                 score_threshold=None):
         """Initialize a new PeerClientOperation.
 
         :param peer:
@@ -452,8 +454,8 @@ class PeerClientOperation(object):
 
         # keep all arguments for retry purpose.
         # not retry if hostport is set.
-        self._hostport = kwargs.get('hostport', None)
-        self._score_threshold = kwargs.get('score_threshold', None)
+        self._hostport = hostport
+        self._score_threshold = score_threshold
         # service name is not stored in peer because the same peer may be
         # used to call multiple services if it's being used for request
         # forwarding
--- a/python/tchannel/tornado/peer.py
+++ b/python/tchannel/tornado/peer.py
@@ -21,13 +21,22 @@
 from __future__ import absolute_import
 
 import logging
+import time
 from collections import deque
 from itertools import chain
 from random import random
 
 from tornado import gen
 
+from ..errors import NoAvailablePeerError
+from ..errors import ProtocolError
+from ..errors import TimeoutError
 from ..handler import CallableRequestHandler
+from ..messages import ErrorCode
+from ..messages.common import StreamState
+from ..theader import ArgSchemeType
+from ..theader import RetryType
+from ..theader import transport_headers
 from ..zipkin.annotation import Endpoint
 from ..zipkin.trace import Trace
 from .connection import StreamConnection
@@ -35,6 +44,7 @@ from .request import Request
 from .stream import InMemStream
 from .stream import Stream
 from .stream import read_full
+from .timeout import timeout
 
 try:
     # included in Tornado 4.2
@@ -169,13 +179,20 @@ class PeerGroup(object):
         :param service_threshold:
             If ``hostport`` was not specified, this specifies the score
             threshold at or below which peers will be ignored.
+        :param blacklist:
+            Peers on the blacklist won't be chosen.
         """
-        return self.choose(
-            hostport=kwargs.pop('hostport', None),
-            score_threshold=kwargs.pop('score_threshold', None),
-        ).request(**kwargs)
+        peer = self.choose(
+            hostport=kwargs.get('hostport', None),
+            score_threshold=kwargs.get('score_threshold', None),
+            blacklist=kwargs.get('blacklist', None),
+        )
+        if peer:
+            return peer.request(**kwargs)
+        else:
+            raise NoAvailablePeerError("Can't find available peer.")
 
-    def choose(self, hostport=None, score_threshold=None):
+    def choose(self, hostport=None, score_threshold=None, blacklist=None):
         """Choose a Peer that matches the given criteria.
 
         The Peer with the highest score will be chosen.
@@ -188,10 +205,14 @@ class PeerGroup(object):
             If specified, Peers with a score equal to or below this will be
             ignored. Defaults to the value specified when the PeerGroup was
             initialized.
+        :param blacklist:
+            Peers on the blacklist won't be chosen.
         :returns:
             A Peer that matches all the requested criteria or None if no such
             Peer was found.
         """
+
+        blacklist = blacklist or set()
         if hostport:
             return self.get(hostport)
 
@@ -199,7 +220,9 @@ class PeerGroup(object):
         chosen_peer = None
         chosen_score = 0
 
-        for host in self.hosts:
+        hosts = self._peers.viewkeys() - blacklist
+
+        for host in hosts:
             peer = self.get(host)
             score = peer.state.score()
 
@@ -221,6 +244,7 @@ class Peer(object):
         'state',
         'host',
         'port',
+        'peer_group',
 
         '_out_conns',
         '_in_conns',
@@ -236,7 +260,7 @@ class Peer(object):
         """Initialize a Peer
 
         :param tchannel:
-            TChannel through which requests will be made
+            TChannel through which requests will be made.
         :param hostport:
             Host-port this Peer is for.
         :param state:
@@ -249,7 +273,6 @@ class Peer(object):
 
         self.tchannel = tchannel
         self.state = state
-
         self.host, port = hostport.rsplit(':', 1)
         self.port = int(port)
 
@@ -403,7 +426,8 @@ class PeerHealthyState(PeerState):
 class PeerClientOperation(object):
     """Encapsulates client operations that can be performed against a peer."""
 
-    def __init__(self, peer, service, **kwargs):
+    def __init__(self, peer, service, arg_scheme=None,
+                 retry=None, parent_tracing=None, **kwargs):
         """Initialize a new PeerClientOperation.
 
         :param peer:
@@ -411,13 +435,28 @@ class PeerClientOperation(object):
         :param service:
             Name of the service being called through this peer. Defaults to
             an empty string.
+        :param arg_scheme
+            arg scheme type
+        :param retry
+            retry type
+        :param parent_tracing
+            tracing span from parent request
         """
         assert peer, "peer must not be None"
         service = service or ''
 
         self.peer = peer
         self.service = service
-        self.parent_tracing = kwargs.get('parent_tracing', None)
+        self.parent_tracing = parent_tracing
+        self.headers = {
+            'as': arg_scheme or ArgSchemeType.DEFAULT,
+            're': retry or RetryType.DEFAULT,
+        }
+
+        # keep all arguments for retry purpose.
+        # not retry if hostport is set.
+        self._hostport = kwargs.get('hostport', None)
+        self._score_threshold = kwargs.get('score_threshold', None)
         # service name is not stored in peer because the same peer may be
         # used to call multiple services if it's being used for request
         # forwarding
@@ -443,6 +482,7 @@ class PeerClientOperation(object):
             Future that contains the response from the peer. If None, an empty
             stream is used.
         """
+
         arg1, arg2, arg3 = (
             maybe_stream(arg1), maybe_stream(arg2), maybe_stream(arg3)
         )
@@ -458,33 +498,113 @@ class PeerClientOperation(object):
             parent_span_id = None
             trace_id = None
 
-        # set 'as' to 'raw if 'as' is empty
+        # set default transport headers
         headers = headers or {}
-        if 'as' not in headers:
-            headers['as'] = 'raw'
+        real_header = {}
+        for h in transport_headers:
+            real_header[h] = headers.get(h, None) or self.headers.get(h, None)
 
-        connection = yield self.peer.connect()
-        message_id = connection.next_message_id()
-        response = yield connection.send_request(
-            Request(
-                service=self.service,
-                argstreams=[InMemStream(endpoint), arg2, arg3],
-                id=message_id,
-                headers=headers,
-                tracing=Trace(
-                    name=endpoint,
-                    trace_id=trace_id,
-                    parent_span_id=parent_span_id,
-                    endpoint=Endpoint(self.peer.host,
-                                      self.peer.port,
-                                      self.service),
-                    traceflags=traceflag,
-                )
+        original_request = Request(
+            service=self.service,
+            argstreams=[InMemStream(endpoint), arg2, arg3],
+            id=0,
+            headers=real_header,
+            tracing=Trace(
+                name=endpoint,
+                trace_id=trace_id,
+                parent_span_id=parent_span_id,
+                endpoint=Endpoint(self.peer.host,
+                                  self.peer.port,
+                                  self.service),
+                traceflags=traceflag,
             )
         )
 
+        response = None
+
+        # black list to record all used peers, so they aren't chosen again.
+        blacklist = set()
+        # only retry on non-stream request
+        if is_streaming_request(arg2, arg3):
+            connection = yield self.peer.connect()
+            message_id = connection.next_message_id()
+            original_request.id = message_id
+            response = yield connection.send_request(original_request)
+        else:
+            start_time = time.time()
+            req = original_request
+            try:
+                peer = self.peer
+                # mac number of times to retry 5.
+                for i in range(Request.MAX_RETRY_TIMES):
+                    try:
+                        req = original_request.clone()
+                        response = yield self.retry_send(
+                            peer, req, start_time)
+                        break
+                    except ProtocolError as e:
+                        if not should_retry_on_error(req, e) or self._hostport:
+                            raise
+
+                        blacklist.add(peer.hostport)
+
+                        # find new peer
+                        peer = self.peer.tchannel.peers.choose(
+                            hostport=self._hostport,
+                            score_threshold=self._score_threshold,
+                            blacklist=blacklist,
+                        )
+                        if not peer or i == Request.MAX_RETRY_TIMES - 1:
+                            raise
+            except TimeoutError:
+                req.set_exception(TimeoutError)
+                raise
+
         log.debug("Got response %s", response)
         raise gen.Return(response)
+
+    @gen.coroutine
+    def retry_send(self, peer, req, start_time):
+        connection = yield peer.connect()
+        req.id = connection.next_message_id()
+        time_left = Request.TIMEOUT - (
+            time.time() - start_time)
+
+        # request timeout 5s
+        response_future = connection.send_request(req)
+        with timeout(response_future, time_left):
+            response = yield response_future
+
+        raise gen.Return(response)
+
+
+def should_retry_on_error(request, error):
+    """rules for retry"""
+
+    retry_flag = request.headers['re']
+
+    if retry_flag == RetryType.NEVER:
+        return False
+
+    if error.code in [ErrorCode.bad_request, ErrorCode.cancelled]:
+        return False
+    elif error.code in [ErrorCode.busy, ErrorCode.declined]:
+        return True
+    elif error.code is ErrorCode.timeout:
+        return retry_flag is not RetryType.CONNECTION_ERROR
+    elif error.code in [ErrorCode.network_error, ErrorCode.unexpected]:
+        return retry_flag is not RetryType.TIMEOUT
+    else:
+        log.error("unknown error type in request retry!")
+        return False
+
+
+def is_streaming_request(arg2, arg3):
+    """check request is stream request or not"""
+    return not ((isinstance(arg2, InMemStream) and
+                isinstance(arg3, InMemStream) and
+                arg2.state == StreamState.completed and
+                arg3.state == StreamState.completed))
 
 
 def maybe_stream(s):
--- a/python/tchannel/tornado/peer.py
+++ b/python/tchannel/tornado/peer.py
@@ -21,7 +21,6 @@
 from __future__ import absolute_import
 
 import logging
-import time
 from collections import deque
 from itertools import chain
 from random import random
@@ -34,9 +33,8 @@ from ..errors import TimeoutError
 from ..handler import CallableRequestHandler
 from ..messages import ErrorCode
 from ..messages.common import StreamState
-from ..theader import ArgSchemeType
-from ..theader import RetryType
-from ..theader import transport_headers
+from ..transport_header import ArgSchemeType
+from ..transport_header import RetryType
 from ..zipkin.annotation import Endpoint
 from ..zipkin.trace import Trace
 from .connection import StreamConnection
@@ -219,7 +217,6 @@ class PeerGroup(object):
         score_threshold = score_threshold or self._score_threshold or 0
         chosen_peer = None
         chosen_score = 0
-
         hosts = self._peers.viewkeys() - blacklist
 
         for host in hosts:
@@ -462,7 +459,12 @@ class PeerClientOperation(object):
         # forwarding
 
     @gen.coroutine
-    def send(self, arg1, arg2, arg3, traceflag=False, headers=None):
+    def send(self, arg1, arg2, arg3,
+             traceflag=False,
+             headers=None,
+             retry_times=Request.MAX_RETRY_TIMES,
+             timeout_per_req=Request.TIMEOUT,
+             retry_delay=Request.RETRY_DELAY):
         """Make a request to the Peer.
 
         :param arg1:
@@ -478,6 +480,12 @@ class PeerClientOperation(object):
             Flag is for tracing.
         :param headers:
             Headers will be put int he message as protocol header.
+        :param retry_times:
+            Number of times to retry.
+        :param timeout_per_req:
+            Timeout for each request.
+        :param retry_delay:
+            Delay between each retry.
         :return:
             Future that contains the response from the peer. If None, an empty
             stream is used.
@@ -500,15 +508,14 @@ class PeerClientOperation(object):
 
         # set default transport headers
         headers = headers or {}
-        real_header = {}
-        for h in transport_headers:
-            real_header[h] = headers.get(h, None) or self.headers.get(h, None)
+        for k, v in self.headers.iteritems():
+            headers.setdefault(k, v)
 
         original_request = Request(
             service=self.service,
             argstreams=[InMemStream(endpoint), arg2, arg3],
             id=0,
-            headers=real_header,
+            headers=headers,
             tracing=Trace(
                 name=endpoint,
                 trace_id=trace_id,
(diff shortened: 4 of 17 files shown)


breerly:
Is this for client & server impl?

junchaowu:
python side is just client

python/tchannel/tornado/peer.py
```diff
@@ -169,13 +179,22 @@ def request(self, **kwargs):
         :param service_threshold:
             If ``hostport`` was not specified, this specifies the score
             threshold at or below which peers will be ignored.
+        :param black_list:
+            Peers on the black_list won't be chosen.
         """
-        return self.choose(
+        kwargs['peer_group'] = self
```
blampe: `**kwargs` is something we should be actively trying to avoid, especially for such a user-facing method.

blampe: Additionally it's generally a bad idea to pass around `self`. This easily leaks memory and blurs class responsibilities.

python/tchannel/tornado/peer.py
```diff
@@ -169,13 +179,22 @@ def request(self, **kwargs):
...
+            black_list=kwargs.pop('black_list', None),
+        )
 
-    def choose(self, hostport=None, score_threshold=None):
+        if peer:
+            return peer.request(**kwargs)
+        else:
+            return None
+
+    def choose(self, hostport=None, score_threshold=None, black_list=None):
```
blampe: `blacklist`

python/tchannel/tornado/peer.py
```diff
@@ -188,11 +207,15 @@ def choose(self, hostport=None, score_threshold=None):
...
+        :param black_list:
+            Peers on the black_list won't be chosen.
         :returns:
             A Peer that matches all the requested criteria or None if no such
             Peer was found.
         """
-        if hostport:
+
+        black_list = black_list or []
+        if hostport and not self.get(hostport) in black_list:
```
blampe: We're doing two `gets` here, does it make more sense for the blacklist to be a list of hostports instead of `Peers`?

python/tchannel/tornado/peer.py
```diff
@@ -418,6 +441,9 @@ def __init__(self, peer, service, **kwargs):
         self.peer = peer
         self.service = service
         self.parent_tracing = kwargs.get('parent_tracing', None)
+        self.peer_group = kwargs['peer_group']
+        kwargs.pop('peer_group', None)
+        self.kwargs = kwargs
```
blampe: Nooooo!

python/tchannel/tornado/peer.py
```diff
@@ -451,6 +478,13 @@ def send(self, arg1, arg2, arg3, traceflag=False, headers=None):
         arg1.close()
         endpoint = yield read_full(arg1)
 
+        is_stream = False
+        if (isinstance(arg2, InMemStream) and
+                isinstance(arg3, InMemStream) and
+                arg2.state == StreamState.completed and
+                arg3.state == StreamState.completed):
+            is_stream = False
```
blampe: What is going on here? The condition is really complicated and doesn't actually change `is_stream`.

python/tchannel/tornado/peer.py
```diff
@@ -169,13 +179,22 @@ def request(self, **kwargs):
...
             score_threshold=kwargs.pop('score_threshold', None),
-        ).request(**kwargs)
+            black_list=kwargs.pop('black_list', None),
+        )
 
-    def choose(self, hostport=None, score_threshold=None):
+        if peer:
+            return peer.request(**kwargs)
+        else:
+            return None
```
abhinav: Also, no matching peer is a major issue. We can't simply return None here. Consumers will be chaining a `.send` call after it so they'll get an unexplainable None-related error in their call. You should be throwing a very specific exception if no peer was found.

python/tchannel/tornado/peer.py
```diff
@@ -188,11 +207,15 @@ def choose(self, hostport=None, score_threshold=None):
...
             initialized.
+        :param black_list:
+            Peers on the black_list won't be chosen.
         :returns:
             A Peer that matches all the requested criteria or None if no such
             Peer was found.
         """
-        if hostport:
+
+        black_list = black_list or []
```
abhinav: Convert blacklist into a set here

python/tchannel/tornado/peer.py
```diff
@@ -458,35 +492,95 @@ def send(self, arg1, arg2, arg3, traceflag=False, headers=None):
...
-                tracing=Trace(
-                    name=endpoint,
-                    trace_id=trace_id,
-                    parent_span_id=parent_span_id,
-                    endpoint=Endpoint(self.peer.host,
-                                      self.peer.port,
-                                      self.service),
-                    traceflags=traceflag,
-                )
+        headers[THeaders.ARG_SCHEME] = headers.get(
```
blampe: `headers.setdefault(THeaders.ARG_SCHEME, ArgSchemeFlag.raw)`

python/tchannel/tornado/peer.py
```diff
@@ -458,35 +492,95 @@ def send(self, arg1, arg2, arg3, traceflag=False, headers=None):
...
+            THeaders.ARG_SCHEME,
+            ArgSchemeFlag.raw,
+        )
+        headers[THeaders.RETRY] = headers.get(
```
blampe: `setdefault`

python/tchannel/tornado/peer.py
```diff
@@ -458,35 +492,95 @@ def send(self, arg1, arg2, arg3, traceflag=False, headers=None):
...
+            THeaders.RETRY,
+            RetryFlag.default,
+        )
+
+        original_request = Request(
+            service=self.service,
+            argstreams=[InMemStream(endpoint), arg2, arg3],
+            id=0,
```
blampe: `id=0` whaaat?

python/tchannel/tornado/peer.py
```diff
@@ -458,35 +492,95 @@ def send(self, arg1, arg2, arg3, traceflag=False, headers=None):
...
+                            raise
+                        black_list.append(peer)
+                        self.kwargs['black_list'] = black_list
+                        peer = self.peer_group.choose(**self.kwargs)
+                        if not peer or i == Request.MAX_RETRY_TIMES - 1:
+                            raise
+            except TimeoutError:
+                req.set_exception(TimeoutError)
+                raise
+
```
blampe: This method is almost 100 lines...

python/tchannel/tornado/request.py
```diff
@@ -54,16 +60,37 @@ def __init__(
...
         self.checksum = checksum
         self.id = id
         self.headers = headers or {}
         self.state = StreamState.init
         self.endpoint = ""
-        self.header = None
-        self.body = None
         self.scheme = scheme
 
+    def clone(self):
```
blampe: Why not just re-use the existing request but give it a new message id?

python/tchannel/tornado/peer.py
breerly: +1 - start by pulling the functionality inside the for loop into its own function.

python/tchannel/tornado/peer.py
```diff
@@ -458,35 +492,95 @@ def send(self, arg1, arg2, arg3, traceflag=False, headers=None):
...
+    if retry_flag == RetryFlag.never:
+        return False
+
+    if error.code in [ErrorCode.bad_request, ErrorCode.cancelled]:
+        return False
+    elif error.code in [ErrorCode.busy, ErrorCode.declined]:
+        return True
+    elif error.code is ErrorCode.timeout:
+        return retry_flag is not RetryFlag.connection_error
```
blampe: "Flag" implies that I can combine these things. If we can only check for equality then these aren't actually flags and we should use a better name.

python/tchannel/tornado/request.py
```diff
@@ -35,6 +39,8 @@ class Request(object):
     Request class is used to represent the CallRequestMessage at User's level.
     This is going to hide the protocol level message information.
     """
+    MAX_RETRY_TIMES = 5
+    TIMEOUT = 5  # 5 second
```
blampe: Need to be able to pass these in.

python/tests/integration/server_manager.py
```diff
@@ -102,7 +131,11 @@ def __init__(self, port, timeout=None, dispatcher=None):
...
-        tornado.ioloop.IOLoop.current().start()
+        try:
+            tornado.ioloop.IOLoop.instance().start()
+            print "stop"
+        except:
+            pass
 
     def shutdown(self):
-        tornado.ioloop.IOLoop.current().stop()
+        tornado.ioloop.IOLoop.instance().stop()
```
blampe: We should always be using `current` over `instance`, why did you need to change these?

python/tests/integration/server_manager.py
```diff
@@ -102,7 +131,11 @@ def __init__(self, port, timeout=None, dispatcher=None):
     def serve(self):
         self.tchannel.host(self.dispatcher).listen()
         self.ready = True
-        tornado.ioloop.IOLoop.current().start()
+        try:
+            tornado.ioloop.IOLoop.instance().start()
+            print "stop"
+        except:
```
blampe: Debugging?

python/tests/integration/test_retry.py
```diff
@@ -0,0 +1,220 @@
...
+        tchannel_server.host(dispatcher).listen()
+        tchannel.peers.get(tchannel_server.hostport)
+
+    hostport = tchannel_server.hostport
+    with (
+        patch(
+            'tchannel.tornado.peer.should_retry_on_error',
+            autospec=True)
+    ) as mock_should_retry_on_error:
+        Request.TIMEOUT = 0.03
```
breerly: This strikes me as odd - changing a class constant seems like something that should never happen.

python/tchannel/glossary.py
```diff
@@ -22,3 +22,24 @@
...
+    connection_error_and_timeout = 'ct'
+    timeout_and_connection_error = 'tc'
+    default = 'c'
+
+
+class ArgSchemeFlag(object):
+    raw = 'raw'
+    json = 'json'
+    http = 'http'
+    thrift = 'thrift'
```
breerly: These should be class constants, since they never change. Please capitalize.

python/tchannel/errors.py
```diff
@@ -82,6 +82,11 @@ class StreamingError(TChannelError):
     pass
 
 
+class PeerError(TChannelError):
+    """Represent error from maintaining the peer"""
```
abhinav: What?

python/tchannel/scheme.py
```diff
@@ -39,23 +39,25 @@ def serialize_body(obj)
...
     def deserialize_body(self, obj):
-        raise NotImplementedError()
+        return obj
+
+    def serialize_body(self, obj):
+        return obj
 
 
-class JsonArgScheme(ArgScheme):
+class JsonArgScheme(RawArgScheme):
```
abhinav: Don't do this. There's no reason for JsonArgScheme to inherit RawArgScheme.

python/tchannel/thead.py
```diff
@@ -0,0 +1,39 @@
+# Copyright (c) 2015 Uber Technologies, Inc.
```
abhinav: Not the best file name.

python/tchannel/thead.py
```diff
@@ -0,0 +1,39 @@
...
+    timeout_and_connection_error = 'tc'
+    default = 'c'
+
+
+class ArgScheme(object):
+    raw = 'raw'
+    json = 'json'
+    http = 'http'
+    thrift = 'thrift'
+    default = 'raw'
```
abhinav: `default` seems unnecessary

python/tchannel/thead.py
```diff
@@ -0,0 +1,39 @@
...
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+
+
+class Retry(object):
```
abhinav: Use a more descriptive name and/or document what this represents and what each of the values means.

python/tchannel/thrift/scheme.py
```diff
@@ -28,7 +28,7 @@
 from .. import scheme
 
 
-class ThriftArgScheme(scheme.ArgScheme):
+class ThriftArgScheme(scheme.RawArgScheme):
```
abhinav: Again, you shouldn't be inheriting RawArgScheme

python/tchannel/tornado/peer.py
junchaowu: That's the original_request. All the actual request is assigned later.

python/tchannel/tornado/request.py
junchaowu: The stream will be consumed.

python/tchannel/tornado/request.py
junchaowu: Should this be instance level or request level or send level?

python/tests/integration/test_retry.py
junchaowu: Here just to reduce the timeout time for the end-to-end test.

python/tchannel/thead.py
junchaowu: tchannel => theader?

python/tchannel/thead.py
junchaowu: if in the future, we change default for any of header's key. We can just change the default

python/tchannel/thrift/scheme.py
junchaowu: But I think it is also make sense. You can image the rawargscheme is the raw base scheme. Any json or thrift arg scheme is a variation of raw arg scheme becuase raw arg scheme basically does nothing.

python/tchannel/tornado/peer.py
```diff
@@ -188,18 +205,23 @@ def choose(self, hostport=None, score_threshold=None):
...
+        blacklist = blacklist or set()
+        if hostport and hostport not in blacklist:
             return self.get(hostport)
 
         score_threshold = score_threshold or self._score_threshold or 0
         chosen_peer = None
         chosen_score = 0
 
-        for host in self.hosts:
+        hosts = set(self.hosts) - blacklist
```
abhinav: [dict.viewkeys](https://docs.python.org/2/library/stdtypes.html#dict.viewkeys) give you a set-like object that contains the keys of the dictionary. Consider using that.

python/tchannel/tornado/peer.py
```diff
@@ -418,6 +440,14 @@ def __init__(self, peer, service, **kwargs):
         self.peer = peer
         self.service = service
         self.parent_tracing = kwargs.get('parent_tracing', None)
+        self.headers = {
+            'as': kwargs.get('arg_scheme', None) or ArgScheme.default,
+            're': kwargs.get('retry', None) or Retry.default,
```
abhinav: Just add these (incl. parent_tracing) as construct arguments with default values of `None`

python/tchannel/tornado/peer.py
```diff
@@ -418,6 +440,14 @@ def __init__(self, peer, service, **kwargs):
...
+        }
+
+        # keep all arguments for retry purpose
+        self.r_hostport = kwargs.get('hostport', None)
+        self.r_score_threshold = kwargs.get('score_threshold', None)
```
abhinav: Use `remote_hostport`

There's only one score_threshold. Remove `r_` prefix.

python/tchannel/tornado/peer.py
abhinav: If this is the point where we actually create the request object, it's better to assign a message ID right away if you haver access to `connection.next_message_id`. Ad-hoc mutation of objects should be avoided.

python/tchannel/tornado/peer.py
```diff
@@ -458,35 +489,110 @@ def send(self, arg1, arg2, arg3, traceflag=False, headers=None):
...
-                    trace_id=trace_id,
-                    parent_span_id=parent_span_id,
-                    endpoint=Endpoint(self.peer.host,
-                                      self.peer.port,
-                                      self.service),
-                    traceflags=traceflag,
-                )
+        real_header = {}
+        for h in transport_headers:
+            real_header[h] = headers.get(h, None) or self.headers.get(h, None)
```
abhinav: Use [`dict.setdefault`](https://docs.python.org/2/library/stdtypes.html#dict.setdefault) on `headers` for all items defined in `self.headers`. This constant `transport_headers` is mostly unnecessary.

python/tchannel/tornado/peer.py
```diff
@@ -458,35 +489,110 @@ def send(self, arg1, arg2, arg3, traceflag=False, headers=None):
...
+
+
+def is_stream(arg2, arg3):
+    """check request is stream request or not"""
+    if (isinstance(arg2, InMemStream) and
+            isinstance(arg3, InMemStream) and
+            arg2.state == StreamState.completed and
+            arg3.state == StreamState.completed):
+        return False
+    else:
```
abhinav: What? We don't support other kinds of streams?

python/tchannel/tornado/peer.py
junchaowu: No. I want to restrict that only transport_headers's key can exist in the dict. If user pass something else, we can't allow that.

python/tchannel/scheme.py
breerly: Agreed. Either the interface is the same by accident, aka duck-typing, or it is the same on purpose. I prefer explicit, so I think you should use an `abc`. The classes that take an `ArgScheme` can enforce that the `abc` was implemented.

python/tchannel/tornado/peer.py
junchaowu: Here is to check if the request contain streaming arg stream. Since retry doesn't support streaming request.

python/tchannel/tornado/peer.py
```diff
@@ -188,18 +205,23 @@ def choose(self, hostport=None, score_threshold=None):
...
+            Peers on the blacklist won't be chosen.
         :returns:
             A Peer that matches all the requested criteria or None if no such
             Peer was found.
         """
-        if hostport:
+
+        blacklist = blacklist or set()
+        if hostport and hostport not in blacklist:
             return self.get(hostport)
```
abhinav: This is completely breaking the semantics of how the method works. If `hostport` is given, it should *always* returns that peer. Blacklist can't apply.

On the Node side, if a hostport is specified, they just don't do retries for that request. We should do the same.

python/tchannel/thead.py
breerly: these are class constants, should be capitalized.

python/tchannel/tornado/peer.py
abhinav: Yes, but what about PipeStream. You're checking only InMemStream.

python/tchannel/thead.py
```diff
@@ -0,0 +1,39 @@
...
+    never = 'n'
+    connection_error = 'c'
+    timeout = 't'
+    connection_error_and_timeout = 'ct'
+    timeout_and_connection_error = 'tc'
+    default = 'c'
```
breerly: these are class constants, should be capitalized.

python/tchannel/tornado/peer.py
abhinav: If you have a static list of keys that are allowed, then don't let the user pass in a dict and use named parameters instead.

python/tchannel/thrift/scheme.py
breerly: I disagree. No need for inheritance here. If you want to define an interface  (`abc` in python), that's a different story.

python/tchannel/tornado/peer.py
abhinav: Also, I disagree that you should restrict to those transport headers. If we add a minor transport-header to the protocol that doesn't change the behavior as far as the client library is concerned, we shouldn't need a new version of the library just to add that to the whitelist of allowed headers.

python/tchannel/tornado/peer.py
junchaowu: All pipestream I assume is streaming request.

abhinav: Your'e not checking for PipeStream. 
`is_stream(PipeStream(..), PipeStream(..))` will be False.

python/tchannel/thrift/scheme.py
breerly: @junchaowu what object benefits from this inheritance structure? You aren't calling into the parent, and `ThriftArgScheme` re-implements all the functionality of the class its extending. This feels a bit more like an interface, which I'm fine with.

python/tchannel/thead.py
breerly: Suggest moving all the argschemes into a file called `schemes.py` and moving this retry class into a class called `retry.py`.

`thead` is really confusing.

python/tchannel/tornado/peer.py
junchaowu: all the request got assigned with the id right behind this and before they call connection.send_stream

python/tchannel/scheme.py
junchaowu: I don't think base class have to always be `abc`.  Base class gives default behavior and subclass makes their own implementation

python/tchannel/thead.py
junchaowu: In that case all the 'field' in the transport header should have their own file? 
name	req	res	description
as	Y	Y	the Arg Scheme
cas	Y	N	Claim At Start
caf	Y	N	Claim At Finish
cn	Y	N	Caller Name
re	Y	N	Retry Flags
se	Y	N	Speculative Execution
fd	Y	Y	Failure Domain

I think move them in central place like 'transport header ' file would be easy to manage.

python/tchannel/thrift/scheme.py
junchaowu: You can image a custom argscheme only implements its own serialization/de-serialization  for header for example, then it doesn't need to care about the body part.

python/tchannel/tornado/peer.py
junchaowu: the opposite case is like what I said, user puts random keys into us. I'd rather to be safe for that instead of flexibility

python/tchannel/tornado/peer.py
junchaowu: what do you mean?

python/tchannel/thrift/scheme.py
breerly: You would just compose in the parts you want to reuse. It's much cleaner that way, and the object hierarchy stays flat. You should avoid inheritance unless you need a contract to do type switching (polymorphism).

python/tchannel/scheme.py
breerly: There isn't any default behavior here, AFAIK that's anticipatory.

Also, creating object hierarchies without a need for type switching is a smell. If the only reason you are making a base class is to share code, you're doing it wrong. You can just as easily compose in like functionality.

Class hierarchies make object responsibilities ambiguous because it tempts you to put anything you want to reuse into the parent. What you should be doing is defining objects with very finite responsibilities, and composing them at runtime.

python/tchannel/tornado/peer.py
junchaowu: discussed offline

python/tchannel/thead.py
junchaowu: Updated

python/tchannel/tornado/peer.py
junchaowu: Change it to not retry when hostport is specified.

python/tchannel/tornado/peer.py
abhinav: As said before: If you have a static list of allowed keys, use actual named arguments for them. "A dictionary with only these 2 keys" is not a valid argument type. Ignoring half the headers in my dictionary silently is not valid behavior.

Additionally, I don't buy the invalid headers argument. If the remote server doesn't know how to handle unexpected protocol headers, they're doing it wrong.

python/tchannel/errors.py
```diff
@@ -82,6 +82,11 @@ class StreamingError(TChannelError):
     pass
 
 
+class NoAvailablePeerError(TChannelError):
+    """Represent error from maintaining the peer"""
```
abhinav: This doc string still does not make sense

python/tchannel/theader.py
```diff
@@ -0,0 +1,39 @@
...
+# all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+
```
abhinav: Follow @breerly's suggestions regarding the file names.

python/tchannel/theader.py
```diff
@@ -0,0 +1,39 @@
...
+
+class RetryType(object):
+    NEVER = 'n'
+    CONNECTION_ERROR = 'c'
+    TIMEOUT = 't'
+    CONNECTION_ERROR_AND_TIMEOUT = 'ct'
+    TIMEOUT_AND_CONNECTION_ERROR = 'tc'
```
abhinav: documentation

python/tchannel/theader.py
```diff
@@ -0,0 +1,39 @@
...
+    DEFAULT = 'c'
```
abhinav: `DEFAULT = CONNECTION_ERROR` is better than copying the value.

python/tchannel/theader.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+class ArgSchemeType(object):
+    RAW = 'raw'
+    JSON = 'json'
+    HTTP = 'http'
+    THRIFT = 'thrift'
+    DEFAULT = 'raw'
```
abhinav: Same as above.

python/tchannel/tornado/peer.py
junchaowu: These are protocol header. I don't think it is a good idea to allow uses to set random things. Can TCP allow you set random header? not

python/tchannel/theader.py
```diff
@@ -0,0 +1,39 @@
...
+class RetryType(object):
+    NEVER = 'n'
+    CONNECTION_ERROR = 'c'
+    TIMEOUT = 't'
+    CONNECTION_ERROR_AND_TIMEOUT = 'ct'
+    TIMEOUT_AND_CONNECTION_ERROR = 'tc'
+    DEFAULT = 'c'
+
+
+class ArgSchemeType(object):
```
abhinav: Wouldn't be the worst idea to use `enum.Enum` here and in `RetryType`. You'll be able to use the string to automatically cast it into the ArgSchemeType object using `ArgSchemeType('http')` by checking `isinstance(s, basestring)` first.

Also, this seems completely disconnected from the other `ArgScheme` stuff.

python/tchannel/tornado/peer.py
```diff
@@ -411,13 +435,28 @@ def __init__(self, peer, service, **kwargs):
...
+        self.parent_tracing = parent_tracing
+        self.headers = {
+            'as': arg_scheme or ArgSchemeType.DEFAULT,
+            're': retry or RetryType.DEFAULT,
+        }
+
+        # keep all arguments for retry purpose.
+        # not retry if hostport is set.
+        self._hostport = kwargs.get('hostport', None)
+        self._score_threshold = kwargs.get('score_threshold', None)
```
abhinav: use keyword arguments with default values

python/tchannel/tornado/peer.py
abhinav: +1

python/tchannel/tornado/peer.py
```diff
@@ -458,34 +498,114 @@ def send(self, arg1, arg2, arg3, traceflag=False, headers=None):
...
+                traceflags=traceflag,
             )
         )
 
+        response = None
+
+        # black list to record all used peers, so they aren't chosen again.
+        blacklist = set()
+        # only retry on non-stream request
+        if is_streaming_request(arg2, arg3):
```
abhinav: Don't specialize `is_streaming_request` to two arguments. Make it accept only one argument and do the `and` here.

python/tchannel/tornado/peer.py
```diff
@@ -458,34 +498,114 @@ def send(self, arg1, arg2, arg3, traceflag=False, headers=None):
...
+        log.error("unknown error type in request retry!")
+        return False
+
+
+def is_streaming_request(arg2, arg3):
+    """check request is stream request or not"""
+    return not ((isinstance(arg2, InMemStream) and
+                isinstance(arg3, InMemStream) and
+                arg2.state == StreamState.completed and
+                arg3.state == StreamState.completed))
```
abhinav: redundant parentheses 

don't specialize to two arguments

python/tchannel/tornado/peer.py
```diff
@@ -458,34 +498,114 @@ def send(self, arg1, arg2, arg3, traceflag=False, headers=None):
...
+        with timeout(response_future, time_left):
+            response = yield response_future
+
+        raise gen.Return(response)
+
+
+def should_retry_on_error(request, error):
+    """rules for retry"""
+
+    retry_flag = request.headers['re']
```
abhinav: `.get` in case the header is unset

python/tchannel/tornado/stream.py
```diff
@@ -98,27 +103,34 @@ def __init__(self, buf=None, auto_close=True):
 
         :param buf: the buffer for the in memory stream
         """
-        self._stream = deque()
+        self.stream = deque()
```
abhinav: No. Consumers don't get access to the `deque`. That's an implementation detail. We've already switched out the underlying implementation before. There's no reason for consumers to need access to this. Add the `_` back.

python/tchannel/tornado/stream.py
```diff
@@ -98,27 +103,34 @@ def __init__(self, buf=None, auto_close=True):
...
         self._condition = Condition()
         self.auto_close = auto_close
 
         self.exception = None
 
+    def clone(self):
+        new_stream = self.__class__()
+        new_stream.state = self.state
+        new_stream.auto_close = self.auto_close
+        new_stream.stream = copy.deepcopy(self.stream)
```
abhinav: What's stopping you from simply doing `deque(self._stream)`? Deques are ordered iterables. Copies will be the same. They contain immutable strings. There's no reason to use `deepcopy`.

python/tchannel/tornado/peer.py
abhinav: 1. TCP doesn't have a concept of transport headers.
2. HTTP has support for arbitrary headers. Although it doesn't play well with the application/protocol headers analogy unless you add the assumption that the application headers and request are part of the HTTP request body.
3. TChannel has a concept of separate application and transport headers. I shouldn't have to update my library just to pass in a new transport header. Further, I should be able to add my own customizations based on new transport headers if I so desire. The low level API should provide a way to specify the headers.
4. I don't know how many times I need to repeat this: If there is a need in the high level API to limit the protocol headers to a specific predefined set of keys, your API should not be accepting a dictionary. It should be accepting actual separate keyword arguments. If I pass in a dictionary to an API that says "headers", I expect **all** key-value pairs from the dictionary to be present in the outgoing packet. Ignoring items in my dictionary silently is wrong behavior.

python/tchannel/tornado/peer.py
junchaowu: it matches the keyword 's default value. from choose method: def choose(self, hostport=None, score_threshold=None, blacklist=None):

abhinav: I mean, since you're consuming those arguments in this method, you should add them to the parameter list of the method instead of using `kwargs.get` for them.

python/tchannel/tornado/peer.py
junchaowu: updated

python/tchannel/errors.py
junchaowu: updated

python/tchannel/tornado/peer.py
```diff
@@ -221,6 +241,7 @@ class Peer(object):
         'state',
         'host',
         'port',
+        'peer_group',
```
abhinav: ?

python/tchannel/tornado/peer.py
blampe: We're still mutating the object here...

python/tchannel/tornado/peer.py
```diff
@@ -458,34 +508,112 @@ def send(self, arg1, arg2, arg3, traceflag=False, headers=None):
...
+                                  self.service),
+                traceflags=traceflag,
             )
         )
 
+        response = None
+        # black list to record all used peers, so they aren't chosen again.
+        blacklist = set()
+        # only retry on non-stream request
+        if is_streaming_request(original_request):
```
blampe: Most of this could probably be moved to the `Request` object.

python/tchannel/tornado/peer.py
```diff
@@ -458,34 +508,112 @@ def send(self, arg1, arg2, arg3, traceflag=False, headers=None):
...
+            connection = yield self.peer.connect()
+            message_id = connection.next_message_id()
+            original_request.id = message_id
+            response = yield connection.send_request(original_request)
+        else:
+            req = original_request
+            # mac number of times to retry 5.
+            peer = self.peer
+            for i in range(retry_times):
```
blampe: Better name than `i`.

python/tchannel/tornado/request.py
```diff
@@ -37,6 +39,9 @@ class Request(object):
     Request class is used to represent the CallRequestMessage at User's level.
     This is going to hide the protocol level message information.
     """
+    MAX_RETRY_TIMES = 3
+    TIMEOUT = 5  # 5 second
+    RETRY_DELAY = 0.3  # 300 ms
```
blampe: Uhhhh this is insane, is node doing this?

python/tchannel/tornado/request.py
```diff
@@ -37,6 +39,9 @@ class Request(object):
     Request class is used to represent the CallRequestMessage at User's level.
     This is going to hide the protocol level message information.
     """
+    MAX_RETRY_TIMES = 3
+    TIMEOUT = 5  # 5 second
```
blampe: Timeout of 1s.

python/tchannel/tornado/stream.py
```diff
@@ -107,6 +111,13 @@ def __init__(self, buf=None, auto_close=True):
 
         self.exception = None
 
+    def clone(self):
+        new_stream = self.__class__()
```
blampe: Don't do this. It looks like you want a `@classmethod`.

python/tchannel/transport_header.py
```diff
@@ -0,0 +1,40 @@
...
+class RetryType(object):
+    """ Retry Type in the protocol header
+    For details, look at protocol definition.
+        https://github.com/uber/tchannel/blob/master/docs/protocol.md
+    """
+    NEVER = 'n'
+    CONNECTION_ERROR = 'c'
+    TIMEOUT = 't'
+    CONNECTION_ERROR_AND_TIMEOUT = 'ct'
+    TIMEOUT_AND_CONNECTION_ERROR = 'tc'
```
blampe: It seems really silly that we define both of these.

python/tchannel/tornado/request.py
junchaowu: Jack just synced with me today. They are doing this.

python/tests/integration/test_retry.py
```diff
@@ -0,0 +1,216 @@
...
+from tchannel.tornado.stream import InMemStream
+
+
+@pytest.mark.gen_test(timeout=10)
+def test_retry_timeout():
+    endpoint = b'tchannelretrytest'
+
+    @tornado.gen.coroutine
+    def handler_error(request, response, proxy):
+        yield tornado.gen.sleep(2)
```
blampe: We shouldn't be writing tests that take multiple seconds, need to figure out how to reduce these timeouts for testing.

python/tests/integration/test_retry.py
```diff
@@ -0,0 +1,216 @@
...
+            ErrorCode.busy,
+            "retry",
+            response.id,
+        )
+        # stop normal response streams
+        response.set_exception(TChannelError("stop stream"))
+
+    tchannel = TChannel()
+    tchannel_server = None
+    for i in range(3):
```
blampe: What is going on here?

python/tchannel/tornado/peer.py
```diff
@@ -458,34 +508,112 @@ def send(self, arg1, arg2, arg3, traceflag=False, headers=None):
...
+
+def should_retry_on_error(request, error):
+    """rules for retry"""
+
+    retry_flag = request.headers.get('re', RetryType.DEFAULT)
+
+    if retry_flag == RetryType.NEVER:
+        return False
+
+    if error.code in [ErrorCode.bad_request, ErrorCode.cancelled]:
```
Raynos: `ErrorCode.protocol_error` ?

python/tests/integration/test_retry.py
```diff
@@ -0,0 +1,216 @@
...
+
+    hostport = tchannel_server.hostport
+    with (
+        patch(
+            'tchannel.tornado.peer.should_retry_on_error',
+            autospec=True)
+    ) as mock_should_retry_on_error:
+
+        mock_should_retry_on_error.return_value = True
+        with pytest.raises(TimeoutError):
```
blampe: It seems like this is mostly just testing our timeout decorator?

python/tests/integration/test_retry.py
junchaowu: Create couple tchannel instance and add them to one's peer group in order to mimic retry

python/tests/integration/test_retry.py
```diff
@@ -0,0 +1,216 @@
...
+            "retry",
+            response.id,
+        )
+
+        # stop normal response streams
+        response.set_exception(TChannelError("stop stream"))
+
+    tchannel = TChannel()
+    tchannel_server = None
+    for i in range(3):
```
blampe: Code duplication, do you want a fixture?

python/tests/integration/test_retry.py
```diff
@@ -0,0 +1,216 @@
...
+        )
+
+        header = yield response.get_header()
+        body = yield response.get_body()
+        assert body == "success"
+        assert header == ""
+
+
+@pytest.mark.gen_test
+@pytest.mark.parametrize('retry_flag, error_code, result', [
```
blampe: Nice!

python/tests/integration/test_retry.py
```diff
@@ -0,0 +1,216 @@
...
+    (RetryType.TIMEOUT_AND_CONNECTION_ERROR, ErrorCode.timeout, True),
+    (RetryType.TIMEOUT, ErrorCode.unexpected, False),
+    (RetryType.TIMEOUT, ErrorCode.network_error, False),
+    (RetryType.CONNECTION_ERROR, ErrorCode.network_error, True),
+    (RetryType.NEVER, ErrorCode.network_error, False),
+    (RetryType.CONNECTION_ERROR_AND_TIMEOUT, ErrorCode.cancelled, False),
+    (RetryType.TIMEOUT_AND_CONNECTION_ERROR, ErrorCode.bad_request, False),
+
+],
+    ids=lambda arg: str(len(arg))
```
blampe: Just print the tuple `lambda arg: arg`

python/tests/integration/test_retry.py
junchaowu: It won'd actually test multiply seconds. Here just needs a number that is greater than the timeout_per_request. So the client will timeout.

python/tchannel/tornado/peer.py
junchaowu: Updated

python/tests/integration/test_retry.py
junchaowu: One test is trying to test retry few times but eventually it fails. Another one is trying to test retry few times and finally find a peer that returns success.

python/tests/integration/test_retry.py
junchaowu: Fixture is not applicable here. TChannel will create instance which need gen_test decorator.

blampe: What? You can put things on the IOLoop outside of a `gen_test` method; inside that method the `IOLoop` will get started and those futures will start being executed. It's still not clear to me why you create the server multiple times, though.

junchaowu: Create separate methods for code reuse

python/tchannel/errors.py
```diff
@@ -82,6 +82,11 @@ class StreamingError(TChannelError):
     pass
 
 
+class NoAvailablePeerError(TChannelError):
+    """Represent Error that TChannel can't find available peer for request"""
```
blampe: """Represents a failure to find any peers for a request."""

python/tchannel/tornado/peer.py
```diff
@@ -169,13 +175,20 @@ def request(self, **kwargs):
...
+            Peers on the blacklist won't be chosen.
         """
-        return self.choose(
-            hostport=kwargs.pop('hostport', None),
-            score_threshold=kwargs.pop('score_threshold', None),
-        ).request(**kwargs)
+        peer = self.choose(
+            hostport=kwargs.get('hostport', None),
+            score_threshold=kwargs.get('score_threshold', None),
+            blacklist=kwargs.get('blacklist', None),
```
blampe: The `**kwargs` interface really needs to be fixed, especially for this and other methods that users will interact with regularly. Save it for a followup diff.

python/tchannel/tornado/peer.py
blampe: This area of code still needs the most work. A lot of `Request` implementation details are leaking into `PeerClientOperation`. A `Request` should know how to send itself and handle retries. The `PCO` should be able to just say `response = yield request.send()` or arguably `response = yield connection.send(request)`.

python/tchannel/tornado/peer.py
```diff
@@ -458,34 +506,81 @@ def send(self, arg1, arg2, arg3, traceflag=False, headers=None):
...
+            message_id = connection.next_message_id()
+            original_request.id = message_id
+            response = yield connection.send_request(original_request)
+        else:
+            req = original_request
+            # mac number of times to retry 5.
+            peer = self.peer
+            for num_of_retry in range(retry_times):
+                try:
+                    req = original_request.clone()
```
blampe: We're cloning the request every time.

junchaowu: only when we are doing retry.

python/tchannel/tornado/peer.py
```diff
@@ -458,34 +506,81 @@ def send(self, arg1, arg2, arg3, traceflag=False, headers=None):
...
+                                  self.service),
+                traceflags=traceflag,
             )
         )
 
+        response = None
+        # black list to record all used peers, so they aren't chosen again.
+        blacklist = set()
+        # only retry on non-stream request
+        if original_request.is_streaming_request():
```
blampe: Why is this not the same logic but with only 1 request attempt?

junchaowu: what do you mean?

python/tchannel/tornado/peer.py
```diff
@@ -458,34 +506,81 @@ def send(self, arg1, arg2, arg3, traceflag=False, headers=None):
...
+                        peer, req, timeout_per_req)
+                    break
+                except (ProtocolError, TimeoutError) as e:
+                    req.set_exception(e)
+                    if not req.should_retry_on_error(e) or self._hostport:
+                        raise
+
+                    if num_of_retry != retry_times - 1:
+                        # delay further retry
+                        yield gen.sleep(retry_delay)
```
blampe: This `sleep` is crazy, why are we doing this.

python/tchannel/tornado/peer.py
blampe: We shouldn't do it if we don't need to.

python/tchannel/tornado/peer.py
blampe: What prevents us from just setting `retry_times = 1` when `is_streaming_request() is True`?

python/tchannel/tornado/request.py
```diff
@@ -37,6 +42,9 @@ class Request(object):
     Request class is used to represent the CallRequestMessage at User's level.
     This is going to hide the protocol level message information.
     """
+    MAX_RETRY_TIMES = 3
+    TIMEOUT = 1  # 1 second
+    RETRY_DELAY = 0.3  # 300 ms
```
blampe: Need to make these configurable in a followup diff.

python/tchannel/tornado/peer.py
junchaowu: Node side they have a retry delay.

python/tchannel/tornado/peer.py
junchaowu: In the retry case, every time before send we will do clone. For streaming request, we don't know the end of stream, we can't do clone of request.

python/tchannel/tornado/peer.py
junchaowu: We don't know if we need to unless we try. So we have to clone first then send, otherwise when we should do retry but we haven't made a copy of request, it will be too late.

python/tchannel/tornado/request.py
junchaowu: They are configurable in the send method. Use can specify those attributes in the send method.

blampe: We're eventually want to be able to override them at the tchannel level.

python/tchannel/tornado/peer.py
```diff
@@ -411,19 +432,39 @@ def __init__(self, peer, service, **kwargs):
...
         # service name is not stored in peer because the same peer may be
         # used to call multiple services if it's being used for request
         # forwarding
 
     @gen.coroutine
-    def send(self, arg1, arg2, arg3, traceflag=False, headers=None):
+    def send(self, arg1, arg2, arg3,
+             traceflag=False,
+             headers=None,
+             retry_times=Request.MAX_RETRY_TIMES,
```
abhinav: This isn't number of retries. This is the number of attempts. Fix the name please.

python/tchannel/tornado/peer.py
```diff
@@ -458,34 +506,79 @@ def send(self, arg1, arg2, arg3, traceflag=False, headers=None):
...
+
+        # mac number of times to retry 3.
+        for num_of_retry in range(retry_times):
+            try:
+                response = yield self._retry_send(
+                    connection, request, timeout_per_req)
+                break
+            except (ProtocolError, TimeoutError) as e:
+                request.set_exception(e)
+                if not request.should_retry_on_error(e) or self._hostport:
```
abhinav: For `hostport` instead of checking here manually, you can do the same thing as `is_streaming_request` that you do above.

python/tchannel/tornado/peer.py
```diff
@@ -458,34 +506,79 @@ def send(self, arg1, arg2, arg3, traceflag=False, headers=None):
...
+                    raise
+                connection = yield peer.connect()
+                # roll back request
+                request.rewind(connection.next_message_id())
+
         log.debug("Got response %s", response)
         raise gen.Return(response)
 
+    @gen.coroutine
+    def _retry_send(self, connection, req, timeout_per_req):
```
abhinav: This function's implementation has nothing to do with retries. Call it something else?

python/tchannel/tornado/peer.py
```diff
@@ -458,34 +506,79 @@ def send(self, arg1, arg2, arg3, traceflag=False, headers=None):
...
+        response_future = connection.send_request(req)
+        with timeout(response_future, timeout_per_req):
+            response = yield response_future
```
abhinav: Where does the timeout get propagated as TTL?

python/tchannel/tornado/peer.py
```diff
@@ -411,19 +432,39 @@ def __init__(self, peer, service, **kwargs):
...
         # used to call multiple services if it's being used for request
         # forwarding
 
     @gen.coroutine
-    def send(self, arg1, arg2, arg3, traceflag=False, headers=None):
+    def send(self, arg1, arg2, arg3,
+             traceflag=False,
+             headers=None,
+             retry_times=Request.MAX_RETRY_TIMES,
+             timeout_per_req=Request.TIMEOUT,
```
abhinav: Just call this `timeout_millis` or `timeout_secs` depending on what the unit of time is.

python/tchannel/tornado/peer.py
```diff
@@ -439,10 +480,17 @@ def send(self, arg1, arg2, arg3, traceflag=False, headers=None):
             Flag is for tracing.
         :param headers:
             Headers will be put int he message as protocol header.
+        :param retry_times:
+            Number of times to retry.
+        :param timeout_per_req:
+            Timeout for each request.
```
abhinav: Document the unit of time

python/tchannel/tornado/request.py
abhinav: Move these constants into the glossary?

python/tchannel/tornado/request.py
```diff
@@ -120,3 +143,38 @@ def get_body_s(self):
...
         return self.argstreams[2]
+
+    def _is_streaming_request(self):
+        """check request is stream request or not"""
+        arg2 = self.argstreams[1]
+        arg3 = self.argstreams[2]
+        return not (isinstance(arg2, InMemStream) and
+                    isinstance(arg3, InMemStream) and
+                    ((arg2.auto_close is True and
+                      arg3.auto_close is True) or (
```
abhinav: `x is True`/`x == True` should be changed to just `x`

python/tchannel/tornado/request.py
```diff
@@ -56,16 +61,34 @@ def __init__(
...
+                self.argstreams[0].clone(),
+                self.argstreams[1].clone(),
+                self.argstreams[2].clone(),
+            ]
+
+    def rewind(self, id=None):
+        self.id = id
+        self.argstreams = [
+            self._copy_argstreams[0].clone(),
+            self._copy_argstreams[1].clone(),
```
abhinav: Why is `copy_argstreams` being cloned twice?

Also, the `_copy_argstreams` attribute won't be set if the request is a streaming request. This will fail with an AttributeError or something.

python/tchannel/tornado/request.py
```diff
@@ -120,3 +143,38 @@ def get_body_s(self):
         :return: the argstream of body
         """
         return self.argstreams[2]
+
+    def _is_streaming_request(self):
```
abhinav: Why not just make this a property instead of recording it separately.

python/tchannel/tornado/request.py
```diff
@@ -120,3 +143,38 @@ def get_body_s(self):
...
+        arg3 = self.argstreams[2]
+        return not (isinstance(arg2, InMemStream) and
+                    isinstance(arg3, InMemStream) and
+                    ((arg2.auto_close is True and
+                      arg3.auto_close is True) or (
+                        arg2.state == StreamState.completed and
+                        arg3.state == StreamState.completed)))
+
+    def should_retry_on_error(self, error):
+        """rules for retry"""
```
abhinav: document the `error` parameter

python/tchannel/tornado/request.py
```diff
@@ -120,3 +143,38 @@ def get_body_s(self):
...
+        if self.is_streaming_request:
+            # not retry for streaming request
+            return False
+
+        retry_flag = self.headers.get('re', RetryType.DEFAULT)
+
+        if retry_flag == RetryType.NEVER:
+            return False
+
+        if error.code in [ErrorCode.bad_request, ErrorCode.cancelled]:
```
abhinav: In a different PR, let's change ErrorCode to an enum and include this information in there.

We can define `__init__` methods on `Enum`s to customize behavior. https://docs.python.org/3/library/enum.html#planet

python/tchannel/tornado/timeout.py
```diff
@@ -37,7 +37,9 @@ def timeout(future, seconds=1, io_loop=None):
     def raise_timeout(*args, **kwargs):
         if future.running():
             future.set_exception(TimeoutError())
-
-    io_loop.call_later(seconds, raise_timeout)
-
-    yield
+    if seconds == 0:
```
abhinav: What? If there is no timeout, why is this function being called? Zero or negative timeout is not valid input.

junchaowu: Keep this hack for now.

python/tchannel/tornado/request.py
junchaowu: Discussed offline. resolved

python/tchannel/tornado/peer.py
```diff
@@ -439,10 +484,17 @@ def send(self, arg1, arg2, arg3, traceflag=False, headers=None):
             Flag is for tracing.
         :param headers:
             Headers will be put int he message as protocol header.
+        :param attempt_times:
+            Number of times to attempts.
```
abhinav: "Maximum number of attempts to send the message"

python/tchannel/tornado/peer.py
```diff
@@ -439,10 +484,17 @@ def send(self, arg1, arg2, arg3, traceflag=False, headers=None):
...
+        :param ttl:
+            Timeout for each request (ms).
+        :param retry_delay:
+            Delay between each retry (s).
```
abhinav: Some consistency between the units would be nice. Change `retry_delay` to `ms` and divide by 1000 before calling `sleep`.

python/tchannel/tornado/peer.py
```diff
@@ -458,34 +510,80 @@ def send(self, arg1, arg2, arg3, traceflag=False, headers=None):
...
+        response = None
+        # black list to record all used peers, so they aren't chosen again.
+        blacklist = set()
+        # only retry on non-stream request
+        if request.is_streaming_request or self._hostport:
+            attempt_times = 1
+            request.ttl = 0
+
+        # mac number of times to retry 3.
+        for num_of_retry in range(attempt_times):
```
abhinav: `s/num_of_retry/attempt/`

abhinav:
:+1:

breerly:
I'm onboard. :ship:

PR #593, opened 2015-05-14
Update Hyperbahn registration system to use peering.

abhinav:
This makes it so that we can continue to use the same TChannel object for all
other requests instead of a HyperbahnClient wrapper around it.

A major advantage of this is that we don't have to implement things like
Thrift support twice.


breerly:
:+1:

PR #595, opened 2015-05-14
add thrift argscheme. Temperary solution

junchaowu:
@blampe @breerly @abhinav


abhinav:
Fixed in a different CR. No longer necessary.

PR #598, opened 2015-05-16
add timeout on connection identification

ShanniLi:
r @jcorbin @Raynos 
cc @kriskowal

### DIFF
--- a/node/connection.js
+++ b/node/connection.js
@@ -85,7 +85,7 @@ TChannelConnection.prototype.setupSocket = function setupSocket() {
     }
 
     function onSocketClose() {
-        self.resetAll(errors.SocketClosedError({reason: 'remote clossed'}));
+        self.resetAll(self.error || errors.SocketClosedError({reason: 'remote clossed'}));
         if (self.remoteName === '0.0.0.0:0') {
             self.channel.peers.delete(self.remoteAddr);
         }
@@ -157,8 +157,9 @@ TChannelConnection.prototype.setupHandler = function setupHandler() {
     }
 };
 
-TChannelConnection.prototype.onTimedOut = function onTimedOut(_arg, self) {
+TChannelConnection.prototype.onTimedOut = function onTimedOut(err, self) {
     self.logger.warn(self.channel.hostPort + ' destroying socket from timeouts');
+    self.error = err;
     self.socket.destroy();
 };
 
--- a/node/connection.js
+++ b/node/connection.js
@@ -85,7 +85,7 @@ TChannelConnection.prototype.setupSocket = function setupSocket() {
     }
 
     function onSocketClose() {
-        self.resetAll(self.error || errors.SocketClosedError({reason: 'remote clossed'}));
+        self.resetAll(errors.SocketClosedError({reason: 'remote clossed'}));
         if (self.remoteName === '0.0.0.0:0') {
             self.channel.peers.delete(self.remoteAddr);
         }
@@ -159,7 +159,7 @@ TChannelConnection.prototype.setupHandler = function setupHandler() {
 
 TChannelConnection.prototype.onTimedOut = function onTimedOut(err, self) {
     self.logger.warn(self.channel.hostPort + ' destroying socket from timeouts');
-    self.error = err;
+    self.resetAll(err);
     self.socket.destroy();
 };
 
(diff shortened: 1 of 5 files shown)


node/connection_base.js
```diff
@@ -113,6 +114,13 @@ TChannelConnectionBase.prototype.onTimeoutCheck = function onTimeoutCheck() {
     }
     if (self.lastTimeoutTime) {
         self.timedOutEvent.emit(self);
+    } else if (!self.remoteName &&
+        self.timers.now() - self.startTime > self.channel.requestDefaults.timeout) {
```
Raynos: we should not use `requestDefaults.timeout`

Let's use `self.channel.initTimeout` and default that to something like 2 seconds.

node/connection_base.js
```diff
@@ -113,6 +114,13 @@ TChannelConnectionBase.prototype.onTimeoutCheck = function onTimeoutCheck() {
...
+        self.timedOutEvent.emit(self, errors.TimeoutError({
```
Raynos: you should just emit an error here instead of a timedOutEvent.

If the connection is not initialized its broken. We already listen on the `close` event in node/request.js so there should be no extra events needed there

ShanniLi: connection only listen on timedOutEvent, so it is the right event to emit. :) 
other issues are fixed.

node/connection_base.js
ShanniLi: fixed.

node/connection_base.js
Raynos: No it should be an error; this connection is dead; you must emit an error. Not getting an init req or init res breaks the connection; its not usable; its dead.

ShanniLi: Who should listen to this error and close the connection? 
Currently, connection listens to the timeout event and closes itself.  Should I make a similar handler for time out error?

Raynos: the channel listens to errors on connections...

node/connection.js
```diff
@@ -157,8 +157,9 @@ TChannelConnection.prototype.setupHandler = function setupHandler() {
     }
 };
 
-TChannelConnection.prototype.onTimedOut = function onTimedOut(_arg, self) {
+TChannelConnection.prototype.onTimedOut = function onTimedOut(err, self) {
     self.logger.warn(self.channel.hostPort + ' destroying socket from timeouts');
+    self.error = err;
```
Raynos: let's explicitely call `self.resetAll(err);` here.

ShanniLi: fixed. :)

node/connection.js
```diff
@@ -157,8 +157,9 @@ TChannelConnection.prototype.setupHandler = function setupHandler() {
     }
 };
 
-TChannelConnection.prototype.onTimedOut = function onTimedOut(_arg, self) {
+TChannelConnection.prototype.onTimedOut = function onTimedOut(err, self) {
     self.logger.warn(self.channel.hostPort + ' destroying socket from timeouts');
+    self.resetAll(err);
```
jcorbin: doesn't resetAll call socket destroy?

Raynos: I don't think it does and that's a bug...

node/connection_base.js
```diff
@@ -113,6 +114,13 @@ TChannelConnectionBase.prototype.onTimeoutCheck = function onTimeoutCheck() {
     }
     if (self.lastTimeoutTime) {
         self.timedOutEvent.emit(self);
+    } else if (!self.remoteName &&
+        self.timers.now() - self.startTime >= self.channel.initTimeout) {
```
jcorbin: we usually line the indent up to hang off the first term in the if clause

node/connection_base.js
```diff
@@ -113,6 +114,13 @@ TChannelConnectionBase.prototype.onTimeoutCheck = function onTimeoutCheck() {
...
+        self.timedOutEvent.emit(self, errors.TimeoutError({
+            start: self.startTime,
+            elapsed: self.timers.now() - self.startTime,
```
jcorbin: call now only once and stash its return please... better even to factor out an `elapesd` quantity that the log and if agree about exactl

node/connection_base.js
ShanniLi: fixed.

node/connection_base.js
ShanniLi: good point. fixed.

jcorbin:
:+1:

Raynos:
lgtm

PR #600, opened 2015-05-17
2.0.0-IDEA: Add a fieldName for as/thrift not ok call response.

Raynos:
Breaking change: instead of having a `nameAsThrift` field on
the body we now have a 4th field on the response in the cb.

You must specify your thrift `fieldName` when sending a not ok
call response in your server-side handler so we can serialize
your body properly.

You must also set the `ok` field on your response shape. This
is to keep symettry with the client.

```js
function helloHandler(opts, req, head, body, cb) {
  cb(null, {
    ok: false,
    fieldName: 'hello',
    body: new Error('hello');
  });
}

client.send(tchan.request({
  serviceName: 'hello'
}), 'hello', null, null, function onResponse(err, res, head, body) {
  if (!res.ok) {
    console.log('hello error, body);
  }
});
```

In the client response handler we will always check the `ok` field.
The ok boolean is an important part of the call response shape and
we should encourage handlers to explicitly set the `ok` field instead
of having it be inferred.

cc @kriskowal @jcorbin


kriskowal:
Looks sensible. :+1:

Willyham:
:+1:

jcorbin:
Can live with it, I'd rather have the field be called simply "type", "thriftType", or "typeName".  My reasoning:
- technically it's not part of a "struct" so it's not logically a "field"
- rather it's part of a union (albeit implicit), over multiple types
- that we then implement or even represent unions using an encoded singleton struct with a "field" on the wire is an implementation detal

Raynos:
Rename to `typeName`

Willyham:
Just merged master into this and opened https://github.com/uber/tchannel/pull/640 to rename and fix test. I'd really like to get this in asap because it's a breaking change for applications.

Raynos:
@willyham all of the 2.0 idea prs are breaking. We want to get them all in; but that will take an entire day or two

Willyham:
That's fine. Most of them will have changes which are pretty lightweight. Some of them (like mandatory 'as' header) will have no changes in the application code. This one has a larger surface area for applications, so I'd like to get this in now and release another beta so that I'm not writing this code against a moving target.

Raynos:
@Willyham the required `"cn"` header and the required threading of `inreq` will have a larger breaking change.

I'd rather get all breaking changes in at once and cut a single new beta.

PR #601, opened 2015-05-18
2.0.0-IDEA: pass in thriftText instead of spec

Raynos:
Breaking change: You must pass in the thrift idl as a raw string
instead of passing a thriftify spec object.

To avoid having two copies of `thriftify` or having drift between
their versions and implementations I've changed the as/thrift
interface to pass in the thrift IDL as text.

I've avoided the `thriftFile` approach; the new interface
encourages people to call `fs.readFileSync()` at startup time
and then instantiate tchannelAsThrift later.

```js
// clients/index.js

var fs = require('fs');
var path = require('path');
var TChannelAsThrift = require('tchannel/as/thrift');

var thriftText = fs.readFileSync(
  path.join(__dirname, '..', 'thrift, 'app.thrift'),
  'utf8'
);

module.exports = ApplicationClients;

function ApplicationClients() {
  this.appThrift = TChannelAsThrift({
    thriftText: thriftText
  });
}
```

cc @kriskowal @jcorbin

### DIFF
--- a/node/as/thrift.js
+++ b/node/as/thrift.js
@@ -23,6 +23,7 @@
 var assert = require('assert');
 var bufrw = require('bufrw');
 var Result = require('bufrw/result');
+var thriftify = require('thriftify');
 
 var errors = require('../errors.js');
 
@@ -37,8 +38,10 @@ function TChannelAsThrift(opts) {
 
     var self = this;
 
-    assert(opts && opts.spec, 'TChannelAsThrift expected spec');
-    self.spec = opts.spec;
+    assert(opts && typeof opts.source === 'string',
+        'must pass source as an argument');
+
+    self.spec = thriftify.parseSpec(opts.source);
 
     // Pulled off of things in `.register` and `.send` rather than passed in
     self.logger = null;
--- a/node/test/as-thrift.js
+++ b/node/test/as-thrift.js
@@ -25,16 +25,16 @@
 
 var path = require('path');
 var TypedError = require('error/typed');
-var thriftify = require('thriftify');
+var fs = require('fs');
 
 var allocCluster = require('./lib/alloc-cluster.js');
 var TChannelAsThrift = require('../as/thrift.js');
 
-var globalSpec = thriftify.readSpecSync(
-    path.join(__dirname, 'anechoic-chamber.thrift')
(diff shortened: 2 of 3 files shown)


Willyham:
I think that having access to the spec might be beneficial for most projects, especially if the output from thriftify or thriftrw ends up having struct/exception type constructors and enum objects. If the parsing *does* move in that direction, it means there's still potential for version mismatch here.

How about adding a `getThriftSpec` function on to the /as/thrift prototype too?

jcorbin:
Yeas the problem here @Willyham is that we don't want integrators to have to take a direct dep on the particular thrift implementation, so that we can easily transition them over the thriftify -> thriftrw -> ??? transition.  The reality here is that consumers are using "tchannel as thrift", not some thrift library directly.

I think we could easily have (building on Jake's example): `this.appThrift.spec`

node/as/thrift.js
```diff
@@ -23,6 +23,7 @@
 var assert = require('assert');
 var bufrw = require('bufrw');
 var Result = require('bufrw/result');
+var thriftify = require('thriftify');
```
jcorbin: I think this means you need to shift a dev dep -> prod dep? or is that already done?

jcorbin:
looks :ok: to me

Willyham:
Yup, totally on board, I just think having a getter method is preferable in this case so that it's much clearer that the spec is available to the implementer.

Raynos:
`source` to be the next.

Raynos:
@jcorbin @kriskowal 

This is ready for review again.

node/as/thrift.js
```diff
@@ -37,8 +38,10 @@ function TChannelAsThrift(opts) {
 
     var self = this;
 
-    assert(opts && opts.spec, 'TChannelAsThrift expected spec');
-    self.spec = opts.spec;
+    assert(opts && typeof opts.source === 'string',
+        'must pass source as an argument');
+
+    self.spec = thriftify.parseSpec(opts.source);
```
kriskowal: @Willyham Has `spec` property. That should suffice.

Raynos: agreed.

PR #602, opened 2015-05-18
2.0.0-IDEA: Make calling .register() and .request() illegal on top channel

Raynos:
It's mandatory for every call request to have a `service` field; This means
for the server we really should use the ServiceNameHandler and we should
use sub channels with their own unique endpoint handlers to handle stuff on
the server.

I've made calling `register()` on a top level channel illegal.

For making outgoing requests; we really should use the hyperbahn client
to create sub channels for outgoing requests and call `request()` on them.

I've made calling `request()` on the top level channel illegal as it allows
people to make mistakes and accidentally make requests that end up at the
wrong peer.

I've personally been bitten by making outgoing requests on the wrong
channel before.

cc @jcorbin @kriskowal

### DIFF
--- a/node/channel.js
+++ b/node/channel.js
@@ -404,14 +404,8 @@ TChannel.prototype.request = function channelRequest(options) {
     assert(!self.destroyed, 'cannot request() to destroyed tchannel');
     var opts = self.requestOptions(options);
 
-    if (!self.serviceName && !opts.host) {
-        if (opts.serviceName &&
-            self.subChannels &&
-            self.subChannels[opts.serviceName]) {
-            return self.subChannels[opts.serviceName].request(opts);
-        } else {
-            throw errors.TopLevelRequestError();
-        }
+    if (!self.topChannel) {
+        throw errors.TopLevelRequestError();
     }
 
     var req = null;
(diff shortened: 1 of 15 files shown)


node/channel.js
```diff
@@ -319,6 +319,11 @@ TChannel.prototype.listen = function listen(port, host, callback) {
 TChannel.prototype.register = function register(name, options, handler) {
     var self = this;
 
+    // If we are the top channel
+    if (!self.topChannel) {
```
jcorbin: I'd rather seen this done by quack than hack:
- check if the handler can register, if so register on it
- otherwise it's an error

jcorbin:
Seems sensible, I think however that we should leave it up to the reality of the handler, rather than a proxy boolean flag.

node/channel.js
Willyham: Apparently my comment on this didn't get submitted. If this logic is correct, it reads extremely counter-intuitively to the comment.

Raynos: `topChannel` is a channel instance that is a reference to the top channel;

all sub channels have a `topChannel` field that is the top channel; the top channel does not have a top channel; i..e `topChannel` is null.

I can do explicit null checks to avoid boolean confusion

Raynos:
Rebased and addressed feedback.

r: @jcorbin @kriskowal

kriskowal:
Seems legit.

jcorbin:
:+1:

PR #603, opened 2015-05-18
2.0.0-IDEA: Make tracing required

Raynos:
- Defaults tracing to true; tracing is opt out
 - Make the parentSpan propagation required
 - Fix relay handler to propagate tracing
 - Default trace agent to be silent.

For zipkin and tracing infrastructure to work we need
tracing always on. It should be opt out not opt in.

For tracing to work at all; we need application developers
to manually propagate parent spans. The only way we can
garantuee that they will propagate spans is to make it
a programmer error if they fail to do so. (it used to be
a warning).

cc @jcorbin @kriskowal

### DIFF
--- a/node/channel.js
+++ b/node/channel.js
@@ -173,7 +173,7 @@ function TChannel(options) {
         self.tracer = new TracingAgent({
             logger: self.logger,
             forceTrace: self.options.forceTrace,
-            serviceName: self.options.tracingServiceName,
+            serviceName: self.options.serviceNameOverwrite,
             reporter: self.options.traceReporter
         });
 
--- a/node/channel.js
+++ b/node/channel.js
@@ -166,7 +166,10 @@ function TChannel(options) {
     self.listening = false;
     self.destroyed = false;
 
-    if (self.options.trace) {
+    var trace = typeof self.options.trace === 'boolean' ?
+        self.options.trace : true;
+
+    if (trace) {
         self.tracer = new TracingAgent({
             logger: self.logger,
             forceTrace: self.options.forceTrace,
--- a/node/package.json
+++ b/node/package.json
@@ -41,6 +41,7 @@
     "debug-logtron": "4.0.0",
     "duplexer": "^0.1.1",
     "faucet": "0.0.1",
+    "format-stack": "4.0.0",
     "istanbul": "^0.3.13",
     "jshint": "^2.5.6",
     "ldjson-stream": "^1.2.1",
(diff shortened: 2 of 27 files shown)


jcorbin:
Calling it "topLevelRequest" seems unfortunate, since it implies more than is going on, rather call it "parentRequest".  The "top level" part makes one think of the original user request, which we may be arbitrary levels down from.

Raynos:
Rename `topLevelRequest` to `hasNoParent`

Raynos:
We should thread through `opts.parent` instead of instead of `opts.span`; `opts.parent` should be an InRequest.

Raynos:
@kriskowal @jcorbin 

I rebased and implemented the `parent` and `hasNoParent` rename. This is ready for review now.

kriskowal:
:ok:

node/channel.js
```diff
@@ -164,11 +164,14 @@ function TChannel(options) {
     self.listening = false;
     self.destroyed = false;
 
-    if (self.options.trace) {
+    var trace = typeof self.options.trace === 'boolean' ?
+        self.options.trace : true;
```
jcorbin: novel spacing snuck in...

node/package.json
```diff
@@ -41,6 +41,7 @@
     "debug-logtron": "4.0.0",
     "duplexer": "^0.1.1",
     "faucet": "0.0.1",
+    "format-stack": "4.0.0",
```
jcorbin: where is this used now?

Raynos:
I'll rebase again after review.

node/trace/agent.js
```diff
@@ -43,7 +44,7 @@ function Agent(options) {
     // incoming reuqests
     self.serviceName = options.serviceName || null;
 
-    self.reporter = options.reporter || self.reporter;
+    self.reporter = options.reporter || nullReporter;
```
jcorbin: I'd mildly prefer to keep the default on the prototype, changing this instead to `if (options.reporter) self.reporter = options.reporter;`

node/package.json
Raynos: this is for `itape` support. I can remove this if wanted.

itape supports `itape test/index.js --format-stack` where it runs your tests with the local `format-stack` library imported. This allows you to not edit the tests to import format-stack for debugging purposes.

itape does require a local copy of the library though because you cant really require global npm modules because of how nodes module system works.

This means `format-stack` is a weird transitive dependency thing.

jcorbin:
:+1:, minor optional feedback

PR #604, opened 2015-05-18
2.0.0-IDEA: Make the "as" transport header mandatory

Raynos:
This is another breaking change. All outgoing requests must
set the "as" header on the way out.

I also fixed a seperate bug in connection about making
assumptions on the `err` object.

I added a bonus commit to reject incoming call requests without
the "as" transport header; however thats a binary breaking change
so I probably want to roll it back as we have no agreement or
concensus on it and it requires everyone in the team to agree.

cc @jcorbin @kriskowal

### DIFF
--- a/node/channel.js
+++ b/node/channel.js
@@ -389,11 +389,31 @@ TChannel.prototype.requestOptions = function requestOptions(options) {
     var opts = {};
     // jshint forin:false
     for (prop in self.requestDefaults) {
+        if (prop === 'headers') {
+            continue;
+        }
+
         opts[prop] = self.requestDefaults[prop];
     }
+    opts.headers = {};
+    if (self.requestDefaults.headers) {
+        for (prop in self.requestDefaults.headers) {
+            opts.headers[prop] = self.requestDefaults.headers[prop];
+        }
+    }
+
     if (options) {
         for (prop in options) {
+            if (prop === 'headers') {
+                continue;
+            }
             opts[prop] = options[prop];
+        }
+    }
+    if (options && options.headers) {
+        opts.headers = opts.headers;
+        for (prop in options.headers) {
+            opts.headers[prop] = options.headers[prop];
         }
     }
     // jshint forin:true
--- a/node/channel.js
+++ b/node/channel.js
@@ -116,7 +116,7 @@ function TChannel(options) {
     self.random = self.options.random || globalRandom;
     self.timers = self.options.timers || globalTimers;
     self.initTimeout = self.options.initTimeout || 2000;
-    self.forceAs = self.options.forceAs;
+    self.requireAs = self.options.requireAs;
 
     // Filled in by the listen call:
     self.host = null;
--- a/node/channel.js
+++ b/node/channel.js
@@ -116,6 +116,7 @@ function TChannel(options) {
     self.random = self.options.random || globalRandom;
     self.timers = self.options.timers || globalTimers;
     self.initTimeout = self.options.initTimeout || 2000;
+    self.forceAs = self.options.forceAs;
 
     // Filled in by the listen call:
     self.host = null;
--- a/node/v2/handler.js
+++ b/node/v2/handler.js
@@ -73,6 +73,8 @@ function TChannelV2Handler(options) {
     self.streamingReq = Object.create(null);
     self.streamingRes = Object.create(null);
     self.writeBuffer = new Buffer(v2.Frame.MaxSize);
+
+    self.forceAs = self.options.forceAs === false ? false : true;
 }
 
 util.inherits(TChannelV2Handler, EventEmitter);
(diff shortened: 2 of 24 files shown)


node/channel.js
```diff
@@ -83,6 +83,8 @@ function TChannel(options) {
     self.requestDefaults = extend({
         timeout: TChannelRequest.defaultTimeout
     }, self.options.requestDefaults);
+    self.transportHeaderDefaults =
```
jcorbin: erm, how was this not already afforded for ala requestDefaults.headers?

node/channel.js
```diff
@@ -356,11 +358,26 @@ TChannel.prototype.requestOptions = function requestOptions(options) {
     for (prop in self.requestDefaults) {
         opts[prop] = self.requestDefaults[prop];
     }
+    if (self.transportHeaderDefaults) {
+        opts.headers = {};
```
jcorbin: let's just always set this empty, and fill it in...

node/channel.js
```diff
@@ -356,11 +358,26 @@ TChannel.prototype.requestOptions = function requestOptions(options) {
...
     if (options) {
         for (prop in options) {
+            if (prop === 'headers') {
+                continue;
+            }
             opts[prop] = options[prop];
         }
     }
+    if (options && options.headers) {
+        opts.headers = opts.headers || {};
```
jcorbin: ...then we don't need to check here

node/v2/handler.js
```diff
@@ -180,6 +181,14 @@ TChannelV2Handler.prototype.handleCallRequest = function handleCallRequest(reqFr
     if (self.remoteHostPort === null) {
         return callback(errors.TChannelCallReqBeforeInitReqError());
     }
+
+    if (!reqFrame.body ||
+        !reqFrame.body.headers ||
+        !reqFrame.body.headers.as
+    ) {
+        return callback(errors.AsHeaderRequired());
```
jcorbin: Not callback(err), that's for low level "should close the socket" type errors.

In this case, this should be an operational error resulting in a `"BadRequest", "missing as header"` error frame to the caller.

jcorbin:
:+1: let's:
- fix the handler guard so that it correctly sends error frame as noted
- then make that opt-outable by a channel option that turns it only into a local warn/error log (maybe stat)
- we had verbale buy in from the python side to make this required, so maybe even the option isn't necessary

Raynos:
@jcorbin @kriskowal ready for review.

kriskowal:
:ok:

From my regular editorial column, “0. naming things”:

> forceAs reads, “regardless of whether `as` is provided, use this one”, whereas requireAs, “ensure that `as` has been provided”.

Raynos:
@kriskowal renamed.

kriskowal:
:+1:

node/v2/handler.js
```diff
@@ -181,7 +184,29 @@ TChannelV2Handler.prototype.handleCallRequest = function handleCallRequest(reqFr
...
+            var err = errors.AsHeaderRequired({
+                frame: 'request'
+            });
+            self.sendErrorFrame(
+                req.res, 'ProtocolError', err.message
+            );
+            return callback();
+        } else {
+            self.logger.error('Expected "as" header for incoming req', {
+                arg1: String(reqFrame.body.args[0]),
```
jcorbin: more caller info? remote addr?

node/v2/handler.js
```diff
@@ -212,6 +237,23 @@ TChannelV2Handler.prototype.handleCallResponse = function handleCallResponse(res
...
+        !resFrame.body.headers.as
+    ) {
+        if (self.requireAs) {
+            var err = errors.AsHeaderRequired({
+                frame: 'response'
+            });
+            return callback(err);
+        } else {
+            self.logger.error('Expected "as" for incoming response', {
+                code: resFrame.body.code
```
jcorbin: moar context plox

node/v2/handler.js
```diff
@@ -379,6 +421,17 @@ TChannelV2Handler.prototype.sendCallRequestFrame = function sendCallRequestFrame
     var reqBody = new v2.CallRequest(
         flags, req.ttl, req.tracing, req.serviceName, req.headers,
         req.checksum.type, args);
+
+    if (self.requireAs) {
+        assert(req.headers && req.headers.as,
+            'Expected the "as" transport header to be set for request');
+    } else {
+        self.logger.error('Expected "as" header to be set for request', {
+            arg1: String(args[0]),
```
jcorbin: y u no moar context

node/v2/handler.js
```diff
@@ -392,6 +445,16 @@ TChannelV2Handler.prototype.sendCallResponseFrame = function sendCallResponseFra
     var resBody = new v2.CallResponse(
         flags, code, res.tracing, res.headers,
         res.checksum.type, args);
+
+    if (self.requireAs) {
+        assert(res.headers && res.headers.as,
+            'Expected the "as" transport header to be set for response');
+    } else {
+        self.logger.error('Expected "as" header to be set for response', {
+            code: code
```
jcorbin: con: text

jcorbin:
:+1: good enough, could be better

Raynos:
Added the context.

PR #608, opened 2015-05-20
As thrift take logger in opts

rf:
Make as thrift take a logger in opts.

If you don't do this and you never register things it won't have a logger, which will cause it to crash when it tries to warn about parse errors.

### DIFF
--- a/node/as/thrift.js
+++ b/node/as/thrift.js
@@ -40,6 +40,7 @@ function TChannelAsThrift(opts) {
     assert(opts && opts.spec, 'TChannelAsThrift expected spec');
     self.spec = opts.spec;
 
+    // Pulled off of things in `.register` and `.send` rather than passed in
     self.logger = null;
 
     var bossMode = opts && opts.bossMode;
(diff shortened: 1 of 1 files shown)


rf:
@Raynos @kriskowal

node/as/thrift.js
```diff
@@ -40,7 +41,7 @@ function TChannelAsThrift(opts) {
     assert(opts && opts.spec, 'TChannelAsThrift expected spec');
     self.spec = opts.spec;
 
-    self.logger = null;
+    self.logger = opts.logger || NullLogtron();
```
Raynos: no; not this.

Raynos: Add `self.logger = self.logger || req.logger` in `send()`

rf:
yes?

node/as/thrift.js
```diff
@@ -40,7 +40,7 @@ function TChannelAsThrift(opts) {
     assert(opts && opts.spec, 'TChannelAsThrift expected spec');
     self.spec = opts.spec;
 
-    self.logger = null;
+    self.logger = opts.logger || null;
```
Raynos: no; no `opts.logger` always null.

rf: Pulling the logger off of things in `.send` and `.register` is a super weird pattern imo

Raynos:
lgtm; maybe squash; otherwise #shippingit

Raynos:
ill steal your branch

PR #610, opened 2015-05-20
Add a waitForIdentified method + friends

Raynos:
This PR adds:

 - waitForIdentified method; used by other libraries that want
    to do direct peer to peer requests without using `node/request.js` 
 - Fix a bug in connection.js
 - Fix hyperbahn to wait for identification
 - Add a timeout to hyperbahn
 - Change the timeout reaper in tchannel to use 100ms

r: @kriskowal @shannili @rf

### DIFF
--- a/node/channel.js
+++ b/node/channel.js
@@ -366,6 +366,13 @@ TChannel.prototype.requestOptions = function requestOptions(options) {
     return opts;
 };
 
+TChannel.prototype.waitForIdentified =
+function waitForIdentified(options, callback) {
+    var self = this;
+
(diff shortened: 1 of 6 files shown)


Raynos:
This is necessary to go in today to unblock autobahn & pals.

kriskowal:
I believe that there is an outstanding issue recommending this feature by
the name withConnection.

Raynos:
I'm open to a rename; i don't care. Whoever renames it has to autobahn and redeploy autobahn though.

ShanniLi:
lgtm :)

PR #611, opened 2015-05-20
Improve timeout specificity

Raynos:
This PR takes on two things

 - refactor the TimeoutError base class into three
    base classes for the three different cases
 - Fix the `lastTimeoutTime` check to be more JIT.

r: @shannili @kriskowal


ShanniLi:
lgtm :+1:

kriskowal:
:ok:

PR #612, opened 2015-05-20
relay: index inreqs by remoteAddr and id

Raynos:
There is an issue in our relay handler implementation
where we were indexing by id rather then by remoteAddr
and id.

r: @kriskowal @shannili

### DIFF
--- a/node/relay_handler.js
+++ b/node/relay_handler.js
@@ -155,7 +155,10 @@ RelayHandler.prototype.type = 'tchannel.relay-handler';
 
 RelayHandler.prototype.handleRequest = function handleRequest(req, buildRes) {
     var self = this;
-    var rereq = self.reqs[req.id];
+
+    var reqKey = getReqKey(req);
+    var rereq = self.reqs[reqKey];
+
     if (rereq) {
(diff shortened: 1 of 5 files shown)


node/relay_handler.js
```diff
@@ -168,12 +169,29 @@ RelayHandler.prototype.handleRequest = function handleRequest(req, buildRes) {
...
-        delete self.reqs[req.id];
+        delete self.sockets[req.remoteAddr].reqs[req.id];
     }
 };
 
+RelayHandler.prototype.getReq = function getReq(req) {
+    var self = this;
+
+    var socket = self.sockets[req.remoteAddr];
+    if (!socket) {
```
ShanniLi: Doing a set in get. It is perfect for the context being called, but potential an issue for reuse. May be better to put it into set. 

This comment is no blocking.

ShanniLi:
lgtm

ShanniLi:
lgtm. Like this version better. 
Just to make sure, req.connection.guid should never contain '~', right?

kriskowal:
Apart from the mis-named GUID, this is :ok:. Recommending that we just call them connection identifiers, which will be unique for the duration of a process.

PR #613, opened 2015-05-20
node: require minimum requests before marking peer as unhealthy

Raynos:
This peer adds a minimum requests counter ( defaults to 5 )
so a peer only goes unhealthy after some minimum number of requests.

This means 50% of requests have to fail in a 1 second interval
after at least 5 requests for it to go unhealthy.

This makes it easy to author tests and do manual testing where
you send two reqs that fail.

r: @shannili @kriskowal

### DIFF
--- a/node/peer.js
+++ b/node/peer.js
@@ -53,6 +53,7 @@ function TChannelPeer(channel, hostPort, options) {
         random: self.channel.random,
         period: self.options.period,
         maxErrorRate: self.options.maxErrorRate,
+        minimumRequests: self.options.minimumRequests,
         probation: self.options.probation,
         stateMachine: self,
         nextHandler: self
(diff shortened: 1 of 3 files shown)


ShanniLi:
lgtm.

kriskowal:
Could alternately set the maxErrorRate to 100% if the intent is just to open the gates for tests. There is a chance we’ll disable the feature in production in just this way.

node/peer.js
```diff
@@ -53,6 +53,7 @@ function TChannelPeer(channel, hostPort, options) {
         random: self.channel.random,
         period: self.options.period,
         maxErrorRate: self.options.maxErrorRate,
+        minimumRequests: self.options.minimumRequests,
```
kriskowal: Naming nit: max is to min as maximum is to minimum. Consistency is my hobgoblin.

Raynos:
I think this change is a bit more friendly for people writing tests or one-off scripts

PR #614, opened 2015-05-20
remove context namedtuple.

junchaowu:
@blampe @breerly @abhinav 
remove the context(message, message_id) namedtuple.
Increase the TTL for request


python/tchannel/messages/common.py
```diff
@@ -33,6 +33,8 @@
 
 PROTOCOL_VERSION = 0x02
 
+DEFAULT_TTL = 1000  # ms
```
breerly: Shouldn't this be in `glossary`?

python/tchannel/handler.py
```diff
@@ -108,17 +104,17 @@ def handle_pre_call(self, message_id, message, connection):
...
-                message_id,
+                message.id,
             )
-        except Exception:
+        except Exception as e:
             connection.send_error(
                 ErrorCode.unexpected,
-                "An unexpected error has occurred!",
-                message_id,
+                "An unexpected error has occurred!" + e.message,
```
abhinav: Probably not a good idea to include service-internal error messages in external responses. Skip doing this. Also, drop the `!`.

(If we were to include the message, concatenation like that doesn't make for a pretty error message.)

python/tchannel/messages/error.py
```diff
@@ -47,6 +47,7 @@ class ErrorMessage(BaseMessage):
     message_type = Types.ERROR
 
     __slots__ = (
+        'id',
```
abhinav: `BaseMessage.__slots__`?

python/tchannel/messages/init_request.py
```diff
@@ -38,7 +38,8 @@ class InitRequestMessage(BaseMessage):
         'headers',
     )
 
-    def __init__(self, version=None, headers=None):
+    def __init__(self, version=None, headers=None, id=0):
+        super(InitRequestMessage, self).__init__(id)
```
abhinav: `BaseMessage.__slots__`?

junchaowu: done

abhinav:
:ok:

PR #617, opened 2015-05-22
[py] Implement endpoint registration API on TChannel.

abhinav:
Given a service,

```thrift
// service.thrift

service Foo {
    i32 bar(1: string a, 2: list<bool> b) throws (1: SomethingWentWrong something);
}
```

You can do:

```python
from service import Foo

app = TChannel()

@app.register(Foo)
def bar(request, response, tchannel):
    a = request.args.a
    b = request.args.b

    if a == "hello":
        raise Foo.SomethingWentWrong()

    response.write_header("Timestamp", current_time())
    response.write_result(42)
```

Instead of ``write_result`` the handler can also just return the result (in a Future or directly).

```python
    response.write_header("Timestamp", current_time())
    return 42
```

For non-Thrift, you'll be able to do:

```python

@app.register('foo', 'json')
def foo(request, response, tchannel):
    # ...

@app.register('bar', 'raw')
def bar(request, response, tchannel):
    # ...

```

@blampe @breerly @junchaowu


python/tests/integration/trace/test_zipkin_trace.py
```diff
@@ -57,24 +56,18 @@ def handler1(request, response, proxy):
...
 
 
 @pytest.yield_fixture
-def trace_server(random_open_port, handlers):
-    with TChannelServerManager(
-        port=random_open_port,
-        dispatcher=handlers
-    ) as manager:
+def trace_server(random_open_port):
+    with TChannelServerManager(random_open_port) as manager:
```
breerly: You're shitting me right.

python/tests/integration/server_manager.py
```diff
@@ -95,15 +109,13 @@ def shutdown(self):
 
 class TChannelServerManager(ServerManager):
```
breerly: Rename to `TChannelTestServer` please. Remove hierarchy.

python/tchannel/tornado/tchannel.py
```diff
@@ -51,6 +56,12 @@ class State(IntEnum):
 class TChannel(object):
     """Manages inbound and outbound connections to various hosts."""
 
+    _SCHEMES = {
+        'raw': scheme.RawArgScheme,
+        'json': scheme.JsonArgScheme,
```
junchaowu: Can we create a class containing fields to represent 'raw' , 'json' .. types?
I think ask use to type string input is dangerous. They may unsure it is capitalized or spell error.

abhinav: Yeah, I was gonna make this an enum or a set of pre-defined constants.

junchaowu: Actually I remember I had it in my retry patch. But it is quite out of date right now.

python/tests/integration/test_server.py
```diff
@@ -25,46 +25,61 @@
...
+    def and_raise(self, exc):
 
-class ServerManager(object):
+        def execute(request, response):
+            raise exc
+
+        self.execute = execute
+
+
+class TestServer(object):
```
breerly: :clap: and the engineer thus managed to achieve code reuse without inheritance :clap:

abhinav: :shrug:

python/examples/handler.py
```diff
@@ -58,19 +57,15 @@ def slow(request, response, proxy):
     response.flush()
 
 
-def get_example_handler():
-    dispatcher = RequestDispatcher()
+def register_example_endpoints(tchannel):
+    tchannel.register("hi", "raw", say_hi)
+    tchannel.register("ok", "raw", say_ok)
+    tchannel.register("echo", "raw", echo)
+    tchannel.register("slow", "raw", slow)
```
breerly: these might be more obvious if we use keyword args:

```
tchannel.register(endpoint="slow", argscheme="raw", handler=slow)
```

abhinav: Updated examples

breerly:
:shipit:

PR #618, opened 2015-05-22
Refactor timeout + ops logic into operations

Raynos:
r: @rf @kriskowal

### DIFF
--- a/node/connection_base.js
+++ b/node/connection_base.js
@@ -27,6 +27,7 @@ var EventEmitter = require('./lib/event_emitter');
 
 var errors = require('./errors');
 var States = require('./reqres_states');
+var Operations = require('./operations');
 
 var DEFAULT_OUTGOING_REQ_TIMEOUT = 2000;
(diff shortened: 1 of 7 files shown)


node/connection_base.js
```diff
@@ -27,6 +27,7 @@ var EventEmitter = require('./lib/event_emitter');
 
 var errors = require('./errors');
 var States = require('./reqres_states');
+var Operations = require('./operations');
```
kriskowal: Operations are CallRequest / CallResponse pairs. We eliminated Operation previously, in favor of using the Request object as a proxy for the corresponding Response. Probably should call this Requests to be consistent with that decision.

kriskowal:
:ok:

node/connection_base.js
Raynos: requests.js next to request.js and in_request.js is just confusion++

operations might not be the best word but it will be less confusing then requests

kriskowal: Compelling argument.

rf:
:+1:

PR #619, opened 2015-05-22
operations: add a connection stale grace period

Raynos:
I think our timeouts are too aggressive because i bumped the
reaper down to 100ms instead of 1s.

I added a 1500ms grace period to see if production becomes
healthier.

r: @rf @kriskowal

### DIFF
--- a/node/operations.js
+++ b/node/operations.js
@@ -224,7 +224,7 @@ function _checkTimeout(ops, direction) {
             if (direction === 'out') {
                 var now = self.timers.now();
                 if (self.lastTimeoutTime &&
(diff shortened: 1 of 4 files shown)


node/operations.js
```diff
@@ -22,6 +22,8 @@
 
 var errors = require('./errors.js');
 
+var CONN_STALE_PERIOD = 1500;
```
kriskowal: A little worried about magic number. We might have to fiddle with this in production.

Raynos: Agreed; i'll thread this through from channel.js

PR #621, opened 2015-05-22
Set the call response "as" transport header

Raynos:
This will make tooling like tcap & tcurl better.

r: @kriskowal @shannili


ShanniLi:
lgtm

kriskowal:
:ok:

PR #623, opened 2015-05-25
remove connection on close

Raynos:
We leak memory in autobahn; I think this is one of the causes.
There may be more.

 - [ ] TODO: add helpers to alloc-cluster to assert that
    there are no more connections / request objects in memory.

r: @kriskowal @shannili


kriskowal:
:ok: good find

ShanniLi:
lgtm

PR #625, opened 2015-05-26
Relay: downgrade certain errors to infos

Raynos:
This PR adds extra context to relay errors. We also downgrade
any EPIPE or ECONNRESET errors to info's.

r: @kriskowal @rf


kriskowal:
:ok:

PR #626, opened 2015-05-26
connection: add more context to connection close errors

Raynos:
This PR just adds more context to these errors. This makes debugging
the logs easier because we will know which serviceName & remoteAddr
these issues are for.

r: @kriskowal @rf


kriskowal:
:ok:

PR #630, opened 2015-05-27
[node] Pin error module version

kriskowal:
Until we are compatible with the latest TypedError (requires message).


kriskowal:
cc @raynos

Raynos:
lgtm. #shipit

PR #631, opened 2015-05-28
package: upgrade to error@6.4.1

Raynos:
This PR updates error to the latest version; this
includes a couple of features.

 - The `fullType` field on `WrappedError` now contains
    the `fullType` for any wrapped error its wrapping.
    I.e. its now recursive.
 - All instances of `TypedError` now have a `fullType`
    field that is an alias for `type`.
 - All constructor functions from `TypedError` and
    `WrappedError` have a static `type` field that can
    be used to see what type of error it will construct.

r: @kriskowal @rf

### DIFF
--- a/node/package.json
+++ b/node/package.json
@@ -22,7 +22,7 @@
   "dependencies": {
     "bufrw": "^0.9.24",
     "crc": "^3.2.1",
-    "error": "6.4.1",
+    "error": "^6.4.1",
     "farmhash": "^0.2.0",
     "hexer": "^1.4.5",
     "json-stringify-safe": "^5.0.0",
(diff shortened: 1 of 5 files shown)


node/package.json
```diff
@@ -22,7 +22,7 @@
   "dependencies": {
     "bufrw": "^0.9.24",
     "crc": "^3.2.1",
-    "error": "6.2.0",
+    "error": "6.4.1",
```
kriskowal: Can we add the carat back?

Raynos: Do you trust me? ;)

kriskowal: I trust that there will be bug fixes.

PR #633, opened 2015-05-29
errors: fully classify all errors

Raynos:
The `classify()` method was not exhaustive. It now is and
there are tests to ensure that it will always be exhaustive

This will be used for improving retries in hyperbahn/handler
and for improving error logging in relay_handler.

r: @kriskowal @rf

### DIFF
--- a/node/errors.js
+++ b/node/errors.js
@@ -462,6 +462,7 @@ module.exports.classify = function classify(err) {
         case 'tchannel-thrift-handler.parse-error.head-failed':
         case 'tchannel.checksum':
         case 'tchannel.duplicate-header-key':
+        case 'tchannel.null-key':
             return 'BadRequest';
 
         case 'tchannel.init.call-request-before-init-request':
@@ -480,7 +481,6 @@ module.exports.classify = function classify(err) {
         case 'tchannel.invalid-error-code':
         case 'tchannel.invalid-frame-type':
         case 'tchannel.missing-init-header':
-        case 'tchannel.null-key':
         case 'tchannel.protocol.read-failed':
         case 'tchannel.protocol.write-failed':
         case 'tchannel.unhandled-frame-type':
--- a/node/errors.js
+++ b/node/errors.js
@@ -411,7 +411,6 @@ module.exports.ThriftHeadStringifyError = WrappedError({
     direction: null
 });
 
-
 module.exports.TopLevelRegisterError = TypedError({
     type: 'tchannel.top-level-register',
     message: 'Cannot register endpoints points on top-level channel.\n' +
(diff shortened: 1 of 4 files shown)


kriskowal:
:ok: When @jcorbin gets back, let’s talk about disseminating this information out to the error constructors.

rf:
cool and neat

node/errors.js
```diff
@@ -445,21 +444,46 @@ module.exports.classify = function classify(err) {
...
+        case 'tchannel-json-handler.stringify-error.body-failed':
+        case 'tchannel-json-handler.parse-error.head-failed':
+        case 'tchannel-json-handler.stringify-error.head-failed':
+        case 'tchannel-thrift-handler.stringify-error.body-failed':
+        case 'tchannel-thrift-handler.stringify-error.head-failed':
+        case 'tchannel.request-frame-state':
+        case 'tchannel.request-already-done':
+        case 'tchannel.response-already-done':
+        case 'tchannel.response-already-started':
+        case 'tchannel.response-frame-state':
```
rf: response already started etc should probably be unexpected errors

node/errors.js
```diff
@@ -445,21 +444,46 @@ module.exports.classify = function classify(err) {
...
         case 'tchannel.connection.timeout':
         case 'tchannel.connection-stale.timeout':
             return 'Timeout';
 
-        case 'tchannel-handler.parse-error.body-failed':
-        case 'tchannel-handler.parse-error.head-failed':
+        case 'tchannel.init.call-request-before-init-request':
+        case 'tchannel.init.call-request-cont-before-init-request':
+        case 'tchannel.init.call-response-before-init-response':
+        case 'tchannel.init.call-response-cont-before-init-response':
```
rf: These should probably be protocol errors

node/errors.js
```diff
@@ -445,21 +444,46 @@ module.exports.classify = function classify(err) {
...
         case 'tchannel.duplicate-header-key':
             return 'BadRequest';
 
+        case 'tchannel.init.duplicate-init-request':
+        case 'tchannel.init.duplicate-init-response':
+        case 'tchannel.init.send-call-request-before-indentified':
+        case 'tchannel.init.send-call-request-cont-before-indentified':
+        case 'tchannel.init.send-call-response-before-indentified':
+        case 'tchannel.init.send-call-response-cont-before-indentified':
+        case 'tchannel.arg1-over-length-limit':
```
rf: Over length limit sounds like a bad request

Raynos:
Updated with some more fixes. @rf @kriskowal

node/errors.js
```diff
@@ -445,34 +444,71 @@ module.exports.classify = function classify(err) {
         case 'tchannel.max-pending-for-service':
             return 'Declined';
 
-        case 'tchannel.timeout':
+        case 'tchannel.socket-local-closed':
+        case 'tchannel.local.reset':
+            return 'Cancelled';
```
kriskowal: Whether local close or reset corresponds to a deliberate cancellation is debatable.

I do suspect that these cases should be candidates for retry.

These cases do not indicate remote health problems, so like cancellation and bad request, they should not trip circuit breaking. Perhaps we need another error class.

Whether deliberate and accidental cancellation should have the same retry semantics is the question.

node/errors.js
```diff
@@ -445,34 +444,71 @@ module.exports.classify = function classify(err) {
...
-        case 'tchannel-handler.parse-error.body-failed':
-        case 'tchannel-handler.parse-error.head-failed':
+        case 'tchannel-handler.json.invalid-body':
+        case 'tchannel-json-handler.parse-error.body-failed':
+        case 'tchannel-json-handler.parse-error.head-failed':
+        case 'tchannel.request-frame-state':
+        case 'tchannel.request-already-done':
+        case 'tchannel-thrift-handler.parse-error.body-failed':
+        case 'tchannel-thrift-handler.parse-error.head-failed':
         case 'tchannel.checksum':
```
kriskowal: Will have to talk to @jcorbin about whether we want to retry if a bit gets flipped in transit. Bad checksum is usually a bad request.

kriskowal:
:ok:

node/errors.js
jcorbin: Same feedback essentially as Kris: this isn't really a cancellation user action, in production I'd expect this to happen most frequently during upgrades / restarts / etc.

node/errors.js
```diff
@@ -445,34 +444,71 @@ module.exports.classify = function classify(err) {
...
         case 'tchannel.invalid-frame-type':
         case 'tchannel.missing-init-header':
-        case 'tchannel.null-key':
         case 'tchannel.protocol.read-failed':
         case 'tchannel.protocol.write-failed':
         case 'tchannel.unhandled-frame-type':
             return 'ProtocolError';
 
+        case 'tchannel.connection.close':
+        case 'tchannel.connection.reset':
```
jcorbin: the mislabeled "cancel" socket closes above should be right here.

node/errors.js
Raynos: The only time these happen is when you do a graceful shutdown of a connection.

This generally only happens when you call channel.close() and that's only really done by our tests; not in production.

I agree that cancellation is not ideal. But manual closing has nothing to do with network error. As far as we know the network is healthy. Maybe it's better marked as declined due to graceful close.

jcorbin: I don't think that word means what you think it means, this is no a "graceful" shutdown.  A graceful shutdown would wait for all pending requests to be finished before closing / shutting down, while at the same time not accepting any more incoming requests.

Raynos: I buy that it should be a network error.

PR #634, opened 2015-05-29
request: use error classify to do retries

Raynos:
Updated the logical request to use the Error.classify()
method to determine whether to retry rather then it's own
seperate switch statement.

This should make the retry logic exhaustive over all
tchannel errors.

r: @kriskowal @rf


kriskowal:
:ok:

rf:
:+1:

PR #635, opened 2015-05-29
relay_handler: improve error logging.

Raynos:
Downgrade all error frames to warn. Downgrade all network errors to warn.


kriskowal:
:ok:

PR #636, opened 2015-05-29
hyperbahn/handler: retry on all network errors

Raynos:
We previously retried on only one type of network error.
Now we retry on all of them.

r: @kriskowal @rf


rf:
:+1:

PR #637, opened 2015-05-29
relay: make exhaustive logging decisions for errors

Raynos:
I tried master in autobahn and it started shouting about
Declined which should not be an error.

I've upgraded this logic to make logging decisions based
on the classifier.

ProtocolError, UnexpectedError are errors if caused by
the relay; they are warns if we are just relaying error frames

NetworkError, Cancelled, Declined, Busy should be rare in
production so I'm keeping them as warnings for now.

BadRequest, Timeout are normal operational things, dropping
them to info() for now so we get insight into operational things
but will be ignored in the future.

r: @rf @kriskowal


kriskowal:
:ok:

PR #638, opened 2015-05-30
fix examples

ShanniLi:
r @kriskowal @Raynos

### DIFF
--- a/node/examples/example1.js
+++ b/node/examples/example1.js
@@ -57,24 +57,32 @@ var listening = ready(function (err) {
     client
         .request({serviceName: 'server'})
         .send('func1', "arg 1", "arg 2", function (err, res) {
-            assert.ifError(err);
-            assert.equal(res.ok, true);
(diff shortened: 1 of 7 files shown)


Raynos:
Lgtm.

node/examples/example1.js
```diff
@@ -45,19 +49,33 @@ var listening = ready(function (err) {
...
+            assert.ifError(err);
+            assert.equal(res.ok, true);
             console.log('normal res: ' + res.arg2.toString() + ' ' + res.arg3.toString());
+            done();
         });
     client
-        .request({host: '127.0.0.1:4040'})
+        .request({serviceName: 'server'})
         .send('func2', "arg 1", "arg 2", function (err, res) {
+            assert.ifError(err);
```
kriskowal: Let’s try not to replicate this pattern anymore. Do `if (err) return assert.end(err)`.

PR #639, opened 2015-06-01
Tchannel node: statsd + inbound stats

ShanniLi:
r @Raynos @kriskowal 
1. add statsd support.
2. add inbound stats.

TODO:
1. stats on req/res sizes
2. connection stats


node/channel.js
```diff
@@ -63,14 +64,8 @@ function TChannel(options) {
...
     self.requestEvent = self.defineEvent('request');
-    self.outboundCallsSentStat = self.defineCounter('outbound.calls.sent');
-    self.outboundCallsSystemErrorsStat = self.defineCounter('outbound.calls.system-errors');
-    self.outboundCallsOperationalErrorsStat = self.defineCounter('outbound.calls.operational-errors');
-    self.outboundCallsSuccessStat = self.defineCounter('outbound.calls.success');
-    self.outboundCallsAppErrorsStat = self.defineCounter('outbound.calls.app-errors');
-    self.outboundCallsRetriesStat = self.defineCounter('outbound.calls.retries');
-    self.outboundCallsLatencyStat = self.defineTiming('outbound.calls.latency');
-    self.outboundCallsPerAttemptLatencyStat = self.defineTiming('outbound.calls.per-attempt-latency');
+    self.setupOutboundStats();
```
Raynos: let's leave this here; having all the fields in the constructor is useful;

let's inline both of these.

node/lib/statsd.js
```diff
@@ -0,0 +1,120 @@
...
+    if (stat.name.indexOf('inbound.') === 0) {
+        return getKeyInbound(common, stat);
+    } else if (stat.name.indexOf('outbound.') === 0) {
+        return getKeyOutbound(common, stat);
+    }
+
+    return '';
+}
+
+TChannelStatsd.prototype.onStat = function onStat(stat) {
```
Raynos: Having a switch statement similar to the one in autobahn where we manually build the key for every stat is a lot easier.

The key has to be build differently for each stat.

node/lib/statsd.js
```diff
@@ -0,0 +1,120 @@
...
+function getKeyOutbound(common, stat) {
+    // var host = common.host || 'host';
+    // var cluster = common.cluster || 'cluster';
+    var version = common.version || 'version';
+    var service = stat.tags.service;
+    var targetService = stat.tags['target-service'];
+
+    // Note: endpoint should have a finite value space
+    var endpoint = stat.tags['target-endpoint'] || 'endpoint';
+    endpoint = endpoint.indexOf('::') !== -1 ?
```
Raynos: do not split on `::` the endpoint should be arg1; no magic.

node/lib/statsd.js
```diff
@@ -0,0 +1,120 @@
...
+    }
+
+    return '';
+}
+
+TChannelStatsd.prototype.onStat = function onStat(stat) {
+    var self = this;
+    var key = getKey(self.channel.statTags, stat);
+    switch (stat.type) {
+        case Counter.type:
```
Raynos: just hard code the strings here; it's simpler.

node/lib/stat_emitter.js
```diff
@@ -69,7 +69,8 @@ function Counter(name, emitter) {
     self.name = name;
     self.emitter = emitter;
 }
-Counter.prototype.type = 'counter';
+Counter.type = 'counter';
```
Raynos: Let's not do this or expose these things. It's unneeded.

node/lib/statsd.js
```diff
@@ -0,0 +1,120 @@
...
+        endpoint.split('::')[1] : endpoint;
+    return [
+        common.app,
+        // host,
+        // cluster,
+        version,
+        callingService,
+        service,
+        endpoint,
+        stat.name
```
Raynos: the static bits of the stat should be earlier in the stat; the dynamic bits should go later.

node/connection_base.js
```diff
@@ -216,8 +216,16 @@ TChannelConnectionBase.prototype.onReqError = function onReqError(req, err) {
 
 TChannelConnectionBase.prototype.runHandler = function runHandler(req) {
     var self = this;
+    self.channel.inboundCallsRecvdStat.increment(1, {
+        'calling-service': req.headers.cn,
+        'service': req.serviceName,
+        'endpoint': String(req.arg1)
+    });
+
+    var latencyStart = self.timers.now();
```
Raynos: this field should be on `inreq` not threaded through options.

node/connection_base.js
```diff
@@ -230,6 +238,12 @@ TChannelConnectionBase.prototype.buildResponse = function buildResponse(req, opt
             state: req.res.state
         }));
     }
+    options = extend({
```
Raynos: just put this logic in `buildOutResponse` itself.

node/out_response.js
```diff
@@ -106,14 +128,24 @@ TChannelOutResponse.prototype.sendParts = function sendParts(parts, isLast) {
...
             // TODO: log warn
             break;
+        default:
+            error = true;
+            assert(false, 'Wrong self state');
+            break;
+    }
+
+    if (isLast && !error) {
+        self._emitLatency();
```
Raynos: we want to emit latency for errors too!

node/out_response.js
```diff
@@ -34,8 +35,13 @@ function TChannelOutResponse(id, options) {
     self.spanEvent = self.defineEvent('span');
     self.finishEvent = self.defineEvent('finish');
 
+    self.channel = options.channel;
+    self.callingService = options.callingService;
+    self.endpoint = options.endpoint;
     self.logger = options.logger;
+    self.latencyStart = options.latencyStart;
```
Raynos: self.latencyEnd = null;

node/out_response.js
```diff
@@ -193,6 +226,20 @@ TChannelOutResponse.prototype.sendError = function sendError(codeString, message
...
         self.message = message;
+        self.channel.inboundCallsSystemErrorsStat.increment(1, {
+            'calling-service': self.callingService,
+            'service': self.serviceName,
+            'endpoint': self.endpoint,
+            'type': self.codeString
+        });
+
+        if (self.codeString === 'ProtocolError') {
+            self.channel.inboundProtocolErrorsStat.increment(1, {
```
Raynos: This is not how the protocol error stat works; its for protocol errors like "init request failed".

node/out_response.js
```diff
@@ -234,9 +281,29 @@ TChannelOutResponse.prototype.sendNotOk = function sendNotOk(res1, res2) {
...
     self.arg3 = res2;
 
+    if (self.ok) {
+        self.channel.inboundCallsSuccessStat.increment(1, {
+            'calling-service': self.callingService,
+            'service': self.serviceName,
+            'endpoint': self.endpoint
+        });
+    } else {
+        var type = [self.arg2, self.arg3].filter(
```
Raynos: this is not the type.

node/out_response.js
```diff
@@ -234,9 +281,29 @@ TChannelOutResponse.prototype.sendNotOk = function sendNotOk(res1, res2) {
...
+            function filter(val) {
+                return val;
+            }).join('.');
+        self.channel.inboundCallsAppErrorsStat.increment(1, {
+            'calling-service': self.callingService,
+            'service': self.serviceName,
+            'endpoint': self.endpoint,
+            'type': type
```
Raynos: default this to `'unknown'` We need some kind of `outResponse.setErrorType()` method to support this properly.

node/test/lib/mock-statsd.js
```diff
@@ -0,0 +1,47 @@
...
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+'use strict';
+
+function MockStatsd() {
```
Raynos: just import `require('statsd-client/null')` instead.

node/test/request-with-statsd.js
```diff
@@ -0,0 +1,149 @@
...
+            cn: 'inPipe'
+        }
+    }).send('Reservoir::get', 'ton', '20', onResponse);
+
+    function onResponse(err, res, arg2, arg3) {
+        assert.ifError(err, 'no error');
+        assert.ok(res.ok, 'res should be ok');
+
+        assert.deepEqual(statsd.parts, [{
+            key: 'pool.1.0.inPipe.reservoir.get.outbound.calls.sent',
```
Raynos: we want a string like `tchannel.outbound.calls.sent.{cn}.{service}.{endpoint}`. let's ignore version, app, host and cluster for now.

Raynos:
This is a good start; we will want to do some cleanup.

I would recommend you add all the `statsd` hookup code & tests in a seperate pull request. That will make this easier to review.

Raynos:
i.e. have one PR for statsd and one for inbound metrics.

node/lib/statsd.js
ShanniLi: which ones are static?

node/test/lib/mock-statsd.js
ShanniLi: do we want to introduce a dependency on statsd-client in tchannel?

node/lib/statsd.js
ShanniLi: Saw your comments below. Understood now. :)

node/lib/statsd.js
ShanniLi: right now. The keys mostly use the format below. There isn't too much of variety. I will change it to switch-based when we want individual keys to have different patterns.  
tchannel.outbound.calls.sent.{cn}.{service}.{endpoint}
or
tchannel.inbound.calls.recvd.{cn}.{service}.{endpoint}

node/test/lib/mock-statsd.js
Raynos: as a dev dependency; who cares.

node/lib/statsd.js
Raynos: I think it's easier to review as a switch. it makes it very obvious the exact mapping of `Stat` object to full statsd path.

The keys are also different; some require a `type`; some dont. Once we add bandwidth they will be different as well.

ShanniLi: sounds good. I will make the fix.

node/test/lib/mock-statsd.js
ShanniLi: cannot find null in 'statsd-client'. :(

ShanniLi:
divided into two separate PRs: #642 & #643

PR #640, opened 2015-06-01
Rename fieldName to typeName

Willyham:
@Raynos updating your PR with the rename


Willyham:
This also fixes the failing test on 'thrift-field-name'.

PR #641, opened 2015-06-01
Node: Bump version of thriftify

Willyham:
@Raynos 

Bump to alpha14 which supports typedefs and sets.


Raynos:
lgtm. Merge at will.

PR #642, opened 2015-06-01
Node: statsd hookup & test 

ShanniLi:
r @Raynos @kriskowal 

First PR for #639 simplification.

### DIFF
--- a/node/lib/statsd.js
+++ b/node/lib/statsd.js
@@ -20,8 +20,6 @@
 
 'use strict';
 
-var assert = require('assert');
-
 function TChannelStatsd(channel, statsd) {
     if (!(this instanceof TChannelStatsd)) {
         return new TChannelStatsd(channel, statsd);
@@ -38,10 +36,14 @@ function TChannelStatsd(channel, statsd) {
 }
 
 function clean(str) {
-    return str
-        .replace(/:/g, '-')
-        .replace(/\./g, '-')
-        .replace(/{|}/g, '-');
+    if (!str) {
+        return 'unknown';
+    } else {
+        return str
+            .replace(/:/g, '-')
+            .replace(/\./g, '-')
+            .replace(/{|}/g, '-');
+    }
 }
 
 function getKey(common, stat) {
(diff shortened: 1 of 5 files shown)


node/lib/statsd.js
```diff
@@ -0,0 +1,176 @@
...
+    var prefix = ['tchannel', stat.name].join('.');
+    switch (stat.name) {
+        // outbound
+        case 'outbound.calls.sent':
+            return [
+                prefix,
+                clean(stat.tags.service),
+                clean(stat.tags['target-service']),
+                clean(stat.tags['target-endpoint'])
+            ].join('.');
```
kriskowal: At some point we figured out that [].join() style interpolation is actually slower than string concatenation in V8.

node/lib/statsd.js
```diff
@@ -0,0 +1,176 @@
...
+            return self.statsd.increment(key, stat.value);
+
+        case 'gauge':
+            return self.statsd.gauge(key, stat.value);
+
+        case 'timing':
+            return self.statsd.timing(key, stat.value);
+
+        default:
+            assert(false, stat.type + ' is not a stats type.');
```
Raynos: let's do `self.channel.logger.error(...)` instead

node/lib/statsd.js
ShanniLi: Is there any reference for this comparison. It is kind of strange since it should be a simple loop through + string concatenation. :)

node/lib/statsd.js
```diff
@@ -0,0 +1,176 @@
...
+}
+
+function getKey(common, stat) {
+    var prefix = ['tchannel', stat.name].join('.');
+    switch (stat.name) {
+        // outbound
+        case 'outbound.calls.sent':
+            return [
+                prefix,
+                clean(stat.tags.service),
```
Raynos: lets default this value to `'unknown'` if its set to undefined;

node/lib/statsd.js
```diff
@@ -0,0 +1,176 @@
...
+                clean(stat.tags['target-service']),
+                clean(stat.tags['target-endpoint'])
+            ].join('.');
+        case 'outbound.calls.success':
+            return [
+                prefix,
+                clean(stat.tags.service),
```
Raynos: ditto for all other `tags.service`

node/lib/statsd.js
```diff
@@ -0,0 +1,176 @@
...
+            ].join('.');
+        case 'inbound.calls.latency':
+            return [
+                prefix,
+                clean(stat.tags['calling-service']),
+                clean(stat.tags.service),
+                clean(stat.tags.endpoint)
+            ].join('.');
+        default:
+        return '';
```
Raynos: indent. Also can we return `'tchannel.bad-stat-object'`

Raynos:
Looking good. Let's get rid of that `assert()` though.

Left other feedback as well.

Raynos:
one last piece of feedback.

Raynos:
lgtm.

node/lib/statsd.js
```diff
@@ -0,0 +1,155 @@
...
+    self.channel.on('stat', onStat);
+
+    function onStat(stat) {
+        self.onStat(stat);
+    }
+}
+
+function clean(str) {
+    if (!str) {
+        return 'unknown';
```
kriskowal: Perhaps take the default as an argument. `outbound.calls.sent.unknown.unknown` is going to be more daunting than `outbound.calls.sent.no-service.no-endpoint`.

node/lib/statsd.js
```diff
@@ -0,0 +1,155 @@
...
+    } else {
+        return str
+            .replace(/:/g, '-')
+            .replace(/\./g, '-')
+            .replace(/{|}/g, '-');
+    }
+}
+
+function getKey(common, stat) {
+    var prefix = ['tchannel', stat.name].join('.');
```
kriskowal: concat

kriskowal: hereafter :cat::cat:

node/lib/statsd.js
```diff
@@ -0,0 +1,155 @@
...
+                clean(stat.tags.service) + '.' +
+                clean(stat.tags.endpoint);
+        default:
+            return 'tchannel.bad-stat-object';
+    }
+}
+
+TChannelStatsd.prototype.onStat = function onStat(stat) {
+    var self = this;
+    var key = getKey(self.channel.statTags, stat);
```
kriskowal: This switch seems superfluous to me. Am I contradicting @raynos? V8 has no trouble with `self.statsd[stat.type](key, stat.value)` and bad on you if it throws because you tried to use a bad stat.type. That’s a programmer error. If you want to be safe, there’s always `(method || missingMethod)`.

Raynos: dynamic dispatch is ok but you still want a default; just do whatever is readable.

I care more about a `switch` for `getKey()`

node/lib/statsd.js
ShanniLi: 'unknown' was @Raynos's suggestion. I extended it to all cases where it could be null/undefined. 
Above being said, I think either approach should be OK. As far as we don't skip fields, it is easy to figure out which field is missing. :)

node/lib/statsd.js
ShanniLi: 'unknown' was @Raynos's suggestion. I extended it to all cases where it could be null/undefined. 
Above being said, I think either approach should be OK. As far as we don't skip fields, it is easy to figure out which field is missing. :)

node/lib/statsd.js
ShanniLi: Missed that. I will fix it. :)

PR #643, opened 2015-06-01
Node: inbound stats & test

ShanniLi:
r @Raynos @kriskowal

Second PR for #639 simplification.

### DIFF
--- a/node/connection_base.js
+++ b/node/connection_base.js
@@ -222,7 +222,6 @@ TChannelConnectionBase.prototype.runHandler = function runHandler(req) {
         'endpoint': String(req.arg1)
     });
 
-    req.latencyStart = self.timers.now();
     self.channel.handler.handleRequest(req, buildResponse);
     function buildResponse(options) {
         return self.buildResponse(req, options);
@@ -239,10 +238,7 @@ TChannelConnectionBase.prototype.buildResponse = function buildResponse(req, opt
     }
     options = extend({
         channel: self.channel,
-        callingService: req.headers.cn,
-        endpoint: String(req.arg1),
-        latencyStart: req.latencyStart,
-        serviceName: req.serviceName
+        inreq: req
     }, options);
     req.res = self.buildOutResponse(req, options);
     req.res.errorEvent.on(onError);
--- a/node/connection_base.js
+++ b/node/connection_base.js
@@ -216,6 +216,13 @@ TChannelConnectionBase.prototype.onReqError = function onReqError(req, err) {
 
 TChannelConnectionBase.prototype.runHandler = function runHandler(req) {
     var self = this;
+    self.channel.inboundCallsRecvdStat.increment(1, {
+        'calling-service': req.headers.cn,
+        'service': req.serviceName,
+        'endpoint': String(req.arg1)
+    });
+
+    req.latencyStart = self.timers.now();
     self.channel.handler.handleRequest(req, buildResponse);
     function buildResponse(options) {
         return self.buildResponse(req, options);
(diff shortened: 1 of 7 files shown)


node/connection_base.js
```diff
@@ -216,6 +216,13 @@ TChannelConnectionBase.prototype.onReqError = function onReqError(req, err) {
 
 TChannelConnectionBase.prototype.runHandler = function runHandler(req) {
     var self = this;
+    self.channel.inboundCallsRecvdStat.increment(1, {
+        'calling-service': req.headers.cn,
+        'service': req.serviceName,
+        'endpoint': String(req.arg1)
+    });
+
+    req.latencyStart = self.timers.now();
```
Raynos: no; this should be on the in_request.js itself as `self.start` field.

node/connection_base.js
```diff
@@ -230,6 +237,13 @@ TChannelConnectionBase.prototype.buildResponse = function buildResponse(req, opt
             state: req.res.state
         }));
     }
+    options = extend({
```
Raynos: move this stuff into `buildOutResponse` instead

Also just pass `inreq` to outResponse as a field if its not already tehre

node/out_response.js
```diff
@@ -34,8 +34,14 @@ function TChannelOutResponse(id, options) {
     self.spanEvent = self.defineEvent('span');
     self.finishEvent = self.defineEvent('finish');
 
+    self.channel = options.channel;
+    self.callingService = options.callingService;
+    self.endpoint = options.endpoint;
     self.logger = options.logger;
+    self.latencyStart = options.latencyStart;
```
Raynos: this class already has `.start` and `.end`; use that.

node/out_response.js
```diff
@@ -34,8 +34,14 @@ function TChannelOutResponse(id, options) {
     self.spanEvent = self.defineEvent('span');
     self.finishEvent = self.defineEvent('finish');
 
+    self.channel = options.channel;
+    self.callingService = options.callingService;
```
Raynos: use an `self.inreq` field instead of `callingService`, `endpoint`, `serviceName`

node/out_response.js
```diff
@@ -89,6 +95,21 @@ TChannelOutResponse.prototype._sendError = function _sendError(codeString, messa
     });
 };
 
+TChannelOutResponse.prototype._emitLatency = function _emitLatency() {
```
Raynos: no need for underscore in method name.

node/out_response.js
```diff
@@ -114,6 +135,15 @@ TChannelOutResponse.prototype.sendParts = function sendParts(parts, isLast) {
...
             break;
+        default:
+            self.channel.logger.error('TChannelOutResponse is in a wrong state', {
+                state: self.state
+            });
+            break;
+    }
+
+    if (isLast) {
+        self._emitLatency();
```
Raynos: not here. just call this in `onFinish()` instead.

node/out_response.js
```diff
@@ -177,6 +207,7 @@ TChannelOutResponse.prototype.sendCallResponseContFrame = function sendCallRespo
 
 TChannelOutResponse.prototype.sendError = function sendError(codeString, message) {
     var self = this;
+    self._emitLatency();
```
Raynos: not here. just call this in `onFinish()`

Raynos:
Finished review; we should shuffle some of this code around.

node/connection_base.js
ShanniLi: Is there any reason to move to buildOutResponse? Note that both self-connection and connection need it.

Raynos: Ah then we can just change it to pass `inreq` in :)

Raynos:
lgtm.

kriskowal:
:ok: Thanks for addressing my feedback about assert.ifError, string concatenation, and default terms in stat names. :+1:

PR #644, opened 2015-06-01
Move outbound call stats into OutRequest

Raynos:
This PR moves a lot of the stats code out of request.js and
directly into out_request.js instead.

It still preserves the semantics of only emitting stats
for a application driven request and not emitting stats
for retries.

This means we get stats for relay handler and for peer to peer
requests like ringpop.

r: @shannili @kriskowal

### DIFF
--- a/node/channel.js
+++ b/node/channel.js
@@ -392,10 +392,13 @@ TChannel.prototype.request = function channelRequest(options) {
         }
     }
 
+    opts.channel = self;
+
     var req = null;
     if (opts.host || // retries are only between hosts
         opts.streamed // streaming retries not yet implemented
     ) {
+        opts.retryCount = 0;
         req = self.peers.request(null, opts);
     } else {
         req = new TChannelRequest(self, opts);
--- a/node/channel.js
+++ b/node/channel.js
@@ -392,8 +392,6 @@ TChannel.prototype.request = function channelRequest(options) {
         }
     }
 
-    opts.channel = self;
-
     var req = null;
     if (opts.host || // retries are only between hosts
         opts.streamed // streaming retries not yet implemented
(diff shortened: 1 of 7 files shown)


node/out_request.js
```diff
@@ -35,9 +36,14 @@ function TChannelOutRequest(id, options) {
     self.responseEvent = self.defineEvent('response');
     self.finishEvent = self.defineEvent('finish');
 
+    assert(options.channel, 'channel required');
+
     self.logger = options.logger;
     self.random = options.random;
     self.timers = options.timers;
+    self.retryCount = options.retryCount;
```
ShanniLi: Why you need to pass in the retryCount? Shouldn't it be set to 0 when out request is initialized?

Raynos: we create one out request for every attempt. So this is either 0, 1, 2 etc. Every retry has a new OutRequest object.

ShanniLi: do we always retry on the same peer?

Raynos: never retry on same peer.

ShanniLi:
lgtm

node/channel.js
```diff
@@ -396,6 +396,7 @@ TChannel.prototype.request = function channelRequest(options) {
     if (opts.host || // retries are only between hosts
         opts.streamed // streaming retries not yet implemented
     ) {
+        opts.retryCount = 0;
```
kriskowal: I suppose there’s no hope of this object ever being a hidden class.

kriskowal:
:ok:

node/channel.js
Raynos: it can if we pre-allocate it and fix all the thingies.

PR #646, opened 2015-06-02
replace / with - in statsd

Raynos:
This is to fix the arg1 for ringpop which uses /.

r: @shannili @kriskowal


ShanniLi:
lgtm

node/lib/statsd.js
```diff
@@ -39,8 +39,10 @@ function clean(str, field) {
     if (!str) {
         return 'no-' + field;
     } else {
+        // TODO generalize
```
kriskowal: `[:/.{}]`

PR #648, opened 2015-06-03
Encode and decode JSON directly rather than using an intermediary buffer

prashantv:


mmihic:
lgtm.  Not sure why it was going through a buffer tbh.

PR #649, opened 2015-06-04
update zipkin trace code based on new thrift idl

junchaowu:
@breerly @abhinav


python/tchannel/zipkin/formatters.py
```diff
@@ -97,6 +98,8 @@ def json_formatter(traces, *json_args, **json_kwargs):
 
 
 def ipv4_to_int(ipv4):
+    if ipv4 == 'localhost':
+        ipv4 = '127.0.0.1'
```
breerly: Does this get converted to an integer before being returned?

junchaowu: inet_aton is a system api to convert api to a binary form. http://linux.die.net/man/3/inet_aton. It doesn't support 'localhost'

python/tchannel/zipkin/tracers.py
```diff
@@ -126,27 +133,25 @@ class TChannelZipkinTracer(object):
...
             try:
-                self.client.send(
-                    InMemStream('span'),
-                    InMemStream(),
-                    InMemStream(base64_thrift_formatter(trace, annotations)),
-                    traceflag=False,  # disable tracing for span uploading
-                )
-            except:
-                pass
+                self.client.submit(thrift_formatter(trace, annotations))
```
abhinav: This is an asynchronous operation. It'll return a future. You won't get exceptions here. Add a done callback to log exceptions in case of failure. (This operation doesn't need to be a coroutine.)

python/tchannel/zipkin/tracers.py
```diff
@@ -126,27 +133,25 @@ class TChannelZipkinTracer(object):
...
+            A list contains hyperbahn instances' ip addresses
         """
         self._tchannel = tchannel
-        self.client = tchannel.request(
-            "tcollector"
-        )
+        self.client = TCollectorClient(self._tchannel)
+
+        if routers:
+            hyperbahn.advertise(self._tchannel, 'tcollector', routers)
```
abhinav: Why are you advertising as 'tcollector'? That's the service you're calling. You only advertise yourself as the service you provide.

abhinav: Also, the user is responsible for advertising. You don't need to do it regardless. You don't need a list of routers either. Let the TChannel pick the peer. For some level of determinism you could accept an optional hostport argument and pass that over to the TCollectorClient (it accepts an optional hostport if you want it to send requests to a specific hostport instead of letting the tchannel pick a peer).

python/tchannel/zipkin/tracers.py
```diff
@@ -114,10 +118,13 @@ def __init__(self, logger):
     def record(self, traces):
         for (trace, annotations) in traces:
             self._logger.info(
-                base64_thrift_formatter(trace, annotations)
+                thrift_formatter(trace, annotations, isbased64=True)
             )
 
 
+TCollectorClient = client_for('tcollector', TCollector)
```
abhinav: I wonder if we want to be able to provide different service names. Like, if there are different implementations of the service in different datacenters, the user may want to provide different names.

Don't do anything about that just yet. This may require a change on the Thrift client implementation side.

Can you add a TODO here?

abhinav:
:+1:

PR #650, opened 2015-06-04
Add support to go to send "as" and "cn" headers.

prashantv:


golang/calloptions.go
```diff
@@ -0,0 +1,48 @@
...
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+// Format is the arg scheme used for a specific call.
+type Format string
+
+// The list of formats supported by tchannel.
+const (
```
Raynos: We support "raw" as well

prashantv: Added support for "raw" and using that as the default if the format is not specified, thanks. I'll update the protocol doc separately.

golang/channel.go
```diff
@@ -166,7 +166,7 @@ func (ch *Channel) PeerInfo() PeerInfo {
 // BeginCall starts a new call to a remote peer, returning an OutboundCall that can
 // be used to write the arguments of the call
 // TODO(mmihic): Support CallOptions such as format, request specific checksums, retries, etc
```
mmihic: You can get rid of this TODO now :)

golang/inbound.go
```diff
@@ -141,6 +143,15 @@ func (call *InboundCall) Operation() []byte {
 	return call.operation
 }
 
+// Headers returns the call headers from the request.
+func (call *InboundCall) Format() Format {
+	return Format(call.headers[ArgScheme])
+}
+
+func (call *InboundCall) CallerName() string {
```
mmihic: godoc

golang/messages.go
```diff
@@ -119,6 +119,17 @@ type initRes struct {
 
 func (m *initRes) messageType() messageType { return messageTypeInitRes }
 
+// Known transport header keys for call requests.
+const (
+	ArgScheme            = "as"
```
mmihic: Godoc or unexport (you should be getting lint warnings on these).

Also, if you leave them exported, do you want to define a type for CallHeaderName?

golang/outbound.go
```diff
@@ -56,12 +56,17 @@ func (c *Connection) beginCall(ctx context.Context, serviceName string) (*Outbou
 		return nil, err
 	}
 
+	headers := callHeaders{
+		CallerName: c.localPeerInfo.ProcessName,
```
mmihic: This should be the name of the service making the call, once you've added the Hyperbahn registration.  It's used for circuit breaking and backrouting, so I don't think we ever want it to be the process name (which is an ephemeral unroutable entity).

mmihic:
lgtm, only minor comments.  I had been thinking we'd have higher level protocol specific APIs a la the nodejs <a href="https://github.com/uber/tchannel/tree/master/node/as">as interfaces</a>, so setting this header would mainly be the responsibility of a tchannel-provided API.  Is that your thought as well?

golang/channel.go
prashantv: Done

golang/inbound.go
prashantv: Updated.

golang/outbound.go
prashantv: Oh interesting, so do we need to have the caller specify this explicitly? What is the value when a client makes this call (and the client does not have any services exposed)?

golang/calloptions.go
mmihic: Oh these need to be godoc'ed as well (we're generally trying to have no golint warnings, and lint warns on exported consts / funcs / types without godoc)

prashantv:
Thanks for the review. Right now, I'm planning to right a TProtocol that can be used for the generated Thrift client, and this TProtocol will be setting the header (and doing all the interaction with tchannel).

golang/calloptions.go
prashantv: golint has no warnings on this file since I have a comment on the block :) Since the values are pretty self explanatory,  I don't think it's worth having separate godoc comments for each value - what do you think?

golang/outbound.go
Raynos: Every service has a name.

Cn, realtime-API, mobile-driver-client, mobile-rider-client

In node we will make this a mandatory option in the tchannel constructor.

mmihic: Once you do the hyperbahn integration, the channel itself will have a service name associated with it that it has advertised to hyperbahn.  That's the value that should go in here.  So it wouldn't come from the caller directly, it'd come from the channel (and is probably an option passed in to the channel - I'd look at the node implementation to see where they pass in the service name to advertise).  Note that this service name is an external route; you can actually have multiple thrift services registered under the same advertised service name (the thrift service that is being targeted gets passed as part of (I think) arg1).

I'm not sure how we are handling pure clients TBH.  These probably still have service names, even though they don't expose any services.  @Raynos?

golang/calloptions.go
mmihic: Ha, I didn't know you could just comment the block.  That totally works for me.

golang/outbound.go
Raynos: We manually pass the service name to the hyperbahn client. I.e. The application developer does it

All node services implement two service names, ring pop and my-app.

We just make the application developers do more configuration.

prashantv: Ok cool, I've added a serviceName string to the TChannel constructor, and I use that in the cn header now.

PR #652, opened 2015-06-04
2.0.0-IDEA: channel: expose AsThrift and AsJSON

Raynos:
This allows us to author "smart client" libraries that
do not depend on a static version of tchannel.

These libraries would take an instance of a sub channel
as an argument and use `channel.TChannelAsThrift` to create
their thrift encoder/decoder.

This allows us to develop applications which only depend
on a single copy of tchannel.

r: @jcorbin @rf @kriskowal

### DIFF
--- a/node/channel.js
+++ b/node/channel.js
@@ -40,6 +40,8 @@ var TChannelRequest = require('./request');
 var TChannelServiceNameHandler = require('./service-name-handler');
 var errors = require('./errors');
 
+var TChannelAsThrift = require('./as/thrift');
+var TChannelAsJSON = require('./as/json');
 var TChannelConnection = require('./connection');
 var TChannelPeers = require('./peers');
 var TChannelServices = require('./services');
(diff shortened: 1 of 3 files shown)


rf:
:ok_hand:

kriskowal:
I don’t get how this is necessary or sufficient to make applications that depend only one one copy of TChannel. Can you elaborate on the before and after for an app? I generally object to this kind of coupling, which forces a dependee to load and execute more than they need. We’re *never* going to Browserify this, right?

Raynos:
@kriskowal 

Previously:

// x-client.js
// some fixed version of tchannel
var TChannelAsThrift = require('tchannel/as/thrift');

function XClient(opts) {
  // some completely different dynamic version of tchannel
  var tchannel = opts.channel
}
```

Because the as/thrift implementation is in tchannel you end up having two copies of tchannel if your doing `npm install x-client`

The only way around this is to manually go into `x-client` and upgrade tchannel. 

The other way around this is to publish tchannel-as-thrift as a seperate package with it's own major version number.

What we want is:

 - only one version of tchannel in the shrinkwrap
 - a non-tedious process. Currently in autobahn an upgrade of tchannel requires first updating 3 dependencies to latest tchannel; then updating autobahn; i.e. i have 4 copies of tchannel and doing manual deduplication by cutting new versions of ringpop, node-client and tcurl.

kriskowal:
Got it. The problem is that the client imports tchannel on your behalf. The problem can be solved either by requiring tchannel and tchannel/as/thrift and passing both into the client, or this because it’s simpler.

:ok:

PR #654, opened 2015-06-05
Egress circuit breaker

kriskowal:
Introduces a circuit health state machine for each callerName::serviceName::endpointName combination, integrated into the egress node for each service name.

- [x] tests

### DIFF
--- a/node/circuits.js
+++ b/node/circuits.js
@@ -76,6 +76,14 @@ EndpointCircuits.prototype.getCircuit = function getCircuit(callerName, serviceN
     return circuit;
 };
 
+EndpointCircuits.prototype.getCircuitTuples = function () {
+    var self = this;
+    return Object.keys(self.circuitsByEndpointName).map(function (endpointName) {
+        var circuit = self.circuitsByEndpointName[endpointName];
+        return [circuit.callerName, circuit.serviceName, circuit.endpointName];
+    });
+};
+
 function ServiceCircuits(root) {
     var self = this;
     self.root = root;
@@ -90,6 +98,14 @@ ServiceCircuits.prototype.getCircuit = function getCircuit(callerName, serviceNa
         self.circuitsByCallerName['$' + callerName] = circuits;
     }
     return circuits.getCircuit(callerName, serviceName, endpointName);
+};
+
+ServiceCircuits.prototype.getCircuitTuples = function getCircuitTuples() {
+    var self = this;
+    var circuits = Object.keys(self.circuitsByCallerName).map(function (callerName) {
+        return self.circuitsByCallerName[callerName].getCircuitTuples();
+    });
+    return Array.prototype.concat.apply([], circuits);
 };
 
 function Circuits(options) {
@@ -115,6 +131,14 @@ Circuits.prototype.getCircuit = function getCircuit(callerName, serviceName, end
         self.circuitsByServiceName['$' + serviceName] = circuits;
     }
     return circuits.getCircuit(callerName, serviceName, endpointName);
+};
+
+Circuits.prototype.getCircuitTuples = function getCircuitTuples() {
+    var self = this;
+    var circuits = Object.keys(self.circuitsByServiceName).map(function (serviceName) {
+        return self.circuitsByServiceName[serviceName].getCircuitTuples();
+    });
+    return Array.prototype.concat.apply([], circuits);
 };
 
 Circuits.prototype.handleRequest = function handleRequest(req, buildRes, nextHandler) {
--- a/node/circuits.js
+++ b/node/circuits.js
@@ -127,7 +127,7 @@ Circuits.prototype.getCircuit = function getCircuit(callerName, serviceName, end
     var self = this;
     var circuits = self.circuitsByServiceName['$' + serviceName];
     if (!circuits) {
-        circuits = new EndpointCircuits(self);
+        circuits = new ServiceCircuits(self);
         self.circuitsByServiceName['$' + serviceName] = circuits;
     }
     return circuits.getCircuit(callerName, serviceName, endpointName);
@@ -151,8 +151,8 @@ Circuits.prototype.handleRequest = function handleRequest(req, buildRes, nextHan
     if (!serviceName) {
         return buildRes().sendError('BadRequest', 'All requests must have a service name');
     }
-    return req.withArg1(function withArg1(endpointName) {
-        var circuit = self.getCircuit(callerName, serviceName, endpointName);
+    return req.withArg1(function withArg1() {
+        var circuit = self.getCircuit(callerName, serviceName, String(req.arg1));
         return circuit.handleRequest(req, buildRes, nextHandler);
     });
 };
--- a/node/circuits.js
+++ b/node/circuits.js
@@ -1,0 +1,213 @@
+// Copyright (c) 2015 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+"use strict";
+
+var inherits = require('util').inherits;
+var States = require('./states');
+var EventEmitter = require('./lib/event_emitter');
+var StateMachine = require('./state_machine');
+var errors = require('./errors');
+
+// The each circuit uses the circuits collection as the "nextHandler" for
+// "shouldRequest" to consult.  Peers use this hook to weight peers both by
+// healthy and other factors, but the circuit only needs to know about health
+// before forwarding.
+
+function AlwaysShouldRequestHandler() { }
+
+AlwaysShouldRequestHandler.prototype.shouldRequest = function shouldRequest() {
+    return true;
+};
+
+var alwaysShouldRequestHandler = new AlwaysShouldRequestHandler();
+
+//  circuit = circuits
+//      .circuitsByServiceName[serviceName]
+//      .circuitsByCallerName[callerName]
+//      .circuitsByEndpointName[endpointName]
+
+function EndpointCircuits(root) {
+    var self = this;
+    self.root = root;
+    self.circuitsByEndpointName = {};
+}
+
+EndpointCircuits.prototype.getCircuit = function getCircuit(callerName, serviceName, endpointName) {
+    var self = this;
+    var circuit = self.circuitsByEndpointName['$' + endpointName];
+    if (!circuit) {
+        circuit = new Circuit();
+        circuit.callerName = callerName;
+        circuit.serviceName = serviceName;
+        circuit.endpointName = endpointName;
+        circuit.shouldRequestOptions = self.root.shouldRequestOptions;
+        var stateOptions = self.root.stateOptions;
+        circuit.stateOptions = {
+            stateMachine: circuit,
+            nextHandler: stateOptions.nextHandler,
+            timers: stateOptions.timers,
+            random: stateOptions.random,
+            period: stateOptions.period,
+            maxErrorRate: stateOptions.maxErrorRate,
+            minRequests: stateOptions.minRequests,
+            probation: stateOptions.probation
+        };
+        circuit.setState(States.HealthyState);
+        self.circuitsByEndpointName['$' + endpointName] = circuit;
+    }
+    return circuit;
+};
+
+function ServiceCircuits(root) {
+    var self = this;
+    self.root = root;
+    self.circuitsByCallerName = {};
+}
+
+ServiceCircuits.prototype.getCircuit = function getCircuit(callerName, serviceName, endpointName) {
+    var self = this;
+    var circuits = self.circuitsByCallerName['$' + callerName];
+    if (!circuits) {
+        circuits = new EndpointCircuits(self.root);
+        self.circuitsByCallerName['$' + callerName] = circuits;
+    }
+    return circuits.getCircuit(callerName, serviceName, endpointName);
+};
+
+function Circuits(options) {
+    var self = this;
+    self.circuitsByServiceName = {};
+    self.stateOptions = {
+        nextHandler: alwaysShouldRequestHandler,
+        timers: options.timers,
+        random: options.random,
+        period: options.period,
+        maxErrorRate: options.maxErrorRate,
+        minRequests: options.minRequests,
+        probation: options.probation
+    };
+    self.shouldRequestOptions = {};
+}
+
+Circuits.prototype.getCircuit = function getCircuit(callerName, serviceName, endpointName) {
+    var self = this;
+    var circuits = self.circuitsByServiceName['$' + serviceName];
+    if (!circuits) {
+        circuits = new EndpointCircuits(self);
+        self.circuitsByServiceName['$' + serviceName] = circuits;
+    }
+    return circuits.getCircuit(callerName, serviceName, endpointName);
+};
+
+Circuits.prototype.handleRequest = function handleRequest(req, buildRes, nextHandler) {
+    var self = this;
+    // Default the caller name.
+    // All callers that fail to specifiy a cn share a circuit for each sn:en
+    // and fail together.
+    var callerName = req.headers.cn || 'yunocn';
+    var serviceName = req.serviceName;
+    if (!serviceName) {
+        return buildRes().sendError('BadRequest', 'All requests must have a service name');
+    }
+    return req.withArg1(function withArg1(endpointName) {
+        var circuit = self.getCircuit(callerName, serviceName, endpointName);
+        return circuit.handleRequest(req, buildRes, nextHandler);
+    });
+};
+
+// Called upon membership change to collect services that the corresponding
+// exit node is no longer responsible for.
+Circuits.prototype.updateServices = function updateServices(managesService) {
+    var self = this;
+    var serviceNames = Object.keys(self.circuitsByServiceName);
+    for (var index = 0; index < serviceNames.length; index++) {
+        var serviceName = serviceNames[index];
+        if (!managesService(serviceName)) {
+            delete self.circuitsByServiceName[serviceName];
+        }
+    }
+};
+
+function Circuit() {
+    var self = this;
+    self.circuitName = null;
+    self.callerName = null;
+    self.serviceName = null;
+    self.endpointName = null;
+    self.shouldRequestOptions = null;
+    self.stateOptions = null;
+    StateMachine.call(self);
+    EventEmitter.call(self);
+    self.stateChangedEvent = self.defineEvent('stateChanged');
+}
+
+inherits(Circuit, EventEmitter);
+
+Circuit.prototype.setState = StateMachine.prototype.setState;
+
+Circuit.prototype.handleRequest = function handleRequest(req, buildRes, nextHandler) {
+    var self = this;
+    if (self.state.shouldRequest(req, self.shouldRequestOptions)) {
+        return self.monitorRequest(req, buildRes, nextHandler);
+    } else {
+        return buildRes().sendError('Declined', 'Service is not healthy');
+    }
+};
+
+Circuit.prototype.monitorRequest = function monitorRequest(req, buildRes, nextHandler) {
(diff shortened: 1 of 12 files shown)


node/circuits.js
```diff
@@ -0,0 +1,154 @@
...
+    }
+    return req.withArg1(function withArg1(endpointName) {
+        var circuit = self.getCircuit(callerName, serviceName, endpointName);
+        return circuit.handleRequest(req, buildRes, nextHandler);
+    });
+};
+
+// Called upon membership change to collect services that the corresponding
+// egress node is no longer responsible for.
+Circuits.prototype.updateServices = function updateServices(serviceNames) {
```
jcorbin: So the API we have now:
```
update :: (names :: List<name>) -> void
```

But we then use names to only ever answer "do we care about this one?" questions.

Can we instead have the api explicate that usage pattern:
```
update :: (owns :: (name) -> Boolean) -> void
```

Where owns is some predicate function?

node/circuits.js
```diff
@@ -0,0 +1,154 @@
...
+    var circuit = self.circuits[circuitName];
+    if (!circuit) {
+        circuit = new Circuit();
+        circuit.circuitName = circuitName;
+        circuit.callerName = callerName;
+        circuit.serviceName = serviceName;
+        circuit.endpointName = endpointName;
+        circuit.shouldRequestOptions = self.shouldRequestOptions;
+        circuit.stateOptions = self.stateOptions;
+        self.circuits[circuitName] = circuit;
```
jcorbin: should also populate `circuitsByServiceName`

node/circuits.js
kriskowal: My intuition is YAGNI. We could circle back and evolve it if necessary. Are you imagining a specific scenario where a different predicate will be useful?

jcorbin:
I question why we have a flat collection, and then a secondary index by service name?  Why not just have `Circuits -> ServiceCircuits -> EndpointCircuits`?

FWIW, adding inreq.withArg1 and the dropping of carried options from hyperbahn service proxy jumped out at me as easily separable ahead of this PR.

node/circuits.js
```diff
@@ -0,0 +1,154 @@
...
+        var selfServiceName = self.circuitsByServiceName[index];
+        if (serviceNames.indexOf(selfServiceNames) < 0) {
+            self.collectService(selfServiceName);
+        }
+    }
+};
+
+Circuits.prototype.collectService = function collectService(serviceName) {
+    var self = this;
+    var circuits = self.circuitsByServiceName[serviceName];
```
kriskowal: Need to delete circuitsByServiceName[serviceName] here.

node/circuits.js
jcorbin: YAGNI only really applies to added complexity.  This is simply a design choice of where to place logic with little added cost (either today, or carrying) that makes the code easier to change later.

kriskowal:
Good points both. I’ll cherry pick some of the separable PR’s out of the leading commits. A three-deep data structure also bypasses the appropriate-escaping question.

node/circuits.js
```diff
@@ -0,0 +1,154 @@
...
+    StateMachine.call(self);
+    self.setState(States.HealthyState);
+}
+
+Circuit.prototype.handleRequest = function handleRequest(req, buildRes, nextHandler) {
+    var self = this;
+    if (self.state.shouldRequest(req, self.shouldRequestOptions)) {
+        return self.monitorRequest(req, buildRes, nextHandler);
+    } else {
+        return buildRes().sendError('Declined', 'Service is not healthy');
```
Raynos: Is this really a declined ? Maybe `Busy` would be better.

node/circuits.js
```diff
@@ -0,0 +1,154 @@
...
+        self.circuits[circuitName] = circuit;
+    }
+    return circuit;
+};
+
+Circuits.prototype.handleRequest = function handleRequest(req, buildRes, nextHandler) {
+    var self = this;
+    var callerName = req.headers.cn;
+    if (!callerName) {
+        return buildRes().sendError('BadRequest', 'All requests must have the cn (caller service name) header');
```
Raynos: This is a breaking change. This cannot be merged until we transition people over to having the cn header.

node/hyperbahn/service_proxy.js
```diff
@@ -166,7 +173,17 @@ function createServiceChannel(serviceName) {
         }
     }
 
-    svcchan.handler = new RelayHandler(svcchan);
+    var handler = new RelayHandler(svcchan);
+
+    // Decorate a circuit health monitor to egress request handlers.
+    if (mode === 'exit' && self.circuits) {
+        handler = new ServiceHealthProxy({
```
Raynos: do we want to go this proxy handler approach ?

We could just have `RelayHandler` itself hook directly into the circuits code. This would avoid a bunch of wrapping and closures.

Raynos:
We cannot make the CN required in this PR;

Making the cn header required should be a seperate PR.

kriskowal:
@raynos, the circuit breaker is only consulted if the service proxy is configured with one. This is not a breaking change, but configuring hyperbahn to use it, certainly would be.

However, we could default the caller name to "yunocn" and all legacy services would share a circuit.

node/circuits.js
kriskowal: Thanks for asking. I reconsulted the :sparkles::book: protocol document :book::sparkles:, and I believe we want to avoid retries, which I believe actually means we should use UnexpectedError.

mmihic: That's unfortunate.  We should add a new code for this.

Raynos:
true; configuring hyperbahn to use it is the breaking change. So its upto autobahn to instantiate a Circuits() and pass it in.

kriskowal:
Per feedback,

I’ve created defaults for cn and sn for the purposes of circuit breaking. These lump services together on a shared circuit if they fail to specify. Specifying thereby just isolates the circuit so you’re less likely to suffer collateral circuit breaking.

Added a tiered data structure for circuits. This makes disposing of all circuits with a shared callee sn trivial, and avoids the need to escape key components.

kriskowal:
This is reviewable, but needs a task. See the circuit-health-test branch if you need to pick up the loose ends before I return.

node/circuits.js
kriskowal: Running with declined. Client should retry, but will continue to be declined but once a circuit checking interval. It is up to the client to back off.

node/hyperbahn/service_proxy.js
kriskowal: investigating

kriskowal: Does not look like this approach would avoid any closures. Using a proxy allows us to conditionally inject the circuit breaker, so we can use RelayHandler for ingress and RelayHandler+ServiceProxy for egress and keeps them loosely coupled. ServiceHealthProxy is very thin.

kriskowal:
r @raynos @jcorbin @breerly

node/test/lib/relay_network.js
```diff
@@ -46,13 +46,18 @@ function RelayNetwork(options) {
     self.serviceNames = options.serviceNames || ['alice', 'bob', 'charlie'];
     self.kValue = options.kValue || 2;
     self.createCircuits = options.createCircuits || noop;
-    self.clusterOptions = options.cluster || {};
```
Raynos: this will break other callers; verify please.

node/test/lib/relay_network.js
```diff
@@ -199,9 +204,11 @@ RelayNetwork.prototype.setCluster = function setCluster(cluster) {
                         cn: serviceName
                     },
                     hasNoParent: true
-                },
-                peers: self.topology[serviceName] // TODO this really ought to be linked directly to the root peers
+                }
             });
+            // The subchannel's peers are linked to the peer list as managed by
+            // incoming connections from the relays.
+            subChannel.peers = channel.peers;
```
Raynos: why this line; this is not something we do in autobahn.

node/test/lib/relay_network.js
kriskowal: Unless RelayNetwork is already used outside of TChannel, all call sites are verified.

node/errors.js
```diff
@@ -580,3 +580,18 @@ module.exports.classify = function classify(err) {
...
+
+var symptoms = {
+    'BadRequest': false, // not an indicator of bad health
+    'Cancelled': false, // not an indicator of bad health
+    'Timeout': true, // TODO throttle
+    'Busy': true, // TODO throttle
+    'Declined': true,
+    'UnexpectedError': true,
+    'NetworkError': true,
+    'FatalProtocolError': true
```
Raynos: UnHealthy ?

Raynos: ProtocolError ?

node/circuits.js
```diff
@@ -0,0 +1,213 @@
...
+    StateMachine.call(self);
+    EventEmitter.call(self);
+    self.stateChangedEvent = self.defineEvent('stateChanged');
+}
+
+inherits(Circuit, EventEmitter);
+
+Circuit.prototype.setState = StateMachine.prototype.setState;
+
+Circuit.prototype.handleRequest = function handleRequest(req, buildRes, nextHandler) {
```
Raynos: Can we move this logic into `choosePeer()` instead ?

node/circuits.js
```diff
@@ -0,0 +1,213 @@
...
+    var self = this;
+    if (self.state.shouldRequest(req, self.shouldRequestOptions)) {
+        return self.monitorRequest(req, buildRes, nextHandler);
+    } else {
+        return buildRes().sendError('Declined', 'Service is not healthy');
+    }
+};
+
+Circuit.prototype.monitorRequest = function monitorRequest(req, buildRes, nextHandler) {
```
Raynos: can this logic be pushed peer.js next to all the other monitoring logic ?

node/hyperbahn/service_proxy.js
```diff
@@ -179,6 +191,14 @@ function updateServiceChannels() {
             self.updateServiceChannel(chan);
         }
     }
+
+    if (self.circuits) {
+        self.circuits.updateServices(managesService);
+    }
+
+    function managesService(serviceName) {
+        return self.egressNodes.isExitFor(serviceName);
```
Raynos: pass `egressNodes` into circuits to avoid closure.

node/test/circuits.js
```diff
@@ -0,0 +1,97 @@
...
+var debugLogtron = require('debug-logtron');
+var RelayNetwork = require('./lib/relay_network.js');
+var Circuits = require('../circuits');
+var States = require('../states');
+var MockTimers = require('time-mock');
+
+RelayNetwork.test('should switch to unhealthy', {
+    timers: new MockTimers(1e9),
+    clusterOptions: {
+        logger: debugLogtron('tchannel', {enabled: false}),
```
Raynos: should not be needed; grab logger from `cluster.logger` from alloc cluster.

node/test/circuits.js
```diff
@@ -0,0 +1,97 @@
...
+
+'use strict';
+
+var debugLogtron = require('debug-logtron');
+var RelayNetwork = require('./lib/relay_network.js');
+var Circuits = require('../circuits');
+var States = require('../states');
+var MockTimers = require('time-mock');
+
+RelayNetwork.test('should switch to unhealthy', {
```
Raynos: We need more tests.

 - Ensure that circuit breaking does not interfere with cancel & bad request
 - Ensure that circuit breaking works for other exceptions like Timeout and Network
 - Ensure that circuits do not interfere with each other. i.e. multiple service names
 - Ensure that circuits do not interfere with each other. i.e. multiple cn
 - Ensure that circuits do not interfere with each other. i.e. multiple endpoints

You are adding a ton of proxies, closures and hash table lookups in the critical path; I will rip through all this code in my perf work.

I strongly recommend you write a bunch of tests to gaurd us from regressions.

node/test/lib/relay_network.js
Raynos: test/lib/hyperbahn-cluster.js`

```
    self.relayNetwork = RelayNetwork({
        numRelays: self.size,
        numInstancesPerService: 1,
        kValue: 5,
        serviceNames: ['bob', 'steve', 'mary'],
        cluster: options.cluster
    });
```

node/test/lib/relay_network.js
kriskowal: We should do this in Hyperbahn. The premise of using a ring is that its shape will adapt. The static list of hosts is only a seed list for bootstrapping. Once a service has found a member of the ring to talk to, the egress affinity peers on the ring will call back and may or may not be members of the seed list. The service should be sending all requests back to the ring on whatever peers the ring creates.

node/errors.js
kriskowal: Good catches.

node/test/lib/relay_network.js
Raynos: Can we just rewrite the peers code to be well behaved. its really confusing.

node/circuits.js
kriskowal: I can’t envision that working. We need a hook somewhere that enables the circuit breaker to observe or intercept the request and response. This will still be true with the pass-through frame relay when we have it. All that I suspect will change is that the content of arg2 and arg3 will not be retained on the request and response, which this adapter does not need anyway.

node/hyperbahn/service_proxy.js
kriskowal: Addressed.

node/circuits.js
kriskowal: We’ll have to pair on that tomorrow. I don’t see a way off hand.

node/circuits.js
```diff
@@ -0,0 +1,238 @@
...
+        };
+        circuit.setState(States.HealthyState);
+        self.circuitsByEndpointName['$' + endpointName] = circuit;
+    }
+    return circuit;
+};
+
+EndpointCircuits.prototype.getCircuitTuples = function () {
+    var self = this;
+    return Object.keys(self.circuitsByEndpointName).map(function (endpointName) {
```
jcorbin: I thought we were discouraging `.map`?

kriskowal: We are. This method is only used in tests.

node/circuits.js
```diff
@@ -0,0 +1,238 @@
...
+    if (!circuits) {
+        circuits = new EndpointCircuits(self.root);
+        self.circuitsByCallerName['$' + callerName] = circuits;
+    }
+    return circuits.getCircuit(callerName, serviceName, endpointName);
+};
+
+ServiceCircuits.prototype.getCircuitTuples = function getCircuitTuples() {
+    var self = this;
+    var circuits = Object.keys(self.circuitsByCallerName).map(function (callerName) {
```
jcorbin: eesh, this looks good and expensive...  I'd think some sort of an injected-iterator api would be better, e.g. each collection has a `.eachCircuitTuple` or even just `.eachCircuit` method.

node/circuits.js
kriskowal: That being said, we’re also discouraging consideration of special cases…

node/circuits.js
```diff
@@ -0,0 +1,238 @@
...
+        return self.circuitsByCallerName[callerName].getCircuitTuples();
+    });
+    return Array.prototype.concat.apply([], circuits);
+};
+
+function Circuits(options) {
+    var self = this;
+    self.circuitsByServiceName = {};
+    self.stateOptions = {
```
jcorbin: Okay, now I've seen it used in two places, should we get a prototype for this? constructor signature `function(stateMachine, options)` and in this case we'd need to set .nextHandler after construction?

node/circuits.js
```diff
@@ -0,0 +1,238 @@
...
+function AlwaysShouldRequestHandler() { }
+
+AlwaysShouldRequestHandler.prototype.shouldRequest = function shouldRequest() {
+    return true;
+};
+
+var alwaysShouldRequestHandler = new AlwaysShouldRequestHandler();
+
+//  circuit = circuits
+//      .circuitsByServiceName[serviceName]
```
jcorbin: add a second column comment noting each class type at each level ... would've really helped explain things first time I had read this... plus a comment above saying: "Here's how the following classes related:"

node/circuits.js
```diff
@@ -0,0 +1,238 @@
...
+    });
+    return Array.prototype.concat.apply([], circuits);
+};
+
+Circuits.prototype.handleRequest = function handleRequest(req, buildRes, nextHandler) {
+    var self = this;
+    // Default the caller name.
+    // All callers that fail to specifiy a cn share a circuit for each sn:en
+    // and fail together.
+    var callerName = req.headers.cn || 'yunocn';
```
jcorbin: while this default greatly amuses me also, why not "unspecified" or "unknown"?

jcorbin: Heck, even "undefined" ;-)

node/circuits.js
```diff
@@ -0,0 +1,238 @@
...
+};
+
+// Called upon membership change to collect services that the corresponding
+// exit node is no longer responsible for.
+Circuits.prototype.updateServices = function updateServices() {
+    var self = this;
+    var serviceNames = Object.keys(self.circuitsByServiceName);
+    for (var index = 0; index < serviceNames.length; index++) {
+        var serviceName = serviceNames[index];
+        if (!self.egressNodes.isExitFor(serviceName)) {
```
jcorbin: If I don't miss my mark, this is the only place where `circuits.egressNodes` is used, it would seem less coupled to have this function simply take an "exists" predicate.

node/circuits.js
```diff
@@ -0,0 +1,238 @@
...
+            delete self.circuitsByServiceName[serviceName];
+        }
+    }
+};
+
+function Circuit() {
+    var self = this;
+    self.circuitName = null;
```
jcorbin: Should we have string defaults for string fields?

node/circuits.js
```diff
@@ -0,0 +1,238 @@
...
+    var serviceNames = Object.keys(self.circuitsByServiceName);
+    for (var index = 0; index < serviceNames.length; index++) {
+        var serviceName = serviceNames[index];
+        if (!self.egressNodes.isExitFor(serviceName)) {
+            delete self.circuitsByServiceName[serviceName];
+        }
+    }
+};
+
+function Circuit() {
```
jcorbin: I'd really prefer to see the constructor take at least the three primary parts of identity (the three names) and fill them in, and push down the caller name default ("yunocn" modulo shed) here.

node/circuits.js
```diff
@@ -0,0 +1,238 @@
...
+    var self = this;
+    self.circuitName = null;
+    self.callerName = null;
+    self.serviceName = null;
+    self.endpointName = null;
+    self.shouldRequestOptions = null;
+    self.stateOptions = null;
+    StateMachine.call(self);
```
jcorbin: Both super constructor calls should frontload before we fill in sub class fields?

node/circuits.js
```diff
@@ -0,0 +1,238 @@
...
+};
+
+Circuit.prototype.monitorRequest = function monitorRequest(req, buildRes, nextHandler) {
+    var self = this;
+
+    self.state.onRequest(req);
+
+    req.errorEvent.on(onError);
+
+    function monitorBuildRes(options) {
```
jcorbin: that's a big function, any utility to pulling out to a method with a one-line binding closure here?

node/circuits.js
kriskowal: Did previously. @Raynos decried the closure allocation.

node/circuits.js
```diff
@@ -0,0 +1,238 @@
...
+        function onError(err) {
+            self.state.onRequestError(err);
+        }
+
+        function onFinish() {
+            // TODO distingiush res.ok?
+            // note that incoming requests do not have responseEvent and clear out
+            // their response upon finish.
+            if (errors.isUnhealthy(res.codeString)) {
+                self.state.onRequestUnhealthy();
```
jcorbin: seems bit of a lack that the res doesn't get passed along for visibility

node/circuits.js
kriskowal: Revised this in working copy. Top level provides getCircuitTuples, underlying layers have collectCircuitTuples(tuples) that mutate the given array. That should address the cost, although again, this is not called outside of tests.

node/errors.js
```diff
@@ -580,3 +580,22 @@ module.exports.classify = function classify(err) {
             return null;
     }
 };
+
+// To determine whether a circuit should break for each response code.
+// TODO consider whether to keep a circuit healthy if a downstream circuit is
+// unhealthy.
+var symptoms = {
```
jcorbin: This doesn't really belong in the errors module... really it should go with where we declare the codes.  Presently that's in `v2/error_response.js`.  Pulling code definition out of the v2 area into a less version-specific one has long been a discussed improvement, but I'm not sure if it's warranted.  However it would force you to take a `.../v2/...` require above, which is Working As Intended as it underscores the coupling to version-specific semantics.

Finally, where-ever it goes, it'd be great if we got static analysis test for "you covered them all" similar to our error classifer.

node/hyperbahn/service_proxy.js
jcorbin: why not just `function ServiceHealthProxy(nextHandler, circuits)` then?

node/peers.js
```diff
@@ -91,8 +92,7 @@ TChannelPeers.prototype.add = function add(hostPort, options) {
         if (self.channel.topChannel) {
             peer = self.channel.topChannel.peers.add(hostPort, options);
         } else {
-            options = options || {};
-            options.initialState = self.initialPeerState;
+            options = options || extend({}, self.peerOptions);
```
jcorbin: `xtend` defaults to a shallow copy, the first empty object is redundant.

node/states.js
```diff
@@ -42,29 +42,21 @@ module.exports.LockedUnhealthyState = LockedUnhealthyState;
  * - channel.random
  */
 
-var symptoms = {
```
jcorbin: Ohh I see, this is where that was tucked away before? Well, errors module is a huge step up at least ;-)

node/test/lib/relay_network.js
jcorbin: Hmm okay so after sleuthing with Kris, I agree that what autobahn-node-client is currently doing is simply wrong... but I'm not sure if this way of doing it is right...

jcorbin:
lgtm, optional feedback imo

PR #658, opened 2015-06-05
Eliminate boolean trap

kriskowal:
Name getOrCreateServiceChannel to reflect its behavior rather than infer its behavior from `, true)`.


kriskowal:
r @raynos

Raynos:
lgtm.

PR #659, opened 2015-06-05
Stand by ring semantics for egress nodes

kriskowal:
If a node is isolated, it becomes egress for all services. Egress nodes must necessarily number >= 1 and <= k, depending on hash collisions and ring sizes near k.

### DIFF
--- a/node/hyperbahn/egress-nodes.js
+++ b/node/hyperbahn/egress-nodes.js
@@ -71,15 +71,7 @@ EgressNodes.prototype.exitsFor = function exitsFor(serviceName) {
     var exitNodes = Object.create(null);
     for (var i = 0; i < k; i++) {
         var shardKey = serviceName + '~' + i;
-
-        // TODO ringpop will return itself if it cannot find
-        // it which is probably the wrong semantics.
         var node = self.ringpop.lookup(shardKey);
-
-        // TODO ringpop can return duplicates. do we want
-        // <= k exitNodes or k exitNodes ?
-        // TODO consider walking the ring instead.
-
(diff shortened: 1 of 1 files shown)


kriskowal:
r @raynos

node/hyperbahn/egress-nodes.js
```diff
@@ -71,15 +71,7 @@ EgressNodes.prototype.exitsFor = function exitsFor(serviceName) {
     var exitNodes = Object.create(null);
     for (var i = 0; i < k; i++) {
         var shardKey = serviceName + '~' + i;
-
-        // TODO ringpop will return itself if it cannot find
```
Raynos: This cannot be removed. What we really want to do is detect whether or not we are "bootstrapping the ring" and if we are we want to send Declined error frames.

We really need to distinquish between "I just started and am bootstrapping" vs "I am isolated in a network partition"

Raynos:
Let's improve the first comment rather then delete it. The other two comments can go

node/hyperbahn/egress-nodes.js
kriskowal: I strongly doubt that this is a useful distinction.

Raynos: It is. When we get a network partition and ringpop converges to be a ring of 1 I want to get an aggressive warning.

Where as being in the bootstrapping phase is normal operation during restarts and redeploys; network partitions are not normal and 90% of the time they are a bug in ringpop instead of a true network partition.

Raynos:
Closing this as these are still useful to reconsider in the future.

PR #660, opened 2015-06-05
2.0.0-IDEA: Make the cn transport header required

Raynos:
This makes it a protocol error to not set the cn header.

There is a `requireCn` option to downgrade to a logger.error()

r: @kriskowal @jcorbin


jcorbin:
:+1:

kriskowal:
:ok:

PR #662, opened 2015-06-06
Node connection metrics

ShanniLi:
r @Raynos @kriskowal @mmihic 
cc @jcorbin 

last big piece of metrics. I will add the missing small pieces next, e.g., bytes sent/received


node/lib/statsd.js
```diff
@@ -127,6 +127,39 @@ function getKey(common, stat) {
...
+
+        case 'connections.errors':
+            return prefix + '.' +
+                clean(stat.tags['peer-host-port'], 'peer-host-port') + '.' +
+                clean(stat.tags.type, 'type');
+
+        case 'connections.closed':
+            return prefix + '.' +
+                clean(stat.tags['peer-host-port'], 'peer-host-port') + '.' +
+                clean(stat.tags.reason, 'reason');
```
kriskowal: I sense a pattern emerging that approximately `sprintf(patterns[stat.name], stat.tags)` might solve.

kriskowal: (but don’t. might be costly)

kriskowal:
:ok:

node/connection_base.js
```diff
@@ -71,6 +73,31 @@ function TChannelConnectionBase(channel, direction, remoteAddr) {
 
     self.tracer = self.channel.tracer;
     self.ops.startTimeoutTimer();
+
+    if (direction === 'out') {
+        self.channel.connectionsInitiatedStat.increment(1, {
```
Raynos: we do not want to report self connections.

Raynos: just move all the code out of connection_base

node/connection_base.js
```diff
@@ -71,6 +73,31 @@ function TChannelConnectionBase(channel, direction, remoteAddr) {
...
+    }
+
+    if (self.reportInterval > 0) {
+        self.reportTimer = self.timers.setTimeout(
+            onReport, self.reportInterval
+        );
+    }
+
+    function onReport() {
+        self.channel.connectionsActiveStat.update(1, {
```
Raynos: That's not how gauges work.

You need to `update(totalNum)`.

The only person that has access to that is the rootChannel. It tracks all the incoming connections in serverConnections and it has access to all the peers so in theory can track outgoing connections.

node/connection_base.js
```diff
@@ -104,6 +131,39 @@ TChannelConnectionBase.prototype.resetAll = function resetAll(err) {
...
     }
 
+    if (!self.remoteName) {
+        if (self.direction === 'out') {
+            self.channel.connectionsConnectErrorsStat.increment(1, {
+                'host-port': self.channel.hostPort || '0.0.0.0:0',
+                'peer-host-port': self.remoteAddr
+            });
+        } else {
+            self.channel.connectionsAcceptedErrorsStat.increment(1, {
```
Raynos: we know the peer; let's add it and just modify the metrics document or ask clarification from @mmihic

node/connection_base.js
```diff
@@ -104,6 +131,39 @@ TChannelConnectionBase.prototype.resetAll = function resetAll(err) {
...
+                'host-port': self.channel.hostPort
+            });
+        }
+    } else {
+        if (err.type !== 'tchannel.socket-local-closed' &&
+            err.type !== 'tchannel.socket-closed') {
```
Raynos: socket-closed is also an error. The only one's that are not errors are local closes.

node/test/connection-stats.js
```diff
@@ -0,0 +1,251 @@
...
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+'use strict';
+
+var os = require('os');
+var TimeMock = require('time-mock');
+
+var allocCluster = require('./lib/alloc-cluster.js');
+var timers = TimeMock(Date.now());
```
Raynos: no; please create a timer per tests.

node/test/connection-stats.js
```diff
@@ -0,0 +1,251 @@
...
+        app: 'server',
+        host: os.hostname()
+    };
+    client.on('stat', function onStat(stat) {
+        stats.push(stat);
+    });
+
+    var subClient = client.makeSubChannel({
+        serviceName: 'server',
+        peers: ['localhost:4040']
```
Raynos: leave a comment about there should be nothing running on this port.

node/test/connection-stats.js
```diff
@@ -0,0 +1,251 @@
...
+                value: 1,
+                tags:
+                {
+                    'host-port': clientHost,
+                    'peer-host-port': serverHost,
+                    app: 'server',
+                    host: os.hostname()
+               }
+            }, {
+                name: 'connections.active',
```
Raynos: write a test where you open 3 connections and gauge is set to 3 :)

Raynos:
Reviewed.

 - We have to fix the gauge. 
 - We should not report self connections

node/test/connection-stats.js
ShanniLi: there is only one test uses that timer.

node/connection_base.js
ShanniLi: it is "The number of currently active outbound connections from the process to the peer.", so it is per peer bases. I can move the connection to peer level. It cannot be at the channel level by definition.

docs/metrics.md
```diff
@@ -218,8 +218,7 @@ The total number of connections from the target peer accepted by this process
 over its lifetime. 
 
 ##### connections.accept-errors
-The total number of accept errors that have occurred on this process.  Will not
-include a peer tag.
+The total number of accept errors that have occurred on this process.
```
Raynos: @mmihic we are including the peer tag in this one because we have access to it.

node/connection.js
```diff
@@ -378,4 +391,111 @@ TChannelConnection.prototype.buildOutResponse = function buildOutResponse(req, o
     return self.handler.buildOutResponse(req, opts);
 };
 
+// this connection is completely broken, and is going away
+// In addition to erroring out all of the pending work, we reset the state
+// in case anybody stumbles across this object in a core dump.
+TChannelConnectionBase.prototype.resetAll = function resetAll(err) {
```
Raynos: XD dont mutate the other prototype bro.

node/connection_base.js
```diff
@@ -75,91 +75,9 @@ function TChannelConnectionBase(channel, direction, remoteAddr) {
 inherits(TChannelConnectionBase, EventEmitter);
 
 TChannelConnectionBase.prototype.close = function close(callback) {
-    var self = this;
-
-    self.resetAll(errors.LocalSocketCloseError());
```
Raynos: you can't change this; Move this `close()` into `TChannelConnection`

Raynos: Then just remove this method.

ShanniLi: cannot do that, also used by self-connection ...

Raynos:
lgtm

PR #664, opened 2015-06-08
go: Remove ReadArgX/WriteArgX. Create ArgWriter/ArgReader.

prashantv:
ArgWriter and ArgReader give some of the convenience of ReadArgX/WriteArgX methods for clients.

### DIFF
--- a/golang/arguments.go
+++ b/golang/arguments.go
@@ -48,8 +48,8 @@ func (r *ArgReader) read(f func() error) error {
 	return r.reader.Close()
 }
 
-// ReadBytes reads from the reader into the byte slice.
-func (r *ArgReader) ReadBytes(bs *[]byte) error {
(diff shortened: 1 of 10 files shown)


mmihic:
I kindof liked the ability for application code to define new format via interfaces vs. having a fixed set of supported formats, but ok.

prashantv:
Yeah, I did like that too. We could add back the Input/Ouptut interfaces, and have WriteOutput and ReadInput methods on the ArgWriter/Reader, that still allows for applications to define their new formats. I just wanted to make the simple use cases (bytes/json) simpler.

mmihic:
Makes sense.  I'd be inclined to leave out the Input/Output interface given the ReadXXX/WriteXXX methods - I think it's better to have just one style of API.

PR #665, opened 2015-06-08
add caller name into protocol header

junchaowu:
@abhinav @blampe

### DIFF
--- a/python/tchannel/tcurl.py
+++ b/python/tchannel/tcurl.py
@@ -289,7 +289,7 @@ def main(argv=None):
     if args.verbose:
         log.setLevel(logging.DEBUG)
 
-    tchannel = TChannel(caller_name='tcurl')
+    tchannel = TChannel(name='tcurl.py')
     if args.in_port:
         create_server(tchannel, args.in_port)
 
--- a/python/tchannel/tcurl.py
+++ b/python/tchannel/tcurl.py
@@ -289,7 +289,7 @@ def main(argv=None):
     if args.verbose:
         log.setLevel(logging.DEBUG)
 
-    tchannel = TChannel()
+    tchannel = TChannel(caller_name='tcurl')
     if args.in_port:
         create_server(tchannel, args.in_port)
 
--- a/python/tchannel/tornado/peer.py
+++ b/python/tchannel/tornado/peer.py
@@ -452,7 +452,7 @@ class PeerClientOperation(object):
         self.headers = {
             'as': arg_scheme or ArgSchemeType.DEFAULT,
             're': retry or RetryType.DEFAULT,
-            'cn': self.peer.tchannel.caller_name,
+            'cn': self.peer.tchannel.name,
         }
 
         # keep all arguments for retry purpose.
--- a/python/tchannel/tornado/peer.py
+++ b/python/tchannel/tornado/peer.py
@@ -452,6 +452,7 @@ class PeerClientOperation(object):
         self.headers = {
             'as': arg_scheme or ArgSchemeType.DEFAULT,
             're': retry or RetryType.DEFAULT,
+            'cn': self.peer.tchannel.caller_name,
         }
 
         # keep all arguments for retry purpose.
--- a/python/tchannel/tornado/tchannel.py
+++ b/python/tchannel/tornado/tchannel.py
@@ -126,7 +126,7 @@ class TChannel(object):
             Name to be register on the hyperbahn.
         :returns:
             A future that resolves to the remote server's response after
-            the first advertise is successful.
+            the first advertise finishes.
         """
         name = name or self.name
         return hyperbahn.advertise(self, name, router)
--- a/python/tchannel/tornado/tchannel.py
+++ b/python/tchannel/tornado/tchannel.py
@@ -112,7 +112,7 @@ class TChannel(object):
             for peer_hostport in known_peers:
                 self.peers.add(peer_hostport)
 
-    def advertise(self, router):
+    def advertise(self, router, name=None):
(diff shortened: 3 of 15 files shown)


python/tchannel/tcurl.py
```diff
@@ -289,7 +289,7 @@ def main(argv=None):
     if args.verbose:
         log.setLevel(logging.DEBUG)
 
-    tchannel = TChannel()
+    tchannel = TChannel(caller_name='tcurl')
```
Raynos: to avoid confusion; it might be nice to call this `tcurl.py`. We can call the node one either `tcurl` or `tcurl.node`.

python/tchannel/tornado/peer.py
```diff
@@ -452,6 +452,7 @@ def __init__(self, peer, service, arg_scheme=None,
         self.headers = {
             'as': arg_scheme or ArgSchemeType.DEFAULT,
             're': retry or RetryType.DEFAULT,
+            'cn': self.peer.tchannel.caller_name,
```
Raynos: in some future you may want to specify the cn on a per request basis.

python/tchannel/tornado/tchannel.py
```diff
@@ -99,15 +103,38 @@ def __init__(self, hostport=None, process_name=None, known_peers=None):
...
+        :param caller_name:
+            Name of the caller behind this TChannel
+        :param routers:
+            Seed list of addresses of Hyperbahn routers
+        :returns:
+            A future that resolves to the remote server's response after
+            the first advertise is successful.
+        """
+
+        self.caller_name = caller_name
```
Raynos: only set this if None ?

python/tchannel/tornado/tchannel.py
```diff
@@ -63,7 +63,8 @@ class TChannel(object):
         # 'http': scheme.HttpArgScheme, TODO
     }
 
-    def __init__(self, hostport=None, process_name=None, known_peers=None):
+    def __init__(self, hostport=None, process_name=None, caller_name=None,
```
abhinav: `caller_name` is a bad name for this. Just use `service`.

python/tchannel/tornado/tchannel.py
```diff
@@ -78,6 +79,9 @@ def __init__(self, hostport=None, process_name=None, known_peers=None):
         :param known_peers:
             A list of host-ports at which already known peers can be reached.
             Defaults to an empty list.
+        :param caller_name:
+            Name of the caller. This will be set in the TChannel
+            protocol header.
```
abhinav: That it will be sent out as a header is not a detail users care about. It's just an argument they have to provide that specifies the name of the service their TChannel represents.

Also, since this is basically a required header at the protocol level, it may be reasonable to make it required.

python/tchannel/tornado/tchannel.py
junchaowu: I think this one used to advertise should override the one used to initialize tchannel anyway. Only in this case we can force the caller_name be consistence not only on caller side but also on the autobahn side.

abhinav: No. You already know the service name. You don't need it as another argument.. And you definitely shouldn't be doing this random mutation in a method whose name gives no indication that it will mutate this state.

python/tchannel/tornado/tchannel.py
```diff
@@ -99,15 +103,38 @@ def __init__(self, hostport=None, process_name=None, known_peers=None):
...
+
+        It also tells the TChannel about the given Hyperbahn routers.
+
+        :param tchannel:
+            TChannel to register with Hyperbahn
+        :param caller_name:
+            Name of the caller behind this TChannel
+        :param routers:
+            Seed list of addresses of Hyperbahn routers
+        :returns:
```
abhinav: Update the docstring

python/tchannel/tornado/tchannel.py
junchaowu: Let me change the name as required in the constructor. Then remove the name in the advertise method.

python/tchannel/tornado/tchannel.py
```diff
@@ -99,15 +102,35 @@ def __init__(self, hostport=None, process_name=None, known_peers=None):
...
+        :param tchannel:
+            TChannel to register with Hyperbahn
+        :param routers:
+            Seed list of addresses of Hyperbahn routers
+        :returns:
+            A future that resolves to the remote server's response after
+            the first advertise is successful.
+        """
+
+        hyperbahn.advertise(self, self.name, router)
```
abhinav: `return`

python/tchannel/tornado/tchannel.py
Raynos: the caller name and the name you advertise as are different.

For example `rt-los_angeles` (advertise) vs `dispatch` (caller name).

python/tchannel/tornado/tchannel.py
```diff
@@ -99,15 +102,35 @@ def __init__(self, hostport=None, process_name=None, known_peers=None):
...
+
+        It also tells the TChannel about the given Hyperbahn routers.
+
+        :param routers:
+            Seed list of addresses of Hyperbahn routers
+        :param name:
+            Name to be register on the hyperbahn.
+        :returns:
+            A future that resolves to the remote server's response after
+            the first advertise is successful.
```
abhinav: it resolves to the remote server's response after the first advertise finished. it may or may not have been successful. just say "after the first advertise finishes"

abhinav:
Fix the comment and then :shipit:

PR #666, opened 2015-06-09
Add required headers to hyperbahn relay handler

Raynos:
r: @kriskowal @jcorbin


node/hyperbahn/handler.js
```diff
@@ -198,7 +205,7 @@ function handleRelayAdvertise(self, req, arg2, arg3, cb) {
 };
 
 HyperbahnHandler.prototype.sendRelayAdvertise =
-function sendRelayAdvertise(hostPort, services, callback) {
+function sendRelayAdvertise(opts, callback) {
```
kriskowal: options is the new opts

kriskowal:
:ok:

Issue #318, opened 2015-04-03
golang: implement Thrift Input/Output


mmihic:

prashantv:
Fixed by PR #667

PR #667, opened 2015-06-09
Go: Thrift support for tchannel (#318)

prashantv:
The client is compatible with node servers, as it skips writing the Thrift RPC call headers.


golang/thrift/output_protocol.go
```diff
@@ -0,0 +1,150 @@
...
+	// ReadArg2 has to be called before checking Response.ApplicationError.
+	reader, err := resp.Arg2Reader()
+	if err != nil {
+		return "", 0, 0, err
+	}
+	if err := reader.Close(); err != nil {
+		return "", 0, 0, err
+	}
+
+	if resp.ApplicationError() {
```
mmihic: Don't you still need to unmarshal arg3 to extract the thrift exception body?

golang/thrift/output_protocol.go
```diff
@@ -0,0 +1,150 @@
...
+
+// TChanOutboundOptions are parameters passed to the underlying tchannel when making requests.
+type TChanOutboundOptions struct {
+	// Context is the context.Context with timeout and deadline information.
+	Context context.Context
+
+	// Dst is the host:port of the service to call.
+	Dst string
+
+	// AutobahnService is the Autobahn service name.
```
mmihic: nit: Hyperbahn

golang/thrift/output_protocol.go
prashantv: Yep, forgot to fix this, thanks

golang/thrift/output_protocol.go
prashantv: Done

prashantv:
Travis is failing since thrift isn't available on it, we'll probably need to check in the thrift generated code. @mmihic can you think of any other alternatives?

prashantv:
+ @weikai77 for comments, this is partially based on your PR.

PR #669, opened 2015-06-09
go: add a connection ID for debug logging

prashantv:


golang/logger.go
```diff
@@ -35,6 +35,9 @@ import (
 // logrus, go-logging, etc).  The SimpleLogger adapts to the standard go log
 // package.
 type Logger interface {
+	// SetPrefix sets a prefix used all logged messages.
+	SetPrefix(prefix string)
```
mmihic: +1.

golang/logger.go
```diff
@@ -35,6 +35,9 @@ import (
 // logrus, go-logging, etc).  The SimpleLogger adapts to the standard go log
 // package.
 type Logger interface {
+	// SetPrefix sets a prefix used all logged messages.
```
mmihic: missed word, s/b: "sets a prefix used for all logged messages"

golang/connection.go
```diff
@@ -175,6 +178,10 @@ func newConnection(conn net.Conn, initialState connectionState, handlers *handle
 		framePool = DefaultFramePool
 	}
 
+	connID := atomic.AddUint32(&nextConnID, 1)
+	log.SetPrefix(fmt.Sprintf("C%v ", connID))
```
mmihic: Log is shared between connections, no?  Seems like this needs to return a logger that has the prefix set, so that each connection gets a unique prefixed logger.

prashantv: Thanks, I totally missed that. I've created a prefixLogger which can be used to wrap loggers with a specific prefix instead.

mmihic:
lgtm

PR #671, opened 2015-06-09
zipkin trace update

junchaowu:
@blampe @abhinav

### DIFF
--- a/python/tchannel/rw.py
+++ b/python/tchannel/rw.py
@@ -347,7 +347,7 @@ class NumberReadWriter(ReadWriter):
         1: '>B',
         2: '>H',
         4: '>I',
-        8: '>Q',
+        8: '>q',
     }
 
     __slots__ = ('_width', '_format')
--- a/python/tchannel/tornado/tchannel.py
+++ b/python/tchannel/tornado/tchannel.py
@@ -66,7 +66,7 @@ class TChannel(object):
     }
 
     def __init__(self, name, hostport=None, process_name=None,
-                 known_peers=None, zipkin=False):
+                 known_peers=None, trace=False):
         """Build or re-use a TChannel.
 
         :param name:
@@ -83,8 +83,9 @@ class TChannel(object):
         :param known_peers:
             A list of host-ports at which already known peers can be reached.
             Defaults to an empty list.
-        :param zipkin:
-            Flag to turn on/off zipkin trace.
+        :param trace:
+            Flag to turn on/off zipkin trace. It can be a bool variable or
+            a function that return true or false.
         """
         self._state = State.ready
         self._handler = RequestDispatcher()
@@ -107,7 +108,7 @@ class TChannel(object):
         )
 
         self.name = name
-        self._zipkin = zipkin
+        self._trace = trace
 
         # register event hooks
         self.event_emitter = EventEmitter()
@@ -119,11 +120,11 @@ class TChannel(object):
                 self.peers.add(peer_hostport)
 
     @property
-    def zipkin(self):
-        if hasattr(self._zipkin, '__call__'):
-            return self._zipkin()
+    def trace(self):
+        if callable(self._trace):
+            return self._trace()
         else:
-            return self._zipkin
+            return self._trace
 
     def advertise(self, router, name=None):
         """Advertise the given TChannel to Hyperbahn.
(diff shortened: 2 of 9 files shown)


python/tchannel/tornado/tchannel.py
```diff
@@ -103,15 +107,24 @@ def __init__(self, name, hostport=None, process_name=None,
...
         self.hooks = EventRegistrar(self.event_emitter)
+        self.hooks.register(ZipkinTraceHook(tchannel=self))
 
         if known_peers:
             for peer_hostport in known_peers:
                 self.peers.add(peer_hostport)
 
+    @property
+    def zipkin(self):
+        if hasattr(self._zipkin, '__call__'):
```
abhinav: if callable(self._zipkin)

junchaowu: http://stackoverflow.com/questions/624926/how-to-detect-whether-a-python-variable-is-a-function

abhinav: It's no longer deprecated.

python/tchannel/tornado/tchannel.py
```diff
@@ -81,6 +83,8 @@ def __init__(self, name, hostport=None, process_name=None,
         :param known_peers:
             A list of host-ports at which already known peers can be reached.
             Defaults to an empty list.
+        :param zipkin:
+            Flag to turn on/off zipkin trace.
```
abhinav: Document that it may be a callable as well.

python/tchannel/tornado/tchannel.py
junchaowu: It is no longer deprecate in v3.2 but as deprecated in 3.0 3.1. and what's wrong with __call__

abhinav: Yeah but we'll never use 3.0/3.1. It exists in 2.X. It exists in 3.2 onwards. When we migrate to 3, we'll use at least 3.4.

python/tchannel/tornado/tchannel.py
```diff
@@ -64,7 +66,7 @@ class TChannel(object):
     }
 
     def __init__(self, name, hostport=None, process_name=None,
-                 known_peers=None):
+                 known_peers=None, zipkin=False):
```
blampe: Zipkin is one instance of a tracer we can turn on, but it's not the only one. I'd refrain from explicitly mentioning Zipkin here and use something like `trace=True` instead.

python/tchannel/tornado/tchannel.py
```diff
@@ -103,15 +108,24 @@ def __init__(self, name, hostport=None, process_name=None,
         )
 
         self.name = name
+        self._trace = trace
 
         # register event hooks
         self.event_emitter = EventEmitter()
         self.hooks = EventRegistrar(self.event_emitter)
+        self.hooks.register(ZipkinTraceHook(tchannel=self))
```
blampe: Hm. I don't like the idea of auto-registering hooks like this. Would this even work if I wanted to use tchannel without a zipkin collector setup?

junchaowu: if the tracingflag set to False, the zipkin hook does nothing. In the trace session of protocol document, it says 
'Trace flags:

flag	description
0x01	tracing enabled for this request
When the enabled flag is not set, the tracing data is still required to be present in the frame, but the tracing data will not necessarily be sent to zipkin.' 
It implies the zipkin tracing is a built in feature.

python/tchannel/tornado/tchannel.py
```diff
@@ -1,3 +1,4 @@
+
```
blampe: remove

python/tchannel/rw.py
```diff
@@ -347,7 +347,7 @@ class NumberReadWriter(ReadWriter):
         1: '>B',
         2: '>H',
         4: '>I',
-        8: '>Q',
+        8: '>q',
```
abhinav: I don't know if you want to change all parsed 64-bit integers to signed. Can we make a separate `rw.number` parameter `signed=True/False`?

junchaowu: in the protocol, it says clearly it is i64. That's why I change it. It will only affect the id which are using 8 bytes.

python/tchannel/tornado/tchannel.py
junchaowu: done

blampe:
I really wish we could do this in a way that was compatible with the Twitter Zipkin IDL but otherwise LGTM.

junchaowu:
@blampe  I totally agree. I've expressed my concern in the morning for the compatibility of stock zipkin.

PR #673, opened 2015-06-09
Add permissions cache

mrbrowning:
This doesn't do much of interest right now, just planting a flag until the actual communication with Jack's vorenus-controller is fleshed out. For now, maintain token buckets in a cache per (directional) service pair that are reset on a one-second basis, removing a token for each attempted call. Going to tackle the gossiping between egress nodes next, which will probably inform a better choice for cache bound than Infinity.


node/hyperbahn/permissions_cache.js
```diff
@@ -0,0 +1,90 @@
...
+    self.channel = self.options.channel;
+    self.logger = self.options.logger;
+    self._intervalId = setInterval(
+        bucketResetCallback(self), BUCKET_RESET_DURATION
+    );
+
+    self.channel.statEvent.addListener(self.increment.bind(self));
+}
+
+inherits(PermissionsCache, LRUCache);
```
Raynos: Probably better to have `self.lru = new LRUCache()` then to inherit.

node/hyperbahn/permissions_cache.js
```diff
@@ -0,0 +1,90 @@
...
+var NUM_TOKENS = 100;
+
+function PermissionsCache(options) {
+    if (!(this instanceof PermissionsCache)) {
+        return new PermissionsCache(options);
+    }
+    var self = this;
+
+    self.options = options;
+    PermissionsCache.super_.call(self, self.options);
```
Raynos: style nitpick: `LRUCache.call(self, options);`

node/hyperbahn/permissions_cache.js
```diff
@@ -0,0 +1,90 @@
...
+// the actual rate limit from vorenus-controller.
+var NUM_TOKENS = 100;
+
+function PermissionsCache(options) {
+    if (!(this instanceof PermissionsCache)) {
+        return new PermissionsCache(options);
+    }
+    var self = this;
+
+    self.options = options;
```
Raynos: `self.options = options;` is bad pattern in tchannel; we should avoid it. just unpack things you need.

node/hyperbahn/permissions_cache.js
```diff
@@ -0,0 +1,90 @@
...
+    PermissionsCache.super_.call(self, self.options);
+
+    self.channel = self.options.channel;
+    self.logger = self.options.logger;
+    self._intervalId = setInterval(
+        bucketResetCallback(self), BUCKET_RESET_DURATION
```
Raynos: style:

```js
self._intervalId = setInterval(bucketReset, BUCKET_RESET_DURATION);

function bucketReset() {
    self.bucketReset();
}
```

node/hyperbahn/permissions_cache.js
```diff
@@ -0,0 +1,90 @@
...
+    );
+
+    self.channel.statEvent.addListener(self.increment.bind(self));
```
Raynos: bind() is banned.

```js
self.channel.statEvent.addListener(increment);

function increment(stat) {
    self.increment(stat);
}
```

node/hyperbahn/permissions_cache.js
```diff
@@ -0,0 +1,90 @@
...
+
+inherits(PermissionsCache, LRUCache);
+
+PermissionsCache.prototype.clearBuckets = function clearBuckets() {
+    var self = this;
+    clearInterval(self._intervalId);
+    self.reset();
+};
+
+PermissionsCache.prototype.increment = function increment(stat) {
```
Raynos: nice trick hooking into the stats.

node/hyperbahn/permissions_cache.js
```diff
@@ -0,0 +1,90 @@
...
+PermissionsCache.prototype.resetBucketTokens = function resetBucketTokens(key) {
+    // A pretty anemic method right now to be sure, but I suspect there might
+    // be more to this logic in the future.
+    var self = this;
+    self.set(key, NUM_TOKENS);
+};
+
+function bucketResetCallback(cache) {
+    return function() {
+        cache.keys().map(cache.resetBucketTokens.bind(cache));
```
Raynos: `bind()` is banned.

```js
var keys = cache.keys();
for (var i = 0; i < keys.length; i++) {
  cache.resetBucketTokens(keys[i]);
}
```

node/hyperbahn/service_proxy.js
```diff
@@ -42,6 +43,10 @@ function ServiceDispatchHandler(options) {
 
     self.egressNodes.on('membershipChanged', onMembershipChanged);
 
+    self.permissionsCache = new PermissionsCache({
```
Raynos: Let's not place this here. Let's just take `self.permissionsCache = options.permissionsCache;`

We can then instantiate it in autobahn/clients/index.js

Raynos:
Mostly node/tchannel style and such; this looks reasonable on the surface. 

We do need to get rid of some of the slow paths though (.bind(), .map())

node/hyperbahn/permissions_cache.js
```diff
@@ -0,0 +1,90 @@
...
+
+PermissionsCache.prototype.resetBucketTokens = function resetBucketTokens(key) {
+    // A pretty anemic method right now to be sure, but I suspect there might
+    // be more to this logic in the future.
+    var self = this;
+    self.set(key, NUM_TOKENS);
+};
+
+function bucketResetCallback(cache) {
+    return function() {
```
Raynos: style: all functions need names

node/hyperbahn/permissions_cache.js
mrbrowning: Cool, I did pick it up from reading elsewhere in the code base, but that sounds preferable.

node/hyperbahn/permissions_cache.js
mrbrowning: Then just treat `self.lru` as a public member and have referrers access that property directly to get call counts?

Raynos: that's upto you; all fields that do not start with `_` are public; it's up to you whether you want to add your own methods or not.

Raynos:
lgtm.

node/test/permissions_cache.js
```diff
@@ -0,0 +1,64 @@
...
+        serviceName: 'server',
+        hasNoParent: true,
+        headers: {
+            cn: 'client',
+            as: 'raw'
+        }
+    }).send('echo', 'a', 'b', onResponse);
+    
+    function onResponse(err, res, arg2, arg3) {
+        assert.ifError(err);
```
kriskowal: `if (err) return assert.end(err);`

kriskowal:
This is good forward motion. Let’s get the flag planted.

PR #674, opened 2015-06-10
hyperbahn: add a grace period for logging

Raynos:
We get a bunch of these logs in prod and tests under
normal operations.

These logs are expected when a hyperbahn node is starting
but are not expected at run-time as they might be a
source of undefined behaviour.

I've added a heuristic to silence the log in the first
5 minutes that the ServiceProxy exists.

r: @kriskowal @rf @jcorbin

### DIFF
--- a/node/hyperbahn/service_proxy.js
+++ b/node/hyperbahn/service_proxy.js
@@ -25,6 +25,7 @@ var RelayHandler = require('../relay_handler');
 
 var REGISTER_GRACE_PERIOD = 1000;
 var REGISTER_TTL = 1000;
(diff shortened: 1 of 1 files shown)


kriskowal:
It might be worthwhile to ask whether the creation of service proxies is log worthy regardless of when it happens. Some amount of these are expected to be created at any time. Might be more useful to just have a gauge for the number of service proxies.

Raynos:
Good point on having a gauge for sub channels ( not service proxies ).

For now I just want to make this info log nicer; I may remove it in the future.

kriskowal:
:ok: looks like it will work as advertised

PR #676, opened 2015-06-10
Cleanup service proxy

Raynos:
I removed some dead code; the config values are no longer used.
The adhoc endpoint stats are now replaced with real endpoint
stats.

r: @jcorbin @kriskowal @rf


kriskowal:
lgtm

PR #688, opened 2015-06-10
go: add log to reqRes* with message ID prefix, and log on failed.

prashantv:


mmihic:
lgtm.  At some point we may want to make the Logger interface tag-aware ala logrus, so that we can emit structured logs that contain these values as tags vs just being prefixes. *shrug*

PR #689, opened 2015-06-10
Go: Pool connections using Peer/SubChannel abstractions

prashantv:
Basic connection pooling/subchannel support:
 - pool connections and reuse them if possible
 - move mutable channel state into a separate mutable struct (to make it easy to avoid accidentally accessing it without the lock)

The OnActive event from connection is kind of hacky, if we need more event support in future, we can create a proper way to register for events.

### DIFF
--- a/golang/peer.go
+++ b/golang/peer.go
@@ -1,0 +1,122 @@
+package tchannel
+
+import "golang.org/x/net/context"
+
+// PeerList maintains a list of Peers.
+type PeerList struct {
+	channel *Channel
+	peers   map[string]*Peer
+}
+
+func newPeerList(channel *Channel) *PeerList {
+	return &PeerList{
+		channel: channel,
+		peers:   make(map[string]*Peer),
+	}
+}
+
+// Add a peer to the peer list.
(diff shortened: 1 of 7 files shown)


golang/peer.go
```diff
@@ -0,0 +1,192 @@
...
+
+	"golang.org/x/net/context"
+)
+
+var (
+	errInvalidConnectionState = errors.New("connection is in an invalid state")
+)
+
+// PeerList maintains a list of Peers.
+type PeerList struct {
```
mmihic: Does this need to be exported?

mmihic:
Nice.  Maybe consider having a test that does concurrent peer / connection setup, to run under the race detector.

I think the OnActive thing is fine, there's precedence for taking this type of simple approach in http.Server#ConnState.

golang/peer.go
prashantv: Yep, exported it, thanks.

prashantv:
Yep, I'm planning to write a stress test with lots of concurrent connections/calls which we can run with the race detector. However, I want to consolidate some of the test setup logic that could be shared as part of that, so I'll do that separately.

PR #690, opened 2015-06-10
Do not confuse socket remoteAddr with init remoteName

Raynos:
This PR fixes a bunch of confusion around remoteAddr
and remoteName.

In this PR we have

 - `conn.socketRemoteAddr`; the remote address of the socket
 - `conn.remoteName`; the remoteName as per init handshake
 - `req/res.remoteAddr`; the remoteName of the connection

We had a bug where we previously non-determinsically set
`req.remoteAddr` to remoteName or socketRemoteAddr

We also fix miscellanious things in this PR; mostly 
to do with not allowing out requests until initialized
and handling ephemeral remote names better.

r: @jcorbin @kriskowal @rf

### DIFF
--- a/node/connection_base.js
+++ b/node/connection_base.js
@@ -78,7 +78,9 @@ inherits(TChannelConnectionBase, EventEmitter);
 TChannelConnectionBase.prototype.request = function connBaseRequest(options) {
     var self = this;
     if (!options) options = {};
-    options.remoteAddr = self.remoteAddr;
+
+    assert(self.remoteName, 'cannot make request unless identified');
+    options.remoteAddr = self.remoteName;
 
     options.channel = self.channel;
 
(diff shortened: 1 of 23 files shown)


node/connection_base.js
```diff
@@ -164,7 +164,9 @@ TChannelConnectionBase.prototype.resetAll = function resetAll(err) {
 TChannelConnectionBase.prototype.request = function connBaseRequest(options) {
     var self = this;
     if (!options) options = {};
-    options.remoteAddr = self.remoteAddr;
+
+    assert(self.remoteName, 'cannot make request unless identified');
+    options.remoteAddr = self.remoteName;
```
Raynos: This is the actual bug fix.

node/hyperbahn/handler.js
```diff
@@ -90,7 +90,7 @@ function handleAdvertise(self, req, arg2, arg3, cb) {
 
     for (var i = 0; i < services.length; i++) {
         var service = services[i];
-        service.hostPort = req.remoteAddr;
+        service.hostPort = req.connection.remoteName;
```
Raynos: This is another bug fix; This makes getting the remoteName in advertise explicit.

kriskowal:
lgtm, rebase needed

Raynos:
Addressed some of the feedback; other naming suggestions I'm going to punt for now.

PR #692, opened 2015-06-10
Add more error logging context to the relay handler

Raynos:

### DIFF
--- a/node/relay_handler.js
+++ b/node/relay_handler.js
@@ -179,6 +179,14 @@ RelayRequest.prototype.logError = function logError(err, codeName) {
         outArg1: String(self.outreq.arg1)
     };
 
+    if (self.outreq.outReqs) {
+        logOptions.allErrors = [];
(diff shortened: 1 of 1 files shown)


node/relay_handler.js
```diff
@@ -175,9 +175,18 @@ RelayRequest.prototype.logError = function logError(err, codeName) {
...
         serviceName: self.outreq.serviceName,
+        callerName: self.inreq.headers.cn,
         outArg1: String(self.outreq.arg1)
     };
 
+    if (self.outreq.outReqs) {
+        logOptions.allErrors = [];
+        for (var i = 0; i < self.outreq.outReqs.length; i++) {
+            var individualReq = self.outreq.outReqs[i];
+            logOptions.allErrors.push(individualReq.err || null);
```
ShanniLi: why pushing null?

Raynos: If you push `undefined` then it wont JSON serialize cleanly.

PR #698, opened 2015-06-11
Propose new error code for unhealthy nodes

kriskowal:
r @mranney @mmihic @jcorbin @prashantv

### DIFF
--- a/docs/protocol.md
+++ b/docs/protocol.md
@@ -463,8 +463,9 @@ code   | name                 | description
 `0x03` | busy                 | Node is too busy, this request is safe to retry elsewhere if desired, prefer other peers for future requests where possible.
 `0x04` | declined             | Node declined request for reasons other than load, the request is safe to retry elsewhere if desired, do not change peer preferencing for future requests.
 `0x05` | unexpected error     | Request resulted in an unexpected error. The request may have been completed before the error, retry only if the request is idempotent or duplicate execution can be handled.
-`0x06` | bad request          | Request args do not match expectations, request will never be satisfied, do not retry
+`0x06` | bad request          | Request args do not match expectations, request will never be satisfied, do not retry.
 `0x07` | network error        | A network error (e.g. socket error) occurred.
+`0x08` | unhealthy            | A relay on the network declined to forward the request to an unhealthy node, do not retry.
 `0xFF` | fatal protocol error | Connection will close after this frame. message ID of this frame should be `0xFFFFFFFF`.
 
 ### id:4


kriskowal:
Per discussion on the circuit breaker with @mattmihic, we need an error code implies that retries should not be attempted, like cancelled or bad request, but without the corresponding meaning.

docs/protocol.md
```diff
@@ -465,6 +465,7 @@ code   | name                 | description
 `0x05` | unexpected error     | Request resulted in an unexpected error. The request may have been completed before the error, retry only if the request is idempotent or duplicate execution can be handled.
 `0x06` | bad request          | Request args do not match expectations, request will never be satisfied, do not retry
 `0x07` | network error        | A network error (e.g. socket error) occurred.
+`0x08` | unhealthy            | Network refused to forward request to unhealthy node.
```
jcorbin: Let's use more consistent nomenclature here, "declined" rather than "refused" I think?

Also maybe we should call the name "forward_unhealthy" rather than just "unhealthy"?

kriskowal: I’ve adjusted the verbiage a little. Avoiding coupling the code to a specific reason for being declined since we may devise any number of systems and nodes along the network for autonomous health detection, all of which should have the same, "declined, don't retry" semantics.

That is, unless I’m wrong about the "don’t retry" semantics. If it’s appropriate to retry on an unhealthy request, the network will usually decline, but will allow a request through periodically to reassess the health.

jcorbin:
sgtm

junchaowu:
Will python client need to care about the new error code?

kriskowal:
All clients would need to understand the new code and know not to retry on such errors.

mranney:
Who would send back such errors, only routers?

kriskowal:
@mranney Yes, only egress relays.

kriskowal:
Oh, no, correction. We also have p2p circuit breakers. They could implement this. Can’t speak for should.

kriskowal:
To be clear, I am not certain that we should do this. A relay could decline without discouraging retries. However, each subsequent retry would go through a relay again, which should be unhealthy or become unhealthy due to the retry, or might be the one lucky request to go through for the circuit’s probation period. Hard to say which will work better in practice without trying both out.

kriskowal:
One thing is sure, while a circuit is unhealthy, if it just declines, the relay will experience elevated load due to retries.

mranney:
I think this is a useful thing to add. I was just hoping there'd be a clean way to specify who should use it and when.

kriskowal:
Will circle back and clarify when and when not to use it when we have some experience with both approaches.

PR #700, opened 2015-06-11
fix traceflag bug

junchaowu:
@blampe @abhinav 
fix circular dependency
temperarily remove local timeout


blampe:
changelog

junchaowu:
Added change log

PR #701, opened 2015-06-12
Thread initial peer state

kriskowal:
This allows us to lock down the peer to peer health monitor, largely for disabling this feature when testing others.


kriskowal:
r @jcorbin @Raynos

jcorbin:
:ok_hand:

PR #702, opened 2015-06-12
go: crc32Checksum.TypeCode returned Crc32 even for Crc32C

prashantv:


mmihic:
pfft.  weak.   Thanks for fixing.

PR #703, opened 2015-06-12
go: Copy arg scheme header to call req response from call req.

prashantv:


mmihic:
lgtm although maybe test?

prashantv:
Added a simple test.

PR #704, opened 2015-06-15
go: add a simple Hyperbahn client to register services.

prashantv:
This works with a local Hyperbahn server + tcurl. It re-registers periodically in the background.

The tests are a little complicated because the logic happens in the background periodically.

### DIFF
--- a/golang/.gitignore
+++ b/golang/.gitignore
@@ -3,6 +3,4 @@
 .DS_Store
 build
 Godeps/_workspace
-messagetype_string.go
-reqresreaderstate_string.go
-reqreswriterstate_string.go
+*_string.go
--- a/golang/channel.go
+++ b/golang/channel.go
@@ -195,6 +195,14 @@ func (ch *Channel) GetSubChannel(serviceName string) *SubChannel {
 	return ch.registerNewSubChannel(serviceName)
 }
 
+// Peers returns the PeerList for the channel.
+func (ch *Channel) Peers() *PeerList {
+	ch.mutable.mut.RLock()
+	defer ch.mutable.mut.RUnlock()
+
+	return ch.peers
+}
+
 // BeginCall starts a new call to a remote peer, returning an OutboundCall that can
 // be used to write the arguments of the call.
 func (ch *Channel) BeginCall(ctx context.Context, hostPort, serviceName, operationName string, callOptions *CallOptions) (*OutboundCall, error) {
--- a/golang/channel.go
+++ b/golang/channel.go
@@ -259,6 +259,11 @@ func (ch *Channel) serve() {
 	}
 }
 
+// Logger returns the logger for this channel.
+func (ch *Channel) Logger() Logger {
+	return ch.log
+}
+
 // Closed returns whether this channel has been closed with .Close()
 func (ch *Channel) Closed() bool {
 	ch.mutable.mut.Lock()
--- a/golang/channel.go
+++ b/golang/channel.go
@@ -197,9 +197,6 @@ func (ch *Channel) GetSubChannel(serviceName string) *SubChannel {
 
 // Peers returns the PeerList for the channel.
 func (ch *Channel) Peers() *PeerList {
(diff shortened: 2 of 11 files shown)


golang/.gitignore
```diff
@@ -3,6 +3,4 @@
 .DS_Store
 build
 Godeps/_workspace
-messagetype_string.go
-reqresreaderstate_string.go
-reqreswriterstate_string.go
+*_string.go
```
abhinav: I was under the impression that you're supposed to check the generated code into the repository?

At least, that's what http://blog.golang.org/generate says.

golang/hyperbahn/registration.go
```diff
@@ -0,0 +1,101 @@
...
+	return registrationInterval + fuzz
+}
+
+// registrationLoops re-registers the service approximately every minute (with some fuzzing).
+func (c *Client) registrationLoop() {
+	sleepFor := c.fuzzedRegistrationInterval()
+	consecutiveFailures := 0
+
+	for {
+		timeSleep(sleepFor)
```
Raynos: One concept you should implement is the notion of `hardFail`

We have a few node programs that want to use "best effort" hyperbahn client and want to degrade by not talking to hyperbahn if it fails.

Other services want a hard failure if they cannot talk to hyperbahn because they only do hyperbahn and no HTTP.

golang/.gitignore
prashantv: I'll let @mmihic comment on this. I'm fine with either approach. If we decide to check in generated files, I'll do it as a separate PR.

golang/hyperbahn/registration.go
prashantv: I like the idea, and I think I want that to be the default. Added a "FailStrategy" option which defaults to Fatal by default which will kill the hosting app. The application has to explicitly opt in to FailStrategyIgnore if it does not want to Fatal.

OnError will be called either way.

golang/.gitignore
mmihic: I have a slight preference for not checking in generated code, but that's mostly a holdover from the old days of terrible IDL.  Whatever you prefer.

golang/channel.go
```diff
@@ -209,6 +209,14 @@ func (ch *Channel) GetSubChannel(serviceName string) *SubChannel {
 	return ch.registerNewSubChannel(serviceName)
 }
 
+// Peers returns the PeerList for the channel.
+func (ch *Channel) Peers() *PeerList {
+	ch.mutable.mut.RLock()
```
mmihic: Do you need the read mut?  I don't see where we ever update this  - looks like it is set in the ctor and never modified after.

golang/hyperbahn/client.go
```diff
@@ -0,0 +1,109 @@
...
+	tchan *tchannel.Channel
+	opts  ClientOptions
+}
+
+// FailStrategy is the strategy to use when registration fails maxRegistrationFailures
+// times consecutively in the background. This is not used if the initial registration fails.
+type FailStrategy int
+
+const (
+	// FailStrategyFatal will call Fatalf on the channel's logger after triggerring handler.OnError.
```
mmihic: nit: several mispellings of triggering as triggerring

golang/hyperbahn/registration.go
```diff
@@ -0,0 +1,104 @@
...
+type adRequest struct {
+	Services []service `json:"services"`
+}
+
+type adResponse struct {
+	ConnectionCount int `json:"connectionCount"`
+}
+
+func (c *Client) sendRegistration() error {
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second*5)
```
mmihic: include timeout in CallOptions?

mmihic:
lgtm

golang/channel.go
prashantv: You're right, no need for mutex!

prashantv:
Fixed the misspellings and other minor issues. Thanks for the late night review Matt

PR #711, opened 2015-06-15
fix the bug that outbound.calls.sent is emitted too late

ShanniLi:
r @Raynos @kriskowal

### DIFF
--- a/node/out_request.js
+++ b/node/out_request.js
@@ -301,6 +301,15 @@ TChannelOutRequest.prototype.send = function send(arg1, arg2, arg3, callback) {
         self.span.name = endpoint;
     }
 
+    if (self.logical === false && self.retryCount === 0) {
+        self.channel.outboundCallsSentStat.increment(1, {
+            'target-service': self.serviceName,
+            'service': self.headers.cn,
+            // TODO should always be buffer
+            'target-endpoint': endpoint
+        });
+    }
+
     if (callback) self.hookupCallback(callback);
 
     self.arg1 = arg1;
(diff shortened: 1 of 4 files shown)


node/out_request.js
```diff
@@ -301,15 +301,6 @@ TChannelOutRequest.prototype.send = function send(arg1, arg2, arg3, callback) {
         self.span.name = endpoint;
     }
 
-    if (self.retryCount === 0) {
```
Raynos: this needs to be changed to `if (self.logical === false) {`

All direct requests ( host option ) and streaming requests go through this code path and not through `request.js` 

See how we emit latency stats for an example.

ShanniLi: fixed.

Raynos:
lgtm.

node/request.js
```diff
@@ -153,6 +153,13 @@ TChannelRequest.prototype.send = function send(arg1, arg2, arg3, callback) {
     self.start = self.timers.now();
     self.resendSanity = self.limit + 1;
 
+    self.channel.outboundCallsSentStat.increment(1, {
+        'target-service': self.serviceName,
+        'service': self.headers.cn,
+        // TODO should always be buffer
```
kriskowal: target-endpoint should always be a string. This TODO is not necessary.

kriskowal:
:accept:

node/request.js
ShanniLi: I remembered @Raynos said that it could be buffer. Can you confirm, Jake?

PR #713, opened 2015-06-15
update tchannel response 'as' header

junchaowu:
@blampe @abhinav


blampe:
a test would be nice but :+1:

PR #715, opened 2015-06-16
Let us spell cancelled one way

kriskowal:
Protocol document says cancelled, so cancelled it shall be throughout.


Raynos:
This is technically a breaking change.

lgtm.

PR #716, opened 2015-06-16
request: cast Buffer to string

Raynos:
r: @kriskowal @shannili

### DIFF
--- a/node/request.js
+++ b/node/request.js
@@ -157,7 +157,7 @@ TChannelRequest.prototype.send = function send(arg1, arg2, arg3, callback) {
         'target-service': self.serviceName,
         'service': self.headers.cn,
         // TODO should always be buffer
-        'target-endpoint': self.arg1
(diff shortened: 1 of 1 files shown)


kriskowal:
:accept:

ShanniLi:
lgtm

node/request.js
```diff
@@ -157,7 +157,7 @@ TChannelRequest.prototype.send = function send(arg1, arg2, arg3, callback) {
         'target-service': self.serviceName,
         'service': self.headers.cn,
         // TODO should always be buffer
-        'target-endpoint': self.arg1
+        'target-endpoint': String(self.arg1)
```
jcorbin: well... that's an unfortunate performance hit, can we maybe then do this stat call after the critical ... resend?

Raynos:
We should store a string arg1 on the object so we do it once instead of for every stat and trace

PR #717, opened 2015-06-16
go: Change Serve to not block on processing sockets.

prashantv:
Serve currently does 2 things:
- Set LocalPeerInfo
- Handle sockets.

Hyperbahn client creation needs the LocalPeerInfo set, but there's no nice way to detect when this step is done. To clean this up, Serve will now return after setting LocalPeerInfo. The actual handling of sockets happens in a goroutine. Tests+examples are updated to not call Serve using a goroutine as it's no longer required.

### DIFF
--- a/golang/channel.go
+++ b/golang/channel.go
@@ -109,37 +109,10 @@ func NewChannel(serviceName string, opts *ChannelOptions) (*Channel, error) {
 	return ch, nil
 }
 
-// Serve serves incoming requests using the provided listener
+// Serve serves incoming requests using the provided listener.
+// The local peer info is set synchronously, but the actual socket listening is done in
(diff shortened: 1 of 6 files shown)


Raynos:
@prashantv 

The common way to do reviews is to add

"""
r: @mmihic 
"""

Which means matt will get an email and a notification in github. I'm pretty sure he doesn't get emailed if he is assigned as a reviewer.

mmihic:
It's cool, I'm subscribed to all of our github repos and I check my notifications pretty obsessively.

mmihic:
This makes sense and I like it better.  My only nit is that it is no longer consistent with `http.Server#ListenAndServe`, but I guess nothing to be done for that.

PR #721, opened 2015-06-16
go: Thrift+TChannel+Hyperbahn guide and example

prashantv:
Add a guide for how to create a simple service hosted on Thrift+TChannel+Hyperbahn
Add code for example server + client.

r: @mmihic @mranney


Raynos:
Read the guide; looks legit to me.

Issue #696, opened 2015-06-11
Node:  Request handler throws TypeError when there is no hostPort


Willyham:
I probably have some tracer misconfiguration. If that is the case then this should be caught further up the stack and at instantiation time, rather than crashing when handling a request.

```
2015-06-11T12:46:04.872Z - fatal: swingline.development.localhost Uncaught Exception: non_object_property_call stack=TypeError: Cannot call method 'split' of null
    at Agent.setupNewSpan (/home/willy/Projects/orion/goodwill/node_modules/tchannel/trace/agent.js:63:44)
    at new TChannelInRequest (/home/willy/Projects/orion/goodwill/node_modules/tchannel/in_request.js:57:36)
    at TChannelV2Handler.buildInRequest (/home/willy/Projects/orion/goodwill/node_modules/tchannel/v2/handler.js:618:16)
    at TChannelV2Handler.handleCallRequest (/home/willy/Projects/orion/goodwill/node_modules/tchannel/v2/handler.js:191:20)
    at TChannelV2Handler.handleFrame (/home/willy/Projects/orion/goodwill/node_modules/tchannel/v2/handler.js:125:25)
    at TChannelConnection.handleReadFrame (/home/willy/Projects/orion/goodwill/node_modules/tchannel/connection.js:231:18)
    at ReadMachine.handleReadFrame [as emit] (/home/willy/Projects/orion/goodwill/node_modules/tchannel/connection.js:174:14)
    at ReadMachine.seek (/home/willy/Projects/orion/goodwill/node_modules/tchannel/node_modules/bufrw/stream/read_machine.js:120:14)
    at ReadMachine.handleChunk (/home/willy/Projects/orion/goodwill/node_modules/tchannel/node_modules/bufrw/stream/read_machine.js:67:28)
    at Socket.onSocketChunk (/home/willy/Projects/orion/goodwill/node_modules/tchannel/connection.js:93:19)
    at Socket.emit (events.js:95:17)
    at Socket.<anonymous> (_stream_readable.js:765:14)
    at Socket.emit (events.js:92:17)
    at emitReadable_ (_stream_readable.js:427:10)
    at emitReadable (_stream_readable.js:423:5)
    at readableAddChunk (_stream_readable.js:166:9)
    at Socket.Readable.push (_stream_readable.js:128:10)
    at TCP.onread (net.js:529:21), message=Cannot call method 'split' of null, _hostname=localhost.localdomain, _pid=27314
Aborted (core dumped)
```

[...]

PR #722, opened 2015-06-16
WIP: demonstrate tracing epehemeral client failure

jcorbin:
Per #696

### DIFF
--- a/node/test/lib/alloc-cluster.js
+++ b/node/test/lib/alloc-cluster.js
@@ -109,14 +109,22 @@ function allocCluster(opts) {
 
     function createChannel(i) {
         var chan = TChannel(extend(channelOptions));
-        var port = opts.listen && opts.listen[i] || 0;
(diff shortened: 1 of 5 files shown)


node/test/lib/alloc-cluster.js
```diff
@@ -109,14 +109,22 @@ function allocCluster(opts) {
 
     function createChannel(i) {
         var chan = TChannel(extend(channelOptions));
-        var port = opts.listen && opts.listen[i] || 0;
-        chan.on('listening', chanReady);
-        chan.listen(port, host);
         cluster.channels[i] = chan;
 
+        var port = opts.listen ? opts.listen[i] : 0;
+        if (port === null) {
```
kriskowal: Should document somewhere that null is a sentinel value that means “do not listen on any port”.

Raynos:
This has probably been fixed at this point; will rebase at some point and verify.

PR #724, opened 2015-06-16
peers: support selfPeer for logical request

Raynos:
This change allows us to make a request to ourselves if
we put ourself in the peers list of a sub channel.

r: @shannili @jcorbin

### DIFF
--- a/node/peers.js
+++ b/node/peers.js
@@ -41,19 +41,24 @@ function TChannelPeers(channel, options) {
     self.options = options || {};
     self.initialPeerState = self.options.initialPeerState;
     self._map = Object.create(null);
-    self.selfPeer = TChannelSelfPeer(self.channel);
+    self.selfPeer = null;
 }
 
 inherits(TChannelPeers, EventEmitter);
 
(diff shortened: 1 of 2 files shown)


Raynos:
Python test failed ^

node/peers.js
```diff
@@ -41,19 +41,24 @@ function TChannelPeers(channel, options) {
...
+        peers.push(self.selfPeer);
+    }
+    var counter = peers.length + 1;
     peers.forEach(function eachPeer(peer) {
         peer.close(onClose);
     });
     self.clear();
+    onClose();
 
     function onClose() {
```
ShanniLi: why this function is necessary?

ShanniLi:
lgtm.

node/peers.js
Raynos: I set counter to be length + 1 and we call `onClose` for each peer; I call an extra `onClose` at the end.

This means that Peers will close if there are zero peers;

The alternative is `if (peers.length === 0) { cb(); }`

PR #726, opened 2015-06-16
Make the tombstone timeout dynamic

Raynos:
Previously it was a static timeout of 5 seconds.

Now it's dynamic based on `req.ttl`.

r: @rf @jcorbin


jcorbin:
:ok: works, the static offset I feel could be replaced with a percentage increase, but static is probably better for now.

PR #728, opened 2015-06-16
statsd: do not add peer for per-attempt latency

Raynos:
We currently send the peer to statsd. This has a huge
cardinality of O(n^2) because ringpop is fully connected.

This makes graphite painfully slow.

r: @shannili @jcorbin

### DIFF
--- a/node/lib/statsd.js
+++ b/node/lib/statsd.js
@@ -96,7 +96,6 @@ function getKey(common, stat) {
                 clean(stat.tags.service, 'service') + '.' +
                 clean(stat.tags['target-service'], 'target-service') + '.' +
                 clean(stat.tags['target-endpoint'], 'endpoint') + '.' +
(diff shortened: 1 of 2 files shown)


ShanniLi:
test failed. :(

ShanniLi:
btw, peer is actually a required field in this stats:

Latency of an individual attempt (initial request or retry) at a call. Measured from the time the first request fragment write is initiated to the time the last response fragment read is received. SHOULD include the peer tag containing the host and port of the peer to whom the call attempt was sent. SHOULD include a retry-count tag indicating the per-request retry count represented by the metric (e.g. the first outgoing attempt should have retry-count=0, the second outgoing attempt should have retry-count=2 etc)

mmihic:
It's optional.  SHOULD = optional but recommended, MUST = required.  See https://www.ietf.org/rfc/rfc2119.txt for spec language.

PR #730, opened 2015-06-16
Add error listeners to connection objects

Raynos:
I've added an error listener for incoming connections; it
previously went directly to uncaught

 - Add error listener for incoming
 - Add error log for outgoing connection errors
 - Fix relay_request timeout race condition.

r: @jcorbin @rf


jcorbin:
:ok: looks sensible

PR #731, opened 2015-06-16
checking header types

rf:
@Raynos

### DIFF
--- a/node/errors.js
+++ b/node/errors.js
@@ -164,6 +164,14 @@ module.exports.InvalidHandlerForRegister = TypedError({
     handler: null
 });
 
+module.exports.InvalidHeaderTypeError = TypedError({
+    type: 'tchannel.invalid-header-type',
+    message: 'invalid header type for header {name}; ' +
+        'expected string, got {headerType}',
+    headerType: null,
+    name: null
+});
+
 module.exports.InvalidJSONBody = TypedError({
     type: 'tchannel-handler.json.invalid-body',
     message: 'Invalid error body, expected a typed-error',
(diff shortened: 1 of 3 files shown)


Raynos:
Lgtm. r: @jcorbin 

This manual check gives a better error then the underlying `bufrw` error. Especially reporting the name of the header that is wrong helps debugging the issue.

node/errors.js
```diff
@@ -164,6 +164,14 @@ module.exports.InvalidHandlerForRegister = TypedError({
     handler: null
 });
 
+module.exports.InvalidHeaderTypeError = TypedError({
+    type: 'tchannel.invalid-header-type',
+    message: 'invalid header type for header {name}; expected string, got' +
```
jcorbin: we don't do line length limit in tchannel, please combine this

node/v2/header.js
```diff
@@ -78,6 +78,13 @@ HeaderRW.prototype.writeInto = function writeInto(headers, buffer, offset) {
         if (res.err) return res;
         offset = res.offset;
 
+        if (typeof headers[key] !== 'string') {
+            return errors.InvalidHeaderTypeError({
```
jcorbin: erm, this needs to return a result... so `return WriteResult.error(errors...);`

jcorbin: also, shouldn't a buffer be valid too, not just a string?

jcorbin: Also, github won't let me comment on it, but you want to do the same check in `HeaderRW#byteLength` above circa L57; there it'll be `LengthResult.error(....)`.

Finally, the fact that any of this didn't fail a test is a bug, please cover these paths and prove they work.

Raynos:
Let's adding a failing test too.

rf:
:+1:

Issue #656, opened 2015-06-05
Node as handler streamlining


Raynos:
Currently the interface for as-x is 

```js
clients.steveThrift.send(clients.steveChannel.request({ 
    timeout: 500,
    hasNoParent: true
}), 'Steve::getSteves', null, {
    request: {
        userUuid: 10
    }
}, function onResponse(err, resp) {
    if (err || !resp.ok) {
        /* endpoint specific error handling */
    }

    var steves = resp.body.steves;
    /* endpoint specific doing stuff with steves */
});
```

I would like to propose

```js
clients.steveThrift.request({ 
    timeout: 500,
    hasNoParent: true
}).send('Steve::getSteves', null, {
    request: {
        userUuid: 10
    }
}, function onResponse(err, resp) {
    if (err || !resp.ok) {
        /* endpoint specific error handling */
    }

    var steves = resp.body.steves;
    /* endpoint specific doing stuff with steves */
});
```

To implement this we would need to pass a sub channel into the `TChannelAsThrift` constructor.

cc @kriskowal @rf @Willyham @jcorbin

kriskowal:
This implies the creation of another layer of Request proxy, RequestAsThrift, etc.

I’m indifferent. I believe @jcorbin already expressed an interest in going this direction.

jcorbin:
:+1:

Raynos:
If people can come up with other proposals that simplify the interface further then I'm :+1: to that too.

jcorbin:
Actually I'm completely :+1: on having explicit RequestAs* and even ResponseAs* layers, I think it will only help.

Willyham:
I think it's definitely an improvement, but I'd imagine most people would break the request object out anyway, so there's not much of a difference between the two:

```
var request = clients.steveChannel.request({ 
  timeout: 500,
  hasNoParent: true
});
clients.steveThrift.send(request, Steve::getSteves' {...}, function() {...})
```

and 
```
var request = { 
  timeout: 500,
  hasNoParent: true
}
clients.steveThrift.request(request).send('Steve::getSteves', {...}, function() {...})
```
If it's clearer from an implementation perspective then sure, this is probably a win-win.

I think an ideal client experience would be something more like:

```
clients.steveThrift.getSteves(params, requestParams, function(){...})
```

Raynos:
This is now implemented.

Willyham:
Is it backwards compatible?

Raynos:
[...]

PR #732, opened 2015-06-16
support TChannelThrift#request helper

Raynos:
This makes it easier to making outgoing requests as you only
need the TChannelAsX instance to making outgoing reqs.

See #656

r: @jcorbin @Willyham @rf


jcorbin:
:ok:

PR #734, opened 2015-06-16
Add function to get a local IP to use when listening

prashantv:


mmihic:
I guess this is hard to test.  lgtm

PR #735, opened 2015-06-16
add inbound.request.size & inbound.response.size to metrics

ShanniLi:
r @Raynos @jcorbin @rf

### DIFF
--- a/node/v2/handler.js
+++ b/node/v2/handler.js
@@ -231,7 +231,6 @@ TChannelV2Handler.prototype.handleCallRequest = function handleCallRequest(reqFr
             });
         }
     }
-
     if (reqFrame.body.args && reqFrame.body.args[0] &&
         reqFrame.body.args[0].length > v2.CallRequest.MaxArg1Size) {
         req.res = self.buildOutResponse(req);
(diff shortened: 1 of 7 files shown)


node/v2/handler.js
```diff
@@ -289,6 +293,14 @@ TChannelV2Handler.prototype.handleCallResponse = function handleCallResponse(res
                 limit: '0x' + v2.CallResponse.MaxArg1Size.toString(16)
         }));
     }
+
+    var req = self.connection.ops.getOutReq(res.id);
+    self.connection.channel.inboundResponseSizeStat.increment(resFrame.size, {
+        'calling-service': !req ? null : req.headers.cn,
```
Raynos: dont use null; use `'unknown'` or something.

Raynos:
Other then the null issue; lgtm.

node/v2/handler.js
ShanniLi: it is handled in statsd.js.

Raynos:
lgtm

node/v2/handler.js
rf: yeah if its null it will be no-calling-service, no-service, etc. unknown-calling-service would probably be better though

rf:
Agree with Jake on the null thing and :+1: otherwise

node/v2/handler.js
ShanniLi: @rf it used to be unknown. Then @kriskowal suggested the current name. :) I don't really have any preference on the default names. They shouldn't be showing up in almost any of these cases.

PR #737, opened 2015-06-17
Add more context to out request errors

Raynos:
This adds more context to the following errors

 - Not setting `parent` or `hasNoParent`
 - Not setting "as" header
 - Not setting "cn" header

These are currently difficult to debug because we have the
init blocking logic and the actual callsite to `request()`
will not be in the error stack.

By adding the serviceName / arg1 it will be easier to upgrade
older services to newer versions of tchannel

r: @jcorbin @shannili


ShanniLi:
lgtm

PR #738, opened 2015-06-17
add metrics for outbound per-attempt errors

ShanniLi:
r @Raynos @rf

### DIFF
--- a/node/out_request.js
+++ b/node/out_request.js
@@ -147,7 +147,8 @@ function emitPerAttemptErrorStat(err) {
             'service': self.headers.cn,
             // TODO should always be buffer
             'target-endpoint': String(self.arg1),
-            'type': err.codeName
+            'type': err.codeName,
+            'retry-count': self.retryCount
         });
     } else {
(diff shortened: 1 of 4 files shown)


node/out_request.js
```diff
@@ -135,6 +137,28 @@ TChannelOutRequest.prototype.onResponse = function onResponse(res, self) {
...
+function emitPerAttemptErrorStat(err) {
+    var self = this;
+
+    if (err.isErrorFrame) {
+        self.channel.outboundCallsPerAttemptSystemErrorsStat.increment(1, {
+            'target-service': self.serviceName,
+            'service': self.headers.cn,
+            // TODO should always be buffer
+            'target-endpoint': String(self.arg1),
+            'type': err.codeName
```
Raynos: let's add retry-count to all of these; like the per attempt latency.

Raynos:
Let's add that retry count; otherwise looking good.

Raynos:
#SHIPPINGIT

Raynos:
lgtm

PR #742, opened 2015-06-18
switching from int to int64

badiib:
This should fix this error I was getting:

src/github.com/uber/tchannel/golang/hyperbahn/registration.go:75: constant 10000000000 overflows int
src/github.com/uber/tchannel/golang/hyperbahn/registration.go:75: constant 20000000000 overflows int


prashantv:
lgtm, feel free to merge.

PR #748, opened 2015-06-19
Enforce that the ttl is non-zero

Raynos:
In the protocol doc the ttl must be non-zero.

cc @abhinav @prashant I hope your python/go implementation
also enforces this.

r: @jcorbin @rf @shannili

### DIFF
--- a/node/v2/call.js
+++ b/node/v2/call.js
@@ -148,6 +148,12 @@ function writeCallReqInto(body, buffer, offset) {
     // flags:1 -- filled in later after argsrw
     offset += bufrw.UInt8.width;
 
+    if (body.ttl === 0) {
+        return bufrw.WriteResult.error(errors.InvalidTTL({
+            ttl: body.ttl
+        }), offset);
+    }
+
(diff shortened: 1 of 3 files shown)


ShanniLi:
lgtm

node/v2/call.js
```diff
@@ -139,6 +148,12 @@ function writeCallReqInto(body, buffer, offset) {
     // flags:1 -- filled in later after argsrw
     offset += bufrw.UInt8.width;
 
+    if (body.ttl === 0) {
```
rf: Is this coerced to a positive integer somewhere? Shouldn't we be checking if it's > 0? What if the user doesn't pass in a ttl at all? Then it's certainly not exactly 0, but it is an invalid ttl.

Raynos: True dat.

Raynos:
Adding gaurding against negative ttl.

abhinav:
@junchaowu what's the Python behavior around this?

PR #749, opened 2015-06-19
go: Require a TTL to be set for all outgoing calls.

prashantv:
r: @mmihic 
cc: @Raynos

### DIFF
--- a/golang/outbound.go
+++ b/golang/outbound.go
@@ -46,9 +46,13 @@ func (c *Connection) beginCall(ctx context.Context, serviceName string, callOpti
 		return nil, err
 	}
 
-	deadline, _ := ctx.Deadline()
+	deadline, ok := ctx.Deadline()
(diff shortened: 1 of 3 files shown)


mmihic:
lgtm

golang/outbound.go
```diff
@@ -46,9 +46,13 @@ func (c *Connection) beginCall(ctx context.Context, serviceName string, callOpti
 		return nil, err
 	}
 
-	deadline, _ := ctx.Deadline()
+	deadline, ok := ctx.Deadline()
```
Raynos: consider defaulting a deadline ? In node we default you into 100ms if you do not set a timeout.

Raynos:
lgtm.

prashantv:
Keeping explicit as it's easy to decide on a timeout when creating a context.

PR #751, opened 2015-06-20
Refactor hyperbahn test to use cluster.dummies

Raynos:
To get the tests to be correct they should use dummies instead of
remotes. Using remotes is incorrect because remotes come
pre-advertised with hyperbahn.

This fixes the tests upstream in autobahn.

cc: @shannili @rf


Raynos:
Self merging. It's just test fixes. It's also in the 2.0.0 critical path.

Raynos:
Travis stamped.

PR #753, opened 2015-06-20
Retry hyperbahn relay-ad on timeouts.

Raynos:
Currently it's marking the relay-ad as an `error()` and failing it.
Instead we should just retry on timeouts.

I'm also making all these magical numbers configurable.

r: @shannili @jcorbin


Raynos:
I am an adult and I want to deploy this change.

ShanniLi:
lgtm

PR #760, opened 2015-06-20
use host instead of hostPort for statsd

ShanniLi:
r @Raynos @jcorbin @rf

### DIFF
--- a/node/lib/statsd.js
+++ b/node/lib/statsd.js
@@ -48,6 +48,18 @@ function clean(str, field) {
     }
 }
 
+function cleanHostPort(str, field) {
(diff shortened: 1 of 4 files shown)


node/lib/statsd.js
```diff
@@ -48,6 +48,18 @@ function clean(str, field) {
     }
 }
 
+function cleanHostPort(str, field) {
+    if (!str) {
+        return 'no-' + field;
+    } else {
+        return str
```
Raynos: leave a comment here saying "assuming ipv4"

Raynos:
lgtm.

PR #761, opened 2015-06-21
hyperbahn: rename register() to advertise()

Raynos:
Whilst writing the tchannel guide I realized we used the word
`register()` twice. Once to mean registrying an endpoint with
a tchannel server and once to mean registrying a server with
hyperbahn.

Internally hyperbahn calls the "ad" method short for advertise.
We've already renamed hyperbahns advertisement protocol to be
around the word "advertise".

This PR changes the hyperbahnClient interface to have an
`advertise()` method. This means that when we document `register()`
it is no longer ambigious.

I've left the old `register()` method in place for 100% back-compat.

r: @shannili @rf @jcorbin

cc: @prashant @abhinav It would be good to rename python and go. It
looks like python already calls this advertise() but it looks like
go uses the word register() twice.

### DIFF
--- a/node/bin/adhoc-multi.js
+++ b/node/bin/adhoc-multi.js
@@ -37,7 +37,7 @@ var fs = require('fs');
 var CountedReadySignal = require('ready-signal/counted');
 
 var TChannel = require('../');
-var AutobahnClient = require('../hyperbahn/');
+var HyperbahnClient = require('../hyperbahn/');
 var DebugLogger = require('debug-logtron');
 
 if (require.main === module) {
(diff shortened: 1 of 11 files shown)


ShanniLi:
yes, advertise is better. It is used only by services.

anson627:
Any implication for xlate?

anson627:
FYI: CI failed.

ShanniLi:
lgtm, please fix the test. :)

Raynos:
@anson627 no implications. 100% back-compat

ShanniLi:
@anson627 this is non-breaking change. The .register() method will still be there (deprecated). xlate just need to change from .register() to .advertise()

anson627:
@ShanniLi cool, I will keep that in mind.

Raynos:
Travis failed because public npm has an outage as of 10 minutes ago.

anson627:
@Raynos sounds fair.

PR #763, opened 2015-06-22
Write the tchannel + hyperbahn guide

Raynos:
This is the the guide for how to node.

It comes with a keyvalue example.

Thank you @prashant for your guide; mine is very
similar :)

r: @rf @shannili @jcorbin

cc @abhinav @prashant @mranney Please check out the
node guide; It's useful to get feedback from people
that don't extensively use tchannel node.


Raynos:
Also cc @blampe would love feedback on the guide as someone who hasn't used node tchannel much.

Raynos:
This is a blocker for 2.0.0. I've merged it to finish the 2.0.0 release

**However:** this really needs review; i would still love feedback from everyone.

node/GUIDE.md
```diff
@@ -0,0 +1,230 @@
...
+    )
+    void put_v1(
+        1: string key,
+        2: string value
+    )
+}
+```
+
+Our service will have two tchannel endpoints that can be.
+Namely `KeyValue::get_v1` and `KeyValue::put_v1`.
```
abhinav: Nit: Personally, I wouldn't encode the version number in the method name. There's nothing stopping you from creating multiple Thrift `service`s behind the same TChannel service since endpoints are namespaced to the service name.

node/GUIDE.md
```diff
@@ -0,0 +1,230 @@
...
+        1: string key
+    )
+    void put_v1(
+        1: string key,
+        2: string value
+    )
+}
+```
+
+Our service will have two tchannel endpoints that can be.
```
prashantv: "that can be." - missing word?

node/GUIDE.md
```diff
@@ -0,0 +1,230 @@
...
+Namely `KeyValue::get_v1` and `KeyValue::put_v1`.
+
+## Writing a tchannel server
+
+To get a server up and running you need to do the following:
+
+1. Create a TChannel and listen on a port
+2. Configure your endpoint handlers using TChannelThrift
+3. Create a hyperbahn client and register with hyperbahn
```
prashantv: s/register with hyperbahn/advertise with hyperbahn/ to match the new name

node/GUIDE.md
```diff
@@ -0,0 +1,230 @@
...
+});
+var ctx = {
+    store: {}
+};
+
+keyThrift.register(keyChan, 'KeyValue::get_v1', ctx, get);
+keyThrift.register(keyChane, 'KeyValue::put_v1', ctx, put);
+
+function get(context, req, head, body, cb) {
+    cb(null, {
```
prashantv: I'm not sure if this is a Node-ism, but I'm not sure what the "null" parameter for callback is. It might be worth adding a comment to the code example so explain it.

I'm extending my Go guide with this, might be worth doing here too - which is to add an expected exception to the thrift file for if the key is not found, and return that here if the key passed to get is not found).

node/GUIDE.md
```diff
@@ -0,0 +1,230 @@
...
+ - Making outgoing call requests using TChannelThrift codec
+
+### Creating a tchannel
+
+TChannel is a bidirectional RPC library; you already know how
+to create a root channel.
+
+A pure client does not need to call `.listen()` but it's
+recommended that you `.listen()` anyway and `advertise()` anyway.
+It's always best to get onto the hyperbahn as early as possible.
```
prashantv: Is there any reason why pure clients should call listen and advertise?

blampe:
Looks great, we need to add something like this for Python.

node/GUIDE.md
mranney: Yes, this is a node-ism. The convention for callbacks it to be (err, res), where the first arg is an Error object or null. This is how node handles the multiple return value thing with Go.

node/GUIDE.md
Raynos: I'd like to recommend encoding the version name somewhere. Putting it in the method name allows us to change the version of one method at a time with maximum granularity.

The idea that you "simply never break back-compat" is a valid idea but thats not how the real world works.

node/GUIDE.md
Raynos: The normal case is for you to listen and use incoming connections. If you do not do this you will use random outgoing connections.

There is no reason for pure clients to advertise; but there should be no pure clients.

PR #765, opened 2015-06-22
Metrics emit & bug fix for connections.active

ShanniLi:
r @Raynos @jcorbin @rf

### DIFF
--- a/node/channel.js
+++ b/node/channel.js
@@ -67,41 +67,6 @@ function TChannel(options) {
     self.connectionEvent = self.defineEvent('connection');
     self.requestEvent = self.defineEvent('request');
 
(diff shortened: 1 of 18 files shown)


node/channel.js
```diff
@@ -111,6 +76,8 @@ function TChannel(options) {
         processName: format('%s[%s]', process.title, process.pid)
     }, options);
 
+    self.emittingStats = self.options.emittingStats;
```
Raynos: why ?

Raynos:
All the tests should close; we should emit stats always.

We cannot run in production without stats.

Please pay the first mover depth of calling close() everywhere.

ShanniLi:
closing this PR. Will create a new one.

PR #771, opened 2015-06-23
go: Fix for partial reads causing connection corruption

prashantv:
When stress testing, I noticed some data corruption. After some
debugging, I found out that we weren't reading some packets
completely and the remaining part of the packet was being read
as the frame header.

Read may not fill the whole buffer, and so we were only partially
reading buffers and then the rest of the buffer was read as a separate
packet.


mmihic:
good catch

prashantv:
Ignoring unrelated python test failure and merging.

PR #776, opened 2015-06-23
peer: report spans through root channel

Raynos:
r @shannili @rf


rf:
if it passes tests then fine

PR #777, opened 2015-06-23
Go: Better support for frame pooling

### LINKED ISSUE #275: golang: implement frame leak detection
golang tchannel support frame pooling to avoid regular large memory allocations.  To work efficiently, frames need to be returned to the pool after they (and they buffers) have been fully used.  We're not currently doing this everywhere.  We should a) make sure we do this and b) add a leak detection pool that verifies that can be used in test runs to verify that all frames are recycled appropriately, and no frame is used after it has been recycled.  The start of this is at  https://github.com/uber/tchannel/tree/mmihic-frame-leak-detector but is blocked on https://github.com/uber/tchannel/issues/274 since we inherently leak pending read frames when a test ends.

Note that failing to return a frame to a pool doesn't actually result in a memory leak (GC will still reclaim it); we're just not getting the benefits of pooling.

prashantv:
- Add a test to track frames and make sure none are leaked.
- Release frames in the successful case

Most of the change is for incoming call req fragments. We need to keep frames till the contents from that frame are read. We have to plumb framepool through mex so fragmenting reader / reqResReader can free frames once they complete parsing all contents in the frame.

Fixes #275 for the successful cases.

### DIFF
--- a/golang/fragmenting_reader.go
+++ b/golang/fragmenting_reader.go
@@ -44,6 +44,7 @@ type readableFragment struct {
 	checksumType ChecksumType
 	checksum     []byte
 	contents     *typed.ReadBuffer
+	done         func()
 }
 
 type fragmentReceiver interface {
@@ -72,6 +73,7 @@ type fragmentingReader struct {
 	curChunk         []byte
 	hasMoreFragments bool
 	receiver         fragmentReceiver
+	curFragment      *readableFragment
 	checksum         Checksum
 	err              error
 }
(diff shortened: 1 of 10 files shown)


golang/fragmenting_reader.go
```diff
@@ -44,6 +44,7 @@ type readableFragment struct {
 	checksumType ChecksumType
 	checksum     []byte
 	contents     *typed.ReadBuffer
+	done         func()
```
mmihic: I like this.

mmihic:
:+1: Thanks for cleaning this up, I apologize to have left you this particular mess.

PR #780, opened 2015-06-24
Peer should choose the identified connection first

ShanniLi:
r @Raynos @rf @jcorbin

### DIFF
--- a/node/peer.js
+++ b/node/peer.js
@@ -149,18 +149,15 @@ TChannelPeer.prototype.getInConnection = function getInConnection() {
     return null;
 };
 
-TChannelPeer.prototype.getOutConnection = function getOutConnection(identified) {
+TChannelPeer.prototype.getOutConnection = function getOutConnection(preferIdentified) {
     var self = this;
     var candidate = null;
     for (var i = self.connections.length - 1; i >= 0; i--) {
         var conn = self.connections[i];
-        if (!conn.closing) {
-            if (!identified || conn.remoteName) {
-                return conn;
-            } else {
-                candidate = candidate || conn;
-            }
-        }
+        if (conn.closing) continue;
+        if (!preferIdentified) return conn; // user doesn't care, take last outgoing
+        if (conn.remoteName) return conn; // user wanted an identified channel, and we found one
+        if (!candidate) candidate = conn; // we'll fallback to returning this if we can't find an identified one
     }
     return candidate;
 };
--- a/node/peer.js
+++ b/node/peer.js
@@ -149,13 +149,20 @@ TChannelPeer.prototype.getInConnection = function getInConnection() {
     return null;
 };
 
-TChannelPeer.prototype.getOutConnection = function getOutConnection() {
+TChannelPeer.prototype.getOutConnection = function getOutConnection(identified) {
     var self = this;
+    var candidate = null;
     for (var i = self.connections.length - 1; i >= 0; i--) {
         var conn = self.connections[i];
-        if (!conn.closing) return conn;
+        if (!conn.closing) {
+            if (!identified || conn.remoteName) {
+                return conn;
+            } else {
+                candidate = candidate || conn;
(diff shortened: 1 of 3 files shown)


node/peer.js
```diff
@@ -149,13 +149,20 @@ TChannelPeer.prototype.getInConnection = function getInConnection() {
...
 
-TChannelPeer.prototype.getOutConnection = function getOutConnection() {
+TChannelPeer.prototype.getOutConnection = function getOutConnection(identified) {
     var self = this;
+    var candidate = null;
     for (var i = self.connections.length - 1; i >= 0; i--) {
         var conn = self.connections[i];
-        if (!conn.closing) return conn;
+        if (!conn.closing) {
+            if (!identified || conn.remoteName) {
```
anson627: any way to make it more readable?

jcorbin: :+1:  please let's do:
```
if (conn.closing) continue;
if (!identified) return conn; // user doesn't care, take last outgoing
if (conn.remoteName) return conn; // user wanted an identified channel, and we found one
if (!candidate) candidate = conn; // we'll fallback to returning this if we can't find an identified one
```

Also please `s/identified/preferIdentified/`, since I'd infer otherwise that if I set `identified` to true, it's a guarantee.

node/test/peer.js
```diff
@@ -0,0 +1,66 @@
...
+        var socket = peer.makeOutSocket();
+        var conn = peer.makeOutConnection(socket);
+        peer.addConnection(conn);
+        assert.doesNotThrow(
+            function noThrow() {
+                peer.request({hasNoParent: true});
+            },
+            'should use the identified connection'
+        );
+
```
jcorbin: please also test the fallback path you've introduced here.

node/peer.js
ShanniLi: :+1:  will do :)

node/peer.js
```diff
@@ -149,13 +149,17 @@ TChannelPeer.prototype.getInConnection = function getInConnection() {
...
     for (var i = self.connections.length - 1; i >= 0; i--) {
         var conn = self.connections[i];
-        if (!conn.closing) return conn;
+        if (conn.closing) continue;
+        if (!preferIdentified) return conn; // user doesn't care, take last outgoing
+        if (conn.remoteName) return conn; // user wanted an identified channel, and we found one
+        if (!candidate) candidate = conn; // we'll fallback to returning this if we can't find an identified one
     }
-    return null;
+    return candidate;
```
Raynos: wont this make `waitForIdentified` blow up ?

Raynos:
We need to add logic to `waitForIdentified` to use `isIdentified()` instead of `isConnected()`

Raynos:
`waitForIdentified` should change

`205:         if (!self.isConnected() && conn.handler) {`

to 

`205:        if (!conn.remoteName) {`

Raynos:
lgtm

anson627:
looks good

node/test/peer.js
jcorbin: I don't see as you tested the fallback path? when you pass true, but there aren't any identified, it should fall back to "best effort, here you go"

ShanniLi: it is tested in the second test: 'peer should return the latest connection when none is identified'

node/test/peer.js
```diff
@@ -0,0 +1,98 @@
...
+        serviceName: 'server',
+        peers: [server.hostPort]
+    });
+
+    var peer = subClient.peers.get(serverHost);
+    var socket = peer.makeOutSocket();
+    var conn = peer.makeOutConnection(socket);
+    peer.addConnection(conn);
+
+    assert.equals(conn, peer.getOutConnection(true), 'should return the latest connection');
```
ShanniLi: @jcorbin the fallback path is tested here, where no connection is identified.

PR #781, opened 2015-06-24
Fix typos in the guide

Raynos:


Raynos:
This is just markdown.

PR #782, opened 2015-06-24
bring back the examples

blampe:
These were broken when we made `name` a required argument.

@abhinav @junchaowu @breerly

### DIFF
--- a/python/examples/options.py
+++ b/python/examples/options.py
@@ -21,7 +21,7 @@
 import argparse
 
 
-def get_args():
+def get_parser():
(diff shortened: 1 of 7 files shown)


breerly:
Almost wonder if we can write integration tests for these inline, then run them as part of our test suite...

breerly:
This'll do for now :+1:

Raynos:
Node spawns the examples as child processes and checks for exit zero as part of tests.

blampe:
Yeah we need something like that.

PR #784, opened 2015-06-25
add outbound stats counters

junchaowu:
@blampe @abhinav @breerly 
excluding
outbound.calls.retries
outbound.request.size
outbound.response.size

### DIFF
--- a/python/tchannel/event.py
+++ b/python/tchannel/event.py
@@ -40,8 +40,11 @@ EventType = enum(
     after_receive_request=0x21,
     before_receive_response=0x30,
     after_receive_response=0x31,
-    after_receive_error=0x40,
-    after_send_error=0x41,
+    after_receive_system_error=0x40,
+    after_receive_system_error_per_attempt=0x41,
+    after_send_system_error=0x42,
+    on_operational_error=0x50,
+    on_operational_error_per_attempt=0x51,
 )
 
 
@@ -86,16 +89,28 @@ class EventHook(object):
         """Not implemented."""
         pass
 
-    def after_receive_response(self, response):
+    def after_receive_response(self, request, response):
         """Called after a ``CALL_RESP`` message is read."""
         pass
 
-    def after_receive_error(self, err):
+    def after_receive_system_error(self, request, err):
         """Called after a ``error`` message is read."""
         pass
 
-    def after_send_error(self, err):
+    def after_send_system_error(self, err):
         """Called after a ``error`` message is sent."""
+        pass
+
+    def after_receive_system_error_per_attempt(self, request, err):
+        """Called after a ``error`` message is read for each attempt."""
+        pass
+
+    def on_operational_error(self, request, err):
+        """Called after an operational error happens per request."""
+        pass
+
+    def on_operational_error_per_attempt(self, request, err):
+        """Called after an operational error happens for each attempt."""
         pass
 
 
--- a/python/tchannel/stats.py
+++ b/python/tchannel/stats.py
@@ -134,4 +134,4 @@ def clean(key, field):
             "{", "-").replace(
             "}", "-").replace(
             ":", "-").replace(
-            " ", "_")
+            " ", "-")
--- a/python/tchannel/stats.py
+++ b/python/tchannel/stats.py
@@ -1,0 +1,137 @@
+# Copyright (c) 2015 Uber Technologies, Inc.
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+
+from __future__ import absolute_import
+
+from .event import EventHook
+from .messages.error import ErrorMessage
+from .tornado.response import StatusCode
+
+
+class StatsHook(EventHook):
+
+    def __init__(self, stats):
+        self._stats = stats
+
+    def before_send_request(self, request):
+        stats_name = "tchannel.outbound.calls.sent"
+        (service, target_service, target_endpoint) = (
+            extract_meta_info_from_request(request))
+
+        key = '.'.join([stats_name,
+                        clean(service, 'service'),
+                        clean(target_service, 'target-service'),
+                        clean(target_endpoint, 'target-endpoint')
+                        ])
+
+        self._stats.count(key, 1)
+
+    def after_receive_response(self, request, response):
+        if response.code == StatusCode.ok:
+            stats_name = "tchannel.outbound.calls.success"
+        else:
+            stats_name = "tchannel.outbound.calls.app-errors"
+        (service, target_service, target_endpoint) = (
+            extract_meta_info_from_request(request))
+        key = '.'.join([stats_name,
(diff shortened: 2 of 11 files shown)


python/tchannel/stats.py
```diff
@@ -0,0 +1,137 @@
...
+        return 'no-' + field
+    else:
+
+        return key.replace(
+            ".", "-").replace(
+            "\/", "-").replace(
+            "{", "-").replace(
+            "}", "-").replace(
+            ":", "-").replace(
+            " ", "_")
```
breerly: Looks like this is an underscore & not a hyphen:

```
>>> "_" == "-"
False
```

python/tchannel/event.py
```diff
@@ -46,8 +46,12 @@ class EventType(IntEnum):
     before_receive_response = 0x30
     after_receive_response = 0x31
 
-    after_receive_error = 0x40
-    after_send_error = 0x41
+    after_receive_system_error = 0x40
```
blampe: I just killed the enum package so you should update with master.

python/tchannel/stats.py
```diff
@@ -0,0 +1,137 @@
...
+from __future__ import absolute_import
+
+from .event import EventHook
+from .messages.error import ErrorMessage
+from .tornado.response import StatusCode
+
+
+class StatsHook(EventHook):
+
+    def __init__(self, stats):
```
blampe: Call it statsd and mention that it's an instance of https://github.com/etsy/statsd?

junchaowu:
rebased on master

python/tchannel/event.py
```diff
@@ -40,8 +40,11 @@
     after_receive_request=0x21,
     before_receive_response=0x30,
     after_receive_response=0x31,
-    after_receive_error=0x40,
-    after_send_error=0x41,
+    after_receive_system_error=0x40,
+    after_receive_system_error_per_attempt=0x41,
```
blampe: The naming here is confusing, what's the difference between `after_recieve` and `after_recieve_per_attempt`?

python/tchannel/statsd.py
```diff
@@ -0,0 +1,143 @@
...
+            statsd_name = "tchannel.outbound.calls.success"
+        else:
+            statsd_name = "tchannel.outbound.calls.app-errors"
+        (service, target_service, target_endpoint) = (
+            extract_meta_info_from_request(request))
+        key = '.'.join([statsd_name,
+                        clean(service, 'service'),
+                        clean(target_service, 'target-service'),
+                        clean(target_endpoint, 'target-endpoint')
+                        ])
```
blampe: Lots of duplicated code, make a helper method.

python/tchannel/statsd.py
```diff
@@ -0,0 +1,143 @@
...
+
+    return (service, target_service, target_endpoint)
+
+
+def clean(key, field):
+    if not key:
+        return 'no-' + field
+    else:
+
+        return key.replace(
```
blampe: Something like `return re.sub('[.{}: ]', '-', key)`.

python/tchannel/event.py
junchaowu: Defined in the doc https://github.com/uber/tchannel/blob/master/docs/metrics.md
anything with per_attempt means considering each attempt situation in the retry. Anything without per_attempt means only consider the number of times to report to user.

python/tchannel/zipkin/zipkin_trace.py
```diff
@@ -87,7 +87,7 @@ def after_receive_error(self, error):
         error.tracing.annotations.append(ann)
         self.tracer.record([(error.tracing, error.tracing.annotations)])
 
-    def after_send_error(self, error):
+    def after_send_system_error(self, error):
```
blampe: Why do some get the request and others don't?

python/tchannel/statsd.py
```diff
@@ -0,0 +1,143 @@
...
+
+
+class StatsdHook(EventHook):
+    """Collect Statsd information in the tchannel req/resp."""
+
+    def __init__(self, statsd):
+        """
+
+        :param statsd: instance of StatsD.
+            `Link text https://github.com/etsy/statsd`
```
abhinav: What? You probably meant:

    instance of `statsd <https://github.com/etsy/statsd>`_ 

The `<`, `>`, and `_` are not optional. `Link text` refers to the text you want to appear in place of the link.

python/tchannel/statsd.py
```diff
@@ -0,0 +1,143 @@
...
+                        clean(ErrorMessage.ERROR_CODES.
+                              get(error.code, None), 'type')
+                        ])
+
+        self._statsd.count(key, 1)
+
+    def on_operational_error(self, request, error):
+        statsd_name = "tchannel.outbound.calls.operational-errors"
+        (service, target_service, target_endpoint) = (
+            extract_meta_info_from_request(request))
```
abhinav: x = (
        foo()
    )

is better than

    x = (
        foo())

python/tchannel/statsd.py
```diff
@@ -0,0 +1,143 @@
...
+                        clean(target_service, 'target-service'),
+                        clean(target_endpoint, 'target-endpoint'),
+                        clean(ErrorMessage.ERROR_CODES.
+                              get(error.code, None), 'type')
+                        ])
+
+        self._statsd.count(key, 1)
+
+
+def extract_meta_info_from_request(request):
```
abhinav: Call this `extract_metadata`. `from_request` is implied; the argument is called request.

python/tchannel/statsd.py
junchaowu: There is a performance comparison http://stackoverflow.com/questions/3411771/multiple-character-replace-with-python

python/tchannel/zipkin/zipkin_trace.py
junchaowu: Now I am only update the outbound events now. For others, will be updated in the inbound events.

python/tchannel/statsd.py
```diff
@@ -0,0 +1,143 @@
...
+        return 'no-' + field
+    else:
+
+        return key.replace(
+            ".", "-").replace(
+            "\/", "-").replace(
+            "{", "-").replace(
+            "}", "-").replace(
+            ":", "-").replace(
+            " ", "-")
```
abhinav: Use a regex-based substitution. 

Precompile it and put it in a constant

    re.compile(r'[{}/\\:\s.]')

And use [`.sub`](https://docs.python.org/2/library/re.html#re.sub) to replace all occurrences with `-`.

An advantage of this is that you can add a `+` to the character class and make it so that multiple adjacent occurrences of the special symbols are replaced with a single dash instead of two.

    >>> r = re.compile(r'[{}/\\:\s.]+')
    >>> r.sub('-', 'foo{}bar')
    'foo-bar'

python/tchannel/tornado/peer.py
```diff
@@ -548,10 +550,40 @@ def send(self, arg1, arg2, arg3,
...
-            request, peer, attempt_times, retry_delay)
+        # event: send_request
+        self.peer.tchannel.event_emitter.fire(
+            EventType.before_send_request,
+            request,
+        )
+
+        try:
+            response = yield self.send_with_retry(
+                request, peer, attempt_times, retry_delay)
```
abhinav: put `)` on separate line. be consistent.

python/tchannel/statsd.py
abhinav: 7 replacements aren't the same as two.

Personally I'd like to work on performance optimization later.

python/tchannel/statsd.py
```diff
@@ -0,0 +1,120 @@
...
+    )
+
+    return '.'.join([statsd_name,
+                     clean(service, 'service'),
+                     clean(target_service, 'target-service'),
+                     clean(target_endpoint, 'target-endpoint')
+                     ])
+
+
+rexp = re.compile(r'[{}/\\:\s.]+')
```
abhinav: Use a better name (preferably all-caps) and put it at the top of the file after the imports, before the definitions.

python/tchannel/tornado/connection.py
```diff
@@ -618,21 +611,19 @@ def send_request(self, request):
...
             response_future.set_result(response)
-            # event: after_receive_response
-            self.tchannel.event_emitter.fire(
-                EventType.after_receive_response,
-                response,
-            )
+
+    def remove_outstanding_request(self, request):
+        """Remove request from pending request list"""
+        self._outstanding.pop(request.id, None)
```
abhinav: A precondition to doing this is that the request has finished, right? That is, the future has stopped running. If that turns out to not be the case, is that a programmer error? Maybe you want an assert statement here before doing the pop? Make sure that `future.running()` is False?

abhinav: Also, depending on where else you `pop` the future, it may be easier to just add a done_callback on the Future that pops it from `_outstanding` regardless. Then you never have to manually have it evicted. When it finished, it leaves.

junchaowu: Keep it for now

python/tchannel/event.py
blampe: This still isn't very clear to me; the docs need to make it obvious.

junchaowu: I am open to find a better name. Any suggestion?

blampe: I think adding some explicit in the docstrings would be best.

PR #785, opened 2015-06-25
take coroutines out of InMemStream

blampe:
This method is really hot and shouldn't be using coroutines.

@abhinav @junchaowu @breerly


junchaowu:
lgtm

PR #786, opened 2015-06-25
remove enum34 package

blampe:
This package is mostly a performance hit and doesn't really buy us anything that we can't achieve by passing integers around.

@abhinav @junchaowu @breerly


junchaowu:
do you miss the enum.py? why the existing one is slow?

abhinav:
Cool :shipit:

PR #790, opened 2015-06-25
do not trace hyperbahn advertise

Raynos:
Tracing the hyperbahn advertise can cause cascading failures
if there is something wrong with tcollector.

We need hyperbahn advertise to always work no matter what so
it is simply less risky to turn tracing off

r: @rf @shannili

cc: @junchaowu @prashantv We should turn tracing off for hyperbahn
advertise in python & go as well.

### DIFF
--- a/node/hyperbahn/handler.js
+++ b/node/hyperbahn/handler.js
@@ -157,6 +157,7 @@ function sendAdvertise(services, options, callback) {
     }
 
     options.serviceName = 'hyperbahn';
+    options.trace = false;
     options.hasNoParent = true;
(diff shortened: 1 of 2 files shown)


ShanniLi:
:+1:

anson627:
what about tchannel? the top tchannel has to set trace to true, which is annoying.

junchaowu:
it's already off in python. (done)

Raynos:
@anson627 the top channel needs tracing; tracing is on by default except for opt-out exceptions.

PR #791, opened 2015-06-25
refactor the retry logic.

junchaowu:
@blampe  @abhinav @breerly 
Take big method into parts

### DIFF
--- a/python/tchannel/tornado/peer.py
+++ b/python/tchannel/tornado/peer.py
@@ -577,8 +577,10 @@ class PeerClientOperation(object):
                 # stop the outgoing request
                 request.set_exception(e)
 
-                should_retry = request.should_retry_on_error(e) and (
-                    num_of_attempt != attempt_times - 1)
+                should_retry = (
+                    request.should_retry_on_error(e) and
+                    num_of_attempt != attempt_times - 1
+                )
 
                 if not should_retry:
                     raise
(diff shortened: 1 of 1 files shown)


junchaowu:
The github CR sucks

python/tchannel/tornado/peer.py
```diff
@@ -551,22 +548,40 @@ def send(self, arg1, arg2, arg3,
...
+        log.debug("Got response %s", response)
+        raise gen.Return(response)
+
+    @gen.coroutine
+    def _send(self, connection, req):
+        response_future = connection.send_request(req)
+        with timeout(response_future, req.ttl):
+            response = yield response_future
+
+        raise gen.Return(response)
```
abhinav: No need for this whole thing to be a coroutine. Just do:

    return gen.with_timeout(connection.send_request(req), req.ttl)

python/tchannel/tornado/peer.py
```diff
@@ -551,22 +548,40 @@ def send(self, arg1, arg2, arg3,
...
-                response = yield self._send(
-                    connection, request)
+                response = yield self._send(connection, request)
                 break
             except (ProtocolError, TimeoutError) as e:
+                # stop the outgoing request
                 request.set_exception(e)
-                if not request.should_retry_on_error(e):
+                if not request.should_retry_on_error(e) or (
+                        num_of_attempt == attempt_times - 1):
```
abhinav: Style nit: If the condition is becoming ungainly, move it into an aptly named variable.

    should_retry = (
        request.should_retry_on_error(e) and
        num_of_attempt <= ...
    )

    if not should_retry:
        ...

python/tchannel/tornado/peer.py
junchaowu: discussed offline. Keep it.

junchaowu:
Update the patch.

abhinav:
:+1:

PR #794, opened 2015-06-25
Python TChannel Synchronous Client

breerly:
Created basic `tchannel.sync.TChannelSyncClient` for making calls to TChannel services in a synchronous environment (uWSGI). Notes:

* Uses https://github.com/breerly/threadloop to allow for fanout.
* Enabled ipdb & `-s` flag for pytest so that we can debug things without having to set this up separately.
* Updated the test server to work with tests that do not have `pytest.mark.gen_test`

cc @blampe @abhinav @junchaowu


blampe:
Filename lgtm :)

abhinav:
Can you rename all file names to be more readable and fix the indentation of everything

breerly:
*merges* :P

python/tests/test_sync.py
```diff
@@ -0,0 +1,27 @@
...
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+
+from __future__ import absolute_import
+
+import pytest
+
+
+def test_works():
+    assert False
```
breerly: test is failing :-1:

breerly:
@junchaowu @blampe @abhinav - what's the reason for having all of the integration tests live `tests/integration` instead of inline with the other tests?

breerly:
Notes:

* I'm using a new ioloop for each request so that we don't have to worry about pooling and get easy thread safety
* I can't get the test server to work in conjunction with the sync client, getting `ioloop already started`. I suspect the ioloop used in the test server is conflicting.
* I'm really sad by the `request().send()` pattern and might avoid it completely here.

python/tchannel/sync.py
```diff
@@ -0,0 +1,133 @@
...
+from collections import namedtuple
+
+from tornado import gen
+from tornado import ioloop
+
+from . import glossary
+from . import tornado as async
+
+
+class TChannelSyncClient(object):
```
junchaowu: Do we need to call this TChannelSyncClient because the client is actually created or got when it calls 'request'? It may be too early to call it client IMO.

python/tchannel/sync.py
```diff
@@ -0,0 +1,133 @@
...
+        """
+        self.async_client = async.TChannel(
+            name,
+            hostport=glossary.EPHEMERAL_HOSTPORT,
+            process_name=process_name,
+            known_peers=known_peers,
+            trace=trace
+        )
+
+    def request(self, *args, **kwargs):
```
junchaowu: can we make this method inside our TChannel class and call it sync_request() so we don't need a separate wrapper?

blampe:
:+1:

junchaowu:
One more thing how does it deal with different schems? We have argschembroker for async, then what do we do with sync case?

breerly:
@junchaowu as it is, this only works with raw. I'm going to add support for thrift :)

python/tchannel/sync.py
breerly: @junchaowu - I could lazily instantiate the tchannel client... what does that buy us? I don't feel strongly either way.

python/tchannel/sync.py
```diff
@@ -0,0 +1,140 @@
...
+
+        @gen.coroutine
+        def make_request():
+
+            response = yield self.operation.send(arg1, arg2, arg3)
+
+            header = yield response.get_header()
+            body = yield response.get_body()
+
+            result = Response(header, body)
```
breerly: I'm not totally sure if we should be converting the TChannel to the stripped down Response here... perhaps it should look more like the actual Response?

python/tchannel/sync.py
```diff
@@ -0,0 +1,140 @@
...
+from tornado import gen
+
+from . import glossary
+from . import tornado as async
+
+
+class TChannelSyncClient(object):
+    """Make synchronous TChannel requests.
+
+    This client does not support incoming connections or requests- this is
```
blampe: Thinking longer term, there's nothing to stop us from implementing an `advertise` method. This opens the door for some interesting things that we could do while we're sitting next to synchronous apps.

python/tchannel/sync.py
```diff
@@ -0,0 +1,140 @@
...
+            'hello', None, json.dumps({"name": "World"})
+        )
+
+    """
+
+    def __init__(self, name, process_name=None, known_peers=None, trace=False):
+        """Initialize a new TChannelClient.
+
+        :param process_name:
+            Name of the calling process. Used for logging purposes only.
```
blampe: I wouldn't downplay the importance of the name, it's used for logging, tracing, stats, etc.

blampe:
:+1:

python/tchannel/sync.py
blampe: This is actually a good argument to put this in a separate library. If we want to make a breaking change to the sync client, does that mean we bump the major on tchannel? Versus a separate package that depends on tchannel.

python/tchannel/sync.py
breerly: :+1:

python/tchannel/sync.py
breerly: Agreed - will change

python/tchannel/sync.py
breerly: Agreed. Once I have a backwards breaking change, I want to move this to https://github.com/hyperbahn

PR #795, opened 2015-06-26
HyperbahnClient apply backoff to advertise

Raynos:
Under failure conditions the edge clients will advertise with
hyperbahn every 200ms instead of every 60 seconds. This
creates extra load on hyperbahn which causes a cascading
failure of the hyperbahn network not being able to recover
in a timely fashion.

To mitigate this the hyperbahn clients should backoff. I've
maxed the backoff to 10seconds.

Secondarily; this PR also fuzzes the 60s retry to be a retry
every (30-90s) to try and even out the load on the hyperbahn
router.

r: @rf @shannili

cc: @prashantv @junchaowu It would be good to get the go
and python hyperbahn clients to also backoff on advertise.

### DIFF
--- a/node/hyperbahn/index.js
+++ b/node/hyperbahn/index.js
@@ -30,12 +30,6 @@ var EventEmitter = require('events').EventEmitter;
 var Reporter = require('../tcollector/reporter.js');
 var TChannelJSON = require('../as/json.js');
 
-var HyperbahnClientInvalidOptionError = TypedError({
-    type: 'hyperbahn-client.invalid-option',
-    message: 'HyperbahnClient {method}: invalid option {name}' +
-        ', expected {expected}, got {actual}'
-});
-
 var AdvertisementTimeoutError = WrappedError({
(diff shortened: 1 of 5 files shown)


ShanniLi:
lgtm

node/hyperbahn/index.js
```diff
@@ -54,7 +48,18 @@ var States = {
     ADVERTISED: 'ADVERTISED'
 };
 var DEFAULT_TTL = 60 * 1000;
-var ADVERTISEMENT_ERROR_DELAY = 200;
+var DEFAULT_ERROR_RETRY_TIMES = [
```
anson627: nice algorithm!

junchaowu:
@Raynos  So this retry solution is only for advertise?

Raynos:
@junchaowu correct; only the advertise done by hyperbahn client has a unique retry strategy external to the tchannel retry strategy.

Raynos:
Python failed the build.

rf:
noice I love it :+1: :+1:

PR #801, opened 2015-06-26
rework write to not be a coroutine

blampe:
@abhinav @breerly @junchaowu

### DIFF
--- a/python/tchannel/tornado/connection.py
+++ b/python/tchannel/tornado/connection.py
@@ -44,6 +44,7 @@ from ..messages.error import ErrorMessage
 from ..messages.types import Types
 from .message_factory import MessageFactory
 from .message_factory import build_protocol_exception
+from .util import chain
 
 try:
     import tornado.queues as queues  # included in 4.2
@@ -294,14 +295,7 @@ class TornadoConnection(object):
 
         fragments = message_factory.fragment(message)
 
-        write_future = self._write(fragments.next())
-
-        for fragment in fragments:
-            write_future.add_done_callback(
-                lambda f: f.exception() or self.write(fragment)
-            )
-
-        return write_future
+        return chain(fragments, self._write)
 
     def _write(self, message):
         """Writes the given message up the wire.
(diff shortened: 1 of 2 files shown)


python/tchannel/tornado/connection.py
```diff
@@ -294,8 +293,15 @@ def write(self, message):
             message_factory = self.response_message_factory
 
         fragments = message_factory.fragment(message)
+
+        write_future = self._write(fragments.next())
+
         for fragment in fragments:
-            yield self._write(fragment)
+            write_future.add_done_callback(
+                lambda f: f.exception() or self.write(fragment)
```
blampe: I think this is wrong, I think I need to `set_exception` on `write_future` to propagate the exception correctly.

junchaowu: lambda f: f.exception() or self.write(fragment)? 

self.write or self._write?

blampe: Yeah should probably be `_write`, is this not tested at all? :(

junchaowu: Well accidentally write() is right but right in a wrong way.

junchaowu: first add_done_callback is not guaranteed the actual callback order.

abhinav: Yeah, this is wrong.

You need to create a result future that you return and if `f.exception()`, you need to `set_exception` (or better, `set_exc_info`) on it. Otherwise, you do `set_result(None)` on it after all chunks are done writing. Currently there's no write-order guarantee either.

Oh and it does weird things with the function scope.

```
>>> def go():
...     functions = []
...     for i in range(5):
...         functions.append(lambda: print(i))
...     return functions
...
>>> for f in go():
...     f()
...
4
4
4
4
4
```

blampe:
ping

abhinav:
whoa 10 days. Don't know how this got missed.

:+1: :shipit:

PR #804, opened 2015-06-26
support for isBusy function

rf:
r @Raynos @jcorbin 

`isBusy` functions take no arguments and return an object with properties `busy` (boolean) and `reason` (string)

### DIFF
--- a/node/service-name-handler.js
+++ b/node/service-name-handler.js
@@ -47,7 +47,7 @@ TChannelServiceNameHandler.prototype.handleRequest = function handleRequest(req,
 
     if (self.isBusy) {
         var busyInfo = self.isBusy();
-        if (busyInfo.busy === true) {
+        if (busyInfo && busyInfo.busy === true) {
             buildRes().sendError('Busy', busyInfo.reason || 'server is busy');
         }
     }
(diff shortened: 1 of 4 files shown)


node/service-name-handler.js
```diff
@@ -21,19 +21,37 @@
...
+    }
 }
 
 TChannelServiceNameHandler.prototype.type = 'tchannel.service-name-handler';
 
 TChannelServiceNameHandler.prototype.handleRequest = function handleRequest(req, buildRes) {
     var self = this;
+
+    if (self.isBusy) {
+        var busyInfo = self.isBusy();
```
Raynos: support returning `null` for performance; creating little objects for every req is going to hurt.

node/test/busy.js
```diff
@@ -0,0 +1,107 @@
+// Copyright (c) 2015 Uber Technologies, Inc.
```
jcorbin: actually include this file from node/test/index.js

node/service-name-handler.js
jcorbin: :+1: , in fact why isn't it just a reason string? YAGNI to the whole info object imo

Raynos:
lgtm

PR #806, opened 2015-06-27
downgrade relay in req timed out to debug

Raynos:
This message is expected under normal conditions. This will
basically happen every time a service times out.

We already have logging and stats for error frames in
general and this creates extra noise.

r: @rf @shannili


rf:
:ok:

ShanniLi:
:+1:

PR #812, opened 2015-06-30
go: Generate Thrift client/service interfaces with context parameter.

prashantv:
- Generate a new client/server interface TChan[Service]
- Implement new interfaces (TChanClient and TChanServer) to integrate with generated code
- Update tests, examples and guide

go-thrift is used to parse the Thrift file, Apache Thrift is used for serialization/deserialization.

@mmihic sorry for the big change - a lot of it is just autogenerated code that can be ignored.


golang/thrift/server.go
```diff
@@ -21,45 +21,106 @@ package thrift
...
-func (s *Server) Register(thriftName string, processorInterface reflect.Type, processor thrift.TProcessor) {
-	handler := &Handler{processor}
-	for i := 0; i < processorInterface.NumMethod(); i++ {
-		methodName := processorInterface.Method(i).Name
-		s.Channel.Register(handler, thriftName+"::"+methodName)
+// Register registers the given TChanServer to be called on any incoming call for its' services.
+// TODO(prashant): Replace Register call with this call.
+func (s *Server) Register(svr TChanServer) {
+	service := svr.Service()
+	s.handlers[service] = svr
```
mmihic: RWMutex?  I know this will typically only be called during initialization so unlikely to be invoked concurrent, but I think it's good for all methods on a given type to have the same concurrency guarantees (either all safe or all unsafe), and the overhead of an RLock is pretty low in a no-write situation.

mmihic:
good stuff.  +1

golang/thrift/server.go
prashantv: Agreed, done.

PR #817, opened 2015-07-01
Most close ephemeral peers or leak peer objects

Raynos:
We were leaking peer objects in prod; peer objects will
report stats in a loop. Because we 10s of thousands of
these objects we were reporting the same stat 10s of
thousands of time and saw a huge spike in about 5k
UDP writes per second.

r: @rf @jcorbin @kriskowal

### DIFF
--- a/node/connection.js
+++ b/node/connection.js
@@ -105,6 +105,10 @@ TChannelConnection.prototype.setupSocket = function setupSocket() {
         }));
 
         if (self.ephemeral) {
+            var peer = self.channel.peers.get(self.socketRemoteAddr);
+            if (peer) {
+                peer.close(noop);
+            }
             self.channel.peers.delete(self.socketRemoteAddr);
         }
     }
@@ -113,6 +117,8 @@ TChannelConnection.prototype.setupSocket = function setupSocket() {
         self.onSocketError(err);
     }
 };
+
+function noop() {}
 
 TChannelConnection.prototype.setupHandler = function setupHandler() {
     var self = this;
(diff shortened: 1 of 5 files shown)


node/connection.js
```diff
@@ -105,6 +105,10 @@ TChannelConnection.prototype.setupSocket = function setupSocket() {
         }));
 
         if (self.ephemeral) {
+            var peer = self.channel.peers.get(self.socketRemoteAddr);
+            if (peer) {
+                peer.close(noop);
+            }
```
kriskowal: This is great and fits our immediate need, but you might be able to knock a non-ephemeral peer out by issuing an ephemeral request from it.

node/test/lib/hyperbahn-cluster.js
```diff
@@ -135,21 +136,24 @@ function HyperbahnApp(opts) {
 
     var self = this;
 
-    self.relayChannel = opts.relayChannel;
-    self.egressNodes = opts.egressNodes;
-    self.hostPort = self.relayChannel.hostPort;
+    self.__relayChannel = opts.relayChannel;
+    self.__egressNodes = opts.egressNodes;
```
kriskowal: What’s the motivation for this style change? Perhaps dunderbar is too aggressive; would one suffice?

kriskowal:
:ok:

rf:
lgtm

node/test/lib/hyperbahn-cluster.js
Raynos: Yes; it's weird; will fix.

PR #818, opened 2015-07-01
go: Support for JSON over tchannel

prashantv:
r: @mmihic 
cc: @slantin ideally you'd use this in ringpop

### DIFF
--- a/golang/json/context.go
+++ b/golang/json/context.go
@@ -1,5 +1,25 @@
 package json
 
+// Copyright (c) 2015 Uber Technologies, Inc.
+
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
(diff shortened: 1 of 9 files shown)


golang/json/context.go
```diff
@@ -0,0 +1,73 @@
...
+	context.Context
+
+	// Headers returns the call request headers.
+	Headers() interface{}
+
+	// ResponseHeaders returns the call response headers.
+	ResponseHeaders() interface{}
+
+	// SetResponseHeaders sets the given response headers on the context.
+	SetResponseHeaders(headers interface{})
```
mmihic: I get why we need to do this, but it smells wrong to have a service-side only method on the shared context.  Probably not worth splitting into separate Context and ServiceContext interfaces though.

mmihic:
lgtm

golang/json/context.go
prashantv: Yeah it's a bit of a pain to break out as well, since if the requester passes a Context which doesn't contain a SetResponseHeaders, then the Call will have to cast it to a ServerContext so it can actually set the response headers it receives. I was hoping to use standard contexts and use WithValue for headers, but unfortunately response headers can't be set that way, since we can't add values to an existing context passed by the user.

PR #820, opened 2015-07-01
initial python guide

blampe:
https://github.com/uber/tchannel/blob/bryce_guide/python/GUIDE.rst

A little rough but I want to be able to give something to people so they can get started working with things.

@abhinav @breerly @junchaowu

### DIFF
--- a/python/GUIDE.rst
+++ b/python/GUIDE.rst
@@ -53,13 +53,13 @@ Create a `Thrift <https://thrift.apache.org/>`_ file under
     }
 
     service KeyValue {
-        string Get(
+        string get(
             1: string key,
         ) throws (
             1: NotFoundError notFound,
         )
 
-        string Set(
+        string set(
             1: string key,
             2: string value,
         )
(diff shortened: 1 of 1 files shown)


python/GUIDE.rst
```diff
@@ -0,0 +1,314 @@
...
+
+
+.. code-block:: thrift
+
+    exception NotFoundError {
+        1: string key,
+    }
+
+    service KeyValue {
+        string Get(
```
abhinav: lower case get and set

python/GUIDE.rst
```diff
@@ -0,0 +1,314 @@
...
+            1: NotFoundError notFound,
+        )
+
+        string Set(
+            1: string key,
+            2: string value,
+        )
+    }
+
+\
```
abhinav: what's with the `\`?

python/GUIDE.rst
```diff
@@ -0,0 +1,314 @@
...
+locally.
+
+
+---------
+Debugging
+---------
+
+Let's spin up the service and make a request to it through Hyperbahn. Python
+provides ``tcurl.py`` script, but we need to use the `Node
+version <https://github.com/uber/tcurl>`_) for now since it has Thrift support.
```
prashantv: stray )

python/GUIDE.rst
```diff
@@ -0,0 +1,314 @@
...
+
+    from service import KeyValue
+
+
+    @gen.coroutine
+    def run():
+        app_name = 'keyvalue-client'
+
+        app = TChannel(app_name)
+        app.advertise(['localhost:23000'], app_name)
```
junchaowu: For the client, should we add example to call server without hyperbahn by using port

python/GUIDE.rst
```diff
@@ -0,0 +1,314 @@
...
+
+        client_class = client_for('keyvalue-server', KeyValue)
```
abhinav: Make this a global?

python/GUIDE.rst
blampe: RST weirdness, without it the line is bold and italicized.

python/GUIDE.rst
blampe: Yeah I was thinking it might make sense to take the hyperbahn stuff out since it's not actually relevant to the outside world yet :)

PR #821, opened 2015-07-01
add backoff logic into advertise method

junchaowu:
@blampe @breerly @abhinav @Raynos

### DIFF
--- a/python/CHANGES.rst
+++ b/python/CHANGES.rst
@@ -1,14 +1,19 @@
 Changelog for tchannel.py
 =========================
 
+0.9.2 (unreleased)
+------------------
+
+- Add exponential backoff to `TChannel.advertise`
+
+  
 0.9.1 (2015-07-09)
 ------------------
 
 - Use threadloop 0.3.* to fix main thread not exiting when `tchannel.sync.TChannelSyncClient` is used.
-- Add exponential backoff to `TChannel.advertise`
 
 
-0.9.0
+0.9.0 (2015-07-07)
 ------------------
 
 - Allow custom handlers for unrecognized endpoints.
--- a/python/CHANGES.rst
+++ b/python/CHANGES.rst
@@ -12,6 +12,7 @@ Changelog for tchannel.py
 
 - Allow custom handlers for unrecognized endpoints.
 - Released `tchannel.sync.TChannelSyncClient` and `tchannel.sync.thrift.client_for`.
+- Support exponential backoff in tchannel advertise method.
 
 
 0.8.5 (2015-06-30)
--- a/python/CHANGES.rst
+++ b/python/CHANGES.rst
@@ -5,7 +5,7 @@ Changelog for tchannel.py
 ------------------
 
 - Use threadloop 0.3.* to fix main thread not exiting when `tchannel.sync.TChannelSyncClient` is used.
-- Support exponential backoff in tchannel advertise method.
+- Add exponential backoff to `TChannel.advertise`
 
 
 0.9.0
--- a/python/CHANGES.rst
+++ b/python/CHANGES.rst
@@ -5,14 +5,14 @@ Changelog for tchannel.py
 ------------------
 
 - Use threadloop 0.3.* to fix main thread not exiting when `tchannel.sync.TChannelSyncClient` is used.
+- Support exponential backoff in tchannel advertise method.
 
 
-0.9.0 (2015-07-07)
+0.9.0
 ------------------
 
 - Allow custom handlers for unrecognized endpoints.
 - Released `tchannel.sync.TChannelSyncClient` and `tchannel.sync.thrift.client_for`.
-- Support exponential backoff in tchannel advertise method.
 
 
 0.8.5 (2015-06-30)
--- a/python/tchannel/tornado/hyperbahn.py
+++ b/python/tchannel/tornado/hyperbahn.py
@@ -22,6 +22,7 @@ from __future__ import absolute_import
 
 import json
 import logging
+import random
 
 from ..errors import ProtocolError
 from ..errors import TimeoutError
@@ -31,20 +32,16 @@ import tornado.ioloop
 
 from .response import StatusCode
 
-DEFAULT_ERROR_RETRY_TIMES = [  # seconds
-    0.2,  # One fast retry
-
-    1,  # Fibonacci backoff
-    1,
-    2,
-    3,
-    5,
-    8,
-
-    10,  # Max timeout at 10 sec
-]
+DEFAULT_EXPO_BASE = 1.4  # try this first
+DEFAULT_MAX_DELAY = 10  # sec
+DEFAULT_MAX_ACCTEMPT = 7  # pow(1.4, 8) > 10
 
 log = logging.getLogger('tchannel')
+
+
+def fuzz(x):
+    tmp = pow(DEFAULT_EXPO_BASE, x) / 2.0
+    return tmp + random.random() * tmp
 
 
 def advertise(tchannel, service, routers):
@@ -99,9 +96,8 @@ def advertise(tchannel, service, routers):
             else:
                 attempt_counter = 0
 
-        delay_time = DEFAULT_ERROR_RETRY_TIMES[
-            min(attempt_counter, len(DEFAULT_ERROR_RETRY_TIMES) - 1)
-        ]
+        delay_time = min(DEFAULT_MAX_DELAY, fuzz(attempt_counter))
+        attempt_counter = min(attempt_counter, DEFAULT_MAX_ACCTEMPT)
         tornado.ioloop.IOLoop.current().call_later(
             delay=delay_time,
             callback=lambda: _register(attempt_counter),
--- a/python/tchannel/tornado/hyperbahn.py
+++ b/python/tchannel/tornado/hyperbahn.py
@@ -26,10 +26,20 @@ import logging
 import tornado.gen
 import tornado.ioloop
 
-from ..messages.error import ErrorCode
+from .response import StatusCode
 
-DEFAULT_TTL = 60  # seconds
+DEFAULT_ERROR_RETRY_TIMES = [  # seconds
+    0.2,  # One fast retry
 
+    1,  # Fibonacci backoff
+    1,
+    2,
+    3,
+    5,
+    8,
+
+    10,  # Max timeout at 10 sec
+]
 
 log = logging.getLogger('tchannel')
 
@@ -59,31 +69,40 @@ def advertise(tchannel, service, routers):
         tchannel.peers.get(router)
 
     @tornado.gen.coroutine
-    def _register():
-        response = yield tchannel.request(service='hyperbahn').send(
-            arg1='ad',  # advertise
-            arg2='',
-            arg3=json.dumps({
-                'services': [
-                    {
-                        'serviceName': service,
-                        'cost': 0,
-                    }
-                ]
-            }),
-            headers={'as': 'json'},
-        )
+    def _register(attempt_counter=0):
 
-        if response.code not in ErrorCode:
-            # re-register every ``DEFAULT_TTL`` seconds
-            tornado.ioloop.IOLoop.current().call_later(
-                delay=DEFAULT_TTL,
-                callback=_register,
+        try:
+            response = yield tchannel.request(service='hyperbahn').send(
+                arg1='ad',  # advertise
+                arg2='',
+                arg3=json.dumps({
+                    'services': [
+                        {
+                            'serviceName': service,
+                            'cost': 0,
+                        }
+                    ]
+                }),
+                headers={'as': 'json'},
+                attempt_times=1,
             )
-            raise tornado.gen.Return(response)
+        except Exception as e:
+            attempt_counter += 1
+            log.error('Failed to register with Hyperbahn: %s', e)
         else:
-            log.error('Failed to register with Hyperbahn: %s', response)
-            raise NotImplementedError  # TODO figure out behavior here
+            if response.code != StatusCode.ok:
+                attempt_counter += 1
+                log.error('Failed to register with Hyperbahn: %s', response)
+            else:
+                attempt_counter = 0
+        finally:
+            delay_time = DEFAULT_ERROR_RETRY_TIMES[
+                min(attempt_counter, len(DEFAULT_ERROR_RETRY_TIMES) - 1)
+            ]
+            tornado.ioloop.IOLoop.current().call_later(
+                delay=delay_time,
+                callback=lambda: _register(attempt_counter),
+            )
 
     return _register()
 
--- a/python/tchannel/tornado/hyperbahn.py
+++ b/python/tchannel/tornado/hyperbahn.py
@@ -23,6 +23,9 @@ from __future__ import absolute_import
 import json
 import logging
 
+from ..errors import ProtocolError
+from ..errors import TimeoutError
+
 import tornado.gen
 import tornado.ioloop
 
@@ -86,7 +89,7 @@ def advertise(tchannel, service, routers):
                 headers={'as': 'json'},
                 attempt_times=1,
             )
-        except Exception as e:
+        except (ProtocolError, TimeoutError) as e:
             attempt_counter += 1
             log.error('Failed to register with Hyperbahn: %s', e)
         else:
@@ -95,14 +98,14 @@ def advertise(tchannel, service, routers):
                 log.error('Failed to register with Hyperbahn: %s', response)
             else:
                 attempt_counter = 0
-        finally:
-            delay_time = DEFAULT_ERROR_RETRY_TIMES[
-                min(attempt_counter, len(DEFAULT_ERROR_RETRY_TIMES) - 1)
-            ]
-            tornado.ioloop.IOLoop.current().call_later(
-                delay=delay_time,
-                callback=lambda: _register(attempt_counter),
-            )
+
+        delay_time = DEFAULT_ERROR_RETRY_TIMES[
+            min(attempt_counter, len(DEFAULT_ERROR_RETRY_TIMES) - 1)
+        ]
+        tornado.ioloop.IOLoop.current().call_later(
+            delay=delay_time,
+            callback=lambda: _register(attempt_counter),
+        )
 
     return _register()
 
--- a/python/tchannel/tornado/hyperbahn.py
+++ b/python/tchannel/tornado/hyperbahn.py
@@ -23,12 +23,12 @@ from __future__ import absolute_import
 import json
 import logging
 import random
+import time
 
 import tornado.gen
 import tornado.ioloop
 
-from tchannel.tornado.timeout import timeout
-
+from ..errors import AdvertiseError
 from ..errors import ProtocolError
 from ..errors import TimeoutError
 from .response import StatusCode
@@ -36,27 +36,10 @@ from .response import StatusCode
 DEFAULT_EXPO_BASE = 1.4  # try this first
 DEFAULT_MAX_DELAY = 10  # sec
 DEFAULT_MAX_ATTEMPT = 7  # pow(1.4, 8) > 10
-DEFAULT_DELAY = 60  # sec delay time for successful advertise
+DEFAULT_DELAY = 60 * 1000  # ms delay time for successful advertise
 DEFAULT_FIRST_ADVERTISE_TIME = 30  # sec
 
 log = logging.getLogger('tchannel')
-
-
-def _advertise(tchannel, service):
-    return tchannel.request(service='hyperbahn').send(
-        arg1='ad',  # advertise
-        arg2='',
-        arg3=json.dumps({
-            'services': [
-                {
-                    'serviceName': service,
-                    'cost': 0,
-                }
-            ]
-        }),
-        headers={'as': 'json'},
-        attempt_times=1,
-    )
 
 
 def _prepare_next_ad(attempt_counter):
@@ -71,48 +54,57 @@ def _prepare_next_ad(attempt_counter):
 
 
 @tornado.gen.coroutine
-def _regular_advertise(tchannel, service, attempt_counter=0):
+def _advertise(tchannel, service):
+    response = None
     try:
-        response = yield _advertise(tchannel, service)
+        response = yield tchannel.request(service='hyperbahn').send(
+            arg1='ad',  # advertise
+            arg2='',
+            arg3=json.dumps({
+                'services': [
+                    {
+                        'serviceName': service,
+                        'cost': 0,
+                    }
+                ]
+            }),
+            headers={'as': 'json'},
+            attempt_times=1,
+        )
     except (ProtocolError, TimeoutError) as e:
-        attempt_counter += 1
         log.error('Failed to register with Hyperbahn: %s', e)
     else:
         if response.code != StatusCode.ok:
-            attempt_counter += 1
             log.error('Failed to register with Hyperbahn: %s', response)
         else:
-            attempt_counter = 0
             log.info('Successfully register with Hyperbahn')
 
-    (attempt_counter, delay_time) = _prepare_next_ad(attempt_counter)
-    tornado.ioloop.IOLoop.current().call_later(
-        delay=delay_time,
-        callback=lambda: _regular_advertise(tchannel, service, attempt_counter)
-    )
+    raise tornado.gen.Return(response)
 
 
 @tornado.gen.coroutine
-def _first_advertise(tchannel, service):
+def _exponential_backoff_advertise(tchannel, service, timeout=None):
+    # first advertise rule apply here.
     attempt_counter = 0
+    start = time.time()
     while True:
-        try:
-            response = yield _advertise(tchannel, service)
-        except (ProtocolError, TimeoutError) as e:
-            attempt_counter += 1
-            log.error('Failed to register with Hyperbahn: %s', e)
-        else:
-            if response.code != StatusCode.ok:
-                attempt_counter += 1
-                log.error('Failed to register with Hyperbahn: %s', response)
-            else:
-                log.info('Successfully registered with Hyperbahn')
-                raise tornado.gen.Return(response)
+        if timeout and time.time() - start > 30000:
+            raise AdvertiseError("Failed to register with Hyperbahn.")
 
+        response = yield _advertise(
+            tchannel,
+            service,
+        )
+        if response is not None and response.code is StatusCode.ok:
+            break
+        attempt_counter += 1
         (attempt_counter, delay_time) = _prepare_next_ad(attempt_counter)
         yield tornado.gen.sleep(delay_time)
 
+    raise tornado.gen.Return(response)
 
+
+@tornado.gen.coroutine
 def advertise(tchannel, service, routers):
     """Advertise the given TChannel to Hyperbahn using the given name.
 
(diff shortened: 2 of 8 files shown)


python/tchannel/tornado/hyperbahn.py
```diff
@@ -26,10 +26,20 @@
...
+    0.2,  # One fast retry
 
+    1,  # Fibonacci backoff
+    1,
+    2,
+    3,
+    5,
+    8,
+
+    10,  # Max timeout at 10 sec
```
blampe: zomg come on dood 13

python/tchannel/tornado/hyperbahn.py
```diff
@@ -59,31 +69,40 @@ def advertise(tchannel, service, routers):
...
+                            'serviceName': service,
+                            'cost': 0,
+                        }
+                    ]
+                }),
+                headers={'as': 'json'},
+                attempt_times=1,
             )
-            raise tornado.gen.Return(response)
+        except Exception as e:
```
blampe: We don't want to retry on *every* possible exception...

junchaowu: The exception caught in this level is for backoff approach. Advertise method will always exist but run in certain period.

python/tests/tornado/test_hyperbahn.py
```diff
@@ -60,5 +61,6 @@ def test_request():
 def test_register():
     channel = TChannel(name='test')
 
-    with pytest.raises(ConnectionClosedError):
+    with patch('tchannel.tornado.hyperbahn.log') as mock_logger:
         yield hyperbahn.advertise(channel, 'foo', ['127.0.0.1:23000'])
+    assert mock_logger.error.called
```
abhinav: Add a proper test for the retry logic. Don't test logging. Test that the send() function was called twice the register payload.

abhinav:
Also, don't forget to update the changelog.

python/tchannel/tornado/hyperbahn.py
junchaowu: That's what @Raynos decided. Maximum is 10s.

python/tchannel/tornado/hyperbahn.py
blampe: Huh? That doesn't change the fact that we should be only retrying on exceptions we expect. For example if we can't find a list of seed hosts on disk and this raises an `IOError`, why waste over 30 seconds attempting to retry?

And what do we do when we fail to register? Do we want the application to continue running? That's what happens currently.

junchaowu: Make retry only for ProtocolError and TimeoutError for now.

junchaowu:
Updated the Changelog

python/CHANGES.rst
```diff
@@ -4,7 +4,7 @@ Changelog for tchannel.py
 0.8.6 (unreleased)
 ------------------
 
-- No changes yet.
+- Add backoff login into advertise method.
```
blampe: Add exponential backoff on errors during ``advertise``.

python/tchannel/tornado/hyperbahn.py
```diff
@@ -22,18 +22,28 @@
...
 import tornado.ioloop
 
-from ..messages.error import ErrorCode
-
-DEFAULT_TTL = 60  # seconds
+from .response import StatusCode
 
+DEFAULT_EXPO_BASE = 1.4  # try this first
+DEFAULT_MAX_DELAY = 10  # sec
+DEFAULT_MAX_ACCTEMPT = 7  # pow(1.4, 8) > 10
```
blampe: ATTEMPT

python/tchannel/tornado/hyperbahn.py
```diff
@@ -59,31 +69,39 @@ def advertise(tchannel, service, routers):
...
-            raise NotImplementedError  # TODO figure out behavior here
+            if response.code != StatusCode.ok:
+                attempt_counter += 1
+                log.error('Failed to register with Hyperbahn: %s', response)
+            else:
+                attempt_counter = 0
+
+        delay_time = min(DEFAULT_MAX_DELAY, fuzz(attempt_counter))
+        attempt_counter = min(attempt_counter, DEFAULT_MAX_ACCTEMPT)
+        tornado.ioloop.IOLoop.current().call_later(
```
blampe: Registration was successful, but then we're calling `_register` again...

python/tchannel/tornado/hyperbahn.py
```diff
@@ -59,31 +69,39 @@ def advertise(tchannel, service, routers):
...
-            }),
-            headers={'as': 'json'},
-        )
-
-        if response.code not in ErrorCode:
-            # re-register every ``DEFAULT_TTL`` seconds
-            tornado.ioloop.IOLoop.current().call_later(
-                delay=DEFAULT_TTL,
-                callback=_register,
+    def _register(attempt_counter=0):
```
blampe: It's impossible to test this because it's a closure, make it a real method.

python/tchannel/tornado/hyperbahn.py
```diff
@@ -22,18 +22,28 @@
...
+from .response import StatusCode
 
+DEFAULT_EXPO_BASE = 1.4  # try this first
+DEFAULT_MAX_DELAY = 10  # sec
+DEFAULT_MAX_ACCTEMPT = 7  # pow(1.4, 8) > 10
 
 log = logging.getLogger('tchannel')
 
 
+def fuzz(x):
```
abhinav: what's x?

python/tchannel/tornado/hyperbahn.py
```diff
@@ -59,31 +69,39 @@ def advertise(tchannel, service, routers):
...
         else:
-            log.error('Failed to register with Hyperbahn: %s', response)
-            raise NotImplementedError  # TODO figure out behavior here
+            if response.code != StatusCode.ok:
+                attempt_counter += 1
+                log.error('Failed to register with Hyperbahn: %s', response)
+            else:
+                attempt_counter = 0
+
+        delay_time = min(DEFAULT_MAX_DELAY, fuzz(attempt_counter))
```
blampe: Keep it simple:
`delay_time = random.uniform(0, min(10, 1.4 ** attempt_counter))`

python/tchannel/tornado/hyperbahn.py
junchaowu: the registration thing is a periodical thing. Tchannel needs periodically register itself in order to let autobahn know it is alive.

python/tchannel/tornado/hyperbahn.py
junchaowu: add docstring

python/tchannel/tornado/hyperbahn.py
junchaowu: In the http://www.awsarchitectureblog.com/2015/03/backoff.html.  EqualJitter works better than the fullJitter.

blampe: I was going by this:
> Of the jittered approaches, “Equal Jitter” is the loser.

junchaowu: You are right. Change it to FullJitter

python/CHANGES.rst
```diff
@@ -5,6 +5,7 @@ Changelog for tchannel.py
 ------------------
 
 - Allow custom handlers for unrecognized endpoints.
+- Add exponential backoff login into advertise method.
```
blampe: Same comment, the wording here still needs work.

python/tchannel/tornado/hyperbahn.py
blampe: Still applies.

junchaowu: add the register to the call_later to remove the closure

junchaowu:
ping

python/tchannel/tornado/hyperbahn.py
```diff
@@ -59,33 +64,46 @@ def advertise(tchannel, service, routers):
...
+        )
 
-    return _register()
+        attempt_counter = min(attempt_counter, DEFAULT_MAX_ATTEMPT)
+        tornado.ioloop.IOLoop.current().call_later(
+            delay=delay_time,
+            callback=lambda: _register(attempt_counter),
+        )
 
+    tornado.ioloop.IOLoop.current().call_later(
```
blampe: You can use `add_callback` if you don't have a delay.

python/tests/tornado/test_hyperbahn.py
```diff
@@ -54,11 +54,3 @@ def test_request():
             arg3='boo',
             headers={'as': 'qux'},
         )
-
-
-@pytest.mark.gen_test
-def test_register():
```
blampe: Why delete this?

junchaowu: We don't `yield hyperbahn.advertise` any longer. When user calls advertise, it just returns.

blampe: Ok, so where are we testing `advertise` if not here?

blampe: s/advertise/register

python/tchannel/tornado/hyperbahn.py
junchaowu: Changed to use add_callback

python/tests/tornado/test_hyperbahn.py
junchaowu: Advertise method is highly tight to autobahn. I tested on dev server with autobahn.
It is not very ideal to unit test here. When we have a mock autobahn, we may test it.

abhinav: Can we create a TChannel server in the test that receives the advertise call? I would rather not find about failure in that logic in production.

python/CHANGES.rst
```diff
@@ -1,7 +1,13 @@
 Changelog for tchannel.py
 =========================
 
-0.8.6 (unreleased)
+0.9.1 (unreleased)
+------------------
+
+- Support exponential backoff in tchannel advertise method.
```
abhinav: - Add exponential backoff to ``TChannel.advertise``

python/tests/tornado/test_hyperbahn.py
```diff
@@ -57,8 +58,16 @@ def test_request():
...
+    @tornado.gen.coroutine
+    def execution(request, response):
+        body = yield response.get_body()
+        assert body['services']['serviceName'] == 'test'
+        assert body['services']['cost'] == 0
+    tchannel_server.expect_call(endpoint, 'json').execution = execution
+    channel = TChannel(name='test')
+    hyperbahn.advertise(channel,
+                        'test', ['localhost:'+str(tchannel_server.port)]
+                        )
```
abhinav: The test is wrong.

- `advertise` returns a future. You're not waiting for the future to return. This function will exit immediately. You won't have tested anything.
- `Expectation.execution` isn't a thing. It's `Expectation.execute`. However you're not supposed to assign it like that anyway. That whole class is an artifact from before we had a convenient way to set up ad-hoc servers.
- `response.get_body`? Should be `request`.

Here's what I think you should be doing instead of trying to shoehorn this behavior into the `TestServer`.

    server = TChannel('test-server')

    @server.register('ad', 'json')
    def ad(req, res):
        # ...

    server.listen()
    client = TChannel('test-client')
    client.advertise(..., [server.hostport])

Also, asserting values inside the request handler is useless if you don't assert that the handler was actually called. If the handler is never called, your test will pass regardless. That's why I told you to use a mock object for that. You could make sure it was actually called:

    call = mock.Mock()

    @tornado.gen.coroutine
    def execute(req, res):
        body = # ...
        call(body)

     call.assert_called_once_with({...})

python/CHANGES.rst
abhinav: ping

python/tests/tornado/test_hyperbahn.py
```diff
@@ -57,8 +58,21 @@ def test_request():
...
+
+    tchannel_server.expect_call(endpoint, 'json').execute = execution
     channel = TChannel(name='test')
+    hyperbahn.advertise(channel,
+                        'test', ['localhost:'+str(tchannel_server.port)]
+                        )
 
-    with pytest.raises(ConnectionClosedError):
-        yield hyperbahn.advertise(channel, 'foo', ['127.0.0.1:23000'])
+    result = yield f
```
abhinav: The future is a nice idea but if the call never goes through, it'll freeze forever/until the pytest-tornado async timeout kicks in.

Also, you want to yield advertise.

python/tests/tornado/test_hyperbahn.py
```diff
@@ -57,8 +58,21 @@ def test_request():
...
-def test_register():
+def test_advertise(tchannel_server):
+    endpoint = b'ad'
+
+    f = Future()
+
+    @tornado.gen.coroutine
+    def execution(request, response):
+        body = yield request.get_body()
+        f.set_result(body)
```
abhinav: You'll want to send some sort of dummy response here. Just an empty dictionary will do.

python/tests/tornado/test_hyperbahn.py
```diff
@@ -57,8 +58,21 @@ def test_request():
...
+
+    tchannel_server.expect_call(endpoint, 'json').execute = execution
```
abhinav: What's the problem with using a TChannel object instead of this `tchannel_server` fixture? As said before, it's a leftover from when we didn't have a convenient interface to register endpoints on the TChannel. We should work towards eliminating its use from the codebase.

Check the old comment on this file for what I'm saying. The code will look about the same but you'll no longer be relying on this unclean assign-to-`execute` approach.

python/tchannel/tornado/hyperbahn.py
```diff
@@ -59,33 +64,43 @@ def advertise(tchannel, service, routers):
...
+        )
 
-    return _register()
+        attempt_counter = min(attempt_counter, DEFAULT_MAX_ATTEMPT)
+        tornado.ioloop.IOLoop.current().call_later(
+            delay=delay_time,
+            callback=lambda: _register(attempt_counter),
+        )
 
+    tornado.ioloop.IOLoop.current().add_callback(callback=lambda: _register(0))
```
abhinav: How is this different from calling `_register(0)` directly?

python/tchannel/tornado/hyperbahn.py
```diff
@@ -59,33 +64,43 @@ def advertise(tchannel, service, routers):
...
+        delay_time = random.uniform(
+            0, min(DEFAULT_MAX_DELAY, DEFAULT_EXPO_BASE ** attempt_counter)
+        )
 
-    return _register()
+        attempt_counter = min(attempt_counter, DEFAULT_MAX_ATTEMPT)
+        tornado.ioloop.IOLoop.current().call_later(
+            delay=delay_time,
+            callback=lambda: _register(attempt_counter),
+        )
```
abhinav: If you `raise Return(response)` here, `_register()` *will* return the response of the first call. That's how `return _register()` was able to make the result of the first call available as a future.

python/tchannel/tornado/hyperbahn.py
```diff
@@ -59,33 +64,43 @@ def advertise(tchannel, service, routers):
...
-            log.error('Failed to register with Hyperbahn: %s', response)
-            raise NotImplementedError  # TODO figure out behavior here
+            if response.code != StatusCode.ok:
+                attempt_counter += 1
+                log.error('Failed to register with Hyperbahn: %s', response)
+            else:
+                attempt_counter = 0
+
+        delay_time = random.uniform(
+            0, min(DEFAULT_MAX_DELAY, DEFAULT_EXPO_BASE ** attempt_counter)
```
abhinav: Doesn't this mean that if the call is successful, you'll be bombarding advertise requests as fast as the system will let you go?

python/tchannel/tornado/hyperbahn.py
```diff
@@ -22,18 +22,98 @@
...
+        )
+    else:
+        delay_time = DEFAULT_DELAY
+    attempt_counter = min(attempt_counter, DEFAULT_MAX_ATTEMPT)
+    return (attempt_counter, delay_time)
+
+
+@tornado.gen.coroutine
+def _regular_advertise(tchannel, service, attempt_counter=0):
```
blampe: This is all copy-pasted from `_first_advertise`...

python/tchannel/tornado/hyperbahn.py
```diff
@@ -58,34 +138,16 @@ def advertise(tchannel, service, routers):
...
+        if f.exception():
+            log.error('Failed to register with Hyperbahn forever.')
         else:
-            log.error('Failed to register with Hyperbahn: %s', response)
-            raise NotImplementedError  # TODO figure out behavior here
-
-    return _register()
+            _regular_advertise(tchannel, service)
 
+    with timeout(ad_future, DEFAULT_FIRST_ADVERTISE_TIME):
```
abhinav: Any reason we're not simply using Tornado's [`with_timeout`](http://tornado.readthedocs.org/en/latest/gen.html#tornado.gen.with_timeout)?

python/tchannel/tornado/hyperbahn.py
junchaowu: There is one different place. In first_advertise, if we get successful response we will return the response. It makes hard to extract common piece of code. There are exception, response not ok, and response ok to compare. Making a common method doesn't gain any thing

python/tchannel/tornado/hyperbahn.py
```diff
@@ -22,18 +22,98 @@
...
+        except (ProtocolError, TimeoutError) as e:
+            attempt_counter += 1
+            log.error('Failed to register with Hyperbahn: %s', e)
+        else:
+            if response.code != StatusCode.ok:
+                attempt_counter += 1
+                log.error('Failed to register with Hyperbahn: %s', response)
+            else:
+                attempt_counter = 0
+                log.info('Successfully register with Hyperbahn')
```
abhinav: > Successfully registered

python/tchannel/tornado/hyperbahn.py
```diff
@@ -22,18 +22,98 @@
...
+            response = yield _advertise(tchannel, service)
+        except (ProtocolError, TimeoutError) as e:
+            attempt_counter += 1
+            log.error('Failed to register with Hyperbahn: %s', e)
+        else:
+            if response.code != StatusCode.ok:
+                attempt_counter += 1
+                log.error('Failed to register with Hyperbahn: %s', response)
+            else:
+                attempt_counter = 0
```
abhinav: Don't need to assign attempt_counter. Returning anyway.

python/tchannel/tornado/hyperbahn.py
junchaowu: We discussed this before. It's because with_timeout throws tornado timeout exception. I use Tchannel timeout exception. If we use tornado's , we need another exception translate.

python/tchannel/tornado/hyperbahn.py
junchaowu: right

python/tests/tornado/test_hyperbahn.py
```diff
@@ -57,8 +59,26 @@ def test_request():
...
+    with mock.patch(
+        'tchannel.tornado.hyperbahn._regular_advertise'
+    ) as mock_regular_advertise:
+        response = yield hyperbahn.advertise(
+            channel,
+            'test', [server.hostport]
+        )
+
+    result = yield response.get_body()
+    assert mock_regular_advertise.called
```
abhinav: You want to assert on the mock inside the context manager.

python/tests/tornado/test_hyperbahn.py
```diff
@@ -57,8 +59,26 @@ def test_request():
...
+        response.write_body(body)
+
+    server.listen()
     channel = TChannel(name='test')
 
-    with pytest.raises(ConnectionClosedError):
-        yield hyperbahn.advertise(channel, 'foo', ['127.0.0.1:23000'])
+    with mock.patch(
+        'tchannel.tornado.hyperbahn._regular_advertise'
+    ) as mock_regular_advertise:
```
abhinav: Use autospec with mock.patch.

python/tchannel/tornado/hyperbahn.py
blampe: I don't buy it.

python/CHANGES.rst
```diff
@@ -1,6 +1,12 @@
 Changelog for tchannel.py
 =========================
 
+0.9.2 (unreleased)
```
breerly: Should be `0.10.0`.

python/tchannel/tornado/hyperbahn.py
```diff
@@ -22,18 +22,80 @@
...
+            0, min(DEFAULT_MAX_DELAY, DEFAULT_EXPO_BASE ** attempt_counter)
+        )
+    else:
+        delay_time = DEFAULT_DELAY
+    attempt_counter = min(attempt_counter, DEFAULT_MAX_ATTEMPT)
+    return (attempt_counter, delay_time)
+
+
+@tornado.gen.coroutine
+def _advertise(tchannel, service, attempt_counter=0, recursion=False):
```
blampe: `_advertise` shouldn't know anything about attempts, and we can also take this recursion bit out. The outer `advertise` method is going to handle the initial retry behavior, and then it's going to schedule a periodic callback.

python/tchannel/tornado/hyperbahn.py
```diff
@@ -58,34 +120,33 @@ def advertise(tchannel, service, routers):
...
-            headers={'as': 'json'},
-        )
-
-        if response.code not in ErrorCode:
-            # re-register every ``DEFAULT_TTL`` seconds
-            tornado.ioloop.IOLoop.current().call_later(
-                delay=DEFAULT_TTL,
-                callback=_register,
+    # first advertise rule apply here.
+    ad_future = Future()
```
blampe: We don't need this `ad_future`, our coroutine's future should be sufficient.

```
attempts = 0
while True:
    response = yield _advertise(tchannel, service)
    if response and response.code is StatusCode.ok:
        break
    attempt_counter += 1
    delay_time = random.uniform(...)
    yield gen.sleep(delay_time)

tornado.ioloop.PeriodicCallback(
    lambda: _advertise(tchannel, service), 
    DEFAULT_DELAY,
).start()
```

For the 30s timeout we can schedule a callback in 30 seconds to `sys.exit` and then cancel it when the `while` loop breaks.

python/CHANGES.rst
junchaowu: I didn't know about this. Whoever cut the release should put on it..

python/tchannel/tornado/hyperbahn.py
junchaowu: we can't do sys.exit. What if we are shadowing, because we can't advertise then we kill the process. No .

blampe: Ok so we could just raise the exception explicitly:

```
while True:
    if been_more_than_30s:
        raise SomeError()
```

The big idea here is that right now `_advertise` is trying to *manage itself*, and we need to move that management into `advertise`.

python/CHANGES.rst
blampe: Is it a breaking change though?

python/tests/tornado/test_hyperbahn.py
```diff
@@ -57,8 +58,21 @@ def test_request():
 
 
 @pytest.mark.gen_test
-def test_register():
+def test_advertise():
```
blampe: Bonus points: test the backoff behavior.

python/CHANGES.rst
junchaowu: The version number should be finalized by the person who actually cut the release.

python/tchannel/tornado/hyperbahn.py
```diff
@@ -22,18 +22,89 @@
...
+        if response.code != StatusCode.ok:
+            log.error('Failed to register with Hyperbahn: %s', response)
+        else:
+            log.info('Successfully register with Hyperbahn')
+
+    raise tornado.gen.Return(response)
+
+
+@tornado.gen.coroutine
+def _exponential_backoff_advertise(tchannel, service, timeout=None):
```
blampe: Nit: maybe `_advertise_with_backoff`?

python/tchannel/tornado/hyperbahn.py
```diff
@@ -22,18 +22,89 @@
...
+    # first advertise rule apply here.
+    attempt_counter = 0
+    start = time.time()
+    while True:
+        if timeout and time.time() - start > 30000:
+            raise AdvertiseError("Failed to register with Hyperbahn.")
+
+        response = yield _advertise(
```
blampe: Can fit this on one line.

python/tchannel/tornado/hyperbahn.py
```diff
@@ -22,18 +22,89 @@
...
+DEFAULT_EXPO_BASE = 1.4  # try this first
+DEFAULT_MAX_DELAY = 10  # sec
+DEFAULT_MAX_ATTEMPT = 7  # pow(1.4, 8) > 10
+DEFAULT_DELAY = 60 * 1000  # ms delay time for successful advertise
+DEFAULT_FIRST_ADVERTISE_TIME = 30  # sec
 
 log = logging.getLogger('tchannel')
 
 
+def _prepare_next_ad(attempt_counter):
```
blampe: I think this can go away.

python/tchannel/tornado/hyperbahn.py
```diff
@@ -22,18 +22,89 @@
...
+            raise AdvertiseError("Failed to register with Hyperbahn.")
+
+        response = yield _advertise(
+            tchannel,
+            service,
+        )
+        if response is not None and response.code is StatusCode.ok:
+            break
+        attempt_counter += 1
+        (attempt_counter, delay_time) = _prepare_next_ad(attempt_counter)
```
blampe: No need to set `attempt_counter` here again.

python/CHANGES.rst
blampe: If you made a breaking change it should be a bump. Personally I don't think this is breaking but you would know better than me.

If it's not a breaking change (just a patch), it'll go under 0.9.2 and diffs will pile up there until we cut a new release.

python/tchannel/tornado/hyperbahn.py
junchaowu: we do. If response is not okay, we need to calculate the delay_time. This also explains why we need _prepare_next_ad

blampe: I see because you have `DEFAULT_MAX_ATTEMPT`. Let's remove the `DEFAULT_DELAY` from `_prepare_next_ad` and we'll be good to go then.

python/tchannel/tornado/hyperbahn.py
junchaowu: explained above

python/CHANGES.rst
breerly: @blampe you're wrong - when you are in 0.x every new feature, even if it is not backwards breaking, is a new minor. In fact, that is the same for 1.x

breerly: @abg @junchaowu @abg just to make sure we're all on the same page:

X.Y.Z - major.minor.patch

Every piece of new functionality should result in a new minor version increment. If the major is still 0, aka we're still in beta, then that functionality can contain breaking changes. This means that applications should lock the version to the minor, and also means we won't be able to ship new functionality to them. 

Once we are 1.0, new functionality still gets shipped in the minor, aka 1.1, 1.2, 1.3, etc... but breaking changes will force us to cut a 2.0.

Patching existing functionality, aka fixing bugs, should always increment the patch version. 

This diff represents new functionality, so the correct next version is `0.10.0`. This is super important because we have customers locked on `tchannel>=0.9,<0.8` that are only expecting bug fixes, no new & possibly breaking changes. 

Makes sense?

PR #823, opened 2015-07-01
Update send method interface in broker.py

junchaowu:
@blampe @abhinav @breerly 
Update the send method interface to match the latest send method interface in the peer.py

### DIFF
--- a/python/tchannel/tornado/broker.py
+++ b/python/tchannel/tornado/broker.py
@@ -64,7 +64,7 @@ class ArgSchemeBroker(object):
         return handler(req, resp, proxy)
 
     @tornado.gen.coroutine
-    def send(self, client, endpoint, header, body, traceflag=False):
+    def send(self, client, endpoint, header, body, **kwargs):
         try:
             if not isinstance(header, Stream):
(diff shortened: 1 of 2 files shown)


abhinav:
:+1:

python/tchannel/tornado/broker.py
```diff
@@ -64,7 +64,7 @@ def handle_call(self, req, resp, proxy):
         return handler(req, resp, proxy)
 
     @tornado.gen.coroutine
-    def send(self, client, endpoint, header, body, traceflag=False):
+    def send(self, client, endpoint, header, body, **kwargs):
```
blampe: No! `**kwargs` is lazy API design.

python/tchannel/tornado/peer.py
```diff
@@ -466,8 +466,8 @@ def __init__(self, peer, service, arg_scheme=None,
 
     @gen.coroutine
     def send(self, arg1, arg2, arg3,
-             traceflag=None,
              headers=None,
+             traceflag=None,
```
blampe: This is a backwards incompatible change.

PR #824, opened 2015-07-01
go: Add header support to Thrift

prashantv:
Works and reuses the binary read/write logic, but at the cost of memory.

r @mmihic


mmihic:
lgtm

PR #826, opened 2015-07-01
remove kwargs from send method in broker.py by use explict arguments

junchaowu:
@blampe @abhinav @breerly


blampe:
If you were concerned about the method signatures falling out of sync, do you want to add a test that prevents that?

junchaowu:
Not testable.

PR #827, opened 2015-07-01
timeout per attempt and retries

ShanniLi:
r @Raynos @kriskowal

### DIFF
--- a/node/request.js
+++ b/node/request.js
@@ -46,14 +46,14 @@ function TChannelRequest(channel, options) {
     if (!self.options.retryFlags) {
         self.options.retryFlags = {
             never: false,
-            onConnectionError: true,
-            onTimeout: false
+            onConnectionError: true
         };
     }
 
     self.triedRemoteAddrs = {};
     self.outReqs = [];
     self.timeout = self.options.timeout || TChannelRequest.defaultTimeout;
+    self.timeoutPerAttempt = self.options.timeoutPerAttempt || self.timeout;
     self.limit = self.options.retryLimit || TChannelRequest.defaultRetryLimit;
     self.start = 0;
     self.end = 0;
@@ -206,6 +206,10 @@ TChannelRequest.prototype.onIdentified = function onIdentified(peer) {
         opts[key] = self.options[key];
     }
     opts.timeout = self.timeout - self.elapsed;
+    if (opts.timeout > self.timeoutPerAttempt) {
+        opts.timeout = self.timeoutPerAttempt;
+    }
+
     opts.retryCount = self.outReqs.length;
     opts.logical = true;
 
@@ -286,6 +290,7 @@ TChannelRequest.prototype.checkTimeout = function checkTimeout(err, res) {
     var now = self.timers.now();
     self.elapsed = now - self.start;
     if (self.elapsed < self.timeout) return false;
+
     if (err) {
         if (!self.err) {
             self.emitError(err);
@@ -329,7 +334,7 @@ TChannelRequest.prototype.shouldRetryError = function shouldRetryError(err) {
                 return true;
 
             case 'Timeout':
-                return !!self.options.retryFlags.onTimeout;
+                return true;
 
             case 'NetworkError':
             case 'ProtocolError':
--- a/node/request.js
+++ b/node/request.js
@@ -46,13 +46,17 @@ function TChannelRequest(channel, options) {
     if (!self.options.retryFlags) {
         self.options.retryFlags = {
             never: false,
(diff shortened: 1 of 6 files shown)


node/request.js
```diff
@@ -329,7 +335,7 @@ TChannelRequest.prototype.shouldRetryError = function shouldRetryError(err) {
                 return true;
 
             case 'Timeout':
-                return !!self.options.retryFlags.onTimeout;
+                return !!err.logical || !!self.options.retryFlags.onTimeout;
```
kriskowal: !! is superfluous before a boolean.

kriskowal:
I believe we should look at per-attempt timeouts as a stopgap, as likely to cause harm as help since they can multiply effort in progress. Speculative trials is the endgame.

node/request.js
ShanniLi: err.logical can be undefined

ShanniLi:
@kriskowal we got user cases where clients time out on services occasionally. In these cases we need some kind of retry. I discussed with @Raynos, seems that this kind of retry is desired. That is reason for this PR. :)

node/request.js
```diff
@@ -206,6 +207,10 @@ TChannelRequest.prototype.onIdentified = function onIdentified(peer) {
         opts[key] = self.options[key];
     }
     opts.timeout = self.timeout - self.elapsed;
+    if (opts.timeout > self.timeoutPerAttempt) {
```
Raynos: This feels wrong.

node/request.js
Raynos: Why check for `err.logical`; just retry on all timeouts.

node/test/timeouts.js
```diff
@@ -120,3 +120,219 @@ allocCluster.test('requests will timeout', {
...
+                timeout: 3000,
+                timeoutPerAttempt: 1000,
+                retryLimit: 2
+            })
+            .send('/timeout', 'h', 'b', done);
+    }
+
+    function done(err, res, arg2, arg3) {
+        assert.ifError(err);
+
```
Raynos: the logical `req` object contains an array outReqs

You should assert on `req.outReq`; namely that there are two out reqs and that the first failed with a timeout.

node/request.js
ShanniLi: For retry on total timeout, we don't want it to be enabled as default, right?

Raynos: the retry logic will check the timeout and early abort if it passed.

node/request.js
```diff
@@ -329,7 +334,7 @@ TChannelRequest.prototype.shouldRetryError = function shouldRetryError(err) {
                 return true;
 
             case 'Timeout':
-                return !!self.options.retryFlags.onTimeout;
+                return true;
```
Raynos: Respect `!!self.options.retryFlags.onTimeout`; dont set this to true.

Sorry; didnt communicate properly.

node/request.js
```diff
@@ -46,14 +46,14 @@ function TChannelRequest(channel, options) {
     if (!self.options.retryFlags) {
         self.options.retryFlags = {
             never: false,
-            onConnectionError: true,
-            onTimeout: false
```
Raynos: leave onTimeout: false here

Raynos: Actually; if `timeoutPerAttempt` is specified then set `onTimeout: true`

node/test/retry.js
```diff
@@ -317,26 +317,12 @@ allocCluster.test('retryFlags work', {
     });
 
     series([
-
-        function defaultToNotRetryingTimeout(next) {
```
Raynos: Why change this code ?

node/test/retry.js
```diff
@@ -356,25 +342,7 @@ allocCluster.test('retryFlags work', {
 
                 next();
             });
-        },
-
-        function canOptOutFully(next) {
```
Raynos: Why change this ?

node/v2/index.js
```diff
@@ -84,11 +84,9 @@ module.exports.parseRetryFlags = function parseRetryFlags(val) {
     val = val || 'c';
     var never = val.indexOf('n') > -1;
     var onConnectionError = !never && val.indexOf('c') > -1;
-    var onTimeout = !never && val.indexOf('t') > -1;
```
Raynos: leave the `onTimeout` flag here.

node/test/timeouts.js
ShanniLi: fixed

Raynos:
lgtm. Thank you @ShanniLi

PR #829, opened 2015-07-02
go: Simpler Context creation + always create root tracing span

prashantv:
Make *correct* context creation easier for clients. It involved too many steps earlier:
 - context.Background, context.WithTimeout, tchannel.NewRootContext, (json|thrift).NewContext

A lot of tests/clients were not calling NewRootContext so NewContext now creates the root trace span.

We may need another version with an options struct in future as we implement more features (retry, failure domains, etc). I want to keep it simple for now.

r @mmihic


mmihic:
lgtm

PR #830, opened 2015-07-02
Line wrap the circuit breaking doc for easier reading

jcorbin:
r @breerly


jcorbin:
The tables are a bit unwieldy still, not sure if GFM supports wrapping within / multi-line cells

Raynos:
ship it

PR #831, opened 2015-07-02
remove test_server

blampe:
@jcorbin I think you added this as part of your test framework but an echo server already lives in `examples`.


abhinav:
:+1:

PR #836, opened 2015-07-02
fix flake test causing by incorrectly set default value

junchaowu:
@blampe @abhinav @breerly

### DIFF
--- a/python/Makefile
+++ b/python/Makefile
@@ -1,7 +1,7 @@
 project := tchannel
 
 flake8 := flake8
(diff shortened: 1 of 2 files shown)


blampe:
Why would this cause flakes?

junchaowu:
@blampe the wrapper method of 'send' in the broker.py will be called with corresponding parameters but the default value are None.

blampe:
Uhhh ok. We *really* need to kill this broker thing.

PR #837, opened 2015-07-02
go: Hyperbahn advertise failures should use FullJitter backoff.

prashantv:
Advertise should fuzz ~60s (not 10s), with a 10s fuzz, so range is 50s-70s.

r @mmihic


mmihic:
lgtm

PR #838, opened 2015-07-02
allow overriding 'not found' behavior

blampe:
This is analogous to a 404 handler. The use case here is a testing tool that would accept requests at any endpoint, replay a response if it's available, or, if not, forward the request to a real server.

@abhinav @breerly @junchaowu

### DIFF
--- a/python/CHANGES.rst
+++ b/python/CHANGES.rst
@@ -4,7 +4,7 @@ Changelog for tchannel.py
 0.8.6 (unreleased)
 ------------------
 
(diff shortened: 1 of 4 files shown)


abhinav:
Looks good except I don't like the idea of None to mean you're specifying the fallback handler. We should avoid special values in general. How about a constant called `FALLBACK`?

junchaowu:
Good

abhinav:
:+1:

PR #839, opened 2015-07-02
add stats for outbound.calls.retries, per-attempt.latency

junchaowu:
@blampe @breerly @abhinav 

Doesn't support fragments or streaming case for latency calculation.

### DIFF
--- a/python/tchannel/event.py
+++ b/python/tchannel/event.py
@@ -63,11 +63,11 @@ class EventHook(object):
 
     """
     def before_send_request(self, request):
-        """Called before any request initiated by the service."""
+        """Called before each request is sent."""
         pass
 
     def before_send_request_per_attempt(self, request, retry_count):
-        """Called before sending each request for each attempt."""
+        """Called before each attempt to send the request."""
         pass
 
     def after_send_request(self, request):
--- a/python/tchannel/event.py
+++ b/python/tchannel/event.py
@@ -33,6 +33,7 @@ log = logging.getLogger('tchannel')
 EventType = enum(
     'EventType',
     before_send_request=0x00,
+    before_send_request_per_attempt=0x02,
     after_send_request=0x01,
     before_send_response=0x10,
     after_send_response=0x11,
@@ -62,7 +63,11 @@ class EventHook(object):
 
     """
     def before_send_request(self, request):
-        """Called before any part of a ``CALL_REQ`` message is sent."""
+        """Called before any request initiated by the service."""
+        pass
+
+    def before_send_request_per_attempt(self, request, retry_count):
+        """Called before sending each request for each attempt."""
         pass
 
     def after_send_request(self, request):
--- a/python/tchannel/statsd.py
+++ b/python/tchannel/statsd.py
@@ -39,7 +39,7 @@ class StatsdHook(EventHook):
         :param statsd: instance of `StatsD <https://github.com/etsy/statsd>`
         """
         self._statsd = statsd
-        self.outbound_attempt = {}
+        self.outbound_attempt = {}  # {span_id: start_time}
 
     def before_send_request(self, request):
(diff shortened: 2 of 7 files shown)


python/tchannel/event.py
```diff
@@ -62,7 +63,11 @@ def before_send_request(self, request):
 
     """
     def before_send_request(self, request):
-        """Called before any part of a ``CALL_REQ`` message is sent."""
+        """Called before any request initiated by the service."""
```
abhinav: "Called before the request is sent."?

python/tchannel/event.py
```diff
@@ -62,7 +63,11 @@ def before_send_request(self, request):
...
+        pass
+
+    def before_send_request_per_attempt(self, request, retry_count):
+        """Called before sending each request for each attempt."""
```
abhinav: Called before each attempt to send the request.

python/tchannel/statsd.py
```diff
@@ -45,6 +47,17 @@ def before_send_request(self, request):
 
         self._statsd.count(key, 1)
 
+    def before_send_request_per_attempt(self, request, retry_count):
+        statsd_name = "tchannel.outbound.calls.retries"
+        retry_count += 1
+        key = common_prefix(statsd_name, request) + '.' + str(retry_count)
```
abhinav: We shouldn't be generating complete metric streams for individual attempt numbers.

python/tchannel/statsd.py
```diff
@@ -45,6 +47,17 @@ def before_send_request(self, request):
...
+
+        self._statsd.count(key, 1)
+
+        # record outbound call start time and retry_count
+        self.outbound_attempt[request.tracing.span_id] = (time.time(),
```
abhinav: I thought we weren't going to key this by span ID?

python/tchannel/statsd.py
```diff
@@ -38,6 +39,7 @@ def __init__(self, statsd):
         :param statsd: instance of `StatsD <https://github.com/etsy/statsd>`
         """
         self._statsd = statsd
+        self.outbound_attempt = {}
```
abhinav: Usually for dictionaries like this, it's a good idea to add a comment indicating what it will map from/to.

python/tchannel/statsd.py
```diff
@@ -91,6 +107,19 @@ def on_operational_error(self, request, error):
...
 
+    def outbound_latency_per_attempt(self, request):
+        if request.tracing.span_id not in self.outbound_attempt:
+            return
+        (start_time, retry_count) = self.outbound_attempt.pop(
+            request.tracing.span_id)
+
+        latency_statsd_name = "tchannel.outbound.calls.per-attempt.latency"
+        key = common_prefix(latency_statsd_name, request) + '.' + str(
+            retry_count)
```
abhinav: Again, we shouldn't be emitting complete stat streams for each attempt count. You don't need to differentiate between the latencies of the different attempts. Just emit a latency metric for each attempt under the same metric name.

python/tchannel/statsd.py
junchaowu: We do. This is the discussion result.

python/tchannel/event.py
```diff
@@ -62,7 +63,11 @@ def before_send_request(self, request):
 
     """
     def before_send_request(self, request):
-        """Called before any part of a ``CALL_REQ`` message is sent."""
+        """Called before each request is sent."""
+        pass
+
+    def before_send_request_per_attempt(self, request, retry_count):
+        """Called before each attempt to send the request."""
```
abhinav: changelog

abhinav:
Update the changelog and :shipit:

blampe:
@junchaowu could you move this to tchannel-python?

PR #843, opened 2015-07-02
Use no-cn as default caller name, and adjust circuit signature

kriskowal:


kriskowal:
r @jcorbin @raynos

Raynos:
lgtm.

PR #844, opened 2015-07-02
Remove the response argument from health state handlers

kriskowal:
And reorder the methods for clarity. These response arguments were not used. We can add them later if necessary.

r @jcorbin @Raynos


Raynos:
lgtm.

PR #845, opened 2015-07-02
State options

kriskowal:
Creates a StateOptions constructor and documents the options inline.


kriskowal:
r @Raynos @jcorbin

Raynos:
lgtm.

PR #846, opened 2015-07-02
Test circuit recovery against failure frequency

kriskowal:


node/test/circuits.js
```diff
@@ -417,6 +417,74 @@ RelayNetwork.test('recovers after five successes', aliceAndBob, function t(netwo
...
 
+RelayNetwork.test('does not recover when successes are non-consecutive', aliceAndBob, function t(network, assert) {
+
+    var count = 0;
+
+    var circuit = network.getCircuit(0, 'alice', 'bob', 'call');
+
+    network.register('call', function (req, res) {
+        count++;
+        if (count > 20 && count % 4 !== 0) {
```
kriskowal: Note that in this case, 1 in 4 fail after the circuit is forced into the unhealthy state.

node/test/circuits.js
```diff
@@ -417,6 +417,74 @@ RelayNetwork.test('recovers after five successes', aliceAndBob, function t(netwo
...
+
+RelayNetwork.test('recovers when failure is periodic but infrequent', aliceAndBob, function t(network, assert) {
+
+    var count = 0;
+
+    var circuit = network.getCircuit(0, 'alice', 'bob', 'call');
+
+    network.register('call', function (req, res) {
+        count++;
+        if (count > 20 && count % 8 !== 0) {
```
kriskowal: Whereas in this case, one in 8. There is some interplay with the number of requests that occur per period, as approximately determined by the exercise count and delay.

kriskowal:
r @jcorbin @Raynos @breerly

Raynos:
lgtm.

PR #849, opened 2015-07-03
relay: Remove retries from relay handler

Raynos:
We remove retry logic from the relay handler.

This should make things faster. It also removes the
cascading retry problem.

r: @kriskowal @shannili @rf

### DIFF
--- a/node/relay_handler.js
+++ b/node/relay_handler.js
@@ -44,16 +44,27 @@ RelayHandler.prototype.handleRequest = function handleRequest(req, buildRes) {
             priorOutReqId: rereq.outreq && rereq.outreq.id
             // TODO more context, like outreq remote addr
         });
-        buildRes().sendError('UnexpectedError', 'request id exists in relay handler');
+        buildRes().sendError(
+            'UnexpectedError', 'request id exists in relay handler'
+        );
         return;
     }
+
     rereq = new RelayRequest(self.channel, req, buildRes);
+
     self.reqs[reqKey] = rereq;
     rereq.finishEvent.on(rereqFinished);
     rereq.createOutRequest();
+
     function rereqFinished() {
-        delete self.reqs[reqKey];
+        self.clearRequest(reqKey);
     }
+};
+
+RelayHandler.prototype.clearRequest = function clearRequest(reqKey) {
+    var self = this;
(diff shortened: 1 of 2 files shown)


kriskowal:
Is this PR labeled correctly? This looks like general performance improvements by factoring closures into methods.

node/relay_request.js
```diff
@@ -51,11 +52,33 @@ RelayRequest.prototype.createOutRequest = function createOutRequest(host) {
         return;
     }
 
+    self.peer = self.channel.peers.choosePeer(null, {
+        host: host
+    });
+    if (!self.peer) {
+        self.onError(errors.NoPeerAvailable());
+        return;
+    }
```
kriskowal: This might be the part that prevents retries. Does erring early when there is no peer available prevent the relay handler from proceeding into retries? Seems like this avoids a very specific case of retrying.

node/relay_handler.js
```diff
@@ -44,18 +44,29 @@ RelayHandler.prototype.handleRequest = function handleRequest(req, buildRes) {
...
     rereq.finishEvent.on(rereqFinished);
     rereq.createOutRequest();
+
     function rereqFinished() {
-        delete self.reqs[reqKey];
+        self.clearRequest(reqKey);
     }
 };
 
+RelayHandler.prototype.clearRequest = function clearRequest(reqKey) {
```
ShanniLi: why creating this function?

node/relay_request.js
Raynos: There are no retries because we call `peer.request()` instead of `channel.request()`

All retries are in the logical request class which is only driven from the `channel`

node/relay_handler.js
Raynos: every closure should be a method. For performance.

node/relay_request.js
kriskowal: :ok:

kriskowal:
:+1:

node/relay_request.js
ShanniLi: @Raynos not sure if it matters, since self.channel.request was passed in opts.host, which will lead to the path that uses self.peers.request.

PR #851, opened 2015-07-03
go: Initial stats reporting support, wire up simple outbound counters

prashantv:
Adds:
- StatsReporter interface which will be used to report stats.
- simple stats reporters (null, simple to print to log)
- recording stats reporter + counter validation

And wires up a couple of outbound stats to be reported.

r @mmihic


mmihic:
I am super excited for this.

PR #854, opened 2015-07-06
Don't write ECONNREFUSED to info log

jwolski:
This log is quite spammy, especially on ringpop startup and ping failures

@Raynos @kriskowal @jcorbin

### DIFF
--- a/node/connection.js
+++ b/node/connection.js
@@ -520,7 +520,8 @@ TChannelConnection.prototype.resetAll = function resetAll(err) {
         self.errorEvent.emit(self, err);
     } else if (
         err.type !== 'tchannel.socket-closed' &&
-        err.type !== 'tchannel.socket-local-closed'
+        err.type !== 'tchannel.socket-local-closed' &&
+        (err.type !== 'tchannel.socket' && err.code !== 'ECONNREFUSED')
     ) {
         logInfo.error = extend(err);
         logInfo.error.message = err.message;


jwolski:
Nevemind. I told tchannel to stfu by not providing it a logger.

Raynos:
@jwolski not passing in a logger is a REALLY BAD IDEA (tm).

jwolski:
I suppose you're right, but tchannel is logging things at a level I don't care for. What is your suggestion? IMO, it should do far less logging than it is or at the very least, since it's an RPC library, should be logging at the debug level more than it is.

Raynos:
@jwolski 

All the tchannel logs are configured for all tchannel services, including hyperbahn. Hyperbahn is silent under steady state.

Because hyperbahn is silent in production under steady state the logger is not noisy under steady state.

The problem is that the logger is noisy in startup & teardown situations; this is a non-trivial problem as those log lines are useful in non-startup and non-teardown situations.

Everyone would benefit from fine tuning the logger in both ringpop & tchannel, I recommend you keep the logger on and that we all work on improving it.

As a rule of thumb; if you have the logger on in production you should have it enabled locally and in your tests. If the logger is too noisy locally or in your tests, then its too noisy in production.

Raynos:
If your concern is testing I highly recommend you use [debug-logtron](https://www.npmjs.com/package/debug-logtron)

It comes with a few features:

 - debug is off by default and enabled with `NODE_DEBUG=x`
 - info/warn are color coded.
 - error/fatal throw exceptions; if they are bugs in prod, they are bugs in tests.
 - All logs go to STDERR; you can always use `2>/dev/null` for reading the actual STDOUT of your tests.

jwolski:
My concern is not testing. My concern is ringpop startup/deploys with tchannel v2. There's a storm of ECONNREFUSED reported in the info log, which are completely expected in that case. And there's a lot more reported in the failed ping / ping-req case.

Raynos:
We should find a way to improve the logger for everyone.

However it should be noted that bringing up an entire ring (1000 nodes) from nothing is not a normal production case, everyone wants to minimize ringpop and tchannel log noise during a rolling restart.

The rolling restart is currently frustratingly noisy and I have not found a good way of making it quieter, all suggestions are welcome.

PR #855, opened 2015-07-06
Improve performance of lib/statsd

Raynos:
This PR mostly optimizes the `clean()` functions in the
statsd integration.

r: @kriskowal @rf @shannili

### DIFF
--- a/node/lib/statsd.js
+++ b/node/lib/statsd.js
@@ -20,6 +20,8 @@
 
 'use strict';
 
+var LENGHT_ARRAYS = {};
+
 function TChannelStatsd(channel, statsd) {
     if (!(this instanceof TChannelStatsd)) {
         return new TChannelStatsd(channel, statsd);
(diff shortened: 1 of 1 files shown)


node/lib/statsd.js
```diff
@@ -36,53 +39,88 @@ function TChannelStatsd(channel, statsd) {
...
-        // TODO generalize
-        return str
-            .replace(/:/g, '-')
-            .replace(/\//g, '-')
-            .replace(/\./g, '-')
-            .replace(/{|}/g, '-');
+        return field;
     }
+
+    copy = LENGHT_ARRAYS[str.length];
```
kriskowal: LENGTH

node/lib/statsd.js
```diff
@@ -36,53 +39,88 @@ function TChannelStatsd(channel, statsd) {
...
+            char === '{' ||
+            char === '}'
+        ) {
+            copy[i] = '-';
+        } else {
+            copy[i] = char;
+        }
+    }
+
+    return copy.join('');
```
kriskowal: I believe the prevailing wisdom is that string concatenation is faster than this.

kriskowal:
:ok:

ShanniLi:
lgtm

node/lib/statsd.js
Raynos: it's not; the flames were like "WHAT IS THIS CONCAT STRING DATA STRUCTURE"

Raynos: wait I didn't try concatenation.

PR #859, opened 2015-07-06
Add Thrift support to Python Sync Client.

breerly:
This PR creates `tchannel.sync.thrift.client_for` function that generates a Thrift client with `TChannelSyncClient` as the transport.

@abhinav @blampe @junchaowu


python/tchannel/sync/thrift.py
```diff
@@ -0,0 +1,123 @@
...
+        from comment import CommentService
+
+        CommentServiceClient = client_for('comment', CommentService)
+
+        tchannel_sync = TChannelSyncClient('my-service')
+        comment_client = CommentServiceClient(tchannel_sync)
+
+        future = comment_client.postComment(
+            articleId,
+            CommendService.Comment("hi")
```
breerly: CommentService*

python/tests/sync/test_thrift.py
```diff
@@ -0,0 +1,57 @@
...
+
+import pytest
+
+from tchannel.sync import TChannelSyncClient
+from tchannel.sync.thrift import client_for
+
+
+@pytest.fixture
+def thrift_sync_client(tchannel_server, thrift_service):
+
```
breerly: this space is annoying me

python/tests/conftest.py
```diff
@@ -61,3 +62,9 @@ def random_open_port():
 def tchannel_server(random_open_port):
     with TestServer(random_open_port) as server:
         yield server
+
+
+@pytest.yield_fixture
+def thrift_service(tmpdir):
```
breerly: This should almost be `thrift_module`.

python/tchannel/sync/client.py
```diff
@@ -25,8 +25,8 @@
 from threadloop import ThreadLoop
 from tornado import gen
 
-from . import glossary
-from . import tornado as async
+from tchannel import glossary
+from tchannel import tornado as async
```
junchaowu: why not relative import? All other part of code uses relative import

python/tchannel/sync/thrift.py
```diff
@@ -0,0 +1,123 @@
...
+            tchannel_sync.async_client,
+            hostport,
+            trace,
+        )
+        self.threadloop = tchannel_sync.threadloop
+
+    init.__name__ = '__init__'
+    methods = {
+        '__init__': init,
+        '__async_client_class__': async_client_for(
```
junchaowu: Why do you need to create __async_client_class__ method? It basically calls the async client_for and is used only once.

python/tchannel/sync/thrift.py
```diff
@@ -0,0 +1,123 @@
...
+    with Hyperbahn under the name "comment", here's how this might be used:
+
+    .. code-block:: python
+
+        from tchannel.sync import TChannelSyncClient
+        from tchannel.sync.thrift import client_for
+
+        from comment import CommentService
+
+        CommentServiceClient = client_for('comment', CommentService)
```
junchaowu: Is it confusing by calling both sync and async `client_for`?

abhinav: I don't think so. They are namespaced differently.You would use this like so:

    from tchannel.sync.thrift import client_for

There's no ambiguity there about which client_for you're referring to.

python/tchannel/sync/thrift.py
```diff
@@ -0,0 +1,123 @@
...
+
+from __future__ import absolute_import
+
+import inspect
+
+from tchannel.thrift.client import client_for as async_client_for
+
+
+def client_for(service, service_module, thrift_service_name=None):
+    """Build a sync client class for the given Thrift service.
```
abhinav: nit: synchronous

python/tchannel/sync/thrift.py
```diff
@@ -0,0 +1,123 @@
...
+        comment_client = CommentServiceClient(tchannel_sync)
+
+        future = comment_client.postComment(
+            articleId,
+            CommendService.Comment("hi")
+        )
+        result = future.result()
+
+    :param service:
+        Name of the HyperBahn service being called.
```
abhinav: nit: we don't capitalize the B in hyperbahn anywhere else

python/tchannel/sync/thrift.py
```diff
@@ -0,0 +1,123 @@
...
+    :param service_module:
+        The Thrift-generated module for that service. This usually has
+        the same name as definied for the service in the IDL.
+    :param thrift_service_name:
+        If the Thrift service has a different name than its module, use
+        this parameter to specify it.
+    :returns:
+        An Thrift-like class, ready to be instantiated and used
+        with TChannelSyncClient.
```
abhinav: The docs need to mention somewhere that the methods will return futures where .result() will block until a result is available.

python/tchannel/sync/client.py
breerly: My understanding is that you don't want to use relative imports to grab packages in parent modules; only children.

python/tchannel/sync/thrift.py
breerly: sure.

python/tchannel/sync/thrift.py
breerly: @junchaowu I thought that at first too... eventually I settled into what @abhinav said.

python/tchannel/sync/thrift.py
breerly: :P

python/tchannel/sync/thrift.py
abhinav: `async_client_for` generates the asynchronous class and stores it as a class-level attribute. You don't want to generate it every time you instantiate a synchronous client. This is essentially equivalent to:

    class CommentServiceSyncClient(object):
        __async_client_class__ = client_for(...)

        # ...

Again, both versions of `client_for` generate new classes at runtime. You don't need to regenerate them because their behavior is static given the previously specified parameters, so it's a good idea to save it as a class-level variable.

breerly: Because I want to instantiate this class in init on `79`, then decorate the methods by calling them off of `self.async_client` on `120`.

blampe:
This is great :)

PR #860, opened 2015-07-06
test: use different ports from examples

Raynos:
This is a simple fixup commit that changes the ports.

Now we can run benchmarks, tests & examples at the
same time.

r: @kriskowal @rf @shannili


kriskowal:
:ok:

ShanniLi:
lgtm

PR #862, opened 2015-07-06
Handle init connection failure edge case.

Raynos:
This is a bug introduced by an earlier relay refactor.
This contains a regression test as well

r: @kriskowal @rf @shannili

### DIFF
--- a/node/peer.js
+++ b/node/peer.js
@@ -283,10 +283,10 @@ TChannelPeer.prototype.addConnection = function addConnection(conn) {
 
 TChannelPeer.prototype.removeConnection = function removeConnection(conn) {
     var self = this;
-    var list = self.connections;
-    var index = list ? list.indexOf(conn) : -1;
+
+    var index = self.connections ? self.connections.indexOf(conn) : -1;
     if (index !== -1) {
-        return list.splice(index, 1)[0];
+        return self.connections.splice(index, 1)[0];
     } else {
         return null;
     }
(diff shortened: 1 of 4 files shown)


node/peer.js
```diff
@@ -283,10 +283,10 @@ TChannelPeer.prototype.addConnection = function addConnection(conn) {
 
 TChannelPeer.prototype.removeConnection = function removeConnection(conn) {
     var self = this;
-    var list = self.connections;
-    var index = list ? list.indexOf(conn) : -1;
+
+    var index = self.connections ? self.connections.indexOf(conn) : -1;
```
ShanniLi: what's the bug here?

Raynos: no bug here; just hard to `git grep` because we aliased as variable.

ShanniLi:
lgtm

node/test/relay-to-dead.js
```diff
@@ -0,0 +1,93 @@
...
+var net = require('net');
+var setTimeout = require('timers').setTimeout;
+
+var allocCluster = require('./lib/alloc-cluster');
+var RelayHandler = require('../relay_handler');
+
+allocCluster.test('relaying to init timeout server', {
+    numPeers: 2
+}, function t(cluster, assert) {
+    var deadServer = net.createServer();
```
kriskowal: Clever. Took me a bit to find the line that makes this server “dead”. Might be good to note that it is passively dead since it does does nothing with incoming connections.

kriskowal:
:ok:

PR #863, opened 2015-07-06
Gaurd against negative ttl in forwarding

Raynos:
This is a bug introduced by a previous PR.
I also added a failing test.

r: @kriskowal @shannili @rf


Raynos:
The test failed because inOps flappy test.

rf:
great PR jake :+1:

ShanniLi:
nice catch! :+1:

PR #866, opened 2015-07-06
Optimize the stats code with fast stats

Raynos:
This adds a new code path of using `channel.buildStat()`
and `channel.emitFastStat()`

This code has a lot less GC churn, anonymous objects and
unnecessary methods.

This creates a non-trivial speed up.

r: @jcorbin @kriskowal @rf @shannili


node/out_response.js
```diff
@@ -256,25 +279,62 @@ TChannelOutResponse.prototype.send = function send(res1, res2) {
...
     return self;
 };
 
+function InboundCallsSuccessTags(cn, serviceName, endpoint) {
+    var self = this;
+
+    self.app = '';
+    self.host = '';
+    self.cluster = '';
+    self.version = '';
```
kriskowal: These might benefit from a shared constructor for the common properties. Also, null might be faster than ''. I don’t believe V8 treats slots differently based on the type to which they are initially set.

Raynos: Interesting ideas; feel free to make a branch and verify with the benchmarks :)

node/out_request.js
```diff
@@ -340,25 +434,22 @@ TChannelOutRequest.prototype.sendCallRequestContFrame = function sendCallRequest
 TChannelOutRequest.prototype.send = function send(arg1, arg2, arg3, callback) {
     var self = this;
 
-    var endpoint = String(arg1);
+    self.endpoint = String(arg1);
```
kriskowal: Let’s all take note of this change. arg1 is coerced to string in the circuit health code as well. That’ll benefit from using this property instead. We may need to introduce the endpoint property to other request variants. cc @jcorbin

kriskowal:
:ok:

node/lib/statsd.js
```diff
@@ -291,22 +222,17 @@ TChannelStatsd.prototype.onStat = function onStat(stat) {
...
-        case 'timing':
-            return self.statsd.timing(key, stat.value);
-
-        default:
-            self.channel.logger.error('Trying to emit an invalid stat object', {
-                statType: stat.type,
-                statName: stat.name
-            });
-            break;
+    if (stat.type === 'counter') {
```
ShanniLi: why not switch?

Raynos: purely because less lines of code. I also think switch is slower but dont have proof.

kriskowal: Because switches are a scourge upon this Earth and Doug Crockford says “no”.

not found

node/test/request-stats.js
```diff
@@ -163,10 +181,12 @@ allocCluster.test('emits stats', {
             type: 'timing',
             value: stats[8].value,
             tags: {
-                'target-service': 'server',
```
ShanniLi: this is also for m3, we also needs to update the metrics spec for that

ShanniLi:
lgtm

rf:
:+1:

Raynos:
This bad boy is going in.

I'm going to pass of the rest of the stats refactoring to @ShanniLi

PR #867, opened 2015-07-06
go: spanID should be equal to traceID for the root request

prashantv:
r: @mmihic

### DIFF
--- a/golang/tracing.go
+++ b/golang/tracing.go
@@ -92,12 +92,17 @@ func (s Span) TracingEnabled() bool { return (s.flags & tracingFlagEnabled) == t
 
 // NewChildSpan begins a new child span in the provided Context
 func (s Span) NewChildSpan() *Span {
-	return &Span{
+	childSpan := &Span{
 		traceID:  s.traceID,
 		parentID: s.spanID,
-		spanID:   uint64(rng.Int63()),
 		flags:    s.flags,
 	}
+	if s.spanID == 0 {
+		childSpan.spanID = childSpan.traceID
+	} else {
+		childSpan.spanID = uint64(rng.Int63())
+	}
+	return childSpan
 }
 
 // CurrentSpan returns the Span value for the provided Context
--- a/golang/tracing.go
+++ b/golang/tracing.go
@@ -23,6 +23,7 @@ package tchannel
 import (
 	"fmt"
 	"math/rand"
+	"sync"
 	"time"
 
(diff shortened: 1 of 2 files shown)


golang/tracing.go
```diff
@@ -92,12 +92,17 @@ func (s Span) TracingEnabled() bool { return (s.flags & tracingFlagEnabled) == t
...
 // NewChildSpan begins a new child span in the provided Context
 func (s Span) NewChildSpan() *Span {
-	return &Span{
+	childSpan := &Span{
 		traceID:  s.traceID,
 		parentID: s.spanID,
-		spanID:   uint64(rng.Int63()),
 		flags:    s.flags,
 	}
+	if s.spanID == 0 {
```
mmihic: This is fine, but isn't technically a requirement according to the <a href="https://twitter.github.io/zipkin/Instrumenting.html#generating-identifiers">zipkin docs</a>.  The only requirement for the initial child span is that it have the same trace id as the parent, and that its parent id be the span id of the parent.  Of course if we want to impose this additional correlation requirement, I guess that's fine.

golang/tracing.go
```diff
@@ -92,12 +92,17 @@ func (s Span) TracingEnabled() bool { return (s.flags & tracingFlagEnabled) == t
...
+		childSpan.spanID = childSpan.traceID
+	} else {
+		childSpan.spanID = uint64(rng.Int63())
```
mmihic: Erp.  I just realized I used a global rng without a protecting mutex.

mmihic:
also, lgtm

golang/tracing.go
prashantv: Ah interesting, thanks for the link. This other page http://itszero.github.io/blog/2014/03/03/introduction-to-twitters-zipkin/ made it seem like traceid = spanid, but you're right, the protocol does not require it. I'll keep the additional constraint since it makes it easy to identify the first span for now.

golang/tracing.go
prashantv: fixed

golang/tracing.go
```diff
@@ -63,9 +65,16 @@ func (s *Span) write(w *typed.WriteBuffer) error {
 
 const tracingFlagEnabled byte = 0x01
 
+// randUint64 returns a random uint64.
+func randUint64() uint64 {
```
mmihic: thanks for fixing this.

PR #868, opened 2015-07-07
test: force flappy test to succeed

Raynos:
This would flap in travis :(

There's no other way to force it other then
just wait for the timeouts to get reaped.

cc @jcorbin @kriskowal @shannili


kriskowal:
:ok:

Raynos:
Another flap; flaps be gone.

PR #869, opened 2015-07-07
refactor tchannel retry logic & Simplify the stats events

junchaowu:
@blampe @breerly @abhinav 

1. reorganize peer send code.
2. simplify retry logic
3. simplify the events types

### DIFF
--- a/python/tchannel/event.py
+++ b/python/tchannel/event.py
@@ -40,11 +40,9 @@ EventType = enum(
     after_receive_request=0x21,
     before_receive_response=0x30,
     after_receive_response=0x31,
-    after_receive_system_error=0x40,
-    after_receive_system_error_per_attempt=0x41,
-    after_send_system_error=0x42,
-    on_operational_error=0x50,
-    on_operational_error_per_attempt=0x51,
+    after_receive_error_response=0x40,
+    after_send_error_response=0x41,
+    on_outbound_error=0x50,
 )
 
 
@@ -93,24 +91,16 @@ class EventHook(object):
         """Called after a ``CALL_RESP`` message is read."""
         pass
 
-    def after_receive_system_error(self, request, err):
+    def after_receive_error_response(self, request, err):
         """Called after a ``error`` message is read."""
         pass
 
-    def after_send_system_error(self, err):
+    def after_send_error_response(self, err):
         """Called after a ``error`` message is sent."""
         pass
 
-    def after_receive_system_error_per_attempt(self, request, err):
-        """Called after a ``error`` message is read for each attempt."""
-        pass
-
-    def on_operational_error(self, request, err):
+    def on_outbound_error(self, request, err):
         """Called after an operational error happens per request."""
-        pass
-
-    def on_operational_error_per_attempt(self, request, err):
-        """Called after an operational error happens for each attempt."""
         pass
 
 
--- a/python/tchannel/event.py
+++ b/python/tchannel/event.py
@@ -100,7 +100,7 @@ class EventHook(object):
         pass
 
     def on_outbound_error(self, request, err):
-        """Called after an operational error happens per request."""
+        """Called after an error happens for outbound request."""
         pass
 
 
--- a/python/tchannel/tornado/peer.py
+++ b/python/tchannel/tornado/peer.py
@@ -20,6 +20,7 @@
 
 from __future__ import absolute_import
 
+import functools
 import logging
 from collections import deque
 from itertools import chain
@@ -422,6 +423,84 @@ class PeerHealthyState(PeerState):
             # control randomness.
 
 
+def retry(func):
+    @gen.coroutine
+    @functools.wraps(func)
+    def func_wrapper(self, endpoint, arg2, arg3, headers,
+                     trace_id, parent_span_id, traceflag,
+                     ttl, attempt_times, retry_delay,):
+        # black list to record all used peers, so they aren't chosen again.
+        blacklist = set()
+        response = None
+        peer = self.peer
+        # max number of times to retry 3.
+        for num_of_attempt in range(attempt_times):
+            try:
+                connection = yield peer.connect()
+                request = Request(
+                    service=self.service,
+                    argstreams=[
+                        InMemStream(endpoint), arg2.clone(), arg3.clone()
+                    ],
+                    id=connection.next_message_id(),
+                    headers=headers,
+                    endpoint=endpoint,
+                    ttl=ttl,
+                    tracing=Trace(
+                        name=endpoint,
+                        trace_id=trace_id,
+                        parent_span_id=parent_span_id,
+                        endpoint=Endpoint(peer.host,
+                                          peer.port,
+                                          self.service),
+                        traceflags=traceflag,
+                    )
+                )
+
+                # event: send_request
+                self.peer.tchannel.event_emitter.fire(
+                    EventType.before_send_request,
+                    request,
+                )
+                response = yield gen.maybe_future(
+                    func(self, connection, request)
+                )
+                # event: receive_response
+                self.peer.tchannel.event_emitter.fire(
+                    EventType.after_receive_response,
+                    request,
+                    response,
+                )
+                break
+            except (ProtocolError, TimeoutError) as error:
+                if hasattr(error, "tracing"):
+                    # event: after_receive_error_response
+                    self.peer.tchannel.event_emitter.fire(
+                        EventType.after_receive_error_response,
+                        request,
+                        error,
+                    )
+                else:
+                    # event: on_outbound_error
+                    self.peer.tchannel.event_emitter.fire(
+                        EventType.on_outbound_error,
+                        request,
+                        error,
+                    )
+                blacklist.add(peer.hostport)
+                peer = yield self.prepare_for_retry(
+                    request, connection, error,
+                    blacklist, retry_delay,
+                    num_of_attempt, attempt_times,
+                )
+                if not peer:
+                    raise
+
+        raise gen.Return(response)
+
+    return func_wrapper
+
+
 class PeerClientOperation(object):
     """Encapsulates client operations that can be performed against a peer."""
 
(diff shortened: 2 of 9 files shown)


python/tchannel/tornado/peer.py
```diff
@@ -422,6 +423,84 @@ def score(self):
             # control randomness.
 
 
+def retry(func):
+    @gen.coroutine
+    @functools.wraps(func)
+    def func_wrapper(self, endpoint, arg2, arg3, headers,
```
blampe: What's `self` an instance of?

python/tchannel/tornado/peer.py
```diff
@@ -422,6 +423,84 @@ def score(self):
             # control randomness.
 
 
+def retry(func):
+    @gen.coroutine
```
blampe: Coroutine within coroutines :(

python/tchannel/tornado/peer.py
junchaowu: Because this decorator will be used on a instance method. The first argument by default is `self`

python/tchannel/tornado/peer.py
junchaowu: Removed this decorator thing. Just to make it as a normal function call.

python/tchannel/event.py
```diff
@@ -40,11 +40,9 @@
     after_receive_request=0x21,
     before_receive_response=0x30,
     after_receive_response=0x31,
-    after_receive_system_error=0x40,
-    after_receive_system_error_per_attempt=0x41,
-    after_send_system_error=0x42,
-    on_operational_error=0x50,
-    on_operational_error_per_attempt=0x51,
```
abhinav: Are these hooks being replaced and consolidated? Add relevant entries to the changelog: "Replace X and Y hooks with Z." Also, this counts as backwards breaking if these hooks were present in the last released version. You'd want to bump the major version.

python/tchannel/tornado/peer.py
```diff
@@ -525,160 +525,123 @@ def send(self, arg1, arg2, arg3,
...
-                    protocol_error,
+                connection = yield peer.connect()
+                request = Request(
+                    service=self.service,
+                    argstreams=[
+                        InMemStream(endpoint), arg2.clone(), arg3.clone()
+                    ],
+                    id=connection.next_message_id(),
+                    headers=headers,
+                    endpoint=endpoint,
```
abhinav: For a different CR: Why is the endpoint being duplicated here? `argstreams` contains it too. We should probably change this since we decided that endpoints cannot be streamed.

python/tchannel/tornado/peer.py
```diff
@@ -525,160 +525,123 @@ def send(self, arg1, arg2, arg3,
...
+                # Error Response will have tracing field.
+                if hasattr(error, "tracing"):
+                    # event: after_receive_error_response
+                    self.peer.tchannel.event_emitter.fire(
+                        EventType.after_receive_error_response, request, error,
+                    )
+                else:
+                    # event: on_outbound_error
+                    self.peer.tchannel.event_emitter.fire(
+                        EventType.on_outbound_error, request, error,
```
abhinav: I don't get the name of this event. The description says it's an operational error.

python/tchannel/tornado/peer.py
```diff
@@ -525,160 +525,123 @@ def send(self, arg1, arg2, arg3,
...
+                                          self.service),
+                        traceflags=traceflag,
+                    )
                 )
 
-                (peer, connection) = yield self.prepare_for_retry(
-                    request, connection, protocol_error,
-                    peer, blacklist, retry_delay,
-                    num_of_attempt, attempt_times,
+                # event: send_request
```
abhinav: "event:" comment outdated here and 6 lines down.

python/tchannel/tornado/peer.py
```diff
@@ -525,160 +525,123 @@ def send(self, arg1, arg2, arg3,
...
         raise gen.Return(response)
 
     @gen.coroutine
+    def _send(self, connection, req):
+        response_future = connection.send_request(req)
+        with timeout(response_future, req.ttl):
+            response = yield response_future
+        raise gen.Return(response)
+
+    def prepare_next_request(self, blacklist):
```
abhinav: Nit: The new version of this method looks simple enough that it probably doesn't need to be a separate method.

python/tchannel/event.py
junchaowu: No. Those affected code is not released yet.

python/tchannel/tornado/peer.py
```diff
@@ -525,160 +525,123 @@ def send(self, arg1, arg2, arg3,
...
                     num_of_attempt, attempt_times,
                 )
-
-                if not connection:
+                if not peer:
                     raise
 
         raise gen.Return(response)
 
     @gen.coroutine
```
blampe: This doesn't need to be a coroutine!!!

```
def _send(self, connection, req):
    response_future = connection.send_request(req)
    timeout(response_future, req.ttl)
    return response_future
```

python/tests/integration/test_retry.py
```diff
@@ -38,7 +38,7 @@
 
 @tornado.gen.coroutine
 def handler_error(request, response, proxy):
-    yield tornado.gen.sleep(0.01)
+    yield tornado.gen.sleep(0.0001)
```
blampe: Why is this happening???

python/tchannel/tornado/peer.py
junchaowu: Fixed with 
@staticmethod
    def _send(connection, req):
        response_future = connection.send_request(req)
        with timeout(response_future, req.ttl):
            return response_future

python/tests/integration/test_retry.py
junchaowu: Removed it. Unnecessary code.

python/tchannel/tornado/peer.py
junchaowu: Endpoint have same value of arg1. I create `endpoint` variable for easy access purpose.

python/tests/integration/test_retry.py
```diff
@@ -88,7 +88,7 @@ def test_retry_timeout():
                 headers={
                     're': RetryType.CONNECTION_ERROR_AND_TIMEOUT
                 },
-                ttl=0.005,
+                ttl=0.00001,
```
blampe: What is going on with all of these magic numbers...?

junchaowu: The number is tweaked to decide whether the test gets a local Timeout based on TTL or receive an Protocol Error Frame from server.

python/tchannel/tornado/peer.py
```diff
@@ -525,160 +525,122 @@ def send(self, arg1, arg2, arg3,
...
-            response = yield response_future
-
-        raise gen.Return(response)
-
-    @gen.coroutine
-    def send_with_retry(self, request, peer, attempt_times, retry_delay):
+    def send_with_retry(self, endpoint, arg2, arg3, headers,
+                        trace_id, parent_span_id, traceflag,
+                        ttl, attempt_times, retry_delay):
         # black list to record all used peers, so they aren't chosen again.
```
abhinav: Can you add an assert here?

    assert attempt_times > 0

Because if attempt_times is 0, we will silently not make a call, and that's programmer error. We need that to fail.

python/tchannel/event.py
abhinav: Okay, then these are new hooks since last release? You still want to add that to the changelog. "Add new hooks: a, b, c"

python/tchannel/tornado/peer.py
abhinav: Yeah, I'm saying that arg1 doesn't need to exist as a stream.

abhinav: Let's discuss this later. If that change happens it'll be in a different diff.

python/tchannel/tornado/peer.py
```diff
@@ -525,160 +525,122 @@ def send(self, arg1, arg2, arg3,
...
-                # event: after_receive_protocol_error_per_attempt
-                self.peer.tchannel.event_emitter.fire(
-                    EventType.after_receive_system_error_per_attempt,
-                    request,
-                    protocol_error,
+                connection = yield peer.connect()
+                request = Request(
+                    service=self.service,
+                    argstreams=[
+                        InMemStream(endpoint), arg2.clone(), arg3.clone()
```
abhinav: Probably add a TODO somewhere about streaming being broken because of this

python/tchannel/tornado/peer.py
abhinav: ping

python/tests/integration/test_retry.py
abhinav: Can't you alter server behavior to do that instead?

python/tchannel/tornado/peer.py
junchaowu: In the send(), it has ' attempt_times = attempt_times or MAX_ATTEMPT_TIMES'

python/tchannel/tornado/peer.py
junchaowu: I have updated the description for error events

python/tchannel/tornado/peer.py
junchaowu: Added TODO.

python/tchannel/tornado/peer.py
abhinav: Yes, that's why it being 0 or something else is a programmer error. Assertion is setting a precondition for calling that function. If that condition is validated, we messed up somewhere.

Also, if this is not part of the public API for that class, prepend a `_` to it.

python/tchannel/tornado/peer.py
```diff
@@ -500,7 +500,7 @@ def send(self, arg1, arg2, arg3,
         arg1, arg2, arg3 = (
             maybe_stream(arg1), maybe_stream(arg2), maybe_stream(arg3)
         )
-
+        assert attempt_times != 0
```
abhinav: No. Not here. This is not what I was saying. This will break things. You call `_send_with_retry` from this function. Add the assert in `_send_with_retry`.

The point of adding the statement there is that if it ever gets called with `attempt_times < 1`, we know that we made a mistake in the code that calls it. Otherwise it'll be very difficult to debug why a request isn't going through.

python/tchannel/event.py
abhinav: Ping. We should attempt to update the changelog as changes are made instead of trying to figure out what changes were made when releasing.

abhinav:
Okay. @blampe?

python/tchannel/tornado/peer.py
```diff
@@ -525,160 +524,124 @@ def send(self, arg1, arg2, arg3,
...
-                endpoint=Endpoint(self.peer.host,
-                                  self.peer.port,
-                                  self.service),
-                traceflags=traceflag,
-            )
-        )
-
         # only retry on non-stream request
-        if request.is_streaming_request or self._hostport:
+        if Request.is_streaming_request(arg2, arg3):
```
blampe: This is ignoring the cached `_is_streaming_request` now? That's an expensive function call to make, this is probably a performance regression.

python/tchannel/tornado/peer.py
```diff
@@ -525,160 +524,124 @@ def send(self, arg1, arg2, arg3,
...
-                if not connection:
+                if not peer:
                     raise
 
         raise gen.Return(response)
 
+    @staticmethod
+    def _send(connection, req):
+        response_future = connection.send_request(req)
+        with timeout(response_future, req.ttl):
```
blampe: This doesn't even need to be a context manager, because the exit block is a no-op.

junchaowu: The way we write the timeout is using yield. If remove the 'with', the it generates a generator. Timeout won't work

python/tchannel/tornado/peer.py
junchaowu: Streaming is broken. I think we shall create a separate api for streaming.

python/tchannel/tornado/peer.py
blampe: Yep you're right :)

python/tchannel/tornado/peer.py
blampe: Broken as in this is breaking it? Or it's been broken and it wasn't tested?

PR #872, opened 2015-07-07
Relay performance cleanup

Raynos:
This is a subset of the perf branch.

 - Remove double id protection to gain performance
 - Simplify code.
 - Re-use peer to gaurd against cannot req until identified edgecase

r: @kriskowal @shannili @rf

### DIFF
--- a/node/connection_base.js
+++ b/node/connection_base.js
@@ -66,7 +66,7 @@ function TChannelConnectionBase(channel, direction, socketRemoteAddr) {
         connection: self
     });
 
-    self.guid = ++CONNECTION_BASE_IDENTIFIER;
+    self.guid = ++CONNECTION_BASE_IDENTIFIER + '~';
 
     self.tracer = self.channel.tracer;
     self.ops.startTimeoutTimer();
--- a/node/relay_request.js
+++ b/node/relay_request.js
@@ -22,10 +22,8 @@
 
 var errors = require('./errors');
 
-function RelayRequest(channel, inreq, buildRes, onfinish) {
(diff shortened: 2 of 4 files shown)


node/connection_base.js
```diff
@@ -66,7 +66,7 @@ function TChannelConnectionBase(channel, direction, socketRemoteAddr) {
         connection: self
     });
 
-    self.guid = ++CONNECTION_BASE_IDENTIFIER;
+    self.guid = String(++CONNECTION_BASE_IDENTIFIER + '~');
```
kriskowal: The string cast is superfluous. Adding a string coerces implicitly.

Raynos: True;

anson627: ++CONNECTION_BASE_IDENTIFIER + '~' is already a string, why call String constructor again? Just out of curiosity.

Raynos:
Fixed.

ShanniLi:
lgtm

node/relay_request.js
```diff
@@ -116,7 +142,7 @@ RelayRequest.prototype.createOutResponse = function createOutResponse(options) {
             remoteAddr: self.inreq.remoteAddr,
             id: self.inreq.id
         });
-        return;
+        return null;
```
anson627: what's benefit of return null instead of undefined?

Raynos: lint.

anson627: make sense.

kriskowal:
:ok:

PR #873, opened 2015-07-07
Node: optimize buildResponse() code path

Raynos:
This one is more non-trivial. Here we rely on not making
any copies of the options passed into buildResponse() to
improve performance.

r: @kriskowal @rf @shannili


ShanniLi:
lgtm

kriskowal:
We do not appear to have a consistent story about whether to abbreviate Response to Res and Request to Req. I previously thought that story was: abbreviate local variables, expand method names.

:ok:

PR #874, opened 2015-07-07
operations: make a copy of tombstones to avoid 100% cpu splice

Raynos:
This is a tricky thing; this fixes two things

 - Turn O(n^2) -> O(n)
 - Avoid splice() which randomly causes a hanging 100% CPU issue

r: @rf @kriskowal @shannili


ShanniLi:
:+1:

anson627:
also cleaner, lgtm

PR #876, opened 2015-07-07
go: Use a random active peer and random active connection rather than first

prashantv:
Use separate random number generators for tracing vs peer/connection selection to avoid contention.

r @mmihic

### DIFF
--- a/golang/peer.go
+++ b/golang/peer.go
@@ -2,14 +2,18 @@ package tchannel
 
 import (
 	"errors"
-	"math/rand"
 	"sync"
+	"time"
 
 	"golang.org/x/net/context"
 )
 
-// ErrInvalidConnectionState indicates that the connection is not in a valid state.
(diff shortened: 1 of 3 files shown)


golang/peer.go
```diff
@@ -30,31 +36,36 @@ func (l *PeerList) Add(hostPort string) *Peer {
...
 
 	p := newPeer(l.channel, hostPort)
-	l.peers[hostPort] = p
+	l.peersByHostPort[hostPort] = p
+	l.peers = append(l.peers, p)
 	return p
 }
 
+func randPeer(peers []*Peer) *Peer {
+	return peers[peerRng.Intn(len(peers))]
```
mmihic: same problem as tracing with unsynchronized global rng.

mmihic: derp, n/m.

mmihic:
lgtm

PR #878, opened 2015-07-07
Cleanup the peers logic

Raynos:
This is another chunk of perf work.

 - Optimize peers to get rid of Object.keys()
 - Make choosePeer() simpler.

r: @kriskowal @shannili @rf

### DIFF
--- a/node/peers.js
+++ b/node/peers.js
@@ -207,9 +207,6 @@ TChannelPeers.prototype.choosePeer = function choosePeer(req, options) {
         return self.add(options.host);
     } else {
         hosts = Object.keys(self._map);
-        if (self.selfPeer) {
-            hosts.push(self.selfPeer.hostPort);
-        }
     }
 
     if (!hosts || !hosts.length) {
--- a/node/peers.js
+++ b/node/peers.js
@@ -198,7 +198,10 @@ TChannelPeers.prototype.choosePeer = function choosePeer(req, options) {
     /*eslint complexity: [2, 15]*/
     var self = this;
 
-    if (!options) options = {};
+    if (!options) {
+        options = {};
+    }
+
     var hosts = null;
     if (options.host) {
         return self.add(options.host);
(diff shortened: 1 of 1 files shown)


ShanniLi:
lgtm

node/peers.js
```diff
@@ -144,23 +147,33 @@ TChannelPeers.prototype.clear = function clear() {
...
         var names = Object.keys(self.channel.subChannels);
         for (var i = 0; i < names.length; i++) {
             var subChannel = self.channel.subChannels[names[i]];
-            delete subChannel.peers._map[hostPort];
+            subChannel.peers.delete(hostPort);
         }
     }
     delete self._map[hostPort];
+    var index = self._keys.indexOf(hostPort);
+    self._keys.splice(index, 1);
```
kriskowal: splice is expensive. need a pop-at-index to go fast.

kriskowal: actually, since the order is unimportant, we can:

```js
self._keys[index] = self._keys[self._keys.index.length - 1];
self._keys.length--;
```

Look Ma', no function calls and O(1).

node/peers.js
```diff
@@ -107,17 +109,18 @@ TChannelPeers.prototype.addPeer = function addPeer(peer) {
...
     if (peer.hostPort !== self.channel.hostPort) {
         self._map[peer.hostPort] = peer;
+        self._keys.push(peer.hostPort);
     }
 };
 
 TChannelPeers.prototype.keys = function keys() {
     var self = this;
-    return Object.keys(self._map);
+    return self._keys.slice();
```
kriskowal: Perhaps we should add `.fauxImmutableKeys()`.

kriskowal:
:ok: with recommendations for further improvement

node/peers.js
Raynos: delete() is not hot path so I'm going to keep the code readable for now.

PR #880, opened 2015-07-07
Operations cleanup for perf

Raynos:
Another chunk of perf work.

This generally moves some code out of connection and
into operations to avoid creating an anonymous
objects unless we really want to log them.

r: @kriskowal @rf @shannili

### DIFF
--- a/node/connection.js
+++ b/node/connection.js
@@ -251,7 +251,10 @@ TChannelConnection.prototype.onWriteError = function onWriteError(err) {
 
 TChannelConnection.prototype.onHandlerError = function onHandlerError(err) {
     var self = this;
-    self.resetAll(err);
+
+    if (err) {
+        self.resetAll(err);
+    }
 };
 
 TChannelConnection.prototype.handlePingResponse = function handlePingResponse(resFrame) {
@@ -273,14 +276,11 @@ TChannelConnection.prototype.handleReadFrame = function handleReadFrame(frame) {
 TChannelConnection.prototype.onCallResponse = function onCallResponse(res) {
     var self = this;
 
-    var called = false;
     var req = self.ops.getOutReq(res.id);
-
     if (res.state === States.Done || res.state === States.Error) {
-        popOutReq();
+        self.popOutReq(res);
     } else {
-        res.errorEvent.on(popOutReq);
-        res.finishEvent.on(popOutReq);
+        self.popOutReqLater(res);
     }
 
     if (!req) {
@@ -295,6 +295,15 @@ TChannelConnection.prototype.onCallResponse = function onCallResponse(res) {
     }
 
     req.emitResponse(res);
+};
+
+TChannelConnection.prototype.popOutReqLater =
+function popOutReqLater(res) {
+    var self = this;
+    var called = false;
+
+    res.errorEvent.on(popOutReq);
+    res.finishEvent.on(popOutReq);
 
     // TODO: move to method
     function popOutReq() {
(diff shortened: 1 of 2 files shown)


node/operations.js
```diff
@@ -46,8 +46,8 @@ function Operations(opts) {
         out: []
     };
     self.requests = {
-        in: {},
-        out: {}
+        in: Object.create(null),
```
ShanniLi: this is faster?

ShanniLi:
lgtm

node/connection.js
```diff
@@ -251,7 +251,10 @@ TChannelConnection.prototype.onWriteError = function onWriteError(err) {
 
 TChannelConnection.prototype.onHandlerError = function onHandlerError(err) {
     var self = this;
-    self.resetAll(err);
+
+    if (err) {
```
kriskowal: Under what circumstances does onHandlerError get called without an err?

node/connection.js
```diff
@@ -305,17 +314,17 @@ TChannelConnection.prototype.onCallResponse = function onCallResponse(res) {
...
         called = true;
-
-        self.ops.popOutReq(res.id, {
-            responseId: res.id,
-            code: res.code,
-            arg1: Buffer.isBuffer(res.arg1) ?
-                String(res.arg1) : 'streamed-arg1',
-            info: 'got call response for unknown id'
-        });
+        self.popOutReq(res);
```
kriskowal: Perhaps we should move this “called” state onto the prototype and perform the check in popOutReq instead, so that it is idempotent.

node/operations.js
kriskowal: no.

```
kris@krismacuber-9:~ ❯❯ node -e 'var start = Date.now(), eta = start + 10000, count = 0; while (Date.now() < eta) { for (var index = 0; index < 1e4; index++) Object.create(null); count += 1e4} console.log(count / (Date.now() - start));'
5001.496856601138
kris@krismacuber-9:~ ❯❯ node -e 'var start = Date.now(), eta = start + 10000, count = 0; while (Date.now() < eta) { for (var index = 0; index < 1e4; index++) Object.create(null); count += 1e4} console.log(count / (Date.now() - start));'
4883.697713886393
kris@krismacuber-9:~ ❯❯ node -e 'var start = Date.now(), eta = start + 10000, count = 0; while (Date.now() < eta) { for (var index = 0; index < 1e4; index++) ({}); count += 1e4} console.log(count / (Date.now() - start));'
299640.79026142485
kris@krismacuber-9:~ ❯❯ node -e 'var start = Date.now(), eta = start + 10000, count = 0; while (Date.now() < eta) { for (var index = 0; index < 1e4; index++) ({}); count += 1e4} console.log(count / (Date.now() - start));'
302802.1155573296
```

node/connection.js
Raynos: it was passed as a callback in places; so this method is a callback and thus err is optional

node/operations.js
kriskowal: However, with the truncated prototype chain, property lookups that miss are slightly faster.

```
kris@krismacuber-9:~ ❯❯ node -e 'var start = Date.now(), eta = start + 10000, count = 0; while (Date.now() < eta) { var o = Object.create(null); for (var index = 0; index < 1e4; index++) o[index]; count += 1e4} console.log(count / (Date.now() - start));'
24977.565061322166
kris@krismacuber-9:~ ❯❯ node -e 'var start = Date.now(), eta = start + 10000, count = 0; while (Date.now() < eta) { var o = Object.create(null); for (var index = 0; index < 1e4; index++) o[index]; count += 1e4} console.log(count / (Date.now() - start));'
24542.003795824592
kris@krismacuber-9:~ ❯❯ node -e 'var start = Date.now(), eta = start + 10000, count = 0; while (Date.now() < eta) { var o = {}; for (var index = 0; index < 1e4; index++) o[index]; count += 1e4} console.log(count / (Date.now() - start));'
21638.54218671992
kris@krismacuber-9:~ ❯❯ node -e 'var start = Date.now(), eta = start + 10000, count = 0; while (Date.now() < eta) { var o = {}; for (var index = 0; index < 1e4; index++) o[index]; count += 1e4} console.log(count / (Date.now() - start));'
20401.317759808328
```

Raynos: we should perf setting and getting properies on `Object.create(null)`;

creating ops is O(n) connections; accessing ops is O(n) requests.

kriskowal:
:ok:

node/connection.js
Raynos: the called boolean is O(n) call response; the conneciton is O(n) connection so it wont work.

PR #881, opened 2015-07-07
Cleanup tracing in in/out req.

Raynos:
This is another chunk of perf work.

We are removing a bunch of trace branching out of 
the host code of these constructors since relay
does not care about tracing.

r: @kriskowal @rf @shannili


ShanniLi:
lgtm

PR #882, opened 2015-07-07
out_response: move onFinish into an emitFinish method

Raynos:
This is another chunk of perf work. I'm simply inline an
event listener to get rid of a closure

r: @kriskowal @rf @shannili


ShanniLi:
lgtm

PR #883, opened 2015-07-07
v2/index: use a RetryFlags constructor

Raynos:
Creates a constructor for an anonymous object on the
hot code path.

r: @kriskowal @rf @shannili

### DIFF
--- a/node/v2/index.js
+++ b/node/v2/index.js
@@ -85,12 +85,19 @@ module.exports.parseRetryFlags = function parseRetryFlags(val) {
     var never = val.indexOf('n') > -1;
     var onConnectionError = !never && val.indexOf('c') > -1;
     var onTimeout = !never && val.indexOf('t') > -1;
-    return {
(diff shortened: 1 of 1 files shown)


ShanniLi:
lgtm

kriskowal:
My impression of this kind of change is that there is a hidden class for the options object regardless of whether you use an object literal or a constructor. You go faster only if code that uses the options object always receives the same hidden class (so it can compile down to accessing properties of a struct by consistent offsets). So, for this kind of change to be beneficial, there must be multiple sites where the same constructor is instantiated and code downstream must always receive that exact hidden class so it can remain optimized. Receiving a different hidden class would derail the optimization.

This optimization by itself should have no impact on performance.

kriskowal:
:ok: in any case since it’s forward progress toward a potential optimization.

Raynos:
@kriskowal The flames show speed up with a constructor.

PR #807, opened 2015-06-28
Node Performance. Making everything faster.

Raynos:
```
raynos at raynos-SVS15127PXB  ~/uber/tchannel/node on perf
$ node benchmarks/compare.js ./benchmarks/relay-master.json ./benchmarks/relay-perf.json 
GET 16KiB, 1000/0  rate: hi-diff:    503.04 ( 30.0%)
GET 16KiB, 10000/0 rate: hi-diff:    -55.12 ( -3.0%)
GET 16KiB, 20000/0 rate: hi-diff:    361.86 ( 22.2%)
GET 4B, 1000/0     rate: hi-diff:   1002.65 ( 50.5%)
GET 4B, 10000/0    rate: hi-diff:   1015.88 ( 52.8%)
GET 4B, 20000/0    rate: hi-diff:    649.46 ( 28.7%)
GET 4KiB, 1000/0   rate: hi-diff:    488.49 ( 22.9%)
GET 4KiB, 10000/0  rate: hi-diff:   -138.16 ( -7.3%)
GET 4KiB, 20000/0  rate: hi-diff:    559.32 ( 29.4%)
PING, 1000/0       rate: hi-diff:    560.39 ( 22.2%)
PING, 10000/0      rate: hi-diff:    523.75 ( 25.2%)
PING, 20000/0      rate: hi-diff:    506.19 ( 23.5%)
SET 16KiB, 1000/0  rate: hi-diff:    451.44 ( 24.5%)
SET 16KiB, 10000/0 rate: hi-diff:    329.22 ( 17.2%)
SET 16KiB, 20000/0 rate: hi-diff:    413.07 ( 23.1%)
SET 4B, 1000/0     rate: hi-diff:    903.13 ( 45.0%)
SET 4B, 10000/0    rate: hi-diff:    780.14 ( 39.2%)
SET 4B, 20000/0    rate: hi-diff:    608.95 ( 26.3%)
SET 4KiB, 1000/0   rate: hi-diff:    554.22 ( 26.9%)
SET 4KiB, 10000/0  rate: hi-diff:    187.90 (  9.2%)
SET 4KiB, 20000/0  rate: hi-diff:    724.21 ( 40.4%)
```

It's start

 - Optimizing all of the stats complexity
 - Purging tracing from the relay
 - Purging retries from the relay
 - Lots of little performance wins left & right
 - Improving the benchmarks scripts included automated torch

The tracing optimization is a big deal; I bet I get a 3x perf
boost in production on this branch (which should be enough
[...]

PR #884, opened 2015-07-07
v2/handler: remove unneeded callbacks

Raynos:
This one is another one from the perf book.

This one is more dubious; we are removing a callback from
the hot path and replacing the entire error path with a bunch
of event emission.

I'm not sure whether this is ideal (cc @jcorbin)

r: @kriskowal @rf @shannili

### DIFF
--- a/node/connection.js
+++ b/node/connection.js
@@ -262,14 +262,12 @@ TChannelConnection.prototype.handlePingResponse = function handlePingResponse(re
 
 TChannelConnection.prototype.handleReadFrame = function handleReadFrame(frame) {
     var self = this;
+
     if (!self.closing) {
         self.ops.lastTimeoutTime = 0;
     }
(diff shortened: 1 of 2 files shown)


anson627:
what perf book? I'm interested in reading it.

ShanniLi:
what does the benchmark say? cannot see why this makes much difference.

Raynos:
@ShanniLi we remove a closure allocation from the hot path.

Allocating closures is expensive; it improved the flames.

if you want to run the benchmarks you can but they take like 10 minutes each (one for master; one for this branch).

Raynos:
@anson627 

Sorry; i'm using weird speech; there is no perf book; this is just a subset of my perf branch.

The perf branch can be found here ( https://github.com/uber/tchannel/pull/807 ) and used to be 3000 loc change.

ShanniLi:
:+1:

Issue #856, opened 2015-07-06
Hyperbhan client should support de-advertise a service


anson627:
this is a feature request
xlate needs a way to stop advertising unhealthy services

ShanniLi:
done

Raynos:
This is not fully implemented.

We just have passive de-advertise which takes 5 minutes to occur instead of active de-advertise that is immediate.

Issue #679, opened 2015-06-10
Node: Peer not marked as unhealthy on connection errors


Raynos:
It looks like the peer does not transition into the unhealthy state if there are connection errors. It looks like it only transitions where there are errors with call requests.

ShanniLi:
done

Raynos:
When was this implemented ?

PR #887, opened 2015-07-07
Hyperbahn unadvertise

ShanniLi:
r @kriskowal @Raynos @rf
cc @anson627 

Reference issues: #856 #679

### DIFF
--- a/node/hyperbahn/index.js
+++ b/node/hyperbahn/index.js
@@ -368,9 +368,7 @@ function unadvertise() {
     self.destroy();
     self.latestAdvertisementResult = null;
     self.state = States.UNADVERTISED;
-    self.tchannel.close(function onClose() {
-        self.emit('unadvertised');
-    });
+    self.emit('unadvertised');
 };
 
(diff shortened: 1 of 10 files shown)


kriskowal:
:ok:

anson627:
lgtm. Thanks for the quick turnaround!

ShanniLi:
makes sense. I will fix it.

Raynos:
Finished review.

node/hyperbahn/index.js
```diff
@@ -358,6 +358,19 @@ function advertise(opts) {
     }
 };
 
+HyperbahnClient.prototype.unadvertise =
+function unadvertise() {
+    var self = this;
+    if (self.state !== States.ADVERTISED) {
+        return;
```
Raynos: you dont want to early return; always close the timer.

Just because `state === UNADVERTISED` does not mean that the timer is gone. because it could be trying in a loop

Raynos:
lgtm.

PR #888, opened 2015-07-07
go: Add simple server/client using raw for benchmarking

prashantv:
r @mmihic

### DIFF
--- a/golang/examples/bench/server/server.go
+++ b/golang/examples/bench/server/server.go
@@ -1,0 +1,95 @@
+package main
+
+import (
+	"errors"
+	"flag"
+	"log"
+	"net/http"
+	_ "net/http/pprof"
+	"sync"
+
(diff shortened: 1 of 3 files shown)


golang/examples/bench/server/server.go
```diff
@@ -0,0 +1,95 @@
...
+
+	ch, err := tchannel.NewChannel("bench-server", &tchannel.ChannelOptions{
+		ProcessName: "bench-server",
+	})
+	if err != nil {
+		log.Fatalf("NewChannel failed: %v", err)
+	}
+
+	handler := raw.Wrap(&kvHandler{vals: make(map[string]string)})
+	ch.Register(handler, "get")
```
Raynos: can you implement ping as well to have compat with the node bench server ?

prashantv: Will do in a separate PR.

PR #889, opened 2015-07-08
Use a constructor function for the options for InRequest

Raynos:
This gets rid of a large anonymous object.

r: @kriskowal @rf @shannili


ShanniLi:
lgtm

kriskowal:
:ok:

PR #890, opened 2015-07-08
Create an OutRequest options constructor

Raynos:
This is the most significant change after the stats
change in the perf branch. This is also the last PR
related to the perf branch

 - Use a constructor for out request options; this should
    give a lot more stability
 - Optimize the request defaults code path by avoiding
    copies and unrolling the loop.

r: @shannili @rf @kriskowal

### DIFF
--- a/node/as/http.js
+++ b/node/as/http.js
@@ -168,12 +168,11 @@ TChannelHTTP.prototype.forwardToTChannel = function forwardToTChannel(tchannel, 
     // - driving peer selection manually therefore
     // TODO: more http state machine integration
 
-    var treq;
     var options = tchannel.requestOptions(extendInto({
         streamed: true,
         hasNoParent: true
     }, tchannelRequestOptions));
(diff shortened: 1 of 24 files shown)


ShanniLi:
lgtm

node/as/http.js
```diff
@@ -181,17 +180,20 @@ TChannelHTTP.prototype.forwardToTChannel = function forwardToTChannel(tchannel,
...
-                self.sendRequest(treq, hreq, forwarded);
-            });
-        } else {
-            treq = peer.request(options);
-            self.sendRequest(treq, hreq, forwarded);
+        peer.waitForIdentified(onIdentified);
+        return null;
+    }
+
+    function onIdentified(err) {
```
kriskowal: long closure

Raynos: The entire as layer is unoptimized and will be improved in a seperate PR

PR #893, opened 2015-07-08
Improve benchmarks to be more like prod

Raynos:
Another attempt at making the local flames and the
production flames look the same

 - Use a real statsd client
 - Use N instances rather then 1.


ShanniLi:
:+1:

kriskowal:
Neat. :ok:

Issue #894, opened 2015-07-08
set code to non-zero when call responses fail


blampe:
We currently always set it to zero.

junchaowu:
This should be user's behavior. Only user can decide this is a failed call response

blampe:
It happens even when the app raises a thrift exception, though.

junchaowu:
On CR now.

PR #895, opened 2015-07-08
Set response code to error when exceptions happen in thrift server

junchaowu:
@blampe @abhinav @breerly 
Fix https://github.com/uber/tchannel/issues/894#issuecomment-119755493

### DIFF
--- a/python/tchannel/thrift/client.py
+++ b/python/tchannel/thrift/client.py
@@ -22,6 +22,7 @@ from __future__ import absolute_import
 
 import inspect
 from collections import namedtuple
+from tchannel.tornado.response import StatusCode
 
 from thrift import Thrift
 from tornado import gen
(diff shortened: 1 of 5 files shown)


blampe:
Needs a test.

junchaowu:
It has been tested. If the code is incorrect , the test_thrift_exception test won't pass

blampe:
If you make code changes without touching a test, then none of our tests were asserting anything about the behavior you changed...

junchaowu:
No. Take a look at the test_thrift_exception test. Previously client code doesn't check the response code that's why the test will pass even if we don't set right response code. Now I add it. So if client throws an thrift exception, it must get 'error' response code.

blampe:
:+1: thanks for the regression test @abhinav

PR #897, opened 2015-07-09
Rate limiting

ShanniLi:
r @kriskowal @Raynos @rf 

Perf: before V.S. after
GET 16KiB, 1000/0  rate: hi-diff:    231.06 (  5.2%)
GET 16KiB, 10000/0 rate: hi-diff:    129.31 (  3.0%)
GET 16KiB, 20000/0 rate: hi-diff:     83.11 (  2.0%)
GET 4B, 1000/0     rate: hi-diff:    178.19 (  2.3%)
GET 4B, 10000/0    rate: hi-diff:    309.22 (  4.3%)
GET 4B, 20000/0    rate: hi-diff:    317.73 (  4.4%)
GET 4KiB, 1000/0   rate: hi-diff:    224.95 (  3.8%)
GET 4KiB, 10000/0  rate: hi-diff:    778.98 ( 15.7%)
GET 4KiB, 20000/0  rate: hi-diff:    252.76 (  4.6%)
PING, 1000/0       rate: hi-diff:    224.51 (  2.8%)
PING, 10000/0      rate: hi-diff:    463.20 (  6.3%)
PING, 20000/0      rate: hi-diff:    126.96 (  1.7%)
SET 16KiB, 1000/0  rate: hi-diff:      6.31 (  0.1%)
SET 16KiB, 10000/0 rate: hi-diff:     95.85 (  2.4%)
SET 16KiB, 20000/0 rate: hi-diff:   -109.33 ( -3.1%)
SET 4B, 1000/0     rate: hi-diff:    440.34 (  5.9%)
SET 4B, 10000/0    rate: hi-diff:    -19.97 ( -0.3%)
SET 4B, 20000/0    rate: hi-diff:    446.07 (  6.4%)
SET 4KiB, 1000/0   rate: hi-diff:   -366.18 ( -5.5%)
SET 4KiB, 10000/0  rate: hi-diff:     87.48 (  1.6%)
SET 4KiB, 20000/0  rate: hi-diff:   -400.05 ( -8.2%)


node/hyperbahn/rate_limiting.js
```diff
@@ -0,0 +1,187 @@
...
+    // check total QPS
+    self.incrementCounter(TOTAL_QPS);
+    var limit = self.shouldRateLimit(TOTAL_QPS);
+    if (limit) {
+        buildRes().sendError('Busy', 'hyperbahn node is rate limited by the total rate ' + limit + ' qps');
+        return true;
+    }
+
+    // check QPS for serviceName
+    if (!self.isExitFor(req.serviceName)) {
```
kriskowal: It should not be necessary to couple this feature to Hyperbahn. In hyperbahn/service_proxy, you can conditionally add a rate limiter to the request handler chain for a particular service.

You do still need to delete tracking on membership change events, and can perform that with a hook.

node/hyperbahn/rate_limiting.js
```diff
@@ -0,0 +1,187 @@
+// Copyright (c) 2015 Uber Technologies, Inc.
```
kriskowal: Like a relay_handler, a rate_limiting_handler is not necessarily coupled to Hyperbahn. Let’s move this into the tchannel/node directory. Let’s also break this into two parts, RateLimiter and RateLimitingHandler, both decoupled from Hyperbahn. It may not even be necessary to look up RateLimiter by service if there is one RateLimiter and RateLimitingHandler per service proxy.

Breaking RateLimiter out will make it easier to experiment, test, and reuse. RateLimiters could be used in other scenarios.

node/hyperbahn/rate_limiting.js
```diff
@@ -0,0 +1,187 @@
...
+// THE SOFTWARE.
+
+'use strict';
+
+var assert = require('assert');
+
+var DEFAULT_SERVICE_QPS_LIMIT = 500;
+var DEFAULT_TOTAL_QPS_LIMIT = 1000;
+var DEFAULT_BUCKETS = 20;
+var TOTAL_QPS = '~~TotalQPS++';
```
kriskowal: Please leave a note that TOTAL_QPS is a sentinel service name. It would be better to just have a RateLimiter object that can be used regardless of the key (serviceName, peer, global), breaking RateLimiting into RateLimiter and RateLimitingHandler (which would use both the global.isBusy() and rateLimiterForService[serviceName].isBusy()).

Implementing the rate limiter as a unit with just an isBusy(), and handleRequest() methods will allow us to more easily compare various approaches with a benchmark harness. Perhaps also create RateLimiterOptions constructor.

node/hyperbahn/rate_limiting.js
```diff
@@ -0,0 +1,187 @@
...
+        return new RateLimiting(options);
+    }
+    var self = this;
+
+    assert(options.serviceDispatchHandler, 'serviceDispatchHandler is required for RateLimiting');
+    self.serviceDispatchHandler = options.serviceDispatchHandler;
+    self.egressNodes = options.serviceDispatchHandler.egressNodes;
+    self.timers = self.serviceDispatchHandler.channel.timers;
+
+    self.qpsLimits = options.qpsLimits || Object.create(null);
```
kriskowal: The Q in QPS is cargo from another domain. We deal in requests here, not queries. Perhaps we should name and discuss these metrics by the names in our domain.

node/hyperbahn/rate_limiting.js
```diff
@@ -0,0 +1,187 @@
...
+        counter.buckets[counter.index] = 0;
+        if (serviceName === TOTAL_QPS) {
+            // total QPS limit never change
+            continue;
+        }
+        var limit = self.qpsLimits[serviceName] || self.defaultServiceQpsLimit;
+        counter.qpsLimit = limit / self.egressNodes.kValueFor(serviceName);
+    }
+
+    self.refreshTimer = self.timers.setTimeout(
```
kriskowal: If we can find a way to reduce the use of timers, we should. I’m sure we can minimize the use of timers.now() or timers.hrnow() and eliminate the use of setTimeout. Also confident we can get away without having multiple buckets for each measurement.

node/hyperbahn/rate_limiting.js
```diff
@@ -0,0 +1,187 @@
...
+        }
+        assert(counter.qps >= 0, 'qps shouls always be larger equal to 0');
+        counter.index = next;
+        counter.buckets[counter.index] = 0;
+        if (serviceName === TOTAL_QPS) {
+            // total QPS limit never change
+            continue;
+        }
+        var limit = self.qpsLimits[serviceName] || self.defaultServiceQpsLimit;
+        counter.qpsLimit = limit / self.egressNodes.kValueFor(serviceName);
```
kriskowal: We do not guarantee an even distribution of RPS on every node of a domain. The max pending limit and pending request based routing ensure that each node receives load in proportion to its rate of response.

I believe that this feature is intended as a fail-safe, not for enforcing SLA’s. To enforce SLA’s, @mrbrowning, @mranney and I can bring you up to speed with the ideas we’ve had for gossiping CN:SN RPS metrics.

node/hyperbahn/rate_limiting.js
```diff
@@ -0,0 +1,187 @@
...
+function handleRequest(req, buildRes) {
+    var self = this;
+    assert(req.serviceName, 'serviceName should not be empty for rate limiting.');
+
+    if (req.serviceName === 'hyperbahn') {
+        return false;
+    }
+
+    // check total QPS
+    self.incrementCounter(TOTAL_QPS);
```
kriskowal: Again, decouple. Confusing to pass a service name to an increment function. Definitely add "service name" to the name of the TOTAL_QPS variable if you decide to keep it anyway.

node/test/hyperbahn/rate-limiting.js
```diff
@@ -0,0 +1,260 @@
...
+                            'should be rate limited');
+                        done();
+                    });
+                }
+            ], function done() {
+                steveHyperbahnClient.destroy();
+                assert.end();
+            });
+        }
+    });
```
kriskowal: Perhaps add a negative test to verify that the rate limiter does not engage when (a) there is smooth and steady load just beneath maxRequestsPerMs (b) there is bursty load just beneath maxRequestsPerMs on average over the window.

node/test/lib/relay_network.js
```diff
@@ -55,6 +55,11 @@ function RelayNetwork(options) {
 
     self.servicePurgePeriod = options.servicePurgePeriod;
 
+    self.qpsLimits = options.qpsLimits;
```
kriskowal: Consider maxRpmsForServiceName.

node/hyperbahn/rate_limiting.js
```diff
@@ -0,0 +1,187 @@
...
+'use strict';
+
+var assert = require('assert');
+
+var DEFAULT_SERVICE_QPS_LIMIT = 500;
+var DEFAULT_TOTAL_QPS_LIMIT = 1000;
+var DEFAULT_BUCKETS = 20;
+var TOTAL_QPS = '~~TotalQPS++';
+
+function RateLimiting(options) {
```
kriskowal: s/RateLimiting/RateLimiter/

node/hyperbahn/rate_limiting.js
```diff
@@ -0,0 +1,187 @@
...
+    if (!(this instanceof RateLimiting)) {
+        return new RateLimiting(options);
+    }
+    var self = this;
+
+    assert(options.serviceDispatchHandler, 'serviceDispatchHandler is required for RateLimiting');
+    self.serviceDispatchHandler = options.serviceDispatchHandler;
+    self.egressNodes = options.serviceDispatchHandler.egressNodes;
+    self.timers = self.serviceDispatchHandler.channel.timers;
```
kriskowal: Rather than reaching into these deep properties (tight coupling), we could require timers etc to be on the options object proper.

node/hyperbahn/rate_limiting.js
```diff
@@ -0,0 +1,187 @@
...
+    self.refresh();
+}
+
+RateLimiting.prototype.type = 'tchannel.hyperbahn.rate-limiting';
+
+RateLimiting.prototype.refresh =
+function refresh() {
+    var self = this;
+
+    // update hte sliding window
```
kriskowal: s/hte/the/

node/hyperbahn/rate_limiting.js
```diff
@@ -0,0 +1,187 @@
...
+    var services = Object.keys(self.counters);
```
kriskowal: s/service/serviceNames/

node/hyperbahn/rate_limiting.js
```diff
@@ -0,0 +1,187 @@
...
+    for (var i = 0; i < services.length; i++) {
+        var serviceName = services[i];
+        var counter = self.counters[serviceName];
+        var next = (counter.index + 1) % self.buckets;
+        if (counter.buckets[next]) {
+            // offset the bucket being moved out
+            counter.qps -= counter.buckets[next];
+        }
+        assert(counter.qps >= 0, 'qps shouls always be larger equal to 0');
```
kriskowal: s/shouls/should/

node/hyperbahn/rate_limiting.js
```diff
@@ -0,0 +1,187 @@
...
+        counter.index = next;
+        counter.buckets[counter.index] = 0;
+        if (serviceName === TOTAL_QPS) {
+            // total QPS limit never change
```
kriskowal: s/change/changes/

node/hyperbahn/rate_limiting.js
```diff
@@ -0,0 +1,187 @@
...
+function createCounter(serviceName) {
+    var self = this;
+
+    var counter = Object.create(null);
+    counter.index = 0;
+    counter.qps = 0;
+    counter.buckets = [];
+    counter.buckets.length = self.buckets;
+    counter.buckets[0] = 0;
+    if (serviceName === TOTAL_QPS) {
```
kriskowal: decouple the qps limit from service name.

node/hyperbahn/rate_limiting.js
```diff
@@ -0,0 +1,187 @@
...
+function shouldRateLimit(serviceName) {
+    var self = this;
+
+    var counter = self.counters[serviceName];
+    assert(counter, 'cannot find counter for ' + serviceName);
+    return counter.qps <= counter.qpsLimit ? 0 : counter.qpsLimit;
+};
+
+RateLimiting.prototype.handleRequest =
+function handleRequest(req, buildRes) {
```
kriskowal: handleRequest should not and doesn’t need to have a return value.

kriskowal:
Great work and great velocity. Changes requested.

Also, perhaps reframe the tests as tests against an arbitrary RateLimiter instance and build scenarios around just that.

node/hyperbahn/rate_limiting.js
ShanniLi: Actually, I think rate limiting should be part of hyperbahn since tchannel is just a RPC call framework and should not have anything to do with rate limiting. @Raynos opinions?

node/hyperbahn/rate_limiting.js
ShanniLi: Same as the comment above. I believe both relay and rate limiting should be part of hyperbahn.

node/hyperbahn/rate_limiting.js
ShanniLi: I just did a search on Google. QPS seems be a well accepted concept for web services and bench marks. Actually, I am OK either way. I will make the change: QPS ==> RPS.

node/hyperbahn/rate_limiting.js
ShanniLi: not sure why though. I am not convinced why having a timer fires 20 times per seconds is slower than call timers.now() thousands of times per seconds. Actually, frequent calls to timers.now() is an anti-pattern in system programming since reading timers are expensive.

node/hyperbahn/rate_limiting.js
ShanniLi: This is best we can do with the information we have at egress nodes. @Raynos can we do better?

kriskowal: I suggest just picking a max throughput that does not need to be scaled or parameterized.

node/hyperbahn/rate_limiting.js
jcorbin: :+1: @kriskowal , please move this out of hyperbahn, it's a general tchannel concern

node/hyperbahn/rate_limiting.js
jcorbin: :+1: @kriskowal , let's keep this dead simple at first, simple static value

node/hyperbahn/rate_limiting.js
```diff
@@ -0,0 +1,187 @@
...
+
+RateLimiting.prototype.createCounter =
+function createCounter(serviceName) {
+    var self = this;
+
+    var counter = Object.create(null);
+    counter.index = 0;
+    counter.qps = 0;
+    counter.buckets = [];
+    counter.buckets.length = self.buckets;
```
jcorbin: uninitiazilezd / undefined new buckets? can we just initialize this with a for loop please?

node/hyperbahn/rate_limiting.js
```diff
@@ -0,0 +1,187 @@
...
+    counter.buckets[0] = 0;
+    if (serviceName === TOTAL_QPS) {
+        counter.qpsLimit = self.totalQpsLimit;
+    } else {
+        var limit = self.qpsLimits[serviceName] || self.defaultServiceQpsLimit;
+        counter.qpsLimit = limit / self.egressNodes.kValueFor(serviceName);
```
jcorbin: drop the scaling

jcorbin:
Good start, but + :100: to decoupling it from hyperbahn for better reuse, composability, and extensability; of which better testing, modeling, etc are one consequent.

In general we prefer to minimize the code in hyperbahn as much as possible and to build it within the tchannel library as simply and loosely coupled as possible.

node/hyperbahn/rate_limiting.js
ShanniLi: Discussed with @kriskowal and @jcorbin offline. It makes sense to make rate limiting generic. I will make the change.

node/hyperbahn/rate_limiting.js
ShanniLi: it is actually guaranteed to be written first before read. I will add the for loop anyway. :)

node/hyperbahn/service_proxy.js
```diff
@@ -84,6 +92,26 @@ function handleRequest(req, buildRes) {
         return;
     }
 
+    // apply rate limiter
+    if (req.serviceName !== 'hyperbahn') {
```
Raynos: Let's apply rate limiting for all services.

We will configure the rate for `hyperbahn`, `ringpop` & `autobahn` seperately.

node/hyperbahn/service_proxy.js
```diff
@@ -84,6 +92,26 @@ function handleRequest(req, buildRes) {
...
+        var isExitNode = self.isExitFor(req.serviceName);
+        self.rateLimiter.incrementTotalCounter();
+        if (isExitNode) {
+            self.rateLimiter.incrementServiceCounter(req.serviceName);
+        }
+
+        if (self.rateLimiter.shouldRateLimitTotalRequest()) {
+            buildRes().sendError('Busy', 'hyperbahn node is rate limited by its total rps');
```
Raynos: Can we put the current total counter in the message.

node/hyperbahn/service_proxy.js
```diff
@@ -84,6 +92,26 @@ function handleRequest(req, buildRes) {
...
+            return;
+        }
+
+        // check RPS for service limit
+        if (isExitNode && self.rateLimiter.shouldRateLimitService(req.serviceName)) {
+            buildRes().sendError('Busy', req.serviceName + ' is rate limited');
```
Raynos: Can we put the current per service counter in the message.

node/hyperbahn/service_proxy.js
```diff
@@ -248,6 +276,7 @@ function updateServiceChannel(svcchan) {
     } else if (!isExit) {
         if (svcchan.serviceProxyMode === 'exit') {
             self.changeToForward(exitNodes, svcchan);
+            self.rateLimiter.removeServiceCounter(svcchan.serviceName);
```
Raynos: Let's put this in `purgeServices()` next to `delete subChannels[serviceName]`

node/test/rate-limiter.js
```diff
@@ -0,0 +1,159 @@
...
+            done();
+        },
+        wait,
+        increment.bind(null, rateLimiter, 'steve', 'bob'),
+        function check2(done) {
+            assert.equals(rateLimiter.totalRequestCounter.rps, 3, 'check2: total request');
+            assert.equals(rateLimiter.counters.steve.rps, 2, 'check2: request for steve');
+            assert.equals(rateLimiter.counters.bob.rps, 1, 'check2: request for bob');
+            done();
+        }], function done() {
```
Raynos: We can place a new line after `}` so we can decrement `done` by one indent

```js
series([
    ...
    }
], function done() { 
    ...
});
```

node/test/hyperbahn/rate-limiter.js
```diff
@@ -0,0 +1,258 @@
...
+                }
+            ], function done() {
+                steveHyperbahnClient.destroy();
+                assert.end();
+            });
+        }
+    });
+
+    HyperbahnCluster.test('total rate limiting works', {
+        size: 1,
```
Raynos: We are going to need to write tests for a size larger then 1

node/hyperbahn/service_proxy.js
Raynos: Rate limiting itself should be behind a flag; we should also have a method for toggling it at run time.

We need to be able to write an admin script to control this.

node/rate_limiter.js
```diff
@@ -0,0 +1,168 @@
...
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+'use strict';
+
+var assert = require('assert');
+
+var DEFAULT_SERVICE_RPS_LIMIT = 5000;
```
Raynos: Let's take a lower limit; like 100.

node/rate_limiter.js
```diff
@@ -0,0 +1,168 @@
...
+var DEFAULT_TOTAL_RPS_LIMIT = 20000;
```
Raynos: We want a lower limit; like 1000.

node/rate_limiter.js
```diff
@@ -0,0 +1,168 @@
...
+
+RateLimiter.prototype.refresh =
+function refresh() {
+    var self = this;
+
+    self.udpateCounter(self.totalRequestCounter);
+
+    var serviceNames = Object.keys(self.counters);
+    for (var i = 0; i < serviceNames.length; i++) {
+        self.udpateCounter(self.counters[serviceNames[i]]);
```
Raynos: `update` typo

node/rate_limiter.js
```diff
@@ -0,0 +1,168 @@
...
+    assert(counter.rps >= 0, 'rps should always be larger equal to 0');
+    counter.index = next;
+    counter.buckets[counter.index] = 0;
+};
+
+RateLimiter.prototype.createCounter =
+function createCounter(rpsLimit) {
+    var self = this;
+
+    var counter = Object.create(null);
```
Raynos: Please create a prototype for `RateLimiterCounter`

Raynos: As an aside; do not use `Object.create(null)` for anything other then dictionaries

node/rate_limiter.js
```diff
@@ -0,0 +1,168 @@
...
+    counter.index = 0;
+    counter.rps = 0;
+    counter.buckets = [];
+    counter.buckets.length = self.buckets;
+    // counter.buckets is read/written in udpateCounter,
+    // where read is always after write on a bucket.
+    counter.buckets[0] = 0;
+    counter.rpsLimit = rpsLimit;
+
+    // if (serviceName === TOTAL_QPS) {
```
Raynos: remove commented out code.

node/hyperbahn/service_proxy.js
Raynos: This is a rare event; let's add a log call for busy-ness

node/hyperbahn/service_proxy.js
Raynos: This is a rare event; let's add a log call for busy-ness

Raynos:
Finished review.

node/rate_limiter.js
ShanniLi: This is fine here. In hyperbahn, I am going to pass in the real low limits as default.

Raynos: ... That's not fine. tchannel should match production rather then having unrealistic defaults.

Unrealistic defaults are a waste of everyone's time.

Raynos:
Still needs a little more work:

 - [ ] Ratelimiting itself needs to be behind a flag; preferably with methods for turning on & off just like the kill switches.
 - [ ] We need tests for cluster larger then size 1. This can be either here or in autobahn itself.

ShanniLi:
@Raynos 
First comment fixed.
Second comment is already implemented in test/hyperbahn/rate-limiter.js
I will also add integration test in hyperbahn.

Raynos:
lgtm.

node/rate_limiter.js
```diff
@@ -0,0 +1,173 @@
...
+};
+
+RateLimiterCounter.prototype.increment =
+function increment() {
+    var self = this;
+    self.buckets[self.index] += 1;
+    self.rps += 1;
+};
+
+function RateLimiter(options) {
```
kriskowal: I am going to say keep this design for now, but note that it limits the flexibility of the interface by coupling the logic of rate limiting to an index of rate limiters by service and by total. Alternately, you could have a RateLimiter instance per service (indexed externally by service name) and another rate limiter instance for the total. This would make it possible to use the rate limiter logic for other scenarios, e.g., indexed by host:port, which we may eventually need to penalize ephemeral clients, or just instantiated independently for testing, benchmarking, and A/B comparisons in isolation.

As @Raynos has observed in other places, even with an instance of a RateLimiter for each service, we would still want a collection of RateLimiters indexed by service to consolidate all of the book keeping under a single timer (as you have with this design).

I say to not make any changes yet because this design may prove better because it has fly-weight rate limiters for each service, reducing memory usage and possibly increasing cache locality. A more flexible design may have a performance cost.

That said, I insist that there is still room to experiment with an alternate design that eliminates the timers and reduces memory usage by using a single statistic per rate limiter and adjusting that statistic on the leading or trailing edge of some sample of requests.

node/rate_limiter.js
```diff
@@ -0,0 +1,173 @@
...
+    if (!(this instanceof RateLimiterCounter)) {
+        return new RateLimiterCounter(options);
+    }
+
+    var self = this;
+    self.index = 0;
+    self.rps = 0;
+    self.numOfBuckets = options.numOfBuckets;
+    self.buckets = [];
+    self.buckets.length = self.numOfBuckets;
```
kriskowal: This is unnecessary turns the buckets array instance into a sparse array. I have seen claims that this turns the array into a hashtable internally, which probably performs poorly in comparison.

- http://www.html5rocks.com/en/tutorials/speed/v8/
- http://thibaultlaurens.github.io/javascript/2013/04/29/how-the-v8-engine-works/

node/rate_limiter.js
```diff
@@ -0,0 +1,173 @@
...
+    // update the sliding window
+    var next = (self.index + 1) % self.numOfBuckets;
+    if (self.buckets[next]) {
+        // offset the bucket being moved out
+        self.rps -= self.buckets[next];
+    }
+
+    assert(self.rps >= 0, 'rps should always be larger equal to 0');
+    self.index = next;
+    self.buckets[self.index] = 0;
```
kriskowal: Note that this will still work, even for a densely packed array, when self.index === self.buckets.length, without turning it into a hashtable, making it unnecessary to declare the array size up front, or “pre-allocating” a memory region by assigning to the array length.

docs/rate-limiting.md
```diff
@@ -0,0 +1,81 @@
+# Rate Limiting
+
+Rate limiting controls the rate of requests that can be forwarded by a Hyperbahn
+process. The motivation behind is that a service should never stress Hyperbahn in
+such an unexpected way that the quality of other services gets affected.
```
kriskowal: such a way that it interferes with other services.

docs/rate-limiting.md
```diff
@@ -0,0 +1,81 @@
...
+In general, rate limiting can be considered as a firewall between services.
+
+
+## Goals
+
+Hyperbahn serves as a communication bus for services. In practice, a service
+can have a surge of  requests due to various of reasons such as sudden increase of
+demands or failures/retries. When such events occur, these requests may saturate the 
+hyperbahn nodes and cause a service quality degradation.
```
kriskowal: Capitalize pronoun Hyperbahn. Please review other occurrences.

docs/rate-limiting.md
```diff
@@ -0,0 +1,81 @@
...
+
+There are two goals of rate limiting:
+
+* Control the RPS (Request Per Second) that a service can have on a hyperbahn node.
+* Control the total RPS that a hyperbahn node should handle.
+
+Non-goals:
+
+* Rate limiting is not used to limit the healthy growth of a service's requests.
```
kriskowal: Rate limiting is not for managing the growth of a service’s usage.

docs/rate-limiting.md
```diff
@@ -0,0 +1,81 @@
...
+* Rate limiting is not [Circuit Breaking](./circuit-breaking.md). It treats all
+  requests the same.
+* Rate limiting is not SLA (Service Level Agreement). It only guarantees to control
```
kriskowal: Rate limiting is not for enforcing SLAs (Service Level Agreements). Plural agreement.

docs/rate-limiting.md
```diff
@@ -0,0 +1,81 @@
...
+
+
+## Design & Implemenation
+
+### The rate
+We use RPS (Request Per Second) to quantify the request rate. At any point of time,
+RPS is the total number of requests processed in the past second. There are various
+ways of counting RPS. We use a sliding window to maintain the RPS counter. As shown below,
+if we divide a second into five buckets, each bucket contains the number of requests during
+that 200ms slot. When time goes on the window moves forward such that the oldest bucket is
```
kriskowal: When time goes on, (comma)

node/benchmarks/relay_server.js
```diff
@@ -76,6 +76,8 @@ function RelayServer(opts) {
     });
     self.relay.handler = ServiceProxy({
         channel: self.relay,
+        totalRpsLimit: 30000,
+        defaultServiceRpsLimit: 30000,
```
kriskowal: Might make a comment that this is 50% over our design goal to avoid interfering with the benchmark.

kriskowal: Infinity is a number in JavaScript and probably better conveys the spirit of this configuration.

kriskowal:
We need something akin to a benchmark for this feature so we can measure the behavior of the rate limiter in response to different idle intervals and burst frequencies and durations. Perhaps open an issue.

kriskowal:
:ok: All comments advisory.

node/rate_limiter.js
ShanniLi: @kriskowal agreed :)

node/rate_limiter.js
ShanniLi: in this case, it is no sparse, the array is used by counter and will be filled within 1 sec.

kriskowal: This line makes V8 turn the array implementation into a hashmap. Does not matter that you fill it in in one second. V8 will have to convert between hashmap and flat array behind the scenes between when you execute this line and one second later, if it even has the logic to detect going from sparse back to compact. Delete this line and it will just go fast. It does not help.

node/rate_limiter.js
ShanniLi: fixed.

node/benchmarks/relay_server.js
ShanniLi: actually, it is not necessary anymore. I removed it.

PR #900, opened 2015-07-09
Node: make connection metrics optional

Raynos:
This PR allows you to turn connection metrics off. Specifically
the bytes recvd/sent and the gauge are expensive.

r: @kriskowal @shannili @rf

### DIFF
--- a/node/v2/handler.js
+++ b/node/v2/handler.js
@@ -220,15 +220,17 @@ function emitBytesRecvd(frame) {
     var self = this;
 
     var channel = self.connection.channel;
-    channel.emitFastStat(channel.buildStat(
-        'connections.bytes-recvd',
-        'counter',
-        frame.size,
-        new ConnectionsBytesRcvdTags(
-            channel.hostPort || '0.0.0.0:0',
-            self.connection.socketRemoteAddr
-        )
-    ));
+    if (channel.emitConnectionMetrics) {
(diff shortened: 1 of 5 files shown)


node/v2/handler.js
```diff
@@ -212,16 +212,25 @@ TChannelV2Handler.prototype.handleCallRequest = function handleCallRequest(reqFr
...
-        new ConnectionsBytesRcvdTags(
-            channel.hostPort || '0.0.0.0:0',
-            self.connection.socketRemoteAddr
-        )
-    ));
+    self.emitBytesRecvd(reqFrame);
+};
 
+TChannelV2Handler.prototype.emitBytesRecvd =
+function emitBytesRecvd(frame) {
```
kriskowal: Is there precedent for compressing "Received" to "Recvd" (I mean, if we ignore C libraries)?

ShanniLi:
lgtm. Good change! Connection metrics seem not that critical in practice anyway.

kriskowal:
:ok:

node/v2/handler.js
Raynos: it's `bytes-recvd` which is the stat name; i don't care what we call this method.

kriskowal: Alright, I’ll shrug.

PR #906, opened 2015-07-10
Provide access to transport metadata in Thrift request handlers.

### LINKED ISSUE #904: Make caller name available in server-side request object for Thrift
Protocol headers are available in raw requests so caller name is accessible there. 

Thrift requests use a different object type. `headers` there refers to the application headers, not the protocol headers. We should make the caller name available as a separate property in there.

abhinav:
Resolves #904.

CC @blampe @breerly @junchaowu @Raynos

### DIFF
--- a/python/CHANGES.rst
+++ b/python/CHANGES.rst
@@ -4,9 +4,11 @@ Changelog for tchannel.py
 0.9.2 (unreleased)
 ------------------
 
-- Add exponential backoff to `TChannel.advertise`
+- Add exponential backoff to ``TChannel.advertise``.
+- Make transport metadata available under ``request.transport`` on the
+  server-side.
 
-  
+
 0.9.1 (2015-07-09)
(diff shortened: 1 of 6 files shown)


python/GUIDE.rst
```diff
@@ -96,6 +99,7 @@ will register handlers against. Open up ``keyvalue/server.py`` and write
 something like this:
 
 .. code-block:: python
+
```
blampe: Thanks :)

blampe:
:+1:

python/CHANGES.rst
```diff
@@ -4,9 +4,11 @@ Changelog for tchannel.py
 0.9.2 (unreleased)
```
breerly: This should be `0.10.0`.

breerly:
Really happy with this :+1:

PR #908, opened 2015-07-10
Allow pulling out new IncomingCall object by a custom key.

pengzhai:
The main purpose is to provide a way to stub out the IncomingCall object in unit tests.

Main changes include:
- Add a new method to allow custom key for pulling out the IncomingCall object.
- Add unit tests.

### DIFF
--- a/golang/context_test.go
+++ b/golang/context_test.go
@@ -24,7 +24,7 @@ func TestWrapContextForTest(t *testing.T) {
 	ctx, cancel := NewContext(time.Second)
 	defer cancel()
 	actual := WrapContextForTest(ctx, call)
-	assert.Equal(t, call, actual.Value(contextKeyCall), "Incorrect call object returned.")
+	assert.Equal(t, call, CurrentCall(actual), "Incorrect call object returned.")
 }
 
 func TestCurrentCallWithNilResult(t *testing.T) {
--- a/golang/context_test.go
+++ b/golang/context_test.go
@@ -1,0 +1,45 @@
+package tchannel_test
+
+import (
+	"testing"
+	"time"
+
+	. "github.com/uber/tchannel/golang"
+
+	"github.com/stretchr/testify/assert"
+	"golang.org/x/net/context"
+)
+
+type mockIncomingCall struct {
+	callerName string
+}
+
+func (mic *mockIncomingCall) CallerName() string {
+	return mic.callerName
+}
+
+var (
+	cn   = "hello"
+	key1 = "foo"
+	key2 = "bar"
+)
+
+func TestCurrentCallWithMatchingKey(t *testing.T) {
+	expected := mockIncomingCall{callerName: cn}
+	ctx, cancel := NewContext(time.Second)
+	defer cancel()
+	ctx = context.WithValue(ctx, key1, &expected)
+
+	actual := CurrentCallWithKey(ctx, key1)
+
+	assert.Equal(t, &expected, actual)
+}
+
+func TestCurrentCallWithoutMatchingKey(t *testing.T) {
+	expected := mockIncomingCall{callerName: cn}
+	ctx, cancel := NewContext(time.Second)
+	defer cancel()
+	ctx = context.WithValue(ctx, key1, &expected)
+
+	assert.Nil(t, CurrentCallWithKey(ctx, key2))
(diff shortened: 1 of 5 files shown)


pengzhai:
Not sure why the coverage regressed. The new unit tests in Golang actually have 100% coverage on the code.

prashantv:
Peng, can you explain how you plan to use this? The incoming Context for your handler methods will still use contextKeyCall, so your handler code has to use CurrentCall.

What I was thinking was that you need to way to create a Context for test that has the call options that you want, so add something like.
NewContextForTest(context.Context ctx, call IncomingCall) context.Context

And then your test would do something like
ctx, cancel := context.WithDeadline(...)
ctx = WrapContextForTest(ctx, fakeCallImplementation)

And then in the testutils, you could add a method
func CreateIncomingCall(opts ...) IncomingCall

That you could use to create an incoming call for your test.

What do you think?

pengzhai:
Hi Prashant,

I was thinking to somehow switch to #CurrentCallWithKey in our code path during unit tests. It wouldn't be a clean solution. 

I think your solution is much cleaner as it creates a parallel setup in the context of testing, without polluting the production code path. I am going to re-do it in this way.

Thanks for the comment! I will be back soon! :-)

pengzhai:
Hi Prashant,

I have updated the change based on your feedback. Please have a look and let me know.

Thanks!

golang/context_test.go
```diff
@@ -0,0 +1,35 @@
...
+var (
+	cn = "hello"
+)
+
+func TestWrapContextForTest(t *testing.T) {
+	call := &mockIncomingCall{callerName: cn}
+	ctx, cancel := NewContext(time.Second)
+	defer cancel()
+	actual := WrapContextForTest(ctx, call)
+	assert.Equal(t, call, actual.Value(contextKeyCall), "Incorrect call object returned.")
```
prashantv: instead of using the .Value, can you use the CurrentCall method to make sure they're the same

golang/testutils/call.go
```diff
@@ -0,0 +1,39 @@
...
+// FakeIncomingCall implements IncomingCall interface.
+type FakeIncomingCall struct {
+	callerName string
+}
+
+func (f *FakeIncomingCall) CallerName() string {
+	return f.callerName
+}
+
+func CreateIncomingCall(callerName string) tchannel.IncomingCall {
```
prashantv: nit: s/CreateIncomingCall/NewIncomingCall
I try to use New[...] instead of Create[...]

golang/thrift/context.go
```diff
@@ -70,6 +70,15 @@ func NewContext(timeout time.Duration) (Context, context.CancelFunc) {
 	}, cancel
 }
 
+// WrapContextForTest returns a Context that is associated with the call.
+// This should be used in units test only.
+func WrapContextForTest(ctx context.Context, call tchannel.IncomingCall) Context {
```
prashantv: optional: instead of exposing a test-specific method here, what about if we exposed
Wrap(ctx context.Context) Context

And then the caller would create the context using tchannel.WrapContextForTest, and then the tchannel/json implementations of context don't need to worry about exposing these interfaces.

golang/context_test.go
pengzhai: Fixed.

golang/testutils/call.go
pengzhai: Fixed.

golang/thrift/context_test.go
```diff
@@ -0,0 +1,15 @@
+package thrift
+
+import (
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestWrapContextForTest(t *testing.T) {
```
prashantv: nit: rename TestWrapContextForTest -> TestWrapContext

golang/thrift/context.go
pengzhai: Really good point. My previous approach felt a bit weird as I had 'wrap...test' at two different places which were not that different...

Fixed.

prashantv:
lgtm, thanks for taking this on Peng!

pengzhai:
Thanks for the help Prashant!

PR #912, opened 2015-07-10
Move Python to separate repo

abhinav:
The Python library is now in the [uber/tchannel-python](https://github.com/uber/tchannel-python) repo and available as a submodule in this repo.


blampe:
:+1:

PR #917, opened 2015-07-12
Add a --multiProc flag to benchmarks

Raynos:
This PR adds a `--multiProc` flag to the benchmarks so that
we can spawn multi client & server processes.

This is useful for any feature branches where the relay is
not running at 100% CPU

 - Refactor benchmark entry into a class
 - Parameterize startServer() and startClient()
 - Implement `--multiProc` flag.


jcorbin:
lgtm, fix the lint

ShanniLi:
lgtm

PR #916, opened 2015-07-12
Hack WIP: In place relay experiment

Raynos:
I wanted to experiment with what is the maximum performance we can get out of an in place relay code.

Joshua is working hard #911 which does in place relaying properly using lazy frames.

This branch will **never be landed**. The only purpose of this hack is to find out what a minimal relay looks like and what else except relaying is slow in tchannel.

 - [x] hack fast in place relay to be replaced by #911 
 - [x] hack fast choosePeer to be replaced by #915 
 - [x] multi process benchmark to find actual req/s
 - [ ] hack faster ops to waste less time in timers
 - [ ] hack faster `net` code to remove wasted time in the tcp code.

PR #918, opened 2015-07-12
Multi process relay benchmark

jcorbin:
r @Raynos 

- broke out the --multiProc from #916
- moved the take tasks into the Makefile
- made the relay take benchmark run with --multiProc


jcorbin:
ahh I missed #917, I'll reduce this down to the take subsumation then once it lands.

jcorbin:
actually, nvm will just slip this into yours ;-)

PR #919, opened 2015-07-12
Fix file output multiplexing bug in benchmarks

Raynos:
The way the benchmarks are written the JSON lines are
being interleaved in output which makes it garbage output

r: @shannili @jcorbin @kriskowal


ShanniLi:
lgtm

PR #921, opened 2015-07-13
WIP: Native call request frame parser

Raynos:
This adds the `tchannel-parser` module to tchannel
which includes a CallRequest parser written in C++

In the flames we have improved the performance of 
the call request parser by 40% and we still have
plenty of C++ tricks to pull.

This is a first pass at seeing what writing the
call request / response frame parser in C++ adds.

r: @Matt-Esch @kriskowal @jcorbin @shannili


Raynos:
RIP

PR #922, opened 2015-07-13
WIP: Use a single timer for timeout reaper

Raynos:
This PR moves the timeout loop out of each connection
into a single one in the rootChannel.

This doubles the performance of the timeout reaper
by spending far less time thrashing lots of timers
per connections

### DIFF
--- a/node/channel.js
+++ b/node/channel.js
@@ -217,8 +217,84 @@ function TChannel(options) {
 
     self.requestDefaults = self.options.requestDefaults ?
         new RequestDefaults(self.options.requestDefaults) : null;
+
+    self.timeoutCheckInterval = self.options.timeoutCheckInterval;
+    self.timeoutFuzz = self.options.timeoutFuzz;
+    self.operationsTimer = null;
+
+    if (self.topChannel === null) {
+        self.startTimeoutTimer();
+    }
 }
 inherits(TChannel, StatEmitter);
+
+TChannel.prototype.startTimeoutTimer =
+function startTimeoutTimer() {
+    var self = this;
+
+    self.operationsTimer = self.timers.setTimeout(
+        onChannelTimeout, self._getTimeoutDelay()
+    );
+
+    function onChannelTimeout() {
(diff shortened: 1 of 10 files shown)


node/channel.js
```diff
@@ -217,9 +217,85 @@ function TChannel(options) {
...
+TChannel.prototype._getTimeoutFuzz =
+function _getTimeoutFuzz() {
+    var self = this;
+
+    var fuzz = self.timeoutFuzz;
+    if (!fuzz) {
+        return 0;
+    }
+
+    return Math.round(Math.floor(self.random() * fuzz)) - (fuzz / 2);
```
kriskowal: Per the paper @jcorbin posted on back-off, fuzz was implement much more simply in Python, like `self.timeoutCheckInterval * Math.random()`, where checkInterval was backing off. May or may not be applicable here, but this math is definitely more complicated than necessary and, when poorly configured, can result in negative timeouts.

node/channel.js
```diff
@@ -159,7 +159,7 @@ function TChannel(options) {
     }
 
     // populated by makeSubChannel
-    self.topChannel = null;
+    self.topChannel = self.options.topChannel || null;
```
kriskowal: :warning: `self.options` should just be `options` to make less work of eliminating `self.options` if possible.

kriskowal:
:ok:ish. I recommend simplifying the timeout fuzz, but otherwise great.

Raynos:
@kriskowal Going to break out this PR into 3 PRs.

kriskowal:
@Raynos Gotta love smaller PRs

node/channel.js
Raynos: This PR just moves code instead of changing its semantics.

ShanniLi: Isn't the purpose of timeout fuzz on the opposite side of single timer?

Raynos:
Ready for re-review: @kriskowal @jcorbin

PR #927, opened 2015-07-14
Ensure all peers exist on root channel

Raynos:
This is part 1 of a timers change

For sanity we enforce that all peers must exist on the
root channel.


jcorbin:
:+1:

kriskowal:
:ok:

PR #929, opened 2015-07-14
move topChannel assignment into constructor

Raynos:
This allows the constructor to determine whether it is
a root channel or not.

This will be useful in a future PR.


jcorbin:
:ok:

ShanniLi:
lgtm. Not sure why it is better?

Issue #934, opened 2015-07-14
consolidating documentation


blampe:
I'd like to propose that we repurpose tchannel.readthedocs.org for high-level TChannel documentation like the protocol, metrics, etc., which are very painful to read on github. Readthedocs supports subprojects, so this can also serve as a landing page to tchannel.readthedocs.org/projects/python, tchannel.readthedocs.org/projects/node, etc.

tchannel.readthedocs.org is currently running outdated Python docs, which have moved to tchannel-python.readthedocs.org. @sectioneight I think you need to add some of us as maintainers on readthedocs so we can start making these changes.

cc @Raynos @jcorbin

Raynos:
I don't mind where we host docs. 

Personally I like reading markdown through the github UI, as long as the docs stay in markdown in the source code and readthedocs is just a pretty GUI then I dont care.

Removing the documentation from tchannel would be bad.

prashantv:
Is readthedocs supposed to be for API documentation or for guides? API
documentation for go tends to live on godoc.org. e.g.
http://godoc.org/github.com/uber/tchannel/golang

Which is automatically generated from the source code. I'm happy to put the
guides up there and point to the godoc for the full API documentation.

sectioneight:
@blampe i added "blampe" as a maintainer on readthedocs. If that's not you, or if there are other people I should add, please let me know the names here

blampe:
tchannel.readthedocs.org

@prashantv I agree golang api docs should continue to live on godoc.org. Take a look at how things are laid out right now. Maybe we should include direct links to every language's guide?

PR #935, opened 2015-07-14
readthedocs compatibility

blampe:
Preview of what this looks like: http://tchannel.readthedocs.org/en/docs/

Refs #934


breerly:
super sweet

abhinav:
I like the idea.

Minor changes:

I would put the protocol spec and all other details under the "Protocol" or "Specification" top-level title and still use a toctree of depth 2. This way our TOC looks like this:

* Specification
  * TChannel V2
  * HTTP over TChannel
  * JSON over TChannel
  * Thrift over TChannel
  * TChannel Traffic Metrics

(Also, yeah, move Thrift before metrics.)

Or maybe drop toctree depth to just 1. We can adjust these details later.

Raynos:
@blampe lgtm.

blampe:
Going to land this. The TOC is crowded but we can iterate on that.

PR #936, opened 2015-07-14
Coordinate all request timeout handling through a min heap on the root channel

jcorbin:
r @Raynos @kriskowal 

Not yet passing tests, early share; see especially the comment block at head of ttl_heap.js

### DIFF
--- a/node/channel.js
+++ b/node/channel.js
@@ -48,6 +48,7 @@ var TChannelPeers = require('./peers');
 var TChannelServices = require('./services');
 var TChannelStatsd = require('./lib/statsd');
 var RetryFlags = require('./retry-flags.js');
+var TimeHeap = require('./time_heap');
 
 var TracingAgent = require('./trace/agent');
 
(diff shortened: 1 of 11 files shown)


Raynos:
I approve; lot's of great simplifications.

node/ttl_heap.js
```diff
@@ -0,0 +1,255 @@
...
+        var now = self.timers.now();
+        self.onTimeout(now);
+    }
+};
+
+TTLHeap.prototype.onTimeout = function onTimeout(now) {
+    var self = this;
+
+    self.timer = null;
+    self.drainExpired(now);
```
kriskowal: Unnecessary O(n) operation if you reframe ttl absolute instead of relative time.

node/ttl_heap.js
```diff
@@ -0,0 +1,255 @@
...
+    self.lastTime = self.timers.now();
+    self.timer = null;
+}
+
+TTLHeap.prototype.clear = function clear() {
+    var self = this;
+
+    self.timers.clearTimeout(self.timer);
+    self.items = [];
+    self.ttls = [];
```
kriskowal: .length = 0 is cheap and eliminates one piece of garbage

jcorbin:
Now ready for review, whilst I write a unit test for the time_heap module

Raynos:
Lets add that 10 second sanity check.

Otherwise lgtm.

jcorbin:
Update: had tests passing last night, now tests don't pass this morning:
- hard to repo flaps local
- however those seemed to stop after an npm install that upgraded debug-logtron
- now I have two actual failures in the new rate limiter test by @ShanniLi , working on those

jcorbin:
Oh also, I switched over to using test/lib/test_search to abstract/aggregate way all the ~70k TimeHeap assertions into one "it worked" pass, or the first failing case if any.

Raynos:
lgtm. ship this.

Raynos:
I noted a memory leak in the benchmark.

All three (client, relay, bench) were sitting at 2GB of memory locally.

jcorbin:
memory leak fixed, landing once travis greens

kriskowal:
:clap: :clap: :clap:

PR #937, opened 2015-07-14
Refactor operation timeout paths ahead of performance work

jcorbin:
r @raynos @kriskowal


jcorbin:
Prep work of #936 pulled out

Raynos:
lgtm.

PR #942, opened 2015-07-14
Add more production dependencies

Raynos:
The `require('tchannel/test/lib/hyperbahn-cluster')` is
now part of the public API for test use cases.

This means some of our dev dependencies are actually
dependencies.

I have fixed all of them except for `tape`. Because tape
is a singleton it cannot be added to dependencies.

### DIFF
--- a/node/package.json
+++ b/node/package.json
@@ -25,6 +25,7 @@
   "dependencies": {
     "bufrw": "^0.9.24",
     "crc": "^3.2.1",
+    "debug-logtron": "5.0.0",
     "error": "^7.0.1",
     "farmhash": "^0.2.0",
     "hexer": "^1.4.5",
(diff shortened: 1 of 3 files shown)


ShanniLi:
lgtm

node/package.json
```diff
@@ -34,7 +35,9 @@
     "ready-signal": "^1.1.1",
     "run-parallel": "^1.1.0",
     "sse4_crc32": "^3.1.0",
+    "tape-cluster": "2.1.0",
     "thriftify": "1.0.0-alpha14",
+    "uber-statsd-client": "1.3.2",
     "xtend": "^4.0.0"
```
kriskowal: These are all pinned to an exact version. Is that intentional?

kriskowal:
:ok: Thanks.

PR #943, opened 2015-07-15
Circuit stats

kriskowal:
r @jcorbin @Raynos @ShanniLi

### DIFF
--- a/node/circuits.js
+++ b/node/circuits.js
@@ -39,9 +39,16 @@ AlwaysShouldRequestHandler.prototype.shouldRequest = function shouldRequest() {
 
 var alwaysShouldRequestHandler = new AlwaysShouldRequestHandler();
 
-//  circuit = circuits                        := Circuits
-//      .circuitsByServiceName[serviceName]   := ServiceCircuits
-//      .circuitsByCallerName[callerName]     := EndpointCircuits
+function CircuitStateChange(circuit, oldState, state) {
+    var self = this;
+    self.circuit = circuit;
+    self.oldState = oldState;
+    self.state = state;
+}
+
+//  circuit = circuits                        : Circuits
+//      .circuitsByServiceName[serviceName]   : ServiceCircuits
+//      .circuitsByCallerName[callerName]     : EndpointCircuits
 //      .circuitsByEndpointName[endpointName]
 
 function EndpointCircuits(root) {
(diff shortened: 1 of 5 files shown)


node/circuits.js
```diff
@@ -66,6 +73,9 @@ EndpointCircuits.prototype.getCircuit = function getCircuit(callerName, serviceN
             minRequests: stateOptions.minRequests,
             probation: stateOptions.probation
         });
+        circuit.stateChangedEvent.on(function circuitStateChanged(states) {
```
Raynos: Putting the logging and stats inline straight in `circuit` and skipping 3 event emissions is going to be faster.

Raynos:
lgtm

node/circuits.js
```diff
@@ -39,9 +39,16 @@ AlwaysShouldRequestHandler.prototype.shouldRequest = function shouldRequest() {
 
 var alwaysShouldRequestHandler = new AlwaysShouldRequestHandler();
 
-//  circuit = circuits                        := Circuits
-//      .circuitsByServiceName[serviceName]   := ServiceCircuits
-//      .circuitsByCallerName[callerName]     := EndpointCircuits
+function CircuitStateChange(circuit, oldState, state) {
```
ShanniLi: maybe better?
oldState ==> fromState
state ==> toState

node/hyperbahn/service_proxy.js
```diff
@@ -424,6 +433,63 @@ function isExitFor(serviceName) {
...
+    var self = this;
+
+    var circuit = change.circuit;
+    var oldState = change.oldState;
+    var state = change.state;
+
+    if (oldState && oldState.healthy !== state.healthy) {
+        // unhealthy -> healthy
+        if (state.healthy) {
+            self.statsd.increment('circuits.healthy.total', 1);
```
ShanniLi: Please use StatEmitter instead of statsd directly. Once m3 is ready we may move away from statsd which has a difference interface compared to m3.

ShanniLi: Please also document these stats in the tchannel metrics doc.

ShanniLi:
All comments are non-blocking.
Would really appreciate fixing the comments about stats. Also, all stats generated through tchannel should be documented in the metrics doc for cross language reference.

node/hyperbahn/service_proxy.js
Raynos: This code is in hyperbahn and doesn't have access to any of the fast stat code in tchannel. For loose coupling we are just going to hard code statsd. 

We could pull fast stats into a seperate library but that will take more work thats out of scope for this PR.

:+1: for metrics doc though.

ShanniLi: If it is for loose coupling, "clean = require('../lib/statsd-clean').clean;" should be removed from this file. :)

PR #944, opened 2015-07-15
make options explicit in connection

ShanniLi:
r @Raynos @kriskowal 

The loop introduces bugs that overwrite valid values by undefined.


Raynos:
lgtm if pass test. Try autobahn tests :D

ShanniLi:
@Raynos autobahn tests is broken. I am fixing it now. :)

PR #945, opened 2015-07-15
hyperbahn-cluster: add namedRemotes option

Raynos:
This allows people to create a hyperbahn cluster with
remotes that have different names then bob and steve.

r: @kriskowal @shannili @rf @jeloi

### DIFF
--- a/node/test/hyperbahn/forward.js
+++ b/node/test/hyperbahn/forward.js
@@ -33,10 +33,12 @@ if (require.main === module) {
 
 function runTests(HyperbahnCluster) {
     HyperbahnCluster.test('advertise and forward', {
-        size: 5
+        size: 5,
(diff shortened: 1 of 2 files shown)


node/test/hyperbahn/forward.js
```diff
@@ -33,10 +33,12 @@ if (require.main === module) {
 
 function runTests(HyperbahnCluster) {
     HyperbahnCluster.test('advertise and forward', {
-        size: 5
+        size: 5,
+        namedRemotes: ['august']
     }, function t(cluster, assert) {
+        console.log('wtf', cluster.namedRemotes);
```
ShanniLi: wtf?

ShanniLi:
please remove the console.log

Otherwise, lgtm

PR #946, opened 2015-07-15
fix torch code in benchmark

Raynos:
This is broken on master. We refactored to an array
of child processes but never tested these branches.

r: @kriskowal @shannili @rf


ShanniLi:
lgtm

PR #948, opened 2015-07-15
rate limiting: fix the test breaks in hyperbahn & support exempt services

ShanniLi:
r @Raynos @kriskowal

### DIFF
--- a/node/test/hyperbahn/rate-limiter.js
+++ b/node/test/hyperbahn/rate-limiter.js
@@ -58,7 +58,11 @@ function runTests(HyperbahnCluster) {
         size: 1,
         timers: timers,
         rateLimiterBuckets: 2,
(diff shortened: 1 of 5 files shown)


node/test/hyperbahn/rate-limiter.js
```diff
@@ -140,7 +148,7 @@ function runTests(HyperbahnCluster) {
                 wait,
                 send.bind(null, opts),
                 function check2(done) {
-                    cluster.relayNetwork.relayChannels.forEach(function (relayChannel, index) {
+                    cluster.getRelayChannels().forEach(function (relayChannel) {
```
Raynos: Let's loop over `cluster.apps` to make it portable with autobahn cluster.

Raynos:
lgtm.

PR #950, opened 2015-07-15
Add endpoint overhead

jcorbin:
Turns out that in reality endpoints don't usually turn around and send / finish a response immediately.

This adds support for adding random-sampled delay into the node benchmark server.


Raynos:
lgtm.

PR #957, opened 2015-07-16
Use correct reference to call req's `service` field

willsalz:
I'm not sure if this field was historically called `service_name` but I was confused when I looked in the call req `Schema` and saw no `service_name`, only `service`. It's a minor naming issue ( if it wasn't intentional ), but still confusing.


Raynos:
lgtm.

PR #959, opened 2015-07-16
Improve tests before TimeHeap work

jcorbin:
Carved out of #936 from @raynos's WIP PR


jcorbin:
Pulled out since it's getting hard to see what the actual problems are since our tests have become rather slow and flappy lately.

kriskowal:
:ok:

PR #961, opened 2015-07-16
Improve rate limiter test

jcorbin:
r @Raynos


node/test/hyperbahn/rate-limiter.js
```diff
@@ -46,8 +46,10 @@ function send(opts, done) {
     });
 }
 
-function wait(done) {
-    setTimeout(done, 500);
+function waitFor(t) {
```
kriskowal: t is unused.

PR #973, opened 2015-07-17
Operations: defer connection timed out emit out of check

jcorbin:
r @Raynos 

Fixes bug through `TChannelOutRequest#onTImeout`:
- connection timed out event calls connection resetAll
- resetAll calls popOutReq
- popOutReq clears req.operations
- which makes line 567 fail in TChannelOutRequest after checkLastTimeoutTime returns


Raynos:
lgtm.

PR #974, opened 2015-07-17
Add explicit section for reserved:1 frame byte

willsalz:
There was a section on the `reserved` padding of 8 bytes in the frame head but the `reserved` frame space of 1 byte was missing from the docs. I'm not sure what the docs should/would say on this lonely byte, so I just copy-pasta'd the blurb from the `reserved:8` section.

### DIFF
--- a/docs/protocol.md
+++ b/docs/protocol.md
@@ -215,6 +215,11 @@ size to 64KiB to allow better interleaving of frames across a shared TCP
 connection. In order to handle most operations implementations will need to
 decode some part of the payload.
 
+### reserved:1
+
+Unused space for future protocol revisions, padded for read alignment, which
+may or may not help on modern Intel processors.
+
 ### id:4
 
 An identifier for this message that is chosen by the sender of a request. This


Raynos:
lgtm.

kriskowal:
:ok:

docs/protocol.md
```diff
@@ -215,6 +215,11 @@ size to 64KiB to allow better interleaving of frames across a shared TCP
 connection. In order to handle most operations implementations will need to
 decode some part of the payload.
 
+### reserved:1
+
+Unused space for future protocol revisions, padded for read alignment, which
```
jcorbin: Saying that this was "for read alignment" is stretching / retconning a bit.  Read alignment wasn't at all a goal when we designed the header.  I think this is more leftover from we used to have a top-level flags single-byte field.

Also the most useful artifact of all 9 of these total reserved bytes is that they make the body start at offset 0x10 which wraps to lines cleanly in a standard 16-byte wide hex dump (e.g. our own tcap, or even just hexer dumping internally for any reason).

PR #976, opened 2015-07-17
fix corrupt svg files

Raynos:
A quick hack in the makefile to fix production flames.

r: @jcorbin @kriskowal


jcorbin:
:+1: , pair debugged / coded in person

PR #977, opened 2015-07-17
add meta.thrift to tchannel

ShanniLi:
r @Raynos @kriskowal @rf

### DIFF
--- a/node/as/meta.thrift
+++ b/node/as/meta.thrift
@@ -1,4 +1,9 @@
+struct HealthStatus {
+    1: required bool ok
+    2: optional string message
+}
+
 service Meta {
-    void health()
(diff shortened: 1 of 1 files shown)


rf:
lgtm

kriskowal:
This is the not the interface I believe emerged from the last offline discussion. I believe we had universal agreement to:

```thrift
struct Health {
   1: required bool ok
   2: optional string message
}
service Meta {
    Healthy health()
}
```

Or words to that effect

ShanniLi:
@kriskowal good catch! Looking ...

ShanniLi:
@kriskowal fixed.

kriskowal:
:ok:

PR #979, opened 2015-07-17
Cleanup TChannelPeers a bit

jcorbin:
- drop dubious `#waitForIdentified`
- drop broken return from `#clear`
- flatten flow of `#add` so it's clearer to break up going forward
- always return null from `#choosePeer` on the top channel

r @Raynos @kriskowal


jcorbin:
need to fix tests still it seems, working on it

Raynos:
lgtm.

kriskowal:
yarr

PR #980, opened 2015-07-17
trace sampling w/ 1% default

rf:
Supports an `options.traceSample` property and defaults it to 1%

r @Raynos @ShanniLi @kriskowal

### DIFF
--- a/node/channel.js
+++ b/node/channel.js
@@ -225,6 +225,8 @@ function TChannel(options) {
         });
     }
 
+    self.traceSample = self.options.traceSample || 0.01; // sample 1% by default
+
     // lazily created by .getServer (usually from .listen)
     self.serverSocket = null;
     self.serverConnections = null;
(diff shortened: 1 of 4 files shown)


node/channel.js
```diff
@@ -225,6 +225,8 @@ function TChannel(options) {
         });
     }
 
+    self.traceSample = self.options.traceSample || 0.01; // sample 1% by default
```
ShanniLi: when you set self.options.traceSample = 0, it will be default to 0.01.

ShanniLi:
lgtm. the comment is optional.

node/channel.js
```diff
@@ -606,6 +612,13 @@ TChannel.prototype.request = function channelRequest(options) {
     var self = this;
 
     options = options || {};
+
+    if (options.hasNoParent && (Math.random() < self.traceSample)) {
```
Raynos: This is incorrect.

```
if (options.hasNoParent) {
    if (Math.random() < self.traceSample) {
        options.trace = true;
    } else {
        options.trace = false;
    }
}
```

Otherwise we turn of tracing for everyone even if they are not the parent request.

kriskowal:
trace and traceSample overlap, since traceSample === 0 is equivalent to trace === false. Consider folding these into a single trace value that varies from 0 to 1, where 0 is equivalent to false, and 1 is equivalent to what was formerly a binary truth.

Raynos:
@kriskowal no. Back compat. No changing the existing `trace` field.

Raynos:
lgtm

PR #981, opened 2015-07-17
Unbreak the tchannel relay for streaming

Raynos:
When circuits is turned on then the circuit breaker will
drain arg1.

To avoid dropping arg1 in the relay we have to favor
the cached buffer if available.

**Note:** We really need to fix the code debt where
arg1 is a stream instead of a buffer. It should always
be a buffer.

r: @kriskowal @jcorbin


jcorbin:
:+1:

kriskowal:
:ok:

PR #982, opened 2015-07-17
Factor out PreferOutgoingHandler in peer

jcorbin:
r @kriskowal


kriskowal:
Looks idemfigurative. :ok:

Raynos:
lgtm

PR #983, opened 2015-07-17
States: pivot API from shouldRequest to willCallNextHandler

jcorbin:
r @kriskowal


kriskowal:
Looks good to me. You’re right that `willCallNextHandler` is a minimum viable name. Seems like `shouldRequest` is the right name, and `shouldRequest` should be `getScore`.

jcorbin:
Punting renames to future work

PR #984, opened 2015-07-17
PreferOutgoingHandler: factor out QOS concept

jcorbin:
r @kriskowal @Raynos

### DIFF
--- a/node/peer.js
+++ b/node/peer.js
@@ -340,11 +340,33 @@ TChannelPeer.prototype.countOutPending = function countOutPending() {
     return pending;
 };
 
+var QOS_UNCONNECTED = 0;
+var QOS_ONLY_INCOMING = 1;
+var QOS_FRESH_OUTGOING = 2;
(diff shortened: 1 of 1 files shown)


node/peer.js
```diff
@@ -340,12 +340,35 @@ TChannelPeer.prototype.countOutPending = function countOutPending() {
     return pending;
 };
 
+var QOS_UNCONNECTED = 0;
```
Raynos: nitpick: defaulting constants to strings is more readable and more debuggable.

I've found it hard to debug `req.state` and such because numbers.

Raynos:
lgtm.

node/peer.js
jcorbin: We can change it going forward... however I'd rather stick with a non-string comparison, so how about instead sentinel objects with nice `.toString()` methods?

kriskowal:
This word, QOS, I do not think it means what you think it means.

PR #915, opened 2015-07-12
Use a binary heap to store peers, making peer choice do less work

jcorbin:
The cost of `TChannelPeers#choosePeer` should run on the order of log-number of peers now.  The choice itself is constant time, but to maintain semantics, a rescore must be done, which implies a log-time heap siftdown.  However it now only runs the score function once, rather than N times, so the scaling amount of work in each request is greatly reduced.

TODO:
- [x] fix existing tests that are broken; probably missed semantics
- [x] implement an out of band rescore hook as noted
- ~~[ ] add a unit test for the heap (an exhaustive search test on all 6-trees should do)~~ (integration coverage proved Good Enough (tm) )

PR #985, opened 2015-07-17
Peers: breakup

jcorbin:
The existing TChannelPeers class already had extensive branching for root vs sub channel use case.

This PR breaks those out explicitly into two subclases of a new shared base class.

Plus I killed use of array splice for fun.

r @kriskowal @Raynos

### DIFF
--- a/node/peers_base.js
+++ b/node/peers_base.js
@@ -1,0 +1,122 @@
+// Copyright (c) 2015 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+'use strict';
+
+var inherits = require('util').inherits;
+var EventEmitter = require('./lib/event_emitter');
+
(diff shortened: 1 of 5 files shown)


node/root_peers.js
```diff
@@ -0,0 +1,147 @@
...
+    self.selfPeer = null;
+}
+
+inherits(TChannelRootPeers, TChannelPeersBase);
+
+TChannelRootPeers.prototype.close = function close(callback) {
+    var self = this;
+
+    var peers = self.values();
+    if (self.selfPeer) {
```
Raynos: Adding an `if (self.selfPeer)` to base class will simplify this class a lot.

`close()` and `sanitySweep()` are not hot; might as well make the base class aware of `selfPeer` and get rid of a bunch of method overloading.

node/peers_base.js
```diff
@@ -0,0 +1,122 @@
...
+
+    var keys = self._keys;
+    var ret = new Array(keys.length);
+    for (var i = 0; i < keys.length; i++) {
+        ret[i] = [keys[i], self._map[keys[i]]];
+    }
+    return ret;
+};
+
+TChannelPeersBase.prototype.delete = function del(hostPort) {
```
Raynos: This delete vs _delete indirection doesn't seem to add much value.

Can just implement `delete()` fully in both RootPeers and SubPeers

Raynos:
Looks good enough.

node/peers_base.js
```diff
@@ -0,0 +1,122 @@
...
+
+    var counter = peers.length + 1;
+    peers.forEach(function eachPeer(peer) {
+        peer.close(onClose);
+    });
+    self.clear();
+    onClose();
+
+    function onClose() {
+        if (--counter <= 0) {
```
kriskowal: Suggest making the counter decrement a statement and flattening the following conditional.

node/root_peers.js
```diff
@@ -0,0 +1,147 @@
...
+    if (array.length === 1) {
+        array.pop();
+        return;
+    }
+
+    var j = array.length - 1;
+    var tmp = array[i];
+    array[i] = array[j];
+    array[j] = tmp;
+}
```
kriskowal: Nice.

node/sub_peers.js
```diff
@@ -0,0 +1,129 @@
...
+    if (array.length === 1) {
+        array.pop();
+        return;
+    }
+
+    var j = array.length - 1;
+    var tmp = array[i];
+    array[i] = array[j];
+    array[j] = tmp;
+}
```
kriskowal: Copy paste makes copy pasta (and makes less hot for v8 optimizer). Factor it out.

kriskowal:
:ok:

node/peers_base.js
jcorbin: Indeed, it's room for currently being developed work, see https://github.com/uber/tchannel/pull/915

node/peers_base.js
jcorbin: pre-existing inlined decrement, keeping as is for now

node/sub_peers.js
jcorbin: I'm not sure what you're asking to have factored out here?

node/root_peers.js
jcorbin: sure, let's punt this to future work, changing enough in this diff as is, and this was the easiest preservation of old logic, so keeping it for now.

node/sub_peers.js
Raynos: The popout function exists in two files; moving it into a seperate file makes maintenance easier ?

PR #988, opened 2015-07-18
channel: do not traceSample if no trace

Raynos:
We must not mutate trace=false -> trace=true due
to the hasNoParent stuff.

r: @rf @shannili @kriskowal

### DIFF
--- a/node/channel.js
+++ b/node/channel.js
@@ -640,17 +640,16 @@ TChannel.prototype.request = function channelRequest(options) {
 
     options = options || {};
 
-    if (options.hasNoParent) {
+    var opts = new RequestOptions(self, options);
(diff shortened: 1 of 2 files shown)


Raynos:
Master is broken. I'm just going to land this and unbreak the rt/hyperbahn test suite.

ShanniLi:
lgtm

node/channel.js
```diff
@@ -640,18 +640,17 @@ TChannel.prototype.request = function channelRequest(options) {
 
     options = options || {};
 
-    if (options.hasNoParent) {
+    var opts = new RequestOptions(self, options);
```
kriskowal: It’s okay to overwrite the options variable. Far preferable to this.

PR #989, opened 2015-07-19
No checksumming for out relay request.

Raynos:
This removes the CRC32 calls in the outrequest and
outresponse path for the relay.

```
GET 16KiB, 1000/0  rate: hi-diff:     65.78 (  3.2%)
GET 16KiB, 10000/0 rate: hi-diff:    -35.42 ( -2.0%)
GET 16KiB, 20000/0 rate: hi-diff:    -45.27 ( -3.0%)
GET 4KiB, 1000/0   rate: hi-diff:    102.29 (  4.2%)
GET 4KiB, 10000/0  rate: hi-diff:    165.04 (  7.8%)
GET 4KiB, 20000/0  rate: hi-diff:    -66.58 ( -3.4%)
SET 16KiB, 1000/0  rate: hi-diff:   -327.10 (-17.6%)
SET 16KiB, 10000/0 rate: hi-diff:     43.05 (  2.7%)
SET 16KiB, 20000/0 rate: hi-diff:     44.40 (  3.2%)
SET 4KiB, 1000/0   rate: hi-diff:    -66.24 ( -2.7%)
SET 4KiB, 10000/0  rate: hi-diff:     74.71 (  3.6%)
SET 4KiB, 20000/0  rate: hi-diff:    120.23 (  7.6%)
```

Benchmarking numbers are spotty but I'm removing
weight from the flamegraphs.

### DIFF
--- a/node/v2/handler.js
+++ b/node/v2/handler.js
@@ -886,7 +886,9 @@ TChannelV2Handler.prototype.buildOutResponse = function buildOutResponse(req, op
     options.tracing = req.tracing;
     options.span = req.span;
     options.checksumType = req.checksum.type;
-    options.checksum = new v2.Checksum(req.checksum.type);
+    if (!options.checksum) {
+        options.checksum = new v2.Checksum(req.checksum.type);
+    }
     if (options.streamed) {
         return new StreamingOutResponse(self, req.id, options);
     } else {
--- a/node/benchmarks/Makefile
+++ b/node/benchmarks/Makefile
@@ -36,7 +36,7 @@ take:
 
 take_relay:
 	node index.js --relay -o relay-$$(git rev-parse --short HEAD).json \
-		-- --relay --skipPing -s 4096,16384 -p 1000,10000,20000
+		-- --relay --skipPing -m 3 -s 4096,16384 -p 1000,10000,20000
 	ln -sf relay-$$(git rev-parse --short HEAD).json relay-$$(basename $$(git symbolic-ref HEAD)).json
 
 take_trace:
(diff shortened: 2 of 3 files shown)


jcorbin:
What is this? It seems to have stat batching mostly / or mixed in with something?

At any rate, can we punt on this incremental change since I'm going to by shaping up the lazy rw path this week which will "skip out relay checksumming" and so much more?

Raynos:
Hmm. This Branch needs rebasing.

There are general cleanup fixes in this branch so it's worth keeping.

Raynos:
This is something we will want long term for retries as well to avoid re-computing checksums in the retry path.

node/v2/handler.js
```diff
@@ -573,9 +573,17 @@ function sendCallRequestFrame(req, flags, args) {
         self.errorEvent.emit(self, errors.SendCallReqBeforeIdentifiedError());
         return;
     }
+
+    var checksum;
+    if (req.checksum.frozen) {
+        checksum = req.checksum;
+    } else {
+        checksum = req.checksum.type;
```
kriskowal: Suspicious. Is `type` the same shape as `checksum`?

kriskowal: Should this be `val` throughout?

Raynos: All the checksumming code goes through `ChecksumOrType : (CheckSum | Number) => CheckSum`

So it should be an object or a numeric type.

Raynos:
Rebased.

ShanniLi:
lgtm

kriskowal:
Seems like a reasonable thing to throw under the bus for speed, though I worry about the integrity of persistent connections. When do we notice a corrupted stream and drop the connection? Perhaps only when the frame header is no longer plausibly valid?

Raynos:
@kriskowal the edges should disconnect connections on protocol errors.

PR #992, opened 2015-07-20
Deterministic retry endpoints

jcorbin:
This disentagles the mocked random from endpoint disposition.  Otherwise it's just too hard to reason about what happens in upcoming changes... plus this seems simpler and more correct anyhow (obvious in retrospect).

r @kriskowal @Raynos

### DIFF
--- a/node/test/retry.js
+++ b/node/test/retry.js
@@ -27,13 +27,6 @@ var TChannel = require('../channel');
 allocCluster.test('request retries', {
     numPeers: 4
 }, function t(cluster, assert) {
-    var random = randSeq([
-        1.0, 0.1, 0.1, 0.1, // .request, chan 1 wins
-             1.0, 0.1, 0.1, // .request, chan 2 wins (1 is skipped)
-                  1.0, 0.1, // .request, chan 3 wins (1-2 are skipped)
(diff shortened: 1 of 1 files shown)


Raynos:
lgtm.

node/test/retry.js
```diff
@@ -27,42 +27,34 @@ var TChannel = require('../channel');
...
+        .register('foo', servedByFoo(4));
 
     var client = TChannel({
         timeoutFuzz: 0,
-        random: random
+        random: randSeq([
+            1.0, 0.1, 0.1, 0.1, // .request, chan 1 wins
+                 1.0, 0.1, 0.1, // .request, chan 2 wins (1 is skipped)
+                      1.0, 0.1, // .request, chan 3 wins (1-2 are skipped)
+                           0.1  // .request, chan 4 wins (only one left)
```
kriskowal: Dyslexics will have an easier time of this if you change 1.0 to 2.0, making it easier to distinguish from 0.1.

kriskowal:
Interesting technique :ok:

node/test/retry.js
jcorbin: 2.0 would be an invalid value for a random function to return, hence the constant

PR #993, opened 2015-07-20
tests/double-response: harden against missing error arg3

jcorbin:
Tests shouldn't crash with an uncaught when possible.

r @Raynos


Raynos:
lgtm.

PR #994, opened 2015-07-20
Sort out relay request log shenanigans

jcorbin:
Whilst doing other work, I managed to provoke a "cannot call method error of undefined" wrt self.logger in RelayRequest from our test suite.

This fixes that, plus:
- factors out the error/code classification as a forward pass to other lazy relay request refactoring work
- adds a defense-in-depth default error log level; the old behavior would have been silent failure
- gets rid of a needless comparison by doing it in its predicated case branch

r @kriskowal @Raynos

### DIFF
--- a/node/relay_request.js
+++ b/node/relay_request.js
@@ -269,6 +269,9 @@ function errorLogLevel(err, codeName) {
         case 'BadRequest':
         case 'Timeout':
             return 'info';
+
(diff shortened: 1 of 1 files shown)


kriskowal:
:ok:

Raynos:
lgtm.

PR #995, opened 2015-07-20
Move connection error vs close handling more towards correct

jcorbin:
Had to back off from making close event node-idiomatic (take only a "was it a clean close" boolean) due to how coupled we are currently to getting the error through close.

r @raynos @kriskowal

### DIFF
--- a/node/connection.js
+++ b/node/connection.js
@@ -463,7 +463,6 @@ TChannelConnection.prototype.resetAll = function resetAll(err) {
     self.closing = true;
     self.closeError = err;
     self.socket.destroy();
-    self.closeEvent.emit(self, err);
 
     var requests = self.ops.getRequests();
     var pending = self.ops.getPending();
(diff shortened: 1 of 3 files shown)


node/connection.js
```diff
@@ -554,16 +537,35 @@ TChannelConnection.prototype.resetAll = function resetAll(err) {
...
             outArg1: String(req.arg1)
         };
-        if (err.type === 'tchannel.socket-local-closed') {
-            err = errors.TChannelLocalResetError(err, info);
+
+        var reqErr = err;
+        if (reqErr.type === 'tchannel.socket-local-closed') {
+            reqErr = errors.TChannelLocalResetError(reqErr, info);
         } else {
-            err = errors.TChannelConnectionResetError(err, info);
```
kriskowal: sonofa

kriskowal:
:ok:

node/peer.js
```diff
@@ -205,30 +205,35 @@ function waitForIdentified(callback) {
...
     if (conn.closing) {
         callback(conn.closeError);
-    } else if (conn.remoteName) {
+        return;
+    }
+
+    if (conn.remoteName) {
         callback(null);
-    } else {
-        self._waitForIdentified(conn, callback);
```
Raynos: There was a reason for this.

I do not want to allocate closures in `waitForIdentified()` and moved the closure allocation into a seperate function.

Raynos:
Let's add the `_waitForIdentified()` method back.

Other then that lgtm.

jcorbin:
added back, travis.then(merge)

PR #996, opened 2015-07-20
Now all code paths in forwardToTChannel are calling callback function

vipulaneja:
Found this during testing my service (xlate) using TChannel on local box. Was not getting TChannel error propagted back to my service


Raynos:
lgtm.

PR #997, opened 2015-07-20
TChannelPeer: count pending identified against random weight

jcorbin:
Needs to maybe have tests written before merging? Up to reviewers: @kriskowal and @Raynos .


kriskowal:
Looking into adding a load balancing test that takes advantage of this change to proceed before identification.

Raynos:
lgtm.

PR #998, opened 2015-07-20
Factor out and make debugable randSeq fixture between test/{send and retry}

jcorbin:
This makes it easier to iterate on peer selection so that you can see how your injected random sequence is behaving and from where.  cc @kriskowal @Raynos

### DIFF
--- a/node/test/lib/peer_score_random.js
+++ b/node/test/lib/peer_score_random.js
@@ -1,0 +1,61 @@
+// Copyright (c) 2015 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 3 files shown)


node/test/lib/peer_score_random.js
```diff
@@ -0,0 +1,61 @@
...
+    var stackLines = stack.split(/\n/);
+    for (var i = 0; i < stackLines.length; i++) {
+        if (/shouldRequest/.test(stackLines[i])) {
+            break;
+        }
+    }
+    stackLines = stackLines
+        .slice(i, i + 5)
+        .filter(function each(line) {
+            if (/_maybeInvalidateScore/.test(line)) {
```
jcorbin: yup, this is a bit jank-coupled to unwritten code as of this branch... such is life

PR #999, opened 2015-07-21
Expose default total rps limit in rate limiter

ShanniLi:
r @Raynos @kriskowal @rf 

otherwise, the following command won't work.
node admin.js rate-limiter total-limit default


Raynos:
lgtm.

PR #1000, opened 2015-07-21
Streaming Thrift proposal

prashantv:


docs/thrift.md
```diff
@@ -208,3 +208,47 @@ message with error code (`code:1`) `0x05` (unexpected error).
...
+
+Streaming Thrift allows requests or responses to be streamed over TChannel.
+
+Streaming Thrift will uses a separate `as` header: `sthrift`, and is based
+on the standard Thrift protocol, with the only exception being the encoding
+of arg3 in both the `call req` and `call res`.
+
+The encoding of data for streaming arg3 is a 4 byte length-prefixed chunk:
+```
+chunk~4 chunk~4 chunk~4 ...
```
junchaowu: How does the receiver translates the chunks to each params for arg3 If there are multi arguments for a thrift call?

docs/thrift.md
```diff
@@ -208,3 +208,47 @@ message with error code (`code:1`) `0x05` (unexpected error).
   [`arg1`]: #arg1
   [`arg3`]: #arg3
 
+
+Streaming Thrift
```
Raynos: Consider documenting in a seperate file. One file per "as" is nice.

docs/thrift.md
```diff
@@ -208,3 +208,47 @@ message with error code (`code:1`) `0x05` (unexpected error).
...
+Streaming Thrift will uses a separate `as` header: `sthrift`, and is based
+on the standard Thrift protocol, with the only exception being the encoding
+of arg3 in both the `call req` and `call res`.
+
+The encoding of data for streaming arg3 is a 4 byte length-prefixed chunk:
+```
+chunk~4 chunk~4 chunk~4 ...
+```
+
+Each chunk is a Thrift encoded payload using TBinaryProtocol.
```
Raynos: Each chunk is a Thrift encoded Struct using TBinaryProtocol

docs/thrift.md
Raynos: each chunk is a struct. so if you have a `method(1: string foo, 2: string bar, 3: string baz)`

Then you should expect each chunk to be `{ foo: ..., bar: ..., baz: ... }`

Raynos: It notes that there is only one arg lower down.

docs/thrift.md
```diff
@@ -208,3 +208,47 @@ message with error code (`code:1`) `0x05` (unexpected error).
...
+There is no limit to the number of chunks that can be streamed.
+
+Streaming Thrift methods have the following restrictions:
+ * If the request is streamed, the method IDL should only specify a single
```
Raynos: Consider supporting multiple args in the form of `method(1: string bar, 2: string baz)` by saying that the request must be a struct for all the arguments.

i.e. `{ bar: '', baz: '' } { bar: '', baz: '' } ...`

docs/thrift.md
Raynos: Never mind. you allowed arbitrary types because you only support the first arg.

docs/thrift.md
prashantv: In compiled languages, we would need to generate a struct for the combination of args. While it's trivial, if the user will be required to use a struct anyway, I'd rather they create a struct explicitly.

docs/thrift.md
prashantv: Decided to only allow structs for now. Clearer how it works that way.

docs/thrift.md
prashantv: Moving to sthrift.md

Raynos:
lgtm.

blampe:
:shipit:

docs/sthrift.md
```diff
@@ -0,0 +1,41 @@
...
+```
+chunk~4 chunk~4 chunk~4 ...
+```
+
+Arguments
+---------
+When streaming request arguments, each `arg3` chunk is the encoded payload for
+the method's first (and only) argument.
+For example, if a method defines a single argument that is a string, then
+each chunk is just a Thrift encoded string.
```
kriskowal: Recommend alternately just supporting streaming argument struct. Thrift methods are already accept only one argument (a struct) so going out of the way to pare out a single argument seems weird.

docs/sthrift.md
```diff
@@ -0,0 +1,41 @@
...
+
+
+Responses
+---------
+When streaming successful (`OK`) responses, each `arg3` chunk is the encoded
+payload for the method's result type.
+For example, if a method returns a string, then each chunk should be a
+Thrift encoded string.
```
kriskowal: Also recommend just streaming the result struct, so {0 success} and all error cases are permitted.

Raynos: Streaming multiple errors is going to be confusing.

docs/sthrift.md
Raynos: @prashantv mentioned that allowing only a single argument makes apache thrift integration simpler.

docs/sthrift.md
kriskowal: Buying it. Paring a single entry from result and arg struct is okay by me.

PR #1004, opened 2015-07-21
Connection cleanup (lazy_relay)

Raynos:
Another piece of cleanup ahead of lazy relaying.

r: @jcorbin @shannili @kriskowal

### DIFF
--- a/node/connection.js
+++ b/node/connection.js
@@ -220,7 +220,7 @@ function sendProtocolError(type, err) {
         });
 
         self.handler.sendErrorFrame({
-            id: protocolError.frameId || 0xFFFFFFFF
+            id: protocolError.frameId || v2.Frame.NullId
         }, 'ProtocolError', protocolError.message);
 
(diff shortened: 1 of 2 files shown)


node/connection.js
```diff
@@ -529,14 +534,11 @@ TChannelConnection.prototype.resetAll = function resetAll(err) {
             return;
         }
 
-        var info = {
+        var info = req.extendLogInfo({
```
ShanniLi: Why do you need to add this extra layer of function call?

ShanniLi:
lgtm with a comment.

node/connection.js
Raynos: We are going to store lazy requests in the operations in the future.

Lazy requests do not have a bunch of fields so we need branching between eager vs lazy requests.

kriskowal:
:ok:

jcorbin:
:+1:

PR #1005, opened 2015-07-21
Factor out periodic state pattern

jcorbin:
Lays the ground work to add timer triggering so that we can maintain periodic behavior in a heaped world.

Also adds only-able support to the peer states test for easier dev.

r @Raynos @kriskowal


Raynos:
LGTM

PR #1006, opened 2015-07-21
Initial lazy frame. (lazy_frame)

Raynos:
This PR adds the actual lazy_frame RW.

I added a bit of cleanup and a unit test as well.

r: @kriskowal @jcorbin @shannili

### DIFF
--- a/node/errors.js
+++ b/node/errors.js
@@ -102,6 +102,12 @@ module.exports.ConnectionTimeoutError = TypedError({
     timeout: null
 });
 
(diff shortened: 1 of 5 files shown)


jcorbin:
:+1: however:
- while I get it that setId needs an imperative test, would it be possible / better to use declarative cases for the simple "dose the RW R" (and missing currently W path)?

Raynos:
@jcorbin I added one declarative test and two imperative tests for methods

PR #1007, opened 2015-07-21
Timed periodic

jcorbin:
Final work so that peer states can maintain functionality beyond being short-polled by request scoring:
- add state deactivation hook
- pass channel time heap to states
- implement and use the heaped timeout api in PeriodicState
- pump period checking from other event trigger paths to maintain expected fast edge triggering (e.g. the probe back to healthy case)

### DIFF
--- a/node/circuits.js
+++ b/node/circuits.js
@@ -117,6 +117,7 @@ function Circuits(options) {
     self.config = options.config || {};
 
     self.stateOptions = new states.StateOptions(null, {
+        timeHeap: options.timeHeap,
         timers: options.timers,
         random: options.random,
         nextHandler: alwaysShouldRequestHandler,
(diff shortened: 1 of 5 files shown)


kriskowal:
How does this behave if an unhealthy handler receives one request every two seconds, with a one second interval between checks? The former semantics allowed the period to be extended to the first received request, so a handler would return to good standing after five consecutive healthy requests over the course of five periods of one second or longer. I concluded there was nothing to be gained from the additional overhead of tracking timers.

kriskowal:
:ok: Need to circle back and make sure we're covering the scenario of a circuit being hit less frequently than the period, to make sure that an unhealthy service still recovers after consecutive successful requests, even if there are empty periods between. Impact would be an alert on a circuit failing to recover, though it would not be an operational problem for a handler receiving less than a request per period being rate limited to a request per period.

PR #1008, opened 2015-07-21
Fix circuits injection

jcorbin:
Claws back instantiation of Circuits from consumers by explicating the config which they supply (e.g. hyperbahn).


Raynos:
lgtm

kriskowal:
legit

PR #1009, opened 2015-07-21
test/peer_states: add test for probing back to health with slower than period

jcorbin:
Per @kriskowal 's observation in #1007

### DIFF
--- a/node/test/peer_states.js
+++ b/node/test/peer_states.js
@@ -252,6 +252,68 @@ testSetup('consecutive success during unhealthy periods restores health', {}, fu
 
 });
 
+testSetup('consecutive success (WITH PAUSES) during unhealthy periods restores health', {}, function t(cluster, assert) {
+    var peer = cluster.client.peers.get(cluster.hosts[1]);
+
+    var Steps = [
+        function failIt(done) {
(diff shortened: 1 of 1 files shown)


jcorbin:
TBD: whether this is actually what we want, my intuition says we should have some sort of decay so that it takes more slow-rate probes to get back to good... but maybe this is actually what we want.

node/test/peer_states.js
```diff
@@ -252,6 +252,68 @@ testSetup('consecutive success during unhealthy periods restores health', {}, fu
...
+
+        function checkBefore(done) {
+            peer.channel.timers.advance(1000);
+            assert.equals(peer.state.shouldRequest(), 1.0, 'expected probe score before try ' + trial);
+            assert.equals(peer.state.type, 'tchannel.unhealthy', 'unhealthy before try ' + trial);
+
+            done();
+        },
+
+        cluster.send('glad'),
```
Raynos: Why does a single glad not transition it back to healthy ?

Raynos:
lgtm.

kriskowal:
woot. ask and you shall indeed receive!

PR #1011, opened 2015-07-21
Improve retry test

jcorbin:
Became large enough to pull out ahead of #915


node/test/retry.js
```diff
@@ -172,32 +186,20 @@ allocCluster.test('request application retries', {
             }
         }
     });
+    withConnectedClient(chan, cluster, onConnected);
```
Raynos: Why this change ? Retries should work without connected clients.

PR #1013, opened 2015-07-21
TChannelRequest: keep triedRemoteAddrs null until needed

jcorbin:
This makes heaped peer selection use its faster non-predicate filtered branch in the more common (first try) case.


Raynos:
lgtm

PR #1014, opened 2015-07-22
Fix peer balancing regression

jcorbin:
The heaped score work in v2.4.0 broke peer balance:
- peers only got rescored once in PeerHeap#choose
- this happens before request allocation (pending count mutation)
- scores weren't being recomputed by the "started a request" path anymore

This PR fixes those issues.

We should add a test for peer balancing so that we don't break this again, and to check this fix.

### DIFF
--- a/node/peer_heap.js
+++ b/node/peer_heap.js
@@ -51,17 +51,6 @@ PeerHeap.prototype.choose = function choose(threshold, filter) {
         return null;
     }
 
-    // TODO: rescore could be deferred:
-    // - store a pendingRescore bit
-    // - set a next tick to rescore and clear bit
-    // - if asked to choose a peer while pendingRescore is still set, take the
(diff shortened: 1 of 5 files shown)


node/peer_heap.js
```diff
@@ -51,17 +51,6 @@ PeerHeap.prototype.choose = function choose(threshold, filter) {
...
 
-    // TODO: rescore could be deferred:
-    // - store a pendingRescore bit
-    // - set a next tick to rescore and clear bit
-    // - if asked to choose a peer while pendingRescore is still set, take the
-    //   hit and rescore then (do not clear bit)
-    //
-    // However such cleverness might just slow down the throughput of a very
-    // busy server rather than help.
-    el.score = el.peer.state.shouldRequest();
```
Raynos: Why was this removed.

Raynos:
Legit.

PR #1017, opened 2015-07-22
PreferOutgoingHandler: unify "not fully ready" tiers

jcorbin:
Fixes an edge case in early peer balancing: when a client library has a batch of requests to kick off with no existing connections, they all get allocated to the first peer to win random selection, since:
- the first score picks a randomized unconnected peer...
- ...and connects to it, changing its tier...
- ...so that any pending counts become irrelevant, preferring the still-connecting peer

Separating these 3 cases as separate tiers really didn't give us anything, since until we've exchanged a first init req/res with a peer we really can't say if it'll be any better than the others.

Clients wishing to control or bound to a set of connecting peers will need to implement such logic manually at present, as the hyperbahn client currently does.  We could choose in the future to instead place a cap on the number of pending new connections, and at that point prefer to select a still-connecting peer in preference to an unconnected one.

r @Raynos @kriskowal 

cc @abhinav @prashantv

### DIFF
--- a/node/peer.js
+++ b/node/peer.js
@@ -462,7 +462,9 @@ PreferOutgoingHandler.prototype.shouldRequest = function shouldRequest() {
         case QOS_UNCONNECTED:
             return 0.1 + random * 0.1;
         case QOS_ONLY_INCOMING:
-            self.peer.connect();
+            if (!self.peer.channel.destroyed) {
+                self.peer.connect();
+            }
             return 0.2 + random * 0.1;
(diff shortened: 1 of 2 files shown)


kriskowal:
:ok:

Raynos:
This will cause issues for hyperbahn clients.

We need some logic in peers akin to "maximum sockets" or otherwise everything becomes fully connected.

This is a breaking change and we have to fix it before cutting a new version.

PR #1019, opened 2015-07-22
Add a prefer incoming strategy

jcorbin:
Intended for use in the Hyperbahn client as the primary first use case.

### DIFF
--- a/node/peer_score_strategies.js
+++ b/node/peer_score_strategies.js
@@ -54,6 +54,8 @@ PreferOutgoing.prototype.getTier = function getTier() {
 PreferOutgoing.prototype.getScore = function getScore() {
     var self = this;
 
+    // TODO: support pruning old incoming conns
+
     // space:
     //   [0.1, 0.4)  peers with no identified outgoing connection
     //   [0.4, 1.0)  identified outgoing connections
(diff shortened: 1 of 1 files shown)


kriskowal:
Strongly advise consolidating this into a biased strategy (with a direction parameter), as opposed to an unbiased strategy (as would be used for Ringpop subpeers)

node/peer_score_strategies.js
```diff
@@ -74,3 +76,57 @@ PreferOutgoing.prototype.getScore = function getScore() {
...
+PreferIncoming.READY_INCOMING = 3;
+
+PreferIncoming.prototype.getTier = function getTier() {
+    var self = this;
+
+    var inconn = self.peer.getInConnection();
+    var outconn = self.peer.getOutConnection();
+
+    if (!inconn && !outconn) {
+        return PreferIncoming.UNCONNECTED;
```
Raynos: Let's favor UNCONNECTED lower then connected.

This leads to bugs where `choosePeer()` picks a `Peer` and then gets an ECONNREFUSED.

Raynos:
We already shipped a prefer incoming strategy in a different PR.

PR #1023, opened 2015-07-23
go: fix init race condition by processing initRes in handleInitReq

prashantv:
Was running into failures when adding state validation before handling call reqs in another branch.
Fixed + made ping check state + added test using ping to validate.

r @mmihic


mmihic:
late, but lgtm

PR #1024, opened 2015-07-23
document Shard key

Raynos:
We currently have a tchannel+ringpop service that implements a ringpop forwarding handler.

This ringpop forwarding handler reads the sharding key out of the transport headers to determine where to forward the request.

I would like to add the shard key to the transport headers to enable efficient forwarding without having to read arg{1..3}.

cc @jcorbin @jwolski @mranney @blampe

### DIFF
--- a/docs/protocol.md
+++ b/docs/protocol.md
@@ -623,6 +623,7 @@ name  | req | res | description
 `re`  | Y   | N   | Retry Flags
 `se`  | Y   | N   | Speculative Execution
 `fd`  | Y   | Y   | Failure Domain
+`sk`  | Y   | N   | Shard key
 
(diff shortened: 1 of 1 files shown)


jwolski:
lgtm

kriskowal:
:ok: by me. Fast relaying within Ringpop is a win.

blampe:
:shipit:

mranney:
There are some other requests for this as well, so I think it's a good idea.

Do you propose rolling this out as part of the current protocol version?

Raynos:
@mranney 

Yes; I think it would be good to add this key to the protocol doc.

v2 currently has forward compatibility for adding as many transport headers as you want. All this does is standardizes on the header name.

PR #1026, opened 2015-07-23
Drop shouldRequest arguments

jcorbin:
r @Raynos @kriskowal 

Dropping them while they're still not used, now that they're not even passed in most cases.


kriskowal:
:ok:

PR #1028, opened 2015-07-23
Timer improvements

jcorbin:
- make periodic states set no timers when idle
- harden OperationTombstone deletion path, and clear its .operations field

r @Raynos @kriskowal


Raynos:
lgtm

PR #1034, opened 2015-07-23
Improve cut

jcorbin:
- creates and publishes a tarball rather than cwd
- therefore longer switches your branch
- accept optional second arg being the npm tag to update (defaults to latest)

r @Raynos


kriskowal:
Sweet, thanks for discovering that npm lets you pass a tarball.

PR #1035, opened 2015-07-24
Add a gaurd and warn for double emitFInish

Raynos:
r: @jcorbin @kriskowal

### DIFF
--- a/node/out_response.js
+++ b/node/out_response.js
@@ -217,11 +217,12 @@ TChannelOutResponse.prototype.sendError = function sendError(codeString, message
 
 TChannelOutResponse.prototype.emitFinish = function emitFinish() {
     var self = this;
+    var now = self.timers.now();
 
     if (self.end) {
         self.logger.warn('out response double emitFinish', {
             end: self.end,
-            now: Date.now(),
+            now: now,
             serviceName: self.inreq.serviceName,
             cn: self.inreq.headers.cn,
             endpoint: String(self.inreq.arg1),
@@ -234,7 +235,7 @@ TChannelOutResponse.prototype.emitFinish = function emitFinish() {
         return;
     }
 
-    self.end = self.timers.now();
+    self.end = now;
 
     var latency = self.end - self.inreq.start;
 
--- a/node/out_response.js
+++ b/node/out_response.js
@@ -218,9 +218,23 @@ TChannelOutResponse.prototype.sendError = function sendError(codeString, message
 TChannelOutResponse.prototype.emitFinish = function emitFinish() {
     var self = this;
 
-    if (!self.end) {
-        self.end = self.timers.now();
+    if (self.end) {
+        self.logger.warn('out response double emitFinish', {
+            end: self.end,
+            now: Date.now(),
+            serviceName: self.inreq.serviceName,
+            cn: self.inreq.headers.cn,
+            endpoint: String(self.inreq.arg1),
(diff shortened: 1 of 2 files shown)


node/out_response.js
```diff
@@ -218,10 +218,24 @@ TChannelOutResponse.prototype.sendError = function sendError(codeString, message
...
 
-    if (!self.end) {
-        self.end = self.timers.now();
+    if (self.end) {
+        self.logger.warn('out response double emitFinish', {
+            end: self.end,
+            now: Date.now(),
+            serviceName: self.inreq.serviceName,
+            cn: self.inreq.headers.cn,
+            endpoint: String(self.inreq.arg1),
```
jcorbin: that already exists as self.inreq.endpoint I found out, something @rf introduced towards tracing... it seems like a useful enough eager parsing to keep around and support for the eager req path, I plan to double down on it when killing arg1 streaming, thoughts?

node/out_response.js
```diff
@@ -218,10 +218,24 @@ TChannelOutResponse.prototype.sendError = function sendError(codeString, message
 TChannelOutResponse.prototype.emitFinish = function emitFinish() {
     var self = this;
 
-    if (!self.end) {
-        self.end = self.timers.now();
+    if (self.end) {
+        self.logger.warn('out response double emitFinish', {
+            end: self.end,
+            now: Date.now(),
```
jcorbin: use channel.timers.now()

jcorbin: in fact hoist a `var now = self.channel.timers.now();` and de-dupe with the update call down below this error check branch

node/out_response.js
```diff
@@ -218,10 +218,24 @@ TChannelOutResponse.prototype.sendError = function sendError(codeString, message
...
+            serviceName: self.inreq.serviceName,
+            cn: self.inreq.headers.cn,
+            endpoint: String(self.inreq.arg1),
+            codeString: self.codeString,
+            errorMessage: self.message,
+            remoteAddr: self.inreq.connection.socketRemoteAddr,
+            state: self.state
```
jcorbin: how about renaming this to 'outResState' and also adding an `inReqState`?

jcorbin:
While we're at it, let's do the exact same thing for [`TChannelInRequest#emitFinish`](https://github.com/uber/tchannel/blob/master/node/in_request.js#L117):
- it's arguably an oversight that inreq has a .start but no .end
- you could at least branch on whether or not state is already done, however if you go this route please audit all state mutations to ensure we're not doing something stupid (like setting state Done, then calling emitFinish, which would inadvertently then be a noop)

jcorbin:
Bonus points if you start another instance of the `extendLogInfo :: (info) -> info` pattern so that the inreq path and outres paths can disentangle all that field augmentation.

node/out_response.js
Raynos: yes lets use `endpoint` everywhere. Just fitting with local style because what even is consistency

node/out_response.js
Raynos: will fix.

node/out_response.js
Raynos: Good idea.

Raynos:
InRequest should not finish twice.. I think InResponse finished twice because we write to it twice.

I could check for OutRequest with emitError and emitResponse

PR #1036, opened 2015-07-24
Add netcat style listener for tchannel

prashantv:
Run a command and pass arg3 as stdin and use output as the response's arg3.

(will add short flags as a separate change, as it requires 3rd party libraries)

cc: @jcorbin @abhinav


abhinav:
:+1:

PR #1037, opened 2015-07-24
Debug tombstone leak

jcorbin:
cc @Raynos @kriskowal


jcorbin:
This helps to demonstrate that the tombstone link is in the relay path, but not the client path; when you uncomment the two OpKindMonitor blocks and run:
```
$ node benchmarks/index.js --relay -- --relay --skipPing -m 3 -s 4,4096 -p 1000,10000,20000
```

We see that each client stays stable (around 3500 for me), while the relay channel leaks.

PR #1038, opened 2015-07-24
Simplify circuits handling api, tighter integration with relay handling

jcorbin:
- dropped the whole handler chaining / nextHandler notion
- instead only buildRes transformers / wrappers
- tighter coupling between relay handler and circuits

r @kriskowal


kriskowal:
:ok:

node/circuits.js
```diff
@@ -202,16 +211,7 @@ inherits(Circuit, EventEmitter);
...
-    var self = this;
-    if (self.state.shouldRequest()) {
-        return self.monitorRequest(req, buildRes, nextHandler);
-    } else {
-        return buildRes().sendError('Declined', 'Service is not healthy');
-    }
-};
-
-Circuit.prototype.monitorRequest = function monitorRequest(req, buildRes, nextHandler) {
+Circuit.prototype.monitorRequest = function monitorRequest(req, buildRes) {
```
kriskowal: It looks like this monitorRequest overwrites the prior monitorRequest

kriskowal: Oh, nope. Circuits vs Circuit

PR #911, opened 2015-07-10
WIP: node: implement lazy frame r/w and relaying

jcorbin:
r @Raynos @kriskowal

PR #1040, opened 2015-07-24
No more arg1 streaming

jcorbin:
Towards #911, and a pre-req of #1038

Drops support for arg1 streaming.

r @kriskowal

### DIFF
--- a/node/argstream.js
+++ b/node/argstream.js
@@ -49,20 +49,15 @@ function ArgStream() {
     self.frameEvent = self.defineEvent('frame');
     self.finishEvent = self.defineEvent('finish');
 
-    self.arg1 = StreamArg();
     self.arg2 = StreamArg();
     self.arg3 = StreamArg();
 
-    self.arg1.on('error', passError);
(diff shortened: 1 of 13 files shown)


jcorbin:
Okay dear reviewers, should I:
- leave `req.withArg1((err, arg1) -> void)` for *hand wave* external compatability
- or should I say "the external layer are the as handlers, so let's drop the old path"?

kriskowal:
@jcorbin leave the thunk. I don’t want surprises.

jcorbin:
Understood, added deprecation comments at least

kriskowal:
:ok: Full of win.

Raynos:
noice.

Raynos:
@jcorbin please update https://github.com/uber/tchannel/blob/master/node/docs/streaming.md

Just need to change docs to mention buffers instead of streams.

Raynos:
Also because `Stability: unstable` this is not a major bump; probably a minor bump though.

PR #1045, opened 2015-07-24
advertise in http ingress example

cacktopus:


node/examples/http_ingress.js
```diff
@@ -77,14 +79,31 @@ var svcChan = tchan.makeSubChannel({
...
 asHTTP.setHandler(svcChan, onRequest);
 
+function advertise() {
+    var hyperbahnClient = HyperbahnClient({
+        tchannel: tchan,
+        serviceName: serviceName,
+        hostPortList: [argv.advertise],
+        hardFail: true
+    });
+    hyperbahnClient.once('advertised', function adv() {
```
jcorbin: maybe also have a `hyperbahnClient.once('error', ...)` and shut things down if it didn't work out (I hope that's possible)?  In other news @Raynos , how come client.advertise doesn't take a (maybe optional) callback?

Raynos: because we advertise in a loop. we emit the advertised event multiple times.

Raynos:
Going to close this out.

PR #1048, opened 2015-07-25
add logging around p2p state transitions

Raynos:
r: @jcorbin @kriskowal @rf @shannili


ShanniLi:
lgtm

ShanniLi:
we may need more logging, e.g., why it is doing state transition

PR #1049, opened 2015-07-25
Add logging for as-http

anson627:
Make as-http logging verbose for now so that we can narrow down issues.
Skip statsd for now to avoid burning cpu

@Raynos @ShanniLi @jcorbin 
cc: @davewhat @vipulaneja

### DIFF
--- a/node/as/http.js
+++ b/node/as/http.js
@@ -67,7 +67,7 @@ function TChannelHTTP(options) {
 
 TChannelHTTP.prototype.sendRequest = function send(treq, hreq, options, callback) {
     assert(treq.streamed, 'as http must have a streamed tchannel request');
-
+    var self = this;
     if (typeof options === 'function') {
         callback = options;
         options = null;
@@ -83,6 +83,9 @@ TChannelHTTP.prototype.sendRequest = function send(treq, hreq, options, callback
     var arg1 = ''; // TODO: left empty for now, could compute circuit names heuristically
     var arg2res = bufrw.toBufferResult(HTTPReqArg2.RW, head);
     if (arg2res.err) {
+        self.logger.error('Buffer write for arg2 failed', {
+            error: arg2res.err
+        });
         callback(arg2res.err, null, null);
         return;
     }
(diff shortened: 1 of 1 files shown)


node/as/http.js
```diff
@@ -156,13 +164,14 @@ TChannelHTTP.prototype.sendResponse = function send(buildResponse, hres, callbac
...
+    self.logger = tchannel.logger;
     tchannel.handler = new AsHTTPHandler(self, tchannel, handler);
     return tchannel.handler;
 };
 
-TChannelHTTP.prototype.forwardToTChannel = function forwardToTChannel(tchannel, hreq, hres, tchannelRequestOptions, callback) {
+TChannelHTTP.prototype.forwardToTChannel = function forwardToTChannel(tchannel, hreq, hres, requestOptions, callback) {
     var self = this;
-
+    self.logger = self.logger || tchannel.logger;
```
ShanniLi: this line seems unnecessary. just pass in the logger at the constructor.

node/as/http.js
```diff
@@ -214,14 +227,16 @@ TChannelHTTP.prototype.forwardToTChannel = function forwardToTChannel(tchannel,
             }
             hres.writeHead(head.statusCode, head.message, headers);
             body.pipe(hres);
+            self.logger.info('Forwarding to tchannel succeeded');
```
Raynos: You will probably regret this one. Consider debug() instead.

node/as/http.js
```diff
@@ -243,7 +258,7 @@ TChannelHTTP.prototype.forwardToHTTP = function forwardToHTTP(ingressChan, optio
     function onResponse(inres) {
         if (!sent) {
             sent = true;
-            // TODO: observable
+            self.logger.info('Forwarding to HTTP succeeded');
```
Raynos: consider debug() instead

node/as/http.js
```diff
@@ -322,13 +346,15 @@ AsHTTPHandler.prototype.handleRequest = function handleRequest(req, buildRespons
...
-        // TODO: observable
+
+        self.logger.debug('Handling request started');
         self.handler.handleRequest(hreq, hres);
     }
 
     function sendResponse(hres) {
         if (!sent) {
             sent = true;
+            self.logger.info('Handling request succeeded');
```
Raynos: consider debug()

ShanniLi:
lgtm. My comments are non-blocking.
You also consider add stats as well.

node/as/http.js
anson627: as-http will always use the same logger as tchannel, so logger is not passed in from TChannelHTTP constructor.

anson627:
will add statsd when it's less costly

Raynos:
@anson627  we have stats enabled for all call req & call res. You already have stats.

ShanniLi:
@Raynos I meant as-http specific ones.

However, I never looked at the call req & res stats for xlate, so what we have may be enough already .

PR #1050, opened 2015-07-25
Wrap around bufrw error

anson627:
Change codeString of sendError to Declined
Wrap around bufrw error
@Raynos @ShanniLi @jcorbin 
cc: @davewhat @vipulaneja

### DIFF
--- a/node/as/http.js
+++ b/node/as/http.js
@@ -83,10 +83,17 @@ TChannelHTTP.prototype.sendRequest = function send(treq, hreq, options, callback
     var arg1 = ''; // TODO: left empty for now, could compute circuit names heuristically
     var arg2res = bufrw.toBufferResult(HTTPReqArg2.RW, head);
     if (arg2res.err) {
+<<<<<<< HEAD
         self.logger.error('Buffer write for arg2 failed', {
             error: arg2res.err
         });
         callback(arg2res.err, null, null);
+=======
+        var toBufferErr = errors.HTTPReqArg2toBufferError(arg2res.err, {
+            head: head
+        });
+        callback(toBufferErr, null, null);
+>>>>>>> Wrap around bufrw error
         return;
     }
     var arg2 = arg2res.value;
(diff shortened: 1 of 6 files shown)


node/as/http.js
```diff
@@ -335,9 +346,8 @@ AsHTTPHandler.prototype.handleRequest = function handleRequest(req, buildRespons
 
     function sendError(err) {
         if (!sent) {
-            // TODO: map type?
             sent = true;
-            buildResponse().sendError('UnexpectedError', err.message);
+            buildResponse().sendError('Declined', err.message);
```
Raynos: This should not be declined. If you want this semantics please `sendError(err, codeName)`

Raynos: Or use `errors.classify(err)`

node/errors.js
```diff
@@ -136,6 +136,34 @@ module.exports.EphemeralInitResponse = TypedError({
     processName: null
 });
 
+module.exports.HTTPReqArg2fromBufferError = WrappedError({
+    type: 'tchannel-http-handler.from-buffer-arg2.req-failed',
```
ShanniLi: maybe, tchannel.http-handler.from-buffer-arg2.req-failed?

ShanniLi:
lgtm. comments are optional.

node/as/http.js
```diff
@@ -361,12 +369,11 @@ AsHTTPHandler.prototype.handleRequest = function handleRequest(req, buildRespons
 
     function sendError(err) {
         if (!sent) {
-            // TODO: map type?
             sent = true;
             self.logger.warn('Handling request failed', {
                 error: err
             });
-            buildResponse().sendError('UnexpectedError', err.message);
+            buildResponse().sendError(errors.classify(err), err.message);
```
Raynos: This returns `null` if it cant be classified.

Please default to `'UnexpectedError'` to avoid problems.

node/errors.js
```diff
@@ -179,7 +207,7 @@ module.exports.InvalidHeaderTypeError = TypedError({
 });
 
 module.exports.InvalidJSONBody = TypedError({
-    type: 'tchannel-handler.json.invalid-body',
+    type: 'tchannel.json-handler.invalid-body',
```
Raynos: No. We never mutate these strings, do not change them.

Raynos:
This breaks master; reverting.

PR #1052, opened 2015-07-27
move rate limiter state mutations to functions

ShanniLi:
r @Raynos @kriskowal @rf

### DIFF
--- a/node/hyperbahn/service_proxy.js
+++ b/node/hyperbahn/service_proxy.js
@@ -512,13 +512,13 @@ function destroy() {
 ServiceDispatchHandler.prototype.enableRateLimiter =
 function enableRateLimiter() {
     var self = this;
-    self.rateLimiter.rateLimiterEnabled = true;
+    self.rateLimiterEnabled = true;
 };
 
 ServiceDispatchHandler.prototype.disableRateLimiter =
 function disableRateLimiter() {
     var self = this;
-    self.rateLimiter.rateLimiterEnabled = false;
+    self.rateLimiterEnabled = false;
 };
 
 // TODO Consider sharding by hostPort and indexing exit exitNodes by hostPort.
--- a/node/hyperbahn/service_proxy.js
+++ b/node/hyperbahn/service_proxy.js
@@ -516,7 +516,7 @@ function enableRateLimiter() {
 };
 
 ServiceDispatchHandler.prototype.disableRateLimiter =
-function enableRateLimiter() {
+function disableRateLimiter() {
     var self = this;
     self.rateLimiter.rateLimiterEnabled = false;
 };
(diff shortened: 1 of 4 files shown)


node/rate_limiter.js
```diff
@@ -124,10 +124,35 @@ function removeServiceCounter(serviceName) {
...
+        self.updateServiceLimit(name, rpsLimitForServiceName[name]);
+    }
+};
+
 RateLimiter.prototype.updateServiceLimit =
 function updateServiceLimit(serviceName, limit) {
     var self = this;
-    self.rpsLimitForServiceName[serviceName] = limit;
+
+    if (limit === undefined) {
```
anson627: when would limit become undefined?

node/hyperbahn/service_proxy.js
```diff
@@ -509,6 +509,18 @@ function destroy() {
...
 };
 
+ServiceDispatchHandler.prototype.enableRateLimiter =
+function enableRateLimiter() {
+    var self = this;
+    self.rateLimiter.rateLimiterEnabled = true;
+};
+
+ServiceDispatchHandler.prototype.disableRateLimiter =
+function enableRateLimiter() {
```
anson627: typo?

node/rate_limiter.js
anson627: can limit be 0? Is that what we are trying to do?

node/hyperbahn/service_proxy.js
ShanniLi: good catch! fixed.

node/rate_limiter.js
ShanniLi: I don't really have strong preference. Allowing 0 let us disable a service completely. That is a flexibility we may want.

anson627: cool, just curious

anson627:
lgtm

node/rate_limiter.js
```diff
@@ -124,10 +124,35 @@ function removeServiceCounter(serviceName) {
...
 };
 
+RateLimiter.prototype.updateExemptServices =
+function updateExemptServices(exemptServices) {
+    var self = this;
+    self.exemptServices = exemptServices;
+};
+
+RateLimiter.prototype.updateRpsLimitForServiceName =
+function updateRpsLimitForServiceName(rpsLimitForServiceName) {
```
Raynos: Let's name this more clearly.

`updateRpsLimitForAllServices`

node/rate_limiter.js
Raynos: it's cleaner to check for `if (limit === null)` and then have `updateServiceLimit('foo', null)`

kriskowal: undefined is appropriate.

kriskowal:
Looks good. I concur with Jake’s recommendation for the method name.

Raynos:
#legit

PR #1055, opened 2015-07-27
Verify that all tests files are imported.

Raynos:
This PR adds a script that verifies that all tests are
imported. It makes a few assumptions

 - all files in test/**/*.js are test files that contain
    `tape` blocks.
 - any files in lib/*.js are non-test files that contain
    NO `tape` blocks.
 - All tests are imported from `test/index.js`.
 - All tests are imported as `require(./PATH.js)`

I've had to inline a couple of folder imports into the
test/index.js for this to work. I've also had to move
a few helper files into lib folder.

I've found two test files that were never imported and
made them pass.

r: @kriskowal @shannili @rf


kriskowal:
:ok:

ShanniLi:
lgtm

node/scripts/verify_included.sh
```diff
@@ -0,0 +1,28 @@
+#!/usr/bin/env bash
+set -e
+# set -x
+
+FILES=$(
+    ls --color=never test/**/*.js test/*.js | \
```
kriskowal: This line is an error on “BSD”. illegal option -- -.

kriskowal:
I tacked another commit on to make this work on “BSD”. Rebase if you wish.

PR #1058, opened 2015-07-28
Remove the peer health state machine

kriskowal:
Still failing a send test. Looks like the pending request counts are not being tracked properly.


kriskowal:
This is ready for review now. This removes the peer health state machine, which bypasses a level of indirection in peer selection. This also makes `peer` a “required option” for requests, replacing peerOptions.

r @raynos @ShanniLi @rf

kriskowal:
Note that this leaves the circuit breaking state machine in tact without alteration—just removing the peer health state machine.

node/channel.js
```diff
@@ -721,17 +720,18 @@ TChannel.prototype._request = function _request(opts) {
         req = opts.peer.request(opts);
     } else if (opts.host) {
         opts.retryCount = 0;
-        req = self.peers.add(opts.host).request(opts);
+        opts.peer = self.peers.add(opts.host);
```
ShanniLi: Not sure why the peer needs to be assigned to opts. It doesn't seem a good idea to change the request parameter.

kriskowal: I am considering a follow-up where the channel and peer are required positional arguments for Request instead of “required options”.

ShanniLi: I see you are using it in peer.request. Will be using extend on opts better here?

kriskowal: I’m sure we should move toward using an options shape instead of an ad hoc dictionary.

node/peer.js
```diff
@@ -460,7 +409,6 @@ PreferOutgoingHandler.prototype.getQOS = function getQOS() {
     }
 };
 
-// Consulted depending on the peer state
```
ShanniLi: There seems to have a bug in PreferOutgoingHandler.prototype.getQOS.
self.peer.getOutConnection() should be called with parameter "true", otherwise, it may return unidentified out connection even there is an identified out connection.

ShanniLi: Also, I am not sure I understand, why out connection is preferred compared to in connection in shouldRequest?

ShanniLi:
Finished review. The question about PreferOutgoingHandler.prototype.getQOS and shouldRequest is a blocking comment.

node/peer.js
kriskowal: Hyperbahn is supposed to prefer to speak to services by its outgoing connections, and services are supposed to prefer to speak to Hyperbahn by their incoming connections. This allows Hyperbahn to manage the set of open connections in response to ring membership changes. This feature is far from complete. We still need to implement "drain and close" for wrong-way connections, when those connections are replaced by connections in the preferred direction.

ShanniLi: Sounds good. Please update the following line:
var outconn = self.peer.getOutConnection();
to 
var outconn = self.peer.getOutConnection(true);

ShanniLi: I talked to @kriskowal. The bug is not in this PR, so I will create a different PR to fix it.

ShanniLi:
lgtm

node/channel.js
Raynos: Make sure `opts.peer` is a field in `RequestOptions()`

Raynos:
lgtm

node/channel.js
kriskowal: Verified

Issue #1060, opened 2015-07-28
Node: Memory leaked caused by `handler.streamingReq` and `handler.streamingRes`


Raynos:
We keep an object of inreq / inres for every streaming request.

This leaks a ton of memory.

We need to remove the req from the object when we are finished streaming. i.e. `req.state === States.DONE`

PR #1067, opened 2015-07-28
fix the memory leak in handler.streamingReq and handler.streamingRes

ShanniLi:
r @Raynos @kriskowal 
#1060

### DIFF
--- a/node/v2/handler.js
+++ b/node/v2/handler.js
@@ -197,6 +197,8 @@ TChannelV2Handler.prototype.handleCallRequest = function handleCallRequest(reqFr
     if (handled) {
         if (req.state === States.Streaming) {
             self.streamingReq[req.id] = req;
+            req.finishEvent.on(onRequestDone);
+            req.errorEvent.on(onRequestDone);
         }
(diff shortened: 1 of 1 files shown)


node/v2/handler.js
```diff
@@ -325,9 +331,15 @@ TChannelV2Handler.prototype.handleCallResponse = function handleCallResponse(res
     if (handled) {
         if (res.state === States.Streaming) {
             self.streamingRes[res.id] = res;
+            res.finishEvent.on(onResponseDone);
+            res.errorEvent.on(onResponseDone);
         }
```
Raynos: You could just do `else { delete self.streamingRes[id]; }`

kriskowal:
:ok: Curious about @raynos’s observation. This approach seems intuitive.

node/v2/handler.js
ShanniLi: if we do else branch, then we need to do it both on if (handled)  and if (res.state === States.Streaming).

And as @kriskowal pointed out, the "else" approach is less intuitive.

Raynos:
lgtm.

PR #1072, opened 2015-07-28
Score within ranges for each pending requests tier

kriskowal:
A peer with fewer pending requests will always be favored over another
with more pending requests, and this provides a uniform probability
distribution for peers with the same number of pending requests.

This algorithm divides the score interval (0, 1) into an infinite series
of equivalence classes per quantity of pending requests. So, (0.5, 1)
for 0 pending requests, (0.25, 0.5) for 1 pending request, (0.125, 0.25)
for 2 pending requests, ad nauseaum.


Raynos:
lgtm

PR #1073, opened 2015-07-28
ALlow us to take the old choose peer logic back

Raynos:
I also fixed a few tests.

r: @kriskowal @shannili

### DIFF
--- a/node/channel.js
+++ b/node/channel.js
@@ -147,6 +147,7 @@ function TChannel(options) {
     self.emitConnectionMetrics =
         typeof self.options.emitConnectionMetrics === 'boolean' ?
         self.options.emitConnectionMetrics : true;
+    self.chooseHeapPeers = self.options.chooseHeapPeers || false;
 
     // Filled in by the listen call:
     self.host = null;
--- a/node/channel.js
+++ b/node/channel.js
@@ -147,7 +147,7 @@ function TChannel(options) {
     self.emitConnectionMetrics =
         typeof self.options.emitConnectionMetrics === 'boolean' ?
         self.options.emitConnectionMetrics : true;
-    self.chooseHeapPeers = self.options.chooseHeapPeers || false;
+    self.choosePeerWithHeap = self.options.choosePeerWithHeap || false;
 
     // Filled in by the listen call:
     self.host = null;
--- a/node/out_request.js
+++ b/node/out_request.js
@@ -83,7 +83,7 @@ function TChannelOutRequest(id, options) {
         self.setupTracing(options);
     }
 
-    self.peer.invalidateScore();
+    // self.peer.invalidateScore();
 }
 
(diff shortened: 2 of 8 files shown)


ShanniLi:
lgtm. It would be nice to provide an interface to update chooseHeapPeers during runtime.

Raynos:
@ShanniLi agreed; we need that interface, in the future :)

node/channel.js
```diff
@@ -147,6 +147,7 @@ function TChannel(options) {
     self.emitConnectionMetrics =
         typeof self.options.emitConnectionMetrics === 'boolean' ?
         self.options.emitConnectionMetrics : true;
+    self.chooseHeapPeers = self.options.chooseHeapPeers || false;
```
kriskowal: useHeap, choosePeerWithHeap

node/out_request.js
```diff
@@ -83,7 +83,7 @@ function TChannelOutRequest(id, options) {
         self.setupTracing(options);
     }
 
-    self.peer.invalidateScore();
+    // self.peer.invalidateScore();
```
kriskowal: deliberate?

Raynos: yes.

kriskowal: Perhaps explain or delete the line. Maybe leave comment on why.

Raynos: delete line; will fix.

node/test/retry.js
```diff
@@ -99,35 +78,33 @@ allocCluster.test('request retries', {
...
-            req.outReqs[0] &&
-            req.outReqs[0].err &&
-            req.outReqs[0].err.type,
-            'tchannel.declined',
-            'expected first request to decline');
+        var reqErrors = req.outReqs.filter(function isErr(r) {
+            return r.err;
+        });
+        var reqResp = req.outReqs.filter(function isResp(r) {
+            return r.res;
```
kriskowal: isRes, hasRes. Let’s not make new abbreviations for response.

node/test/retry.js
```diff
@@ -99,35 +78,33 @@ allocCluster.test('request retries', {
...
+                'expected error request'
+            );
+        }
+
+        assert.ok(reqResp[0] &&
+                  reqResp[0].res, 'expected to have 4th response');
+        assert.deepEqual(reqResp[0] &&
+                         reqResp[0].res.arg2, arg2, 'arg2 came form 4th response');
+        assert.deepEqual(reqResp[0] &&
+                         reqResp[0].res.arg3, arg3, 'arg3 came form 4th response');
```
kriskowal: Although, @jcorbin will likely shed a tear, this does seem less fragile going forward, as we test the intended behavior and not implementation details that alternate algorithms might not exhibit.

kriskowal:
:ok:

PR #1074, opened 2015-07-28
Document new peer score random variable

kriskowal:
Unfortunately, it is likely to be short lived. Working this out, we have discovered a fatal flaw of the heap. It tends to squeeze unlucky peers out of the selection pool. Receiving a low score also decreases its probability of being invalidated (because we only invalidate score when a peer takes a request). We are going to explore an alternate approach that keeps peers in an array with subranges dedicated to equivalence classes, and then choose a request from the best equivalence class with a uniform random variable.


kriskowal:
r @jcorbin @raynos @shannili

Raynos:
lgtm.

ShanniLi:
:+1:

PR #1088, opened 2015-07-30
stats for rate limiting

ShanniLi:
r @Raynos @kriskowal

### DIFF
--- a/node/channel.js
+++ b/node/channel.js
@@ -145,11 +145,7 @@ function TChannel(options) {
 
     self.statsd = self.options.statsd;
     if (self.statsd) {
-        self.channelStatsd = new TChannelStatsd({
-            statEmitter: self,
-            statsd: self.statsd,
-            logger: self.logger
-        });
+        self.channelStatsd = new TChannelStatsd(self, self.statsd);
         self.options.statsd = null;
     }
 
--- a/node/channel.js
+++ b/node/channel.js
@@ -127,17 +127,6 @@ function TChannel(options) {
         processName: format('%s[%s]', process.title, process.pid)
     }, options);
 
-    // required: 'app'
-    // optional: 'host', 'cluster', 'version'
-    assert(!self.options.statTags || self.options.statTags.app, 'the stats must have the "app" tag');
-    self.statTags = new StatTags(self.options.statTags || {});
-
-    self.statsd = self.options.statsd;
-    if (self.statsd) {
-        self.channelStatsd = new TChannelStatsd(self, self.statsd);
-        self.options.statsd = null;
-    }
-
     self.logger = self.options.logger || nullLogger;
     self.random = self.options.random || globalRandom;
     self.timers = self.options.timers || globalTimers;
@@ -148,6 +137,21 @@ function TChannel(options) {
         typeof self.options.emitConnectionMetrics === 'boolean' ?
         self.options.emitConnectionMetrics : true;
     self.choosePeerWithHeap = self.options.choosePeerWithHeap || false;
+
+    // required: 'app'
+    // optional: 'host', 'cluster', 'version'
+    assert(!self.options.statTags || self.options.statTags.app, 'the stats must have the "app" tag');
+    self.statTags = new StatTags(self.options.statTags || {});
+
+    self.statsd = self.options.statsd;
+    if (self.statsd) {
+        self.channelStatsd = new TChannelStatsd({
+            statEmitter: self,
+            statsd: self.statsd,
+            logger: self.logger
+        });
+        self.options.statsd = null;
+    }
 
     // Filled in by the listen call:
     self.host = null;
--- a/node/lib/statsd.js
+++ b/node/lib/statsd.js
@@ -23,18 +23,20 @@
 var Cleaner = require('./statsd-clean');
 var clean = Cleaner.clean;
 var cleanHostPort = Cleaner.cleanHostPort;
+var nullLogger = require('../null-logger.js');
 
-function TChannelStatsd(channel, statsd) {
+function TChannelStatsd(options) {
     if (!(this instanceof TChannelStatsd)) {
-        return new TChannelStatsd(channel, statsd);
+        return new TChannelStatsd(options);
     }
 
     var self = this;
 
-    self.statsd = statsd;
-    self.channel = channel;
+    self.statsd = options.statsd;
+    self.statEmitter = options.statEmitter;
+    self.logger = options.logger || nullLogger;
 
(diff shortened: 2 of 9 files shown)


node/lib/statsd.js
```diff
@@ -23,18 +23,20 @@
...
+        return new TChannelStatsd(options);
     }
 
     var self = this;
 
-    self.statsd = statsd;
-    self.channel = channel;
+    self.statsd = options.statsd;
+    self.statEmitter = options.statEmitter;
+    self.logger = options.logger || nullLogger;
```
Raynos: no null logger; make it required.

node/channel.js
```diff
@@ -149,6 +138,21 @@ function TChannel(options) {
...
     self.choosePeerWithHeap = self.options.choosePeerWithHeap || false;
 
+    // required: 'app'
+    // optional: 'host', 'cluster', 'version'
+    assert(!self.options.statTags || self.options.statTags.app, 'the stats must have the "app" tag');
+    self.statTags = new StatTags(self.options.statTags || {});
+
+    self.statsd = self.options.statsd;
+    if (self.statsd) {
+        self.channelStatsd = new TChannelStatsd({
```
Raynos: no; let's just pass in channel.

node/lib/statsd.js
```diff
@@ -23,18 +23,20 @@
 var Cleaner = require('./statsd-clean');
 var clean = Cleaner.clean;
 var cleanHostPort = Cleaner.cleanHostPort;
+var nullLogger = require('../null-logger.js');
 
-function TChannelStatsd(channel, statsd) {
+function TChannelStatsd(options) {
```
Raynos: lets keep this simple and just pass in channel.

node/rate_limiter.js
```diff
@@ -72,7 +75,18 @@ function RateLimiter(options) {
...
+    self.rateLimiterEmptyTags = new stat.RateLimiterEmptyTags();
+    self.serviceRpsStat = self.statEmitter.defineGauge('tchannel.rate-limiting.service-rps');
+    self.serviceRpsLimitStat = self.statEmitter.defineGauge('tchannel.rate-limiting.service-rps-limit');
+    self.totalRpsStat = self.statEmitter.defineGauge('tchannel.rate-limiting.total-rps');
+    self.totalRpsLimitStat = self.statEmitter.defineGauge('tchannel.rate-limiting.total-rps-limit');
+    self.totalBusyStat = self.statEmitter.defineCounter('tchannel.rate-limiting.total-busy');
+    self.serviceBusyStat = self.statEmitter.defineCounter('tchannel.rate-limiting.service-busy');
+
     self.timers = options.timers;
+    self.logger = options.logger || nullLogger;
```
Raynos: no null logger; make it required.

node/rate_limiter.js
```diff
@@ -72,7 +75,18 @@ function RateLimiter(options) {
     }
     var self = this;
 
+    // stats
+    self.statEmitter = options.statEmitter || new StatEmitter();
```
Raynos: let's not add a new stats emitter. Let's do it all through `channel.buildStat()`

node/rate_limiter.js
```diff
@@ -72,7 +75,18 @@ function RateLimiter(options) {
...
+    self.rateLimiterEmptyTags = new stat.RateLimiterEmptyTags();
+    self.serviceRpsStat = self.statEmitter.defineGauge('tchannel.rate-limiting.service-rps');
```
Raynos: gauges are incredibly expensive.

node/rate_limiter.js
```diff
@@ -97,17 +111,22 @@ function RateLimiter(options) {
...
 }
 
-RateLimiter.prototype.type = 'tchannel.hyperbahn.rate-limiting';
+RateLimiter.prototype.type = 'tchannel.rate-limiting';
 
 RateLimiter.prototype.refresh =
 function refresh() {
     var self = this;
 
+    self.totalRpsStat.update(self.totalRequestCounter.rps, self.rateLimiterEmptyTags);
```
Raynos: This is too frequent and high rate. We can't do this at scale.

Raynos:
This adds too many stats; these stats are all very expensive.

node/rate_limiter.js
ShanniLi: yes, I used a wrong sample rate. I will make it every one second. Then the cost should be low, since it will be just 4 stats per second per process.

Raynos: gauges are expensive because we dont aggregate them due to a stat bug. 

but 4 per second should be ok.

Raynos:
Let's see what the impact is on the benchmark.

node/rate_limiter.js
ShanniLi: Since the code uses fast stats already, why do I need to use channel.buildStat()?

node/channel.js
```diff
@@ -149,6 +138,17 @@ function TChannel(options) {
         self.options.emitConnectionMetrics : true;
     self.choosePeerWithHeap = self.options.choosePeerWithHeap || false;
 
+    // required: 'app'
```
Raynos: Why did this block move ?

node/rate_limiter.js
```diff
@@ -72,10 +73,16 @@ function RateLimiter(options) {
     }
     var self = this;
 
-    self.timers = options.timers;
+    // stats
+    self.channel = options.channel;
+    assert(self.channel && !self.channel.topChannel, 'RateLimiter requires top channel');
+    self.rateLimiterEmptyTags = new stat.RateLimiterEmptyTags();
```
Raynos: technically this is not safe to do; lets just create a new empty object every time.

node/channel.js
ShanniLi: logger was not ready

node/rate_limiter.js
```diff
@@ -97,17 +104,55 @@ function RateLimiter(options) {
...
+    }
+
     self.totalRequestCounter.refresh();
 
     var serviceNames = Object.keys(self.counters);
     for (var i = 0; i < serviceNames.length; i++) {
-        self.counters[serviceNames[i]].refresh();
+        var counter = self.counters[serviceNames[i]];
+        if (self.cycle === self.numOfBuckets) {
+            var statsTag = new stat.RateLimiterServiceTags(serviceNames[i]);
```
Raynos: create a statsTag for each one; a unique one.

Raynos:
Need to create a tag per call; other then that lgm

PR #1089, opened 2015-07-30
change protocol doc to point to external docs

Raynos:
The definition of the "Arg Scheme" in the protocol document was
stale.

To avoid confusion I rewrote this section to point to other
documents that explain each arg scheme in detail

r: @blampe @breerly @abhinav


kriskowal:
:ok:

abhinav:
:+1:

PR #1091, opened 2015-07-31
Rate limiter should allow traffic within the limit

ShanniLi:
r @Raynos @kriskowal @blampe

### DIFF
--- a/node/test/hyperbahn/rate-limiter.js
+++ b/node/test/hyperbahn/rate-limiter.js
@@ -211,6 +211,7 @@ function runTests(HyperbahnCluster) {
             series([
                 send.bind(null, opts),
                 send.bind(null, opts),
+                send.bind(null, opts),
                 function sendError(done) {
                     var tchannelJSON = TChannelJSON({
                         logger: cluster.logger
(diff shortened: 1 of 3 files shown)


node/test/hyperbahn/rate-limiter.js
```diff
@@ -270,6 +271,7 @@ function runTests(HyperbahnCluster) {
                 send.bind(null, opts),
                 send.bind(null, opts),
                 waitFor(500),
+                send.bind(null, opts),
```
blampe: Can you explain why this is testing the new behavior? Is this asserting that we've made a call even after triggering the rate limit?

ShanniLi: @blampe the old behavior is:
1. increment the counters
2. rate limiting check
where all requests are considered for the sake of rate limiting. Therefore, once the rps is beyond the limit, no request can be served.

In the new behavior, it is:
1. rate limiting check. if passes, goto step 2.
2. increment the counters
where only served requests are considered in the counter.

Raynos:
lgtm.

PR #1094, opened 2015-07-31
Improve all edge cases around double timer issues

Raynos:
I think this also fixes the stale tombstones issue we see
in production.

r @kriskowal @jcorbin @shannili

### DIFF
--- a/node/argstream.js
+++ b/node/argstream.js
@@ -153,10 +153,12 @@ function OutArgStream() {
         self._handleFrameChunk(2, null);
     });
     self.arg3.on('finish', function onArg3Finish() {
-        self._handleFrameChunk(3, null);
-        self._flushParts(true);
-        self.finished = true;
-        self.finishEvent.emit(self);
+        if (!self.finished) {
+            self._handleFrameChunk(3, null);
+            self._flushParts(true);
+            self.finished = true;
+            self.finishEvent.emit(self);
+        }
     });
 }
 
--- a/node/out_request.js
+++ b/node/out_request.js
@@ -500,6 +500,20 @@ TChannelOutRequest.prototype.hookupCallback = function hookupCallback(callback) 
 TChannelOutRequest.prototype.onTimeout = function onTimeout(now) {
     var self = this;
 
+    if (self.err) {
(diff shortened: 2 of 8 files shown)


node/out_request.js
```diff
@@ -486,13 +500,27 @@ TChannelOutRequest.prototype.hookupCallback = function hookupCallback(callback)
 TChannelOutRequest.prototype.onTimeout = function onTimeout(now) {
     var self = this;
 
+    if (self.err) {
+        self.channel.logger.warn('weird onTimeout for errored out request', {
```
kriskowal: s/weird/unexpected

node/out_response.js
```diff
@@ -217,9 +217,10 @@ TChannelOutResponse.prototype.emitFinish = function emitFinish() {
             codeString: self.codeString,
             errorMessage: self.message,
             remoteAddr: self.inreq.connection.socketRemoteAddr,
-            state: self.state
-        });
+            state: self.state,
 
+            isOk: self.ok
```
kriskowal: In one sentence, clarify when `isOk` should be used instead of merely `ok`.

kriskowal:
:ok:

node/out_response.js
Raynos: No reason. I wrote hasResponse and isOk at the same time

node/argstream.js
```diff
@@ -153,10 +153,12 @@ function OutArgStream() {
         self._handleFrameChunk(2, null);
     });
     self.arg3.on('finish', function onArg3Finish() {
-        self._handleFrameChunk(3, null);
-        self._flushParts(true);
-        self.finished = true;
-        self.finishEvent.emit(self);
+        if (!self.finished) {
```
ShanniLi: this smells bad. :(
arg3 shouldn't be finished twice at the first place ...

node/out_request.js
ShanniLi: +1

node/out_response.js
ShanniLi: yes, "ok" is better. Needs to keep the consistency.

ShanniLi:
:+1:

PR #1096, opened 2015-07-31
fix streamingReq memory leak

Raynos:
We tried to fix this but we didnt get the check right.
We should check the `isLast` predicate.

r: @shannili @kriskowal

### DIFF
--- a/node/v2/handler.js
+++ b/node/v2/handler.js
@@ -195,7 +195,8 @@ TChannelV2Handler.prototype.handleCallRequest = function handleCallRequest(reqFr
 
     var handled = self._handleCallFrame(req, reqFrame);
     if (handled) {
-        if (req.state === States.Streaming) {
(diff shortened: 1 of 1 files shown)


ShanniLi:
lgtm

node/v2/handler.js
```diff
@@ -195,7 +195,8 @@ TChannelV2Handler.prototype.handleCallRequest = function handleCallRequest(reqFr
 
     var handled = self._handleCallFrame(req, reqFrame);
     if (handled) {
-        if (req.state === States.Streaming) {
+        var isLast = !(reqFrame.body.flags & v2.CallFlags.Fragment);
+        if (!isLast) {
```
kriskowal: double negative is confusing. reversing the order of the then/else clauses.

kriskowal:
:ok:

PR #1099, opened 2015-08-02
Error backoff

ShanniLi:
r @jcorbin @Raynos @kriskowal @rf 
cc @junchaowu @prashantv @anson627 

The implementation is described in the design doc. Only linear backoff is implemented.
If this makes sense, we will need to implement it in python and go clients also.

### DIFF
--- a/docs/error-backoff.md
+++ b/docs/error-backoff.md
@@ -1,33 +1,19 @@
 # Error Backoff
 
 Error backoff is a mechanism that throttles outgoing TChannel traffic in order to improve the
-health of the Hyperbahn cluster and its hosted services. By definition, error backoff monitors
-the status of outgoing TChannel traffic. If it determines that 1) the Hyperbahn cluster/services
-are unhealthy, and 2) the unheathy state can be improved by throttling outgoing
-traffic, error backoff will effectively throttle outgoing traffic by returning error responses
-corresponding to the out requests.
-
-There are many reasons why error backoff can help. For example, when responding to `tchannel.busy`,
-it means that a service is stressed so that requests to the service should be reduced. As another
-example, when `tchannel.declined` is received, it may indicate the absence of any healthy host
-for the service, so there is no reason to send more requests to the same service in the next short
-amount of time period.
+health of the Hyperbahn cluster and its hosted services.
 
 
 ## Goals
 
 Error backoff works passively to defend Hyperbahn cluster and services from flooding requests and retries.
-When a service is unhealth or over stressed, there will be an increasing number of error responses.
-Such an unhealthy state should be handled by reducing the requests to the service. On the other hand,
-it is very important that error backoff never affects healthy traffic.
+It is very important that error backoff never affects healthy traffic.
 
 Goals:
 
-* Monitor the health state of an Hyperbahn edge, i.e., callerName ==> serviceName. Most effectively,
-  such information should be observed locally through call responses.
-* Throttle out requests if the current unheathy state can be relieved by reducing the requests.
-* The throttling should grow if the health state continues to decline.
-* The throttling should decline if the health state improves.
+* Throttle out requests when busy frame is received.
+* The throttling should grow if we have a growth in busy frames.
+* The throttling should decline if the number of busy frames decline.
 * Error backoff should converge. In other words, if the traffic stablizes in terms of failure/success
   ratios, the level of throttling should also be stable.
 * No healthy traffic should be affected at any time. 
@@ -45,29 +31,19 @@ Non-goals:
 
 ## Design & Implemenation
 
-### Errors sensitive to backoff 
-Only issues that can be relieved by throttling should be considered for error backoff. Errors sensitive to
-backoff are also called eligible errors. There are three types of eligible errors handled by error backoff. 
-
-`tchannel.timeout` is received when a request is not completed within its timeout limit. The error indicates
-that the Hyerbahn cluster and/or the service is becoming slow.
+### Eligible errors
 
 `tchannel.busy` is returned by Rate Limiting when one of the relaying nodes decides that the RPS limit for this
-request has been reached.
-
-`tchannel.declined` is returned when 1) there is no peer available for this service or 2) the request is declined
-by Circuit Breaking. 
+request has been reached. It can also be returned by the edge server when it has high event loop lag, RAM usage,
+etc.
 
 
-### Health monitor 
-Error backoff is specific to Hyperbahn edges, i.e., callerName ==> serviceName. Therefore, the measurement of
-traffic health is associated with the edges.
+### Monitor 
+Error backoff is specific to Hyperbahn edges, i.e., callerName ==> serviceName.
 
 When an error is received, error backoff should take the error type and the frequency of the error into account.
-Only eligible errors should be considered towarding an unhealthy state. The frequency of the error is used to
-identify the change of the service quality such as degradation or recovery. In other words, the more errors received,
-the unhealthier the service is considered.
-
+Only eligible errors should be considered for backoff. The frequency of the error is used to
+identify the change of the service quality such as degradation or recovery.
 
 ### Throttling
 When a request should be throttled, error backoff returns `tchannel.busy` immediately rather than sending the request
--- a/docs/error-backoff.md
+++ b/docs/error-backoff.md
@@ -1,0 +1,94 @@
+# Error Backoff
+
+Error backoff is a mechanism that throttles outgoing TChannel traffic in order to improve the
+health of the Hyperbahn cluster and its hosted services. By definition, error backoff monitors
+the status of outgoing TChannel traffic. If it determines that 1) the Hyperbahn cluster/services
+are unhealthy, and 2) the unheathy state can be improved by throttling outgoing
+traffic, error backoff will effectively throttle outgoing traffic by returning error responses
+corresponding to the out requests.
+
+There are many reasons why error backoff can help. For example, when responding to `tchannel.busy`,
+it means that a service is stressed so that requests to the service should be reduced. As another
+example, when `tchannel.declined` is received, it may indicate the absence of any healthy host
+for the service, so there is no reason to send more requests to the same service in the next short
+amount of time period.
+
+
+## Goals
+
+Error backoff works passively to defend Hyperbahn cluster and services from flooding requests and retries.
+When a service is unhealth or over stressed, there will be an increasing number of error responses.
+Such an unhealthy state should be handled by reducing the requests to the service. On the other hand,
+it is very important that error backoff never affects healthy traffic.
+
+Goals:
+
+* Monitor the health state of an Hyperbahn edge, i.e., callerName ==> serviceName. Most effectively,
+  such information should be observed locally through call responses.
+* Throttle out requests if the current unheathy state can be relieved by reducing the requests.
+* The throttling should grow if the health state continues to decline.
+* The throttling should decline if the health state improves.
+* Error backoff should converge. In other words, if the traffic stablizes in terms of failure/success
+  ratios, the level of throttling should also be stable.
+* No healthy traffic should be affected at any time. 
+
+Non-goals:
+
+* Error backoff doesn't do congestion control. [Rate Limiting](./rate-limiting.md) controls traffic and
+  relieves the impact of congestions. In comparison, error backfoff reacts to rate limiting error frames
+  in order to reduce the number of requests sent to the Hyperbahn cluster or even forwarded by the
+  Hyperbahn cluster.
+* Error backoff is not [Circuit Breaking](./circuit-breaking.md). When enabled on the client side, 
+  error backoff throttles the requests sent by the client to the Hyperbahn cluster. When enabled
+  on the Hyperbahn nodes, error backoff throttles the requests that can be forwarded through.
+
+
+## Design & Implemenation
+
+### Errors sensitive to backoff 
+Only issues that can be relieved by throttling should be considered for error backoff. Errors sensitive to
+backoff are also called eligible errors. There are three types of eligible errors handled by error backoff. 
+
+`tchannel.timeout` is received when a request is not completed within its timeout limit. The error indicates
+that the Hyerbahn cluster and/or the service is becoming slow.
+
+`tchannel.busy` is returned by Rate Limiting when one of the relaying nodes decides that the RPS limit for this
+request has been reached.
+
+`tchannel.declined` is returned when 1) there is no peer available for this service or 2) the request is declined
+by Circuit Breaking. 
+
+
+### Health monitor 
+Error backoff is specific to Hyperbahn edges, i.e., callerName ==> serviceName. Therefore, the measurement of
+traffic health is associated with the edges.
+
+When an error is received, error backoff should take the error type and the frequency of the error into account.
+Only eligible errors should be considered towarding an unhealthy state. The frequency of the error is used to
+identify the change of the service quality such as degradation or recovery. In other words, the more errors received,
+the unhealthier the service is considered.
+
+
+### Throttling
+When a request should be throttled, error backoff returns `tchannel.busy` immediately rather than sending the request
+out.
+
+
+### Backoff mechanisms
+
+#### Linear backoff
+Linear backoff is a straightforward approach that throttles requests based on the number of error responses received.
+Given a backoff rate of `R` and an eligible error count of `E`, linear backoff throttles the next `R * E` requests.
+Linear backoff is simple to implement and easy to test. The recovery rate is linear as well. 
+
+#### Exponential backoff
+Exponential backoff is similar to linear backoff only except that the number of throttled requests is an exponential
+function of `R * E`.
+
+### Service alerts
+Warning should be logged when a request is throttled by error backoff.
+
+
+## Test plan
+* Unit tests
+* Integration tests in hyperbhan cluster
--- a/docs/error-backoff.md
+++ b/docs/error-backoff.md
@@ -1,7 +1,7 @@
 # Error Backoff
 
 Error backoff is a mechanism that throttles outgoing TChannel traffic in order to improve the
-health of the Hyperbahn cluster and its hosted services.
+availability of the Hyperbahn cluster and its hosted services.
 
 
 ## Goals
--- a/node/channel.js
+++ b/node/channel.js
@@ -40,6 +40,8 @@ var TChannelRequest = require('./request');
 var TChannelServiceNameHandler = require('./service-name-handler');
 var errors = require('./errors');
 
+var ErrorBackoff = require('./error_backoff.js');
+
 var BaseStat = require('./lib/stat.js').BaseStat;
 var TChannelAsThrift = require('./as/thrift');
 var TChannelAsJSON = require('./as/json');
@@ -253,6 +255,13 @@ function TChannel(options) {
     if (!self.topChannel) {
         self.sanityTimer = self.timers.setTimeout(doSanitySweep, SANITY_PERIOD);
         self.flushStats();
+    }
+
+    if (!self.topChannel) {
+        self.errorBackoff = new ErrorBackoff({
+            channel: self,
+            backoffRate: self.options.backoffRate
+        });
     }
 
     function doSanitySweep() {
--- a/node/channel.js
+++ b/node/channel.js
@@ -260,6 +260,7 @@ function TChannel(options) {
     if (!self.topChannel) {
         self.errorBackoff = new ErrorBackoff({
             channel: self,
+            enabled: self.options.enableErrorBackoff,
             backoffRate: self.options.backoffRate
         });
     }
--- a/node/connection.js
+++ b/node/connection.js
@@ -346,6 +346,8 @@ TChannelConnection.prototype.onCallError = function onCallError(err) {
 
         req.emitError(err);
     }
+
+    self.channel.errorBackoff.handleError(err, req.headers.cn, req.serviceName);
 };
 
 TChannelConnection.prototype.start = function start() {
--- a/node/connection_base.js
+++ b/node/connection_base.js
@@ -86,7 +86,13 @@ function connBaseRequest(options) {
     var req = self.buildOutRequest(options);
     self.ops.addOutReq(req);
     req.peer.invalidateScore();
+    req.errorEvent.on(onReqError);
+
     return req;
+
+    function onReqError(err) {
+        self.ops.popOutReq(req.id, req);
+    }
 };
 
 TChannelConnectionBase.prototype.handleCallRequest = function handleCallRequest(req) {
--- a/node/error_backoff.js
+++ b/node/error_backoff.js
@@ -64,7 +64,7 @@ function handleError(err, cn, serviceName) {
     }
 
     if (!err || !cn || !serviceName) {
-        self.logger.warn('ErrorBackoff.handleError called wiht invalid parameters', {
+        self.logger.warn('ErrorBackoff.handleError called with invalid parameters', {
             error: err,
             cn: cn,
             serviceName: serviceName
@@ -108,7 +108,7 @@ function getBackoffError(cn, serviceName) {
     }
 
     if (!cn || !serviceName) {
-        self.logger.warn('ErrorBackoff.getBackoffError called wiht invalid parameters', {
+        self.logger.warn('ErrorBackoff.getBackoffError called with invalid parameters', {
             cn: cn,
             serviceName: serviceName
         });
--- a/node/error_backoff.js
+++ b/node/error_backoff.js
@@ -34,6 +34,11 @@ var BackoffError = TypedError({
     serviceName: null
 });
 
+function getEdgeName(cn, serviceName) {
+    return cn + '~~' + serviceName;
+}
+
+
 function ErrorBackoff(options) {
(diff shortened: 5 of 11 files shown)


node/error_backoff.js
```diff
@@ -0,0 +1,125 @@
...
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+'use strict';
+
+var assert = require('assert');
+var TypedError = require('error/typed');
+
+var BackoffError = TypedError({
```
anson627: move to errors.js?

node/error_backoff.js
```diff
@@ -0,0 +1,125 @@
...
+            serviceName: serviceName
+        });
+        return;
+    }
+
+    if (!self.shouldConsider(err)) {
+        return;
+    }
+
+    var edge = cn + '~~' + serviceName;
```
anson627: refactoring to a one line method, so that we duplicate the code.

node/request.js
```diff
@@ -106,7 +106,7 @@ TChannelRequest.prototype.emitResponse = function emitResponse(res) {
     self.responseEvent.emit(self, res);
 };
 
-TChannelRequest.prototype.hookupStreamCallback = function hookupCallback(callback) {
+TChannelRequest.prototype.hookupStreamCallback = function hookupStreamCallback(callback) {
```
anson627: unrelated change?

node/error_backoff.js
ShanniLi: It is a v2 specific error. Not sure errors.js is a good place. Ideally, it should be in v2/error_response.js, but then we will need to reference that file from the top level ...

node/request.js
ShanniLi: it is in a separate commit. :)

node/channel.js
```diff
@@ -255,6 +257,13 @@ function TChannel(options) {
         self.flushStats();
     }
 
+    if (!self.topChannel) {
+        self.errorBackoff = new ErrorBackoff({
```
kriskowal: I recommend CongestionAvoidance (AIMD, controlled by errors), to be later complemented by FlowControl (controlled by latency). AIMD is "additive increase, multiplicative decrease". I see the multiplicative decrease part, but no mechanism for an edge to recover from failure. Be sure to trap successes as well as failures to complete the control loop.

node/error_backoff.js
```diff
@@ -0,0 +1,125 @@
...
+
+ErrorBackoff.prototype.getBackoffError =
+function getBackoffError(cn, serviceName) {
+    var self = this;
+    if (!self.backoffRate) {
+        return null;
+    }
+
+    if (!cn || !serviceName) {
+        self.logger.warn('ErrorBackoff.getBackoffError called wiht invalid parameters', {
```
kriskowal: s/wiht/with/

node/error_backoff.js
```diff
@@ -0,0 +1,125 @@
...
+            cn: cn,
+            serviceName: serviceName
+        });
+        return null;
+    }
+
+    var edge = cn + '~~' + serviceName;
```
kriskowal: I know @raynos encourages this pattern as an expedient, but there is nothing at all, not even a doc, that prevents cn or sn from having this delimiter. This is why the circuit breaker tracks these in a two-level data structure. I encourage you to solve this problem by any means.

node/test/error_backoff.js
```diff
@@ -0,0 +1,204 @@
...
+
+'use strict';
+
+var TimeMock = require('time-mock');
+var allocCluster = require('./lib/alloc-cluster.js');
+var timers = TimeMock(Date.now());
+var ErrorBackoff = require('../error_backoff.js');
+var series = require('run-series');
+
+function send(channel, endpoint, check, done) {
```
kriskowal: check arg is unused, probably also not useful since you can do the same in a callback thunk.

node/test/error_backoff.js
```diff
@@ -0,0 +1,204 @@
...
+        assert.ok(err, 'there should be an error');
+        assert.equal(err.type, 'tchannel.busy', 'error type as expected');
+    }
+    function check2(err, res) {
+        assert.ok(err, 'there should be a backoff error');
+        assert.equal(err.type, 'tchannel.busy', 'backoff error type as expected');
+        assert.equal(err.subType, 'tchannel.backoff.error', 'backoff error subType as expected');
+    }
+    function check3(err, res) {
+        assert.ifError(err, 'there should be no error');
```
kriskowal: I encourage everyone to use the pattern: `if (err) return assert.end(err);` instead of this err and continue pattern.

node/error_backoff.js
```diff
@@ -0,0 +1,125 @@
...
+
+    if (!self.shouldConsider(err)) {
+        return;
+    }
+
+    var edge = cn + '~~' + serviceName;
+    if (!self.reqErrors[edge]) {
+        self.reqErrors[edge] = self.backoffRate;
+    } else {
+        self.reqErrors[edge] += self.backoffRate;
```
kriskowal: What does `reqErrors` represent? What are the units of `backoffRate`? This looks like linear accumulating backoff instead of multiplicative, but depends on what the unit are. Consider a more illuminating name, like `errorCountByEdge` (if count goes up by 1 for each error), or `delayByEdge` if the unit is an artificial delay (milliseconds) between requests.

(Please only abbreviate request to req in local variables.)

node/error_backoff.js
```diff
@@ -0,0 +1,125 @@
...
+        });
+        return null;
+    }
+
+    var edge = cn + '~~' + serviceName;
+    if (!self.reqErrors[edge] || self.reqErrors[edge] < 1) {
+        return null;
+    }
+
+    self.reqErrors[edge] -= 1;
```
kriskowal: I’m still reading and haven’t figured out the units, but this constant should probably have a name and be configurable, since backoffRate has a name and isn’t necessarily 1.

Is the idea that `getBackoffError` is depleting a penalty each time it’s called? `get` is usually reserved for pure accessors. Try `take` or `nextBackoffError` if it’s consuming something as a side-effect.

node/test/error_backoff.js
```diff
@@ -0,0 +1,204 @@
...
+        channel: channel,
+        backoffRate: 1
+    });
+    channel.logger.whitelist(
+        'warn',
+        'ErrorBackoff.handleError called wiht invalid parameters'
+    );
+    channel.logger.whitelist(
+        'warn',
+        'ErrorBackoff.getBackoffError called wiht invalid parameters'
```
kriskowal: s/wiht/with/

node/error_backoff.js
kriskowal: We keep all errors in one module per package.

kriskowal:
I am going to wait for a doc before commenting further. Intuition is that we either have per request back-off that forgets its state upon timeout, or per-edge back-off that has a corresponding back-on so we don’t lock a cn:sn pair into a degraded state.

docs/error-backoff.md
```diff
@@ -0,0 +1,94 @@
+# Error Backoff
+
+Error backoff is a mechanism that throttles outgoing TChannel traffic in order to improve the
+health of the Hyperbahn cluster and its hosted services. By definition, error backoff monitors
+the status of outgoing TChannel traffic. If it determines that 1) the Hyperbahn cluster/services
```
Raynos: "the Hyperbahn cluster/services are busy, and 2) the busy state can be improved by throttling outgoing traffic"

Let's only throttle on Busy as first pass.

docs/error-backoff.md
```diff
@@ -0,0 +1,94 @@
...
+are unhealthy, and 2) the unheathy state can be improved by throttling outgoing
+traffic, error backoff will effectively throttle outgoing traffic by returning error responses
+corresponding to the out requests.
+
+There are many reasons why error backoff can help. For example, when responding to `tchannel.busy`,
+it means that a service is stressed so that requests to the service should be reduced. As another
+example, when `tchannel.declined` is received, it may indicate the absence of any healthy host
```
Raynos: tchannel.declined means "try me again somewhere else".

docs/error-backoff.md
```diff
@@ -0,0 +1,94 @@
...
+for the service, so there is no reason to send more requests to the same service in the next short
+amount of time period.
+
+
+## Goals
+
+Error backoff works passively to defend Hyperbahn cluster and services from flooding requests and retries.
+When a service is unhealth or over stressed, there will be an increasing number of error responses.
```
Raynos: The unhealthy case is completely different; that should be handled by the circuit breaker.

docs/error-backoff.md
```diff
@@ -0,0 +1,94 @@
...
+Such an unhealthy state should be handled by reducing the requests to the service. On the other hand,
+it is very important that error backoff never affects healthy traffic.
+
+Goals:
+
+* Monitor the health state of an Hyperbahn edge, i.e., callerName ==> serviceName. Most effectively,
```
Raynos: You are not the circuit breaker.

You should be monitoring the load on any given host:port; either by total or by service.

docs/error-backoff.md
```diff
@@ -0,0 +1,94 @@
...
+  such information should be observed locally through call responses.
+* Throttle out requests if the current unheathy state can be relieved by reducing the requests.
```
Raynos: You should throttle requests if you have determined that the host:port is busy, either by total or by service name.

docs/error-backoff.md
```diff
@@ -0,0 +1,94 @@
...
+* The throttling should grow if the health state continues to decline.
```
Raynos: The throttling should grow if we have a growth in busy frames.

docs/error-backoff.md
```diff
@@ -0,0 +1,94 @@
...
+* The throttling should decline if the health state improves.
```
Raynos: You are not health state; you are not circuit breaker. Backoff is purely about Busy frames.

docs/error-backoff.md
```diff
@@ -0,0 +1,94 @@
...
+* Error backoff should converge. In other words, if the traffic stablizes in terms of failure/success
```
Raynos: This is a great point; we definitely want this; if Backoff works correctly then the server should have at most 0/1 Busy's and the client should have a high rate of "ClientSide Busy"

docs/error-backoff.md
```diff
@@ -0,0 +1,94 @@
...
+  ratios, the level of throttling should also be stable.
+* No healthy traffic should be affected at any time.
```
Raynos: When a connection to hyperbahn has a single Busy service it should not affect any other services on said connection.

It should also not affect the caller as a whole.

docs/error-backoff.md
```diff
@@ -0,0 +1,94 @@
...
+  on the Hyperbahn nodes, error backoff throttles the requests that can be forwarded through.
+
+
+## Design & Implemenation
+
+### Errors sensitive to backoff 
+Only issues that can be relieved by throttling should be considered for error backoff. Errors sensitive to
+backoff are also called eligible errors. There are three types of eligible errors handled by error backoff. 
+
+`tchannel.timeout` is received when a request is not completed within its timeout limit. The error indicates
```
Raynos: timeout does not indicate that the service is slow; it indicates that a dependency somewhere is slow. The best solution is to try again or use speculative execution.

I dont think we need backoff on timeouts, we just need to smooth over them.

If a service is running at high CPU or at high event loop lagging and is leaking latency instead of spiking it; it needs to return busy frames.

docs/error-backoff.md
```diff
@@ -0,0 +1,94 @@
...
+that the Hyerbahn cluster and/or the service is becoming slow.
+
+`tchannel.busy` is returned by Rate Limiting when one of the relaying nodes decides that the RPS limit for this
+request has been reached.
+
+`tchannel.declined` is returned when 1) there is no peer available for this service or 2) the request is declined
```
Raynos: > Node declined request for reasons other than load, the request is safe to retry elsewhere if desired, do not change peer preferencing for future requests.

Declined means its safe to try elsewhere; circuit breaker should never have been returning declineds; we have an Unhealthy type for that.

docs/error-backoff.md
```diff
@@ -0,0 +1,94 @@
...
+## Design & Implemenation
+
+### Errors sensitive to backoff 
+Only issues that can be relieved by throttling should be considered for error backoff. Errors sensitive to
+backoff are also called eligible errors. There are three types of eligible errors handled by error backoff. 
+
+`tchannel.timeout` is received when a request is not completed within its timeout limit. The error indicates
+that the Hyerbahn cluster and/or the service is becoming slow.
+
+`tchannel.busy` is returned by Rate Limiting when one of the relaying nodes decides that the RPS limit for this
```
Raynos: Tchannel.busy is also returned by the edge server when it has high event loop lag / RAM / etc.

docs/error-backoff.md
```diff
@@ -0,0 +1,94 @@
...
+request has been reached.
+
+`tchannel.declined` is returned when 1) there is no peer available for this service or 2) the request is declined
+by Circuit Breaking. 
+
+
+### Health monitor 
+Error backoff is specific to Hyperbahn edges, i.e., callerName ==> serviceName. Therefore, the measurement of
```
Raynos: it is not specific to edges. It's either total or per serviceName. 

You may in the hyperbahn entry node case want to penalize incoming requests by callerName weighting.

docs/error-backoff.md
```diff
@@ -0,0 +1,94 @@
...
+traffic health is associated with the edges.
+
+When an error is received, error backoff should take the error type and the frequency of the error into account.
+Only eligible errors should be considered towarding an unhealthy state. The frequency of the error is used to
```
Raynos: Let's just stop using the word healthy; it's very confusing. Use the word "busy" or something.

node/v2/handler.js
```diff
@@ -548,15 +549,17 @@ function sendCallRequestFrame(req, flags, args) {
         req.checksum.type, args
     );
 
-    if (!self.verifyCallRequestFrame(req, args)) {
-        return;
+    var result;
+    if (self.verifyCallRequestFrame(req, args)) {
```
Raynos: Why this change; let's avoid if possible.

docs/error-backoff.md
ShanniLi: if there is no peer available right now, does it actually mean try again later?

docs/error-backoff.md
ShanniLi: Yes, heathy state is confusing here. I will think about a better word.

The point I want express here is: the system is unheathy and error backoff my improve it. Error backoff doesn't really tracks the actually healthy state.

docs/error-backoff.md
ShanniLi: I don't think error backoff is congestion control. It should simply backoff on errors. It is rate limiter's job to monitor the load.

On the other hand, when enabled on Hyperbahn node, I think error backoff works as a mild "circuit breaker", i.e., reduce the traffic on a given edge.

docs/error-backoff.md
kriskowal: Busy and available. Backing off when busy, and recovering when available are the right words. Healthy and unhealthy are more general, covering all classes of error except deliberate cases like cancellation. What you are describing *is* flow control. Please consider treating this as a flow control problem, where busy errors are a signal of oversaturation, where a *multiplicative* back-off and successful delivery produces an *linear* recovery. Linear back-off won’t suffice.

Consider writing a *quick* simulation to verify the approach.

node/test/error_backoff.js
ShanniLi: not sure if I understand. I will need to check the response at each series call, how can I do that without a check function?

node/channel.js
ShanniLi: it is actually error backoff only

docs/error-backoff.md
ShanniLi: Linear backoff is a good start for testing and evaluation. We can easily switch to exponential backoff when the framework of error backoff works.

docs/error-backoff.md
```diff
@@ -0,0 +1,70 @@
+# Error Backoff
+
+Error backoff is a mechanism that throttles outgoing TChannel traffic in order to improve the
+health of the Hyperbahn cluster and its hosted services.
```
Raynos: Improve availability

node/connection.js
```diff
@@ -346,6 +346,8 @@ TChannelConnection.prototype.onCallError = function onCallError(err) {
 
         req.emitError(err);
     }
+
+    self.channel.errorBackoff.handleError(err, req.headers.cn, req.serviceName);
```
Raynos: cleaner to just pass `req` in as a whole.

node/connection_base.js
```diff
@@ -86,7 +86,13 @@ function connBaseRequest(options) {
     var req = self.buildOutRequest(options);
     self.ops.addOutReq(req);
     req.peer.invalidateScore();
+    req.errorEvent.on(onReqError);
+
     return req;
+
+    function onReqError(err) {
+        self.ops.popOutReq(req.id, req);
```
Raynos: Why is this needed ? Do this in a seperate PR ?

node/errors.js
```diff
@@ -72,6 +72,13 @@ module.exports.AsHeaderRequired = TypedError({
     frame: null
 });
 
+module.exports.BackoffError = TypedError({
+    type: 'tchannel.backoff.error',
+    message: 'Backoff error triggered due to previous requset failures {cn} ==> {serviceName}',
```
Raynos: `request`

node/error_backoff.js
```diff
@@ -0,0 +1,112 @@
...
+    if (!(this instanceof ErrorBackoff)) {
+        return new ErrorBackoff(options);
+    }
+    var self = this;
+
+    self.channel = options.channel;
+    assert(self.channel && !self.channel.topChannel, 'ErrorBackoff requires top channel');
+    self.backoffRate = options.backoffRate || 0;
+    assert(self.backoffRate >= 0, 'ErrorBackoff requires a nonnegative backoffRate');
+    self.reqErrors = {};
```
Raynos: `Object.create(null)`

node/channel.js
```diff
@@ -255,6 +257,13 @@ function TChannel(options) {
...
+            channel: self,
+            backoffRate: self.options.backoffRate
```
Raynos: Favor an enabled boolean over `backoffRate: 0`

node/error_backoff.js
```diff
@@ -0,0 +1,112 @@
...
+
+ErrorBackoff.prototype.handleError =
+function handleError(err, cn, serviceName) {
+    var self = this;
+    if (!self.backoffRate) {
+        return;
+    }
+
+    if (!err || !cn || !serviceName) {
+        self.logger.warn('ErrorBackoff.handleError called with invalid parameters', {
```
Raynos: `self.logger.error()` please.

node/error_backoff.js
```diff
@@ -0,0 +1,112 @@
...
+
+ErrorBackoff.prototype.nextBackoffError =
+function nextBackoffError(cn, serviceName) {
+    var self = this;
+    if (!self.backoffRate) {
+        return null;
+    }
+
+    if (!cn || !serviceName) {
+        self.logger.warn('ErrorBackoff.nextBackoffError called with invalid parameters', {
```
Raynos: `logger.error()`

node/out_request.js
```diff
@@ -351,6 +351,12 @@ TChannelOutRequest.prototype.sendCallRequestFrame = function sendCallRequestFram
     var self = this;
     switch (self.state) {
         case States.Initial:
+            var errorBackoff = self.channel.errorBackoff || self.channel.topChannel.errorBackoff;
+            var err = errorBackoff.nextBackoffError(self.headers.cn, self.serviceName);
```
Raynos: favor `errorBackoff.nextBackoffError(self)`

node/error_backoff.js
Raynos: +1 on named constant instead of magic number.

node/error_backoff.js
```diff
@@ -0,0 +1,112 @@
...
+        return null;
+    }
+
+    var edge = getEdgeName(cn, serviceName);
+    if (!self.reqErrors[edge] || self.reqErrors[edge] < 1) {
+        return null;
+    }
+
+    self.reqErrors[edge] -= 1;
+    return errors.BackoffError({
```
Raynos: let's rename this to `shouldBackoff()` and construct the error in `out_request.js`

Then in out_request you can add way more useful context to this error.

Raynos:
Let's get our good friend @jcorbin to chip in on this beast.

node/connection.js
ShanniLi: actually, this is clearer interface and easier for unit testing

node/connection_base.js
ShanniLi: without this fix, this PR will fail the test

node/out_request.js
ShanniLi: the same reason, I want to keep the interface simple and easy for unit testing.

node/error_backoff.js
```diff
@@ -0,0 +1,125 @@
...
+ErrorBackoff.prototype.type = 'tchannel.error-backoff';
+
+ErrorBackoff.prototype.handleError =
+function handleError(err, cn, serviceName) {
+    var self = this;
+    if (!self.enabled || !self.backoffRate) {
+        return;
+    }
+
+    if (!err || !cn || !serviceName) {
```
jcorbin: these look like programing errors to me, use assert

node/error_backoff.js
```diff
@@ -0,0 +1,125 @@
...
+ErrorBackoff.prototype.disable =
+function disable() {
+    var self = this;
+    self.enabled = false;
+    self.reqErrors = Object.create(null);
+};
+
+ErrorBackoff.prototype.shouldConsider =
+function shouldConsider(err) {
+    if (err.type === 'tchannel.busy') {
```
jcorbin: Adding custom error classification logic here seems wrong to me.  Rather I think we want the exact same "is it (un)healthy" logic used by circuits, aka `errors.isUnhealthy(errors.classify(err))`

node/error_backoff.js
```diff
@@ -0,0 +1,125 @@
...
+};
+
+ErrorBackoff.prototype.shouldBackoff =
+function shouldBackoff(cn, serviceName) {
+    var self = this;
+    if (!self.enabled || !self.backoffRate) {
+        return false;
+    }
+
+    if (!cn || !serviceName) {
```
jcorbin: again, this seems a programming error to me, should be an assert?

node/request.js
jcorbin: please break it out into separate pr, there's too much going on here as it is

node/connection_base.js
jcorbin: please pull this out into a separate pr first then, it is a separate bug

jcorbin: also in that separate pr, please add a regression test, or have some existing test cover this prior bug so that we don't un-fix it ;-)

ShanniLi: This bug can only be triggered with this PR's, because in this PR, we emit an error in TChannelOutRequest.prototype.sendCallRequestFrame at State.Initial

node/error_backoff.js
ShanniLi: :+1: Good point!

kriskowal:
We need to come to a consensus about the problem we are solving with this, and do more to vet whether this is the appropriate solution. The problem is congestion avoidance, and the solution should be a congestion avoidance algorithm. This implementation is shaped like a rate limiting / token bucket / leaky bucket solution, but it is not implemented in such terms and has at least one flaw (no cap on the penalty) that would be revealed by modeling the problem. We need to take this to the drawing board and get consensus before diving in.

Please,

1.) Extract the markdown document you’ve contributed and open a PR for just that. As written, it still needs work because it does not explain the proposed problem or solution. At least one of the non-goals is definitely the goal: congestion avoidance.
2.) Write a small simulation and verify the behavior of the proposed solution for the problems it is intended to solve. In this case, reducing traffic from ingress, through egress, and to a saturated service, regardless of where the bottleneck manifests, and recovering from a degraded state in a reasonable amount of time.
3.) Implement, test, deploy, repeat.

This particular problem is tricky, has a lot of prior-art, and deserves diligence.

jcorbin:
:+1: @kriskowal , this feature requires more deliberate design, and especially modeling / simulation in a reboot.

ShanniLi:
I was planning to implement a simple solution just like lb_pool or airlock. Not sure how the idea of congestion control gets in. Yes, it will have the effect of congestion control, but it is just one of its benefit.

Raynos:
:+1:

This feature needs more thought; let's open a PR similar to @kriskowal PR for the hyperbahn worker selection logic that has a markdown and a simulation.

Let's get consensus on the algorithm and vet the simulation; once that's done we will move onto implementing; deploying to ingress; verification and more.

ShanniLi:
Sounds good, I will create a separate PR with a MD to get a consensus about the design first.

PR #1104, opened 2015-08-03
fix a log bug in rate limiting

ShanniLi:
r @Raynos @kriskowal

### DIFF
--- a/node/hyperbahn/service_proxy.js
+++ b/node/hyperbahn/service_proxy.js
@@ -141,6 +141,7 @@ function rateLimit(req, buildRes) {
     if (self.rateLimiter.shouldRateLimitTotalRequest(req.serviceName)) {
         var totalLimit = self.rateLimiter.totalRequestCounter.rpsLimit;
         self.logger.info('hyperbahn node is rate-limited by the total rps limit', {
+            cn: req.headers && req.headers.cn,
             rpsLimit: totalLimit
         });
         buildRes().sendError('Busy', 'hyperbahn node is rate-limited by the total rps of ' + totalLimit);
@@ -149,8 +150,9 @@ function rateLimit(req, buildRes) {
 
     // check RPS for service limit
     if (isExitNode && self.rateLimiter.shouldRateLimitService(req.serviceName)) {
-        var serviceLimit = self.rateLimiter.rpsLimitForServiceName[req.serviceName];
+        var serviceLimit = self.rateLimiter.getRpsLimitForService(req.serviceName);
         self.logger.info('hyperbahn service is rate-limited by the service rps limit', {
+            cn: req.headers && req.headers.cn,
             serviceName: req.serviceName,
             rpsLimit: serviceLimit
         });
--- a/node/hyperbahn/service_proxy.js
+++ b/node/hyperbahn/service_proxy.js
@@ -141,8 +141,11 @@ function rateLimit(req, buildRes) {
     if (self.rateLimiter.shouldRateLimitTotalRequest(req.serviceName)) {
         var totalLimit = self.rateLimiter.totalRequestCounter.rpsLimit;
         self.logger.info('hyperbahn node is rate-limited by the total rps limit', {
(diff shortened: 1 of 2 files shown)


node/hyperbahn/service_proxy.js
```diff
@@ -141,6 +141,7 @@ function rateLimit(req, buildRes) {
     if (self.rateLimiter.shouldRateLimitTotalRequest(req.serviceName)) {
         var totalLimit = self.rateLimiter.totalRequestCounter.rpsLimit;
         self.logger.info('hyperbahn node is rate-limited by the total rps limit', {
+            cn: req.headers && req.headers.cn,
```
Raynos: req.headers.cn, req.serviceName, req.endpoint, req.remoteAddr

node/hyperbahn/service_proxy.js
```diff
@@ -149,8 +150,9 @@ function rateLimit(req, buildRes) {
 
     // check RPS for service limit
     if (isExitNode && self.rateLimiter.shouldRateLimitService(req.serviceName)) {
-        var serviceLimit = self.rateLimiter.rpsLimitForServiceName[req.serviceName];
+        var serviceLimit = self.rateLimiter.getRpsLimitForService(req.serviceName);
         self.logger.info('hyperbahn service is rate-limited by the service rps limit', {
+            cn: req.headers && req.headers.cn,
```
Raynos: req.headers.cn, req.serviceName, req.endpoint, req.remoteAddr

node/hyperbahn/service_proxy.js
```diff
@@ -141,7 +141,11 @@ function rateLimit(req, buildRes) {
     if (self.rateLimiter.shouldRateLimitTotalRequest(req.serviceName)) {
         var totalLimit = self.rateLimiter.totalRequestCounter.rpsLimit;
         self.logger.info('hyperbahn node is rate-limited by the total rps limit', {
-            rpsLimit: totalLimit
+            rpsLimit: totalLimit,
```
Raynos: Let's add currentRps

node/hyperbahn/service_proxy.js
```diff
@@ -149,10 +153,13 @@ function rateLimit(req, buildRes) {
 
     // check RPS for service limit
     if (isExitNode && self.rateLimiter.shouldRateLimitService(req.serviceName)) {
-        var serviceLimit = self.rateLimiter.rpsLimitForServiceName[req.serviceName];
+        var serviceLimit = self.rateLimiter.getRpsLimitForService(req.serviceName);
         self.logger.info('hyperbahn service is rate-limited by the service rps limit', {
+            rpsLimit: serviceLimit,
```
Raynos: Let's add currentRps

ShanniLi: there is no need since currentRps will be equals to serviceLimit at that point.

node/hyperbahn/service_proxy.js
ShanniLi: there is no need since currentRps will be equals to serviceLimit at that point.

jcorbin: that is the hope, but we should still log it because reality

node/hyperbahn/service_proxy.js
jcorbin: same argument, log it because reality

jcorbin:
Has anyone stopped to think that invoking the slow logging path for every rate limited request is perhaps a Really Bad Idea (tm)?  I'd expect that we'd be able to take down hyperbahn exit nodes with sufficient level of this...

ShanniLi:
@jcorbin good point! The only reason to have a log there is to help debugging issues. Should we do a sampling?

Maybe removing the log is OK, since we already have stats coverage.

Raynos:
@jcorbin we already log every error frame. 

Error frames can ddos Hyperbahn. This is a larger issue and there is an entire logging section on the routing roadmap.

PR #1105, opened 2015-08-04
Fix streamingReq leak once and for all.

Raynos:
Third time lucky.

Note: I'm using assertCleanState() in alloc-cluster;
we should use this for other resource leaks.


ShanniLi:
lgtm. Consider using finishEvent if it doesn't work this time. :)

PR #1106, opened 2015-08-04
Unadvertise now

anson627:
Close peer and delete it from peer heap when unadvertise
Not destroy hyperbahn client so that it can re-advertise
@ShanniLi @Raynos @kriskowal 
cc: @davewhat @vipulaneja

### DIFF
--- a/node/hyperbahn/handler.js
+++ b/node/hyperbahn/handler.js
@@ -90,19 +90,21 @@ HyperbahnHandler.prototype.type = 'hyperbahn.advertisement-handler';
         connectionCount: Number
     }
 */
+
 HyperbahnHandler.prototype.handleAdvertise =
 function handleAdvertise(self, req, arg2, arg3, cb) {
-    self.sendRelays(self, req, arg2, arg3, 'relay-ad', cb);
+    self.sendRelays(req, arg2, arg3, 'relay-ad', cb);
 };
 
 HyperbahnHandler.prototype.handleUnadvertise =
 function handleUnadvertise(self, req, arg2, arg3, cb) {
-    self.sendRelays(self, req, arg2, arg3, 'relay-unad', cb);
+    self.sendRelays(req, arg2, arg3, 'relay-unad', cb);
 };
 
 HyperbahnHandler.prototype.sendRelays =
-function sendRelays(self, req, arg2, arg3, endpoint, cb) {
-    /*eslint max-statements: [2, 25], max-params: [2, 5]*/
+function sendRelays(req, arg2, arg3, endpoint, cb) {
+    /*eslint max-statements: [2, 25], max-params: [2, 6]*/
+    var self = this;
     var services = arg3.services;
 
     var servicesByExitNode = {};
@@ -196,17 +198,17 @@ function sendAdvertise(services, options, callback) {
 
 HyperbahnHandler.prototype.handleRelayAdvertise =
 function handleRelayAdvertise(self, req, arg2, arg3, cb) {
-    self.handleRelay(self, req, arg2, arg3, cb, self.advertise);
+    self.handleRelay(req, arg2, arg3, cb, self.advertise);
 };
 
 HyperbahnHandler.prototype.handleRelayUnadvertise =
 function handleRelayUnadvertise(self, req, arg2, arg3, cb) {
-    self.handleRelay(self, req, arg2, arg3, cb, self.unadvertise);
+    self.handleRelay(req, arg2, arg3, cb, self.unadvertise);
 };
 
 HyperbahnHandler.prototype.handleRelay =
-function handleRelay(self, req, arg2, arg3, cb, func) {
-    /*eslint max-params: [2, 5]*/
+function handleRelay(req, arg2, arg3, cb, func) {
+    var self = this;
     var services = arg3.services;
     var logger = self.channel.logger;
 
--- a/node/hyperbahn/handler.js
+++ b/node/hyperbahn/handler.js
@@ -90,7 +90,6 @@ HyperbahnHandler.prototype.type = 'hyperbahn.advertisement-handler';
         connectionCount: Number
     }
 */
-
 HyperbahnHandler.prototype.handleAdvertise =
 function handleAdvertise(self, req, arg2, arg3, cb) {
     self.sendRelays(req, arg2, arg3, 'relay-ad', cb);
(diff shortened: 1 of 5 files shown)


node/hyperbahn/index.js
```diff
@@ -362,12 +362,22 @@ function advertise(opts) {
...
     var self = this;
-    self.destroy();
+    self.sendRequest(opts, 'unad', unadvertiseInternalCb);
+    timers.clearTimeout(self._advertisementTimer);
+    timers.clearTimeout(self.advertisementTimeoutTimer);
     self.latestAdvertisementResult = null;
     self.state = States.UNADVERTISED;
     self.emit('unadvertised');
+    function unadvertiseInternalCb(error, result) {
+        self.logger.debug('HyperbahnClient: unadvertisement complete', {
```
ShanniLi: please use logger.warn

node/hyperbahn/index.js
```diff
@@ -387,7 +397,7 @@ HyperbahnClient.prototype.getHealthyRetryTime = function getHealthyRetryTime() {
 };
 
 HyperbahnClient.prototype.advertiseAgain =
-function advertiseAgain(delay) {
+    function advertiseAgain(delay) {
```
ShanniLi: why space here?

node/test/hyperbahn/unadvertise.js
```diff
@@ -35,7 +35,6 @@ if (require.main === module) {
 function runTests(HyperbahnCluster) {
     HyperbahnCluster.test('advertise and forward', {
         size: 5,
-        servicePurgePeriod: 50
```
ShanniLi: Please keep this test and create a new one in this file for unadvertise endpoint.

node/test/hyperbahn/unadvertise.js
```diff
@@ -75,8 +74,7 @@ function runTests(HyperbahnCluster) {
     });
 
     HyperbahnCluster.test('advertise, unadvertise and forward', {
-        size: 5,
-        servicePurgePeriod: 50
```
ShanniLi: Please keep this test and create a new one in this file for unadvertise endpoint.

node/hyperbahn/handler.js
```diff
@@ -178,9 +195,20 @@ function sendAdvertise(services, options, callback) {
 
     res: {}
 */
+
 HyperbahnHandler.prototype.handleRelayAdvertise =
 function handleRelayAdvertise(self, req, arg2, arg3, cb) {
-    /*eslint max-params: [2, 5]*/
+    self.handleRelay(req, arg2, arg3, cb, self.advertise);
```
ShanniLi: This is generally bad. The reason it works here is self.advertise doesn't use this pointer.

node/hyperbahn/index.js
anson627: you mean warn on error, otherwise debug?

node/hyperbahn/index.js
anson627: will remove

node/test/hyperbahn/unadvertise.js
anson627: will do

node/test/hyperbahn/unadvertise.js
anson627: will do

node/hyperbahn/handler.js
```diff
@@ -296,3 +324,8 @@ HyperbahnHandler.prototype.advertise =
 function advertise(service) {
     throw new Error('not implemented');
 };
+
+HyperbahnHandler.prototype.unadvertise =
```
ShanniLi: This should work:
self.channel.topChannel.handler.removeServicePeer(service.serviceName, service.hostPort);

anson627: oh, so you can get top channel from sub channel? let me try.

node/hyperbahn/handler.js
anson627: I don't understand the purpose for passing self as a parameter.

node/hyperbahn/index.js
ShanniLi: Oh, I missed this this for all cases. You may want to do logger.warn for error cases.

node/hyperbahn/handler.js
ShanniLi: in this case, when func is called, "this" pointer will be null. When you do callback binding, the first parameter is "this".

node/hyperbahn/index.js
anson627: will do

node/hyperbahn/handler.js
anson627: I see, what do you suggest doing here? I want to share most of the code for advertise and unadvertise.

node/hyperbahn/handler.js
anson627: it works.

node/hyperbahn/handler.js
ShanniLi: self.advertise.bind(self) is a correct way, but @Raynos said it is not efficient.

Another way is to pass in the endpoint name and do an if-else in self.handleRelay

ShanniLi:
:+1: 
you will also need to update hyperbahn for .advertise

anson627:
@ShanniLi I can just remove the advertise function override from hyperbahn, right?

ShanniLi:
yes

node/hyperbahn/handler.js
```diff
@@ -294,5 +329,12 @@ function logError(err, opts, response) {
 
 HyperbahnHandler.prototype.advertise =
 function advertise(service) {
-    throw new Error('not implemented');
+    var self = this;
+    self.channel.topChannel.handler.refreshServicePeer(service.serviceName, service.hostPort);
```
Raynos: This is not ideal; Would be better to do `self.serviceProxy.refreshServicePeer()`

node/hyperbahn/index.js
```diff
@@ -362,12 +362,29 @@ function advertise(opts) {
...
+    function unadvertiseInternalCb(error, result) {
+        if (error) {
+            self.logger.warn('HyperbahnClient: unadvertisement failure', {
+                error: error,
+                serviceName: self.serviceName,
+                hostPort: self.hostPort
+            });
+            return;
+        }
+        self.logger.debug('HyperbahnClient: unadvertisement success', {
```
Raynos: No `debug()`.

Raynos:
@anson627 Let's open a new PR to remove the `logger.debug()`.

We generally do not want any debug logs in tchannel. Logging that you did what you were told is generally not very useful,

Raynos:
Other wise looks good.

node/hyperbahn/index.js
anson627: Will do. Good to know our logging strategy.

node/hyperbahn/handler.js
anson627: will do in my next change.

anson627: HyperbahnHandler doesn't have reference to ServiceDispatchHandler(service_proxy)

PR #1108, opened 2015-08-04
[WIP] go: Thrift streaming prototype

prashantv:
Prototype for Thrift streaming. Still to do before this can be merged:
- [ ] Clean up server registration API
- [ ] Thrift client should be refactored to share code
- [ ] Streaming server should not require Thrift client
- [ ] Add tests for all 9 combinations of req/res possibilities (req/res can be void, static args, stream)
- [ ] Add support for primitive Thrift types (bool, i32, etc) to work as return types in streaming code generator.
- [ ] Support Thrift exceptions


prashantv:
Moved to tchannel-go.

PR #1109, opened 2015-08-04
go: use int8 for byte to match Thrift compiler

prashantv:
Even though Go has a byte type which should be used, the Thrift compiler uses int8. We need to match the Thrift compiler, so use int8 till we can fix the Thrift compiler.

### DIFF
--- a/golang/thrift/thrift-gen/names.go
+++ b/golang/thrift/thrift-gen/names.go
@@ -113,8 +113,10 @@ func goPublicFieldName(name string) string {
 }
 
 var thriftToGo = map[string]string{
-	"bool":   "bool",
-	"byte":   "byte",
+	"bool": "bool",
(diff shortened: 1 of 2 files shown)


kriskowal:
Thrift defines `byte` as a signed 8 bit integer. https://thrift.apache.org/docs/types

Go defines `byte` as *unsigned* and `int8` as signed. http://golang.org/pkg/builtin/#byte http://golang.org/pkg/builtin/#in8

It sounds like int8 is the way to go, although the mismatch between Thrift and Go is annoying. Go’s semantics make more sense to me.

prashantv:
Ahh, thanks Kris, that makes more sense. I guess they thought of C `char` but called it `byte`, while `byte` in Go is used when dealing with binary, where you want unsigned.

I'll remove my TODO.

kriskowal:
Blame Java. :trollface:

mmihic:
Always

PR #1110, opened 2015-08-05
downgrade sse4 for node 0.10.26

Raynos:
Latest version of sse4 depends on nan@2 which refuses
to compile in node 0.10.26

r: @kriskowal @shannili


ShanniLi:
lgtm

PR #1114, opened 2015-08-05
go: Temporary fix for message exhange leak on timeout

prashantv:
This is a temporary fix to avoid leaking goroutines (required for another change).

### DIFF
--- a/golang/inbound.go
+++ b/golang/inbound.go
@@ -166,6 +166,15 @@ func (c *Connection) dispatchInbound(call *InboundCall) {
 		return
 	}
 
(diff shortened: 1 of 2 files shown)


golang/inbound.go
```diff
@@ -166,6 +166,15 @@ func (c *Connection) dispatchInbound(call *InboundCall) {
 		return
 	}
 
+	// TODO(prashant): This is an expensive way to check for cancellation, and is not thread-safe.
```
mmihic: I was thinking at the time that maybe we'd maintain a heap of exchanges ordered by deadline, and have a ticker thread that just checked the heap and purged any exchanges that had expired.

PR #1115, opened 2015-08-05
go: Fix goroutine leaks and add tests to detect leaks

prashantv:
Add a test which checks all stacks for goroutines started by tchannel and makes sure there are none. The verification is run as the last test to make sure no previous tests have left any leaks.

Fixes the following leaks:
 - When a connection is closed on the remote side, the writeFrames goroutine did not end as it blocked on sendCh forever. Wire up Close to close the sendCh.
 - Fix race condition when a channel is closed and an outgoing call is made where a connection may not be closed even though the channel is closing.
 - On a protocol error, close the channel after the queued errorMessage is sent
 - Fix tests that created channels and did not close them before the end of the test

r: @mmihic

### DIFF
--- a/golang/channel.go
+++ b/golang/channel.go
@@ -308,7 +308,7 @@ func (ch *Channel) serve() {
 
 		// Register the connection in the peer once the channel is set up.
 		onActive := func(c *Connection) {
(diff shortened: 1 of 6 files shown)


mmihic:
lgtm

PR #1118, opened 2015-08-05
Remove duplicate 're' flag table

willsalz:
I'm assuming that this was just a copy-pasta error.

r @Raynos @kriskowal

### DIFF
--- a/docs/protocol.md
+++ b/docs/protocol.md
@@ -661,15 +661,6 @@ Value is name of the service making the call.
 
 ### Transport Header `re` -- Retry Flags
 
-Flags are encoded as a UTF-8 string, as are all header values. Each flag is
-represented by a single character:
-
-| flag | meaning
-| ---- | -------
-| `n`  | no retry, expose all errors immediately - cancels `c` and `t`
-| `c`  | retry on connection error. The specific mechanism of retry is not specified by tchannel. Actual retries are handled by  the routing layer.
-| `t`  | retry on timeout
-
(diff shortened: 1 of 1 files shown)


prashantv:
It's actually not a duplicate - the first table lists the different flags, and the second table shows valid combinations of these flags.

willsalz:
ok...that's not super clear.

docs/protocol.md
```diff
@@ -661,15 +661,6 @@ Value is name of the service making the call.
...
-Flags are encoded as a UTF-8 string, as are all header values. Each flag is
-represented by a single character:
-
-| flag | meaning
-| ---- | -------
-| `n`  | no retry, expose all errors immediately - cancels `c` and `t`
-| `c`  | retry on connection error. The specific mechanism of retry is not specified by tchannel. Actual retries are handled by  the routing layer.
-| `t`  | retry on timeout
-
 Valid values for `re` are:
```
jcorbin: Perhaps it would help to re-word this to something like "Valid combinations of these flags are:"

docs/protocol.md
```diff
@@ -661,15 +661,6 @@ Value is name of the service making the call.
 
 ### Transport Header `re` -- Retry Flags
 
-Flags are encoded as a UTF-8 string, as are all header values. Each flag is
```
jcorbin: Maybe a lead in such as: "The value of this field is one or more of the following flags:"

PR #1122, opened 2015-08-06
Remote config control for circuits feature

kriskowal:
Alternate solution to circuit enable/disable that uses just one flag.

r @jcorbin @raynos @ShanniLi

### DIFF
--- a/node/hyperbahn/service_proxy.js
+++ b/node/hyperbahn/service_proxy.js
@@ -51,7 +51,8 @@ function ServiceDispatchHandler(options) {
     self.permissionsCache = options.permissionsCache;
     self.serviceReqDefaults = options.serviceReqDefaults || {};
 
-    self.circuits = options.circuitsConfig && options.circuitsConfig.enabled ? new Circuits({
+    self.circuitsEnabled = options.circuitsConfig && options.circuitsConfig.enabled || false;
+    self.circuits = self.circuitsEnabled ? new Circuits({
         timeHeap: self.channel.timeHeap,
         timers: self.channel.timers,
         random: self.random,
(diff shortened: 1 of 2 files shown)


Raynos:
:-1: I've seen bugs with the `LockedHealthy` state in production.

node/hyperbahn/service_proxy.js
```diff
@@ -252,7 +258,7 @@ function createServiceChannel(serviceName) {
     var handler = new RelayHandler(svcchan);
 
     // Decorate a circuit health monitor to egress request handlers.
-    if (mode === 'exit' && self.circuits) {
+    if (mode === 'exit') {
```
Raynos: Let's keep the if here.

node/hyperbahn/service_proxy.js
```diff
@@ -536,6 +540,18 @@ function destroy() {
     self.rateLimiter.destroy();
 };
 
+ServiceDispatchHandler.prototype.enableCircuits =
+function enableCircuits() {
```
Raynos: Let's mutate all the exit handlers here.

node/hyperbahn/service_proxy.js
```diff
@@ -536,6 +540,18 @@ function destroy() {
...
+    var self = this;
+    self.circuits.reset(states.HealthyState);
+};
+
+ServiceDispatchHandler.prototype.disableCircuits =
+function disableCircuits() {
```
Raynos: Let's mutate all the exit handlers here.

ShanniLi: +1, the current approach is a little confusing.

kriskowal:
Alternate solution, now atop the simplified circuit thru relay code @jcorbin has since landed, that turns enables and disables circuits with a single flag on the service dispatch handler, to which relays refer.

r @jcorbin @rf

jcorbin:
lgtm

PR #1125, opened 2015-08-07
Update as-http monitorable

anson627:
Add callback to forwardToHTTP so that xlate can log and measure the latency of ingress http traffic
Remove all debug level log entries based on the logging convention
@Raynos @ShanniLi @jcorbin 
cc: @davewhat @vipulaneja


Raynos:
@anson627 can you also update the examples ?

jcorbin:
Okay once examples are updated

anson627:
done.

PR #1128, opened 2015-08-07
configurable k value

ShanniLi:
r @Raynos @kriskowal

### DIFF
--- a/node/hyperbahn/egress-nodes.js
+++ b/node/hyperbahn/egress-nodes.js
@@ -59,6 +59,13 @@ EgressNodes.prototype.kValueFor = function kValueFor(serviceName) {
         self.defaultKValue;
 };
 
+EgressNodes.prototype.setDefaultKValue = function setDefaultKValue(kValue) {
+    var self = this;
+    if (typeof kValue === 'number' && kValue > 0 && self.defaultKValue !== kValue) {
+        self.defaultKValue = kValue;
+    }
+};
+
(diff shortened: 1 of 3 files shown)


node/hyperbahn/egress-nodes.js
```diff
@@ -59,6 +59,13 @@ EgressNodes.prototype.kValueFor = function kValueFor(serviceName) {
         self.defaultKValue;
 };
 
+EgressNodes.prototype.setDefaultKValue = function setDefaultKValue(kValue) {
+    var self = this;
+    if (typeof kValue === 'number' && kValue > 0 && self.defaultKValue !== kValue) {
```
Raynos: no if check. do an `assert()` if you must.

ShanniLi: If someone mess up with the config file, or the config file is corrupted, assert will cause uncaught exceptions. It should be avoided.

jcorbin: exactly, that's called a programmer error, break fast, break hard, break early, use an assert

jcorbin:
This PR seems incomplete?  You add a new method but don't use it?  Context in comments, docs, or something more please.

Raynos:
It's used in rt/hyperbahn as part of flipr default k value.

Raynos:
lgtm.

Raynos:
Actually lets add a test.

PR #1130, opened 2015-08-07
RelayRequest: make double response logs consistent

jcorbin:
Noticed while reviewing #1094

r @Raynos


Raynos:
lgtm

node/relay_handler.js
```diff
@@ -199,14 +199,15 @@ RelayRequest.prototype.createOutResponse = function createOutResponse(options) {
     // It is also possible that the out request gets repead with a timeout
     // Both the in & out req try to create an outgoing response
     if (self.inreq.res && self.inreq.res.codeString === 'Timeout') {
-        self.logger.debug('relay: in request already timed out', {
+        self.logger.debug('relay request already timed out', {
```
Raynos: can you also upgrade to `info()` ?

I think `debug()` level is useless.

jcorbin: good idea, can be done in a future pr

PR #1132, opened 2015-08-07
Greatly increase probability of egress set having length k

kriskowal:
Got this idea from @davewhat. We can almost always find k nodes. That will become important for controlling k dynamically.


kriskowal:
r @jcorbin @Raynos

Raynos:
How does this behave when rings disagree.

How does the original logic behave when rings disagree

Should we just walk the ring for k peers instead?

Raynos:
Old PR. Going to close this out.

Let's revisit if needed.

PR #1134, opened 2015-08-07
fix the as-http test: wrong ready signal was used

ShanniLi:
r @jcorbin @Raynos @kriskowal 

Ready got over signaled in this bug ...


jcorbin:
nice find, lgtm

PR #1135, opened 2015-08-07
StreamingIn{Request,Response}: error if arg1 is fragmented

jcorbin:
Making it an error ahead of dropping support in #1040 

r @kriskowal @Raynos

### DIFF
--- a/node/errors.js
+++ b/node/errors.js
@@ -25,6 +25,11 @@ var WrappedError = require('error/wrapped');
 
 // All exported errors must be in sorted order
 
+module.exports.Arg1Fragmented = TypedError({
+    type: 'tchannel.arg1-fragmented',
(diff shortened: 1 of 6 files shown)


Raynos:
Can we get a regression test ?

kriskowal:
:ok:

anson627:
Can we fix this as well?

as/http.js line 294 ~ 299
    // TODO: real soon arg1 won't be streamable / fragmentable
    if (req.streamed) {
        req.arg1.onValueReady(onArg1);
    } else {
        onArg1(null, req.arg1);
    }

jcorbin:
Good point @anson627 , I'll fix that in #1040 when I actually drop support for it entirely and deprecate the arg1 waiting functions.

kriskowal:
:ok:

PR #1123, opened 2015-08-06
Proposal for scaling down from fully connected affinity

kriskowal:
cc @mranney @jcorbin @Raynos

PR #1137, opened 2015-08-08
WIP: Distribute peers evenly among affine relays

kriskowal:
Requesting early review. This implements #1123.

TODO

- [x] make work
- [ ] test
- [ ] test
- [ ] test

r @raynos @jcorbin @ShanniLi

### DIFF
--- a/node/hyperbahn/service_proxy.js
+++ b/node/hyperbahn/service_proxy.js
@@ -30,6 +30,8 @@ var Circuits = require('../circuits');
 
 var DEFAULT_LOG_GRACE_PERIOD = 5 * 60 * 1000;
 var SERVICE_PURGE_PERIOD = 5 * 60 * 1000;
+var DEFAULT_MIN_PEERS_PER_WORKER = 5;
+var DEFAULT_MIN_PEERS_PER_RELAY = 5;
 
 function ServiceDispatchHandler(options) {
     if (!(this instanceof ServiceDispatchHandler)) {
@@ -73,6 +75,10 @@ function ServiceDispatchHandler(options) {
         numOfBuckets: options.rateLimiterBuckets
     });
     self.rateLimiterEnabled = options.rateLimiterEnabled;
+
+    self.partialAffinityEnabled = options.partialAffinityEnabled;
+    self.minPeersPerWorker = options.minPeersPerWorker || DEFAULT_MIN_PEERS_PER_WORKER;
+    self.minPeersPerRelay = options.minPeersPerRelay || DEFAULT_MIN_PEERS_PER_RELAY;
 
     self.egressNodes.on('membershipChanged', onMembershipChanged);
 
@@ -275,11 +281,65 @@ ServiceDispatchHandler.prototype.refreshServicePeer =
 function refreshServicePeer(serviceName, hostPort) {
     var self = this;
 
+    // Add hostPort to service peers, decision pending on whether to connect.
     var peer = self.getServicePeer(serviceName, hostPort);
-    peer.connect(true);
 
+    // The old way: fully connect every egress to all affine peers.
+    if (!self.partialAffinityEnabled) {
+        peer.connectTo();
+        return;
+    }
+
+    // Make a note not to collect this service
     var time = self.channel.timers.now();
     self.exitServices[serviceName] = time;
+
+    // Obtain and sort the affine worker and relay lists.
+    var relays = Object.keys(self.egressNodes.exitsFor(serviceName));
+    relays.sort();
+    var serviceChannel = self.getOrCreateServiceChannel(serviceName);
+    var workers = serviceChannel.peers.keys();
+    workers.sort();
+
+    // Find our position within the affine relay set so we can project that
+    // position into the affine worker set.
+    var relayIndex = relays.indexOf(self.channel.hostPort);
+    // istanbul ignore if
+    if (relayIndex < 0) {
+        // This should only occur if an advertisement loses the race with a
+        // relay ring membership change.
+        self.logger.warn('Relay could not find itself in the affinity set for service', {
+            serviceName: serviceName,
+            relayHostPort: self.channel.hostPort,
+            workerHostPort: hostPort,
+            relays: relays,
+            workers: workers
+        });
+        return;
+    }
+
+    // Compute the range of workers that this relay should be connected to.
+    var ratio = workers.length / relays.length;
+    var start = (relayIndex * ratio) % workers.length;
+    var length = Math.max(self.minPeersPerRelay, self.minPeersPerWorker * ratio);
+    var stop = (start + length) % workers.length;
+
+    // Adjust connections as necessary.
+    for (var workerIndex = 0; workerIndex < workers.length; workerIndex++) {
+        var shouldConnect;
+        if (start <= stop) { // ...start||||stop...
+            shouldConnect = workerIndex >= start && workerIndex < stop;
+        } else { // |||stop...start|||
+            shouldConnect = workerIndex >= start || workerIndex < stop;
+        }
+        if (shouldConnect) {
+            peer = self.getServicePeer(serviceName, workers[workerIndex]);
+            peer.connectTo();
+        }
+        // TODO Drop peers that no longer have affinity for this service, such
+        // that they may be elligible for having their connections reaped.
+    }
+
 };
 
 ServiceDispatchHandler.prototype.removeServicePeer =
@@ -644,6 +704,18 @@ ServiceDispatchHandler.prototype.disableRateLimiter =
 function disableRateLimiter() {
     var self = this;
     self.rateLimiterEnabled = false;
+};
+
+ServiceDispatchHandler.prototype.enablePartialAffinity =
+function enablePartialAffinity() {
+    var self = this;
+    self.partialAffinityEnabled = true;
+};
+
+ServiceDispatchHandler.prototype.disablePartialAffinity =
+function disablePartialAffinity() {
+    var self = this;
+    self.partialAffinityEnabled = false;
 };
 
 ServiceDispatchHandler.prototype.extendLogInfo =
--- a/node/test/lib/relay_network.js
+++ b/node/test/lib/relay_network.js
@@ -58,6 +58,10 @@ function RelayNetwork(options) {
     self.defaultServiceRpsLimit = options.defaultServiceRpsLimit;
     self.rateLimiterBuckets = options.rateLimiterBuckets;
     self.rateLimiterEnabled = options.rateLimiterEnabled;
+    self.circuitsEnabled = options.circuitsEnabled;
+    self.partialAffinityEnabled = options.partialAffinityEnabled;
+    self.minPeersPerWorker = options.minPeersPerWorker;
+    self.minPeersPerRelay = options.minPeersPerRelay;
 
     self.numPeers = self.numRelays + self.serviceNames.length * self.numInstancesPerService;
     self.clusterOptions.numPeers = self.numPeers;
@@ -169,6 +173,10 @@ RelayNetwork.prototype.setCluster = function setCluster(cluster) {
             defaultServiceRpsLimit: self.defaultServiceRpsLimit,
             rateLimiterBuckets: self.rateLimiterBuckets,
             rateLimiterEnabled: self.rateLimiterEnabled,
+            circuitsEnabled: self.circuitsEnabled,
+            partialAffinityEnabled: self.partialAffinityEnabled,
+            minPeersPerWorker: self.minPeersPerWorker,
+            minPeersPerRelay: self.minPeersPerRelay,
             circuitsConfig: self.circuitsConfig
         });
 


ShanniLi:
I was thinking do we need to do similar thing for ingress nodes as well? When there are more services on boarding, each large service will also has a high k value. It is lots of connections as well.

Raynos:
Hyperbahn is fully connected; ingress already has open ringpop sockets to egress.

ShanniLi:
That is what I meant. The number of these sockets can grow big as well.

jcorbin:
Those sockets are the same @ShanniLi, let's discuss offline if you need further clarification.

node/hyperbahn/service_proxy.js
```diff
@@ -295,11 +295,54 @@ ServiceDispatchHandler.prototype.refreshServicePeer =
...
 
+    // Add hostPort to service peers, decision pending on whether to connect.
     var peer = self.getServicePeer(serviceName, hostPort);
-    peer.connect();
 
+    // Make a note not to collect this service
     var time = self.channel.timers.now();
     self.exitServices[serviceName] = time;
+
+    var minWorkerPeers = 5; // TODO config
```
Raynos: `self.minWorkersPeer` please.

node/hyperbahn/service_proxy.js
```diff
@@ -295,11 +295,54 @@ ServiceDispatchHandler.prototype.refreshServicePeer =
...
+    var minRelayPeers = 5; // TODO config
```
Raynos: `self.minRelayPeers` please.

node/hyperbahn/service_proxy.js
```diff
@@ -295,11 +295,54 @@ ServiceDispatchHandler.prototype.refreshServicePeer =
...
+
```
Raynos: Let's have an `if (x) { fullyConnectOldSemantics() }`

node/hyperbahn/service_proxy.js
```diff
@@ -295,11 +295,54 @@ ServiceDispatchHandler.prototype.refreshServicePeer =
...
+    var serviceChannel = self.getOrCreateServiceChannel(serviceName);
+    var workers = serviceChannel.peers.keys();
+    workers.sort();
+
+    // Find our position within the affine relay set so we can project that
+    // position into the affine worker set.
+    var relayIndex = relays.indexOf(self.channel.hostPort);
+    if (relayIndex < 0) {
+        // This should not be possible.
+        self.logger.error('Relay find itself within the affine relays for an ad it received', {
```
Raynos: Message: 'Relay could not find itself in the affinity set for serviceName`

node/hyperbahn/service_proxy.js
```diff
@@ -295,11 +295,54 @@ ServiceDispatchHandler.prototype.refreshServicePeer =
...
+            serviceName: serviceName,
+            relayHostPort: self.channel.hostPort,
```
Raynos: let's also print `relays`, `workers`.

node/hyperbahn/service_proxy.js
```diff
@@ -295,11 +295,54 @@ ServiceDispatchHandler.prototype.refreshServicePeer =
...
+            workerHostPort: hostPort
+        });
+        return;
+    }
+
+    // Compute the range of workers that this relay should be connected to.
+    var ratio = workers.length / relays.length;
+    var start = relayIndex * ratio;
```
Raynos: Should we `floor()` the start here ?

node/hyperbahn/service_proxy.js
```diff
@@ -295,11 +295,54 @@ ServiceDispatchHandler.prototype.refreshServicePeer =
...
+    var stop = start + Math.max(minRelayPeers, minWorkerPeers * ratio) + 0.5;
```
Raynos: Let's clamp `stop` to be less then `workers.length`

Raynos: Why the + 0.5; smells to me.

Raynos: need to mod `stop` by relays.length for roll over to work.

node/hyperbahn/service_proxy.js
```diff
@@ -295,11 +295,54 @@ ServiceDispatchHandler.prototype.refreshServicePeer =
...
+
+    // Adjust connections as necessary.
+    for (var workerIndex = 0; workerIndex < workers.length; workerIndex++) {
+        var shouldConnect = workerIndex >= start && workerIndex < stop ||
```
Raynos:
```js
if (stop <= start) {
  tmp = stop;
  stop = start;
  start = tmp;
}

shouldConnect = (start <= workerIndex && workerIndex < stop);
```

Let's simplify this condition by swapping stop & start if they are in the wrong order because of roll around.

node/hyperbahn/service_proxy.js
```diff
@@ -295,11 +295,54 @@ ServiceDispatchHandler.prototype.refreshServicePeer =
...
+            workerIndex > stop && workerIndex <= start;
+        peer = self.getServicePeer(serviceName, workers[workerIndex]);
+        if (shouldConnect) {
+            peer.connect();
+        } else {
+            peer.close(noop);
+            self.channel.peers.delete(workers[workerIndex]);
```
Raynos: Let's not delete any peers; if all workers for this exit node die we want to connect to the other ones.

node/hyperbahn/service_proxy.js
```diff
@@ -295,11 +295,54 @@ ServiceDispatchHandler.prototype.refreshServicePeer =
...
+
+    // Adjust connections as necessary.
+    for (var workerIndex = 0; workerIndex < workers.length; workerIndex++) {
+        var shouldConnect = workerIndex >= start && workerIndex < stop ||
+            workerIndex > stop && workerIndex <= start;
+        peer = self.getServicePeer(serviceName, workers[workerIndex]);
+        if (shouldConnect) {
+            peer.connect();
+        } else {
+            peer.close(noop);
```
Raynos: `close()` is not the semantics we want; we want the semantics of `peer.closeConnections()` knowing that it's valid to re-open connections in the future on this peer object.

Raynos:
Let's get @jcorbin to chip in as well.

node/hyperbahn/service_proxy.js
kriskowal: May I raise you `minPeersPerWorker`?

node/hyperbahn/service_proxy.js
kriskowal: Consider `workers.length = 10` and `stop = 2` (because 12 % 10 is 2) and `start = 8`. 0 should be in the interval. 4 should not. If you swap start and stop, you get the exact inverse of the appropriate interval and would have to negate shouldConnect. You would also have to invert the inclusive vs exclusive boundary conditions.

node/hyperbahn/service_proxy.js
kriskowal: Turned out to be easier to check both normal ranges and inverted ranges.

I’ll swap out the 0.5 and use Math.round. The simulation used the `int(x + 0.5)` trick since Python doesn’t have round.

I’ll get the modulo in there somewhere. Thanks for catching that.

node/hyperbahn/service_proxy.js
kriskowal: This is a tricky case and I don’t fully understand the implications here. Need to get @jcorbin’s eyes on this case. My intuition is that in the static stable case, the peer list should be exactly and only the peers we determine here. When these change, the old peers should be removed from the pool, drained, closed (unless shared by another subchannel), perhaps restored to active duty later if they are refed by a peer before they get a chance to close. Retaining a peer forever just in case all the others stop advertising or close seems like a memory leak to me.

node/hyperbahn/service_proxy.js
kriskowal: Ah, that seems like a close enough approximation of the right thing to do. Channels that share peers might interfere with each other this way, but they can just reopen.

kriskowal:
Putting modulos in (oops) gets me farther along in the tests. Thanks for catching that.

node/hyperbahn/service_proxy.js
Raynos: Sure :)

node/hyperbahn/service_proxy.js
Raynos: You can't swap... :( 

In that case can we do:

```js
if (start <= stop) {
  shouldConnect = ...;
} else {
  shouldConnect = ...;
}
```

Rather then one expression that handles all permutations.

node/hyperbahn/service_proxy.js
Raynos: We have a reaper for reaping dead sub channels. We should extend the reaper for reaping dead peers.

I think keeping peers around in case all the affinity workers die is a good idea. But let's get a second opinion from @jcorbin

node/hyperbahn/service_proxy.js
kriskowal: Since we’re talking about methods that don’t exist, in the long run, I believe the semantics we need here are: remove the peer from the subchannel’s peers, and after no subchannels retain the peer (perhaps in a reaper cycle), close all connections, dropping any pending requests on the floor. This may hurt some pending requests, but they can retry.

As an expedient, I am going to submit a version of this feature that never collects peers if they have ever been connected. This should reduce the number of connected peers, but will still leak toward fully connected over time, especially while receiving initial advertisements.

node/hyperbahn/service_proxy.js
```diff
@@ -74,6 +74,10 @@ function ServiceDispatchHandler(options) {
     });
     self.rateLimiterEnabled = options.rateLimiterEnabled;
 
+    self.partialAffinityEnabled = options.partialAffinityEnabled;
+    self.minPeersPerWorker = options.minPeersPerWorker || 5;
```
Raynos: nitpick: constants at top of file.

node/hyperbahn/service_proxy.js
```diff
@@ -295,11 +299,60 @@ ServiceDispatchHandler.prototype.refreshServicePeer =
...
+
+    // Compute the range of workers that this relay should be connected to.
+    var ratio = workers.length / relays.length;
+    var start = (relayIndex * ratio) % workers.length;
+    var length = Math.max(self.minPeersPerRelay, self.minPeersPerWorker * ratio);
+    var stop = (start + length) % workers.length;
+
+    // Adjust connections as necessary.
+    for (var workerIndex = 0; workerIndex < workers.length; workerIndex++) {
+        var shouldConnect = workerIndex >= start && workerIndex < stop ||
```
Raynos:
```js
if (start <= stop) {
    shouldConnect = index >= start && index < stop;
} else {
    shouldConnect = index >= start || index < stop;
}
```

Raynos:
I'll take this feature over; add a bunch of configuration commands and hook it up to flipr.

node/hyperbahn/service_proxy.js
jcorbin: :+1: hoist constant pls k thx

node/hyperbahn/service_proxy.js
```diff
@@ -295,11 +299,60 @@ ServiceDispatchHandler.prototype.refreshServicePeer =
...
-    peer.connect();
 
+    // The old way: fully connect every egress to all affine peers.
+    if (!self.partialAffinityEnabled) {
+        peer.connect();
+        return;
+    }
+
+    // Make a note not to collect this service
     var time = self.channel.timers.now();
```
jcorbin: we used to always do this, now we don't in the "off" case, please restore behavior?

node/hyperbahn/service_proxy.js
```diff
@@ -295,11 +299,60 @@ ServiceDispatchHandler.prototype.refreshServicePeer =
...
+    // Obtain and sort the affine worker and relay lists.
+    var relays = Object.keys(self.egressNodes.exitsFor(serviceName));
+    relays.sort();
+    var serviceChannel = self.getOrCreateServiceChannel(serviceName);
+    var workers = serviceChannel.peers.keys();
+    workers.sort();
+
+    // Find our position within the affine relay set so we can project that
+    // position into the affine worker set.
+    var relayIndex = relays.indexOf(self.channel.hostPort);
```
jcorbin: could be binary bisection... #justsayin... not sure where the inflection point is for node indexOf vs bisect

node/hyperbahn/service_proxy.js
```diff
@@ -295,11 +299,60 @@ ServiceDispatchHandler.prototype.refreshServicePeer =
...
+    if (relayIndex < 0) {
+        // This should only occur if an advertisement loses the race with a
+        // relay ring membership change.
+        self.logger.warn('Relay could not find itself in the affinity set for service', {
+            serviceName: serviceName,
+            relayHostPort: self.channel.hostPort,
```
jcorbin: You Touched A Log Site --> You Factor Out A foo.extendLogInfo

node/hyperbahn/service_proxy.js
jcorbin: :+1: please clarify this logic, not using a network of only ands and ors

node/hyperbahn/service_proxy.js
```diff
@@ -295,11 +299,60 @@ ServiceDispatchHandler.prototype.refreshServicePeer =
...
+    }
+
+    // Compute the range of workers that this relay should be connected to.
+    var ratio = workers.length / relays.length;
+    var start = (relayIndex * ratio) % workers.length;
+    var length = Math.max(self.minPeersPerRelay, self.minPeersPerWorker * ratio);
+    var stop = (start + length) % workers.length;
+
+    // Adjust connections as necessary.
+    for (var workerIndex = 0; workerIndex < workers.length; workerIndex++) {
```
jcorbin: Please don't do a o(workers) loop with a branch in it; instead please have a separate passage for "connect to these", and separate the disconnect / reaping concern.  +:100: if we find a way to do that disconnect in less than o(workers).

I'm imagining something like:
```
for (var i = start, stop = start + length; i < stop; i++) {
  connectTo(workers[i % workers.length]);
  update(someFastDataStructureToMakeTheReapingNotSuck);
}
```

jcorbin:
promising... looking forward to:
- [ ] those #testsplease
- [ ] preserving old behavior For Sure
- [ ] algorithmic wins that let us beat o(service workers)

PR #1140, opened 2015-08-11
Cleanup handler validation

jcorbin:
r @kriskowal @Raynos

### DIFF
--- a/node/v2/handler.js
+++ b/node/v2/handler.js
@@ -191,7 +191,12 @@ TChannelV2Handler.prototype.handleCallRequest = function handleCallRequest(reqFr
 
     var req = self.buildInRequest(reqFrame);
 
-    if (self.incomingRequestInvalid(reqFrame, req)) {
+    var err = self.checkCallReqFrame(reqFrame);
+    if (err) {
+        req.res = self.buildOutResponse(req);
+        self.sendErrorFrame(
+            req.res, 'ProtocolError', err.message
+        );
         return;
     }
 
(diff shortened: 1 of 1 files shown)


node/v2/handler.js
```diff
@@ -183,14 +183,21 @@ TChannelV2Handler.prototype.handleInitResponse = function handleInitResponse(res
...
 
     var req = self.buildInRequest(reqFrame);
 
-    if (self.incomingRequestInvalid(reqFrame, req)) {
-        return null;
+    var err = self.checkCallReqFrame(reqFrame);
+    if (err) {
+        req.res = self.buildOutResponse(req);
+        self.sendErrorFrame(
+            req.res, 'ProtocolError', err.message
```
Raynos: We can switch here on the type of `err` and send an appropriate error type.

We can use `Errors.classify(err) || 'ProtocolError'`

Let's verify that the arg1 over length error is properly classified.

Raynos:
Let's avoid the BadRequest regression; otherwise lgtm.

node/v2/handler.js
jcorbin: Will do, good call; it's something I handle better later by simply emitting the err on req.errorEvent, which gets hooked up to such classification.  However I didn't want to add all of that in this PR.

Confirmed classification of arg1 over length, as of 4b45dec errors.js:578 has it as a `BadRequest`.

Raynos:
lgtm.

Issue #1138, opened 2015-08-08
Hyperbahn: Fix rate limiter stats to line up with request volume


Raynos:
The total rate per service number is wildly different from the inbound request volume number.

We should probably remove all the guages.

Raynos:
cc @ShanniLi

ShanniLi:
Yes, gauges may not be the right stats there. Let me try counter.

ShanniLi:
@Raynos I looked at the dashboard, the stats is correct. You cannot use sum with gauges, because it is just a sampling every one second. If you want to use sum, we need to use counter. In this case, there will be a little perf degradation.

Raynos:
@ShanniLi I want to use sum; we need to use a counter. 

Counters should be more performant then guages.

ShanniLi:
OK, I will make the change

PR #1141, opened 2015-08-11
Rate limiter stats

ShanniLi:
r @Raynos @kriskowal @rf 
for issue #1138

### DIFF
--- a/node/lib/statsd.js
+++ b/node/lib/statsd.js
@@ -120,7 +120,7 @@ TChannelStatsd.prototype.onStat = function onStat(stat) {
     var key = getKey(stat);
 
     if (stat.type === 'counter') {
-        return self.statsd.counter(key, stat.value);
+        return self.statsd.increment(key, stat.value);
     } else if (stat.type === 'gauge') {
         return self.statsd.gauge(key, stat.value);
     } else if (stat.type === 'timing') {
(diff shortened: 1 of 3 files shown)


node/lib/statsd.js
```diff
@@ -120,7 +120,7 @@ TChannelStatsd.prototype.onStat = function onStat(stat) {
     var key = getKey(stat);
 
     if (stat.type === 'counter') {
-        return self.statsd.increment(key, stat.value);
+        return self.statsd.counter(key, stat.value);
```
Raynos: do we really want this change ?

node/rate_limiter.js
```diff
@@ -112,7 +112,7 @@ function refresh() {
     if (self.cycle === self.numOfBuckets) {
         self.channel.emitFastStat(self.channel.buildStat(
             'tchannel.rate-limiting.total-rps',
-            'gauge',
+            'counter',
             self.totalRequestCounter.rps,
```
Raynos: I think this works because `refreshDelay` is 1000.

node/lib/statsd.js
Raynos: let's not make this unless needed.

ShanniLi: fixed

Raynos:
lgtm :D

PR #1144, opened 2015-08-11
Classify more errors

jcorbin:
- typed and classified all `TODO typed errors` in conn and handler
- typed and classified retry limit exceeded error

r @kriskowal @Raynos

### DIFF
--- a/node/errors.js
+++ b/node/errors.js
@@ -551,6 +551,11 @@ module.exports.UnimplementedMethod = TypedError({
     methodName: null
 });
 
+module.exports.UnknownConnectionReset = TypedError({
+    type: 'tchannel.connection.unknown-reset',
+    message: 'unknown connection reset'
+});
+
 // utilities
 
 module.exports.classify = function classify(err) {
@@ -628,6 +633,7 @@ module.exports.classify = function classify(err) {
         case 'tchannel-thrift-handler.stringify-error.body-failed':
         case 'tchannel-thrift-handler.stringify-error.head-failed':
         case 'tchannel.argstream.unknown-frame-handling-state':
+        case 'tchannel.connection.unknown-reset':
         case 'tchannel.http-handler.from-buffer-arg2.req-failed':
         case 'tchannel.http-handler.from-buffer-arg2.res-failed':
         case 'tchannel.hydrated-error.default-type':
--- a/node/errors.js
+++ b/node/errors.js
@@ -603,11 +603,13 @@ module.exports.classify = function classify(err) {
     }
 
     switch (err.type) {
+        case 'tchannel.request.retry-limit-exceeded':
+            return 'Cancelled';
+
         case 'tchannel.max-pending':
         case 'tchannel.max-pending-for-service':
         case 'tchannel.no-peer-available':
         case 'tchannel.no-service-handler':
-        case 'tchannel.request.retry-limit-exceeded':
             return 'Declined';
 
         case 'tchannel.connection-stale.timeout':
--- a/node/errors.js
+++ b/node/errors.js
@@ -556,6 +556,13 @@ module.exports.TransportHeaderTooLong = TypedError({
     endOffset: null
 });
 
+module.exports.UnexpectedCallFrameAfterDone = TypedError({
+    type: 'tchannel.call.frame-unexpected.after-done',
+    message: 'got call frame (type {frameType}) in done state',
+    frameId: null,
+    frameType: null
+});
+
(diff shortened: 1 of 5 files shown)


node/errors.js
```diff
@@ -563,6 +607,7 @@ module.exports.classify = function classify(err) {
         case 'tchannel.max-pending-for-service':
         case 'tchannel.no-peer-available':
         case 'tchannel.no-service-handler':
+        case 'tchannel.request.retry-limit-exceeded':
```
Raynos: Note: If we return `Cancelled` for retry-limit-exceeded we could enable retries at exit node; this would propagate a Cancelled when retries run out and the entry node or edge clients would not retry in a cascading fashion.

node/errors.js
```diff
@@ -628,6 +688,7 @@ module.exports.classify = function classify(err) {
         case 'tchannel-thrift-handler.stringify-error.body-failed':
         case 'tchannel-thrift-handler.stringify-error.head-failed':
         case 'tchannel.argstream.unknown-frame-handling-state':
+        case 'tchannel.connection.unknown-reset':
```
Raynos: Should this be a NetworkError ?

node/v1/handler.js
```diff
@@ -70,7 +70,7 @@ TChannelHandler.prototype.handleFrame = function handleFrame(frame) {
 TChannelHandler.prototype.handleInitRequest = function handleInitRequest(reqFrame) {
     var self = this;
     if (self.remoteName !== null) {
-        self.emit('error', new Error('duplicate init request')); // TODO typed error
+        self.emit('error', new Error('duplicate init request'));
```
Raynos: this is not typed; please do not remove TODO.

node/v1/handler.js
```diff
@@ -82,7 +82,7 @@ TChannelHandler.prototype.handleInitRequest = function handleInitRequest(reqFram
 TChannelHandler.prototype.handleInitResponse = function handleInitResponse(resFrame) {
     var self = this;
     if (self.remoteName !== null) {
-        self.emit('error', new Error('duplicate init response')); // TODO typed error
+        self.emit('error', new Error('duplicate init response'));
```
Raynos: this is not typed; please do not remove TODO.

node/v1/handler.js
```diff
@@ -101,7 +101,7 @@ TChannelHandler.prototype.handleCallRequest = function handleCallRequest(reqFram
     }
 
     if (self.remoteName === null) {
-        self.emit('error', new Error('call request before init request')); // TODO typed error
+        self.emit('error', new Error('call request before init request'));
```
Raynos: this is not typed; please do not remove TODO.

node/v1/handler.js
```diff
@@ -125,7 +125,7 @@ TChannelHandler.prototype.handleCallResponse = function handleCallResponse(resFr
     }
 
     if (self.remoteName === null) {
-        self.emit('error', new Error('call response before init response')); // TODO typed error
+        self.emit('error', new Error('call response before init response'));
```
Raynos: this is not typed; please do not remove TODO.

Raynos:
Request changes: Lets fix comments

Otherwise, lgtm.

node/v1/handler.js
jcorbin: it's v1, we're not changing v1, so its "TODOs" end in "WONTFIX" essentially ;-)

node/v1/handler.js
jcorbin: it's v1, we're not changing v1, so its "TODOs" end in "WONTFIX" essentially ;-)

node/v1/handler.js
jcorbin: it's v1, we're not changing v1, so its "TODOs" end in "WONTFIX" essentially ;-)

node/v1/handler.js
jcorbin: it's v1, we're not changing v1, so its "TODOs" end in "WONTFIX" essentially ;-)

node/errors.js
jcorbin: Good idea, let's do this and see how it goes.

node/errors.js
jcorbin: I don't think we can say that with confidence, all this means is "someone forgot to pass an error to conn.resetAll"

jcorbin:
Not going to leave the TODOs in v1, since v1 code is essentially frozen.

PR #1145, opened 2015-08-11
Improve req/res error handling and visibility

jcorbin:
- add req error handling to send error frames out of handler
- add in-res error handling to re-emit errors on the corresponding out-req
- emit classified non-protocol errors from request handling paths on the corresponding request

TODO:
- [x] maybe kill off TChannelConnectionBase#onReqError
- [x] pass tests

r @Raynos @kriskowal

### DIFF
--- a/node/v2/handler.js
+++ b/node/v2/handler.js
@@ -906,7 +906,9 @@ TChannelV2Handler.prototype.onReqError = function onReqError(err, req) {
     var self = this;
 
     var codeName = errors.classify(err);
-    if (codeName) {
+    if (codeName &&
+        codeName !== 'ProtocolError'
+    ) {
         // TODO: move req to error state?
         if (!req.res) {
             req.res = self.buildOutResponse(req);
--- a/node/v2/handler.js
+++ b/node/v2/handler.js
@@ -193,16 +193,7 @@ TChannelV2Handler.prototype.handleCallRequest = function handleCallRequest(reqFr
 
     var err = self.checkCallReqFrame(reqFrame);
     if (err) {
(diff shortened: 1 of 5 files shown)


node/v2/handler.js
```diff
@@ -857,10 +849,49 @@ TChannelV2Handler.prototype.buildInRequest = function buildInRequest(reqFrame) {
...
+
+TChannelV2Handler.prototype.onReqError = function onReqError(err, req) {
+    var self = this;
+
+    var codeName = errors.classify(err);
+    if (codeName &&
+        codeName !== 'ProtocolError'
+    ) {
+        // TODO: move req to error state?
+        req.res = self.buildOutResponse(req);
```
Raynos: I think this entire code path is an "anti pattern".

We should favor `res.sendError()` which will correctly manage the entire OutResponse state machine.

node/v2/handler.js
```diff
@@ -857,10 +849,49 @@ TChannelV2Handler.prototype.buildInRequest = function buildInRequest(reqFrame) {
...
+TChannelV2Handler.prototype.onResError = function onResError(err, res) {
+    var self = this;
+
+    var codeName = errors.classify(err);
+    if (codeName &&
+        codeName !== 'ProtocolError'
+    ) {
+        // TODO: wrap errors to clarify "errors on responses to req..." ?
+        var req = self.connection.ops.getOutReq(res.id);
+        req.errorEvent.emit(req, err);
```
Raynos: Why emit on the `OutRequest` rather then emitting on the `InResponse` ?

kriskowal:
:ok: Looks good to me.

PR #1147, opened 2015-08-12
Update protocol documentation for errors

prashantv:
Error messages contain the id, not the trace ID as the summary shows.

r @Raynos


prashantv:
ready for review

Raynos:
lgtm.

PR #1148, opened 2015-08-12
go: errorFrame contains tracing span, and uses ID from the header 

prashantv:
Related: #1147 for information about how tracing information should be filled.


prashantv:
cc @jsu1212 This should fix the error you saw (misunderstanding of the protocol spec that I cleared up).

PR #1149, opened 2015-08-12
Improve logging around stale tombstone operations

Raynos:
Trying to track down stale tombstones. Turns out that
the relay was a dead end.

The relay has more tombstones then the clients because
the clients re-cycle connections and thus clear tombstones.

If you we do 200k requests instead of 20k requests per test
batch you find that relay & clients have same number of
tombstones.

I also fixed a race condition between heap & stale tombstone
reaper that would prematurely warn about stale tombstones.

r: @jcorbin @shannili

### DIFF
--- a/node/monitor.js
+++ b/node/monitor.js
@@ -32,6 +32,7 @@ function ChanConnMonitor(channel, options) {
     self.channel = channel;
     self.interval = options.interval;
     self.byBucket = options.byBucket;
+    self.byTTL = options.byTTL;
     self.timers = channel.timers;
     self.timer = null;
     self.running = false;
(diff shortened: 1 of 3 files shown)


ShanniLi:
:+1:

node/monitor.js
```diff
@@ -165,21 +167,98 @@ OpKindMonitor.prototype.summary = function summary() {
     if (self.inCounts.length) {
         self.log('= %s IN COUNTS: %j',
                     self.options.desc,
-                    self.inCounts.reduce(sumCounts));
+                    self.inCounts.reduce(sumCounts, {}));
```
kriskowal: Should probably pop these uses of array methods regardless of how cold these paths are.

node/time_heap.js
```diff
@@ -77,6 +77,7 @@ function TimeHeap(options) {
     self.lastTime = self.timers.now();
     self.timer = null;
     self.end = 0;
+    self.lastRun = 0;
```
kriskowal: null is a more sensible default value since it is unlikely that this would have run in the 70’s.

kriskowal:
:ok:

node/time_heap.js
jcorbin: disagree, #typestability please

node/operations.js
```diff
@@ -85,9 +87,25 @@ OperationTombstone.prototype.extendLogInfo = function extendLogInfo(info) {
...
+
+    self.destroyed = true;
+
+    self.onTimeout(now);
+};
+
 OperationTombstone.prototype.onTimeout = function onTimeout(now) {
     var self = this;
 
+    if (!self.destroyed && now < self.timeout + self.time) {
```
jcorbin: Hmm, if they truly are timing out early, that's a bug in the time heap.  Can you please pull out this "fix" into a separate PR from the instrumentation changes?  I'd then rather try to see if we can fix the heap bug (or prove me wrong in labelling it such), but this kind of logic (and its corollary below) is exactly the sort of thing we're supposed to have encapsulated away behind the time heap.

jcorbin:
Great find, but please break it up so we can focus more on the cause of the problem and hopefully fix it, rather than the symptom ("early" timeouts shouldn't be a thing).

node/operations.js
Raynos: Early time-outs are a thing in the destroy() path. Which is why I have a destroyed gaurd.

This explains why the client and relay have different number of tombstones in the benchmark. 

I don't think we have a heap bug.

node/time_heap.js
kriskowal: Would you buy -Infinity? It’s legit, but not a smallint. I won’t insist. The 70’s are not likely to occur again and might as well be infinity ago.

Raynos:
Closing because three more PRs.

 - #1153 
 - #1152
 - #1151

PR #1150, opened 2015-08-12
[tchannel-go] - Allow subchannel handler registrations

aravindvs:
@prashantv 

This patch allows the subchannels to register their own set of handlers
independent of the topchannel. They will still be using the topchannel's
host/port for all the traffic.

The connection management involves maintaining a pointer to the
subchannel map (similar to handler map). When a request comes in, we
first look in the handler map to find the handler, if it fails, we try
the subchannel map to see if we can get one.
(In case of a non-existant handler, we will effectively be doing 3 O(1)
lookups. If all the traffic comes through subchannel servicename/op map we will
have one unnecessary lookup on the topchannel)

This shouldn't break any existing APIs. The ping test uses this new feature to register its
own handler on a subchannel.
TODO:
----
* Add peer support for subchannels - right now they are using the same
  peer list as the topchannel

### DIFF
--- a/golang/channel.go
+++ b/golang/channel.go
@@ -96,15 +96,15 @@ type Channel struct {
 	connectionOptions ConnectionOptions
 	handlers          *handlerMap
 	peers             *PeerList
+	subChannels       *subChannelMap
 
 	// mutable contains all the members of Channel which are mutable.
 	mutable struct {
-		mut         sync.RWMutex // protects members of the mutable struct.
-		state       ChannelState
-		peerInfo    LocalPeerInfo // May be ephemeral if this is a client only channel
-		l           net.Listener  // May be nil if this is a client only channel
-		subChannels map[string]*SubChannel
-		conns       []*Connection
+		mut      sync.RWMutex // protects members of the mutable struct.
+		state    ChannelState
+		peerInfo LocalPeerInfo // May be ephemeral if this is a client only channel
+		l        net.Listener  // May be nil if this is a client only channel
+		conns    []*Connection
 	}
 }
 
@@ -142,6 +142,7 @@ func NewChannel(serviceName string, opts *ChannelOptions) (*Channel, error) {
 		statsReporter:     statsReporter,
 		traceReporter:     traceReporter,
 		handlers:          &handlerMap{},
+		subChannels:       &subChannelMap{},
 	}
 	ch.mutable.peerInfo = LocalPeerInfo{
 		PeerInfo: PeerInfo{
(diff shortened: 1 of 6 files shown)


golang/channel.go
```diff
@@ -231,34 +231,19 @@ func (ch *Channel) createCommonStats() {
...
-	mutable.subChannels[serviceName] = sc
-	return sc
-}
-
 // GetSubChannel returns a SubChannel for the given service name. If the subchannel does not
 // exist, it is created.
 func (ch *Channel) GetSubChannel(serviceName string) *SubChannel {
-	mutable := &ch.mutable
-	mutable.mut.RLock()
+	subChMap := ch.subChannels
```
prashantv: This logic should live inside of subChannelMap. I'd probably just make a "getOrAdd" style function this uses.

golang/json/handler.go
```diff
@@ -117,6 +117,33 @@ func Register(ch *tchannel.Channel, funcs Handlers, onError func(context.Context
 	return nil
 }
 
+func RegisterSub(ch *tchannel.SubChannel, funcs Handlers, onError func(context.Context, error)) error {
```
prashantv: Rather than do this, let's create a separate interface in tchannel that only contains "Register". That way, both tchannel.Channel and tchannel.Subchannel can be passed to the same Register function, rather than duplicating logic.

Maybe something like:
```go
type Registrar interface {
  func Register(h Handler, operationName string)
}
```

golang/subchannel.go
```diff
@@ -29,10 +43,47 @@ func (c *SubChannel) BeginCall(ctx context.Context, operationName string, callOp
...
+// Register registers a handler on the subchannel for a service+operation pair
+func (c *SubChannel) Register(h Handler, operationName string) {
+	c.handlers.register(h, c.ServiceName(), operationName)
+}
+
+// Find if a handler for the given service+operation pair exists
+func (subChMap *subChannelMap) find(serviceName string, operation []byte) Handler {
+	subChMap.mut.RLock()
+
+	if sc, ok := subChMap.subchannels[serviceName]; ok {
```
prashantv: getOrAdd should be able to reuse find.

prashantv:
Thanks for starting this @aravindvs

golang/channel.go
aravindvs: moved the meat of the logic within subChannelMap..

golang/json/handler.go
aravindvs: @prashantv - thanks for the suggestion.. removed code duplication now by doing this.. the recent commit should address this..

golang/subchannel.go
aravindvs: Done..

aravindvs:
@prashantv - thanks for the feedback.. I addressed the comments in the recent commit.

prashantv:
lgtm, thanks @aravindvs, feel free to merge.

PR #1151, opened 2015-08-12
Add more features to monitor for debugging purposes

Raynos:
Here I added a feature to view tombstones by:

 - TTL
 - Time bucket per 5s.

This allows us to determine whether or the tombstones are
inbalanced per timestamp or evenly spread out.


jcorbin:
lgtm, we can address any perf concerns going forward.

PR #1152, opened 2015-08-12
Fix tombstone + timeheap race + more logging

Raynos:
This PR improves the stale tombstone detection by avoiding a 
tombstone reaper + timeHeap race condition.

It also improves logging to get more visibility into the
production stale tombstones.


Raynos:
^ flap :(

kriskowal:
Looks good to me.

PR #1155, opened 2015-08-13
Workaround http response status message issue

anson627:
this is a workaround for issue that http response writeHead doesn't handle status message well
since our current node version is 0.10, so head.message is an empty string and the else block will always be executed.
it's not clear to me if this issue will be resolved in 0.12 or not:
https://github.com/joyent/node/issues/25490
@jcorbin @Raynos @ShanniLi 
cc: @davewhat @vipulaneja

### DIFF
--- a/node/as/http.js
+++ b/node/as/http.js
@@ -229,8 +229,7 @@ TChannelHTTP.prototype.forwardToTChannel = function forwardToTChannel(tchannel, 
                 headers[pair[0]] = pair[1];
             }
             // work-arround a node issue where default statusMessage is missing
-            // from the client side when server side try to set it
-            // https://github.com/joyent/node/issues/25490
+            // from the client side when server side set as optional parameter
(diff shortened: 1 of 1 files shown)


node/as/http.js
```diff
@@ -228,7 +228,12 @@ TChannelHTTP.prototype.forwardToTChannel = function forwardToTChannel(tchannel,
                 var pair = head.headerPairs[i];
                 headers[pair[0]] = pair[1];
             }
-            hres.writeHead(head.statusCode, head.message, headers);
+            // work-around https://github.com/joyent/node/issues/25490
```
ShanniLi: may be a good idea to expand this url to a short comment explaining why the code is necessary

ShanniLi:
left one comment. Otherwise, lgtm.

anson627:
go build failed. checking in.

PR #1158, opened 2015-08-14
Fix as raw docs

breerly:
We incorrectly say that `arg1` is raw bytes instead of endpoint, leading @willsalz to think there can only be 1 raw endpoint per service and myself to create this https://github.com/uber/tchannel-python/issues/113

### DIFF
--- a/docs/raw.md
+++ b/docs/raw.md
@@ -10,6 +10,6 @@ before using `as=raw`.
 
 ## Arguments
 
- - `arg1` : raw bytes
+ - `arg1` : endpoint
  - `arg2` : raw bytes
  - `arg3` : raw bytes


docs/raw.md
```diff
@@ -10,6 +10,6 @@ before using `as=raw`.
 
 ## Arguments
 
- - `arg1` : raw bytes
+ - `arg1` : endpoint
```
willsalz: shouldn't this be 'UTF-8' encoded String?

breerly: Probably :)

Raynos:
If your going to work with Hyperbahn then `arg1` needs to be a small UTF8 string of small cardinality of metrics will find you.

jcorbin:
We should leave this as "raw bytes" only and say no more; limiting it to be a utf-8 string seems wrong to me.  This is simply a "bag of bytes on which dispatch of (shrug) form gets done".  Whether that maps at all to something resembling "endpoints" is up to the raw implementors.

willsalz:
cc @truncs

Raynos:
old Pr. Going to close this out.

Feel free to re-open if still applicable.

willsalz:
This is a huge contradiction from protocol.md and needs to be resolved.

Raynos:
@willsalz 

Why is it a huge contradiction ? 

The thrift documentation clearly documents how arg1 works for thrift ( https://github.com/uber/tchannel/blob/master/docs/thrift.md#arg1 ).

I don't see whats wrong with raw bytes.

willsalz:
Hmm, I should have been more careful with my words: so, this PR *should* be closed, but what was unresolved is whether Raw Arg Schemes can treat arg1 with any semantics. The Python, possibly other implementations, have given `arg1` semantic meaning. Maybe this PR isn't the correct forum for discussion this issue. Anyway, just wanted to bring that up. If you feel like this discussion should go in Issues feel free to close.

Not sure what this has to do with Thrift, however.

Raynos:
@willsalz 

The way I think about is that "raw" has "undefined" semantics for arg1.

There exist helper functions in the TChannel node and python library that allow you to treat arg1 as a UTF8 string and do static endpoint dispatching on it for raw endpoints.

However that's simply an implementation detail of that application / server. It would be equally valid to write a custom raw RequestHandler that treated arg1 as a user UUID or something like that.

Raynos:
I don't think this protocol doc is the right place to talk about it.

I think we need a seperate venue (we have internal google docs for this somewhere) to talk about cross language interface consistency for json / raw / thrift / etc endpoint handlers.

PR #1159, opened 2015-08-14
Proposal: throttling on busy responses

ShanniLi:
r @jcorbin @Raynos @kriskowal 

This is a new PR per suggestions from #1099 

First step: added the proposal doc.
Will add a simulation program shortly.

### DIFF
--- a/docs/busy-throttling.md
+++ b/docs/busy-throttling.md
@@ -1,0 +1,26 @@
+# Throttling on Busy Responses
+
+In TChannel, a busy response is sent to indicate that a service is overloaded. Upon
+receiving busy responses, the client should reduce its requests to the service so
+that the service can recover to a healthy state.
+
+In order to do throttling, it is necessary to measure how fast the client is sending
+the requests. We use RPS (Request Per Second), i.e., the number of request sent every
+second. Given the client `clientA` that sends requests to the sevice `serviceB` with a
+RPS of `RpsToB`, there are three scenairos that we care about:
+
+1. All requests responeded without any busy responses. In this case, no request
+   should be throttled. `clientA` should continue to send all requests to `serviceB`. 
+2. If a busy response is received, `clientA` should throttle its sent requests to `serviceB`.
+   We use an exponential throttling rate, i.e., for every busy response received,
+   we set the RPS limit to `RpsToB / 2`. This will
+   actively cut down half of the request volume from `clientA` to `serviceB`.
+3. When requests to `serviceB` is under throttling, there should be a gradual increase
+   of the RPS limit to allow some requests coming through. Such a recover rate should
+   be linear. For example, the RPS limit can increase at a 5% rate per second.
+
+For throttling on busy responses, it is important that normal traffic is not affected. When
+there is a busy response, requests will get throttled in an exponential rate. On the other 
+hand, the recover rate makes sure that the throttling condition can be gradually lifted when
+the service's throughput is improved.  
+
--- a/docs/busy-throttling.md
+++ b/docs/busy-throttling.md
@@ -4,23 +4,182 @@ In TChannel, a busy response is sent to indicate that a service is overloaded. U
 receiving busy responses, the client should reduce its requests to the service so
 that the service can recover to a healthy state.
 
+## Goals
+There are three goals we want to achieve:
+
+* When a busy response is received, we want to exponentially reduce the outgoing requests
+  to the service.
+* The longer the busy responses are received, the longer it should take to recover the outgoing
+  requests. This is to reduce the flappy scenarios where the service is back and forth in good
+  and bad states.
+* When there is no busy reponses and the service is back to healthy. Traffic should not take too
+  long to recover.
+
+
+## Design
+
 In order to do throttling, it is necessary to measure how fast the client is sending
 the requests. We use RPS (Request Per Second), i.e., the number of request sent every
 second. Given the client `clientA` that sends requests to the sevice `serviceB` with a
-RPS of `RpsToB`, there are three scenairos that we care about:
+RPS of `RpsToB` and a RPS limit of `RpsLimitToB`, there are three scenairos that we
+care about:
 
 1. All requests responeded without any busy responses. In this case, no request
    should be throttled. `clientA` should continue to send all requests to `serviceB`. 
 2. If a busy response is received, `clientA` should throttle its sent requests to `serviceB`.
    We use an exponential throttling rate, i.e., for every busy response received,
-   we set the RPS limit to `RpsToB / 2`. This will
+   we set the RPS limit `RpsLimitToB` to `RpsToB / 2`. This will
    actively cut down half of the request volume from `clientA` to `serviceB`.
 3. When requests to `serviceB` is under throttling, there should be a gradual increase
-   of the RPS limit to allow some requests coming through. Such a recover rate should
-   be linear. For example, the RPS limit can increase at a 5% rate per second.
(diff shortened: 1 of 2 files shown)


docs/busy-throttling.md
```diff
@@ -0,0 +1,26 @@
...
+that the service can recover to a healthy state.
+
+In order to do throttling, it is necessary to measure how fast the client is sending
+the requests. We use RPS (Request Per Second), i.e., the number of request sent every
+second. Given the client `clientA` that sends requests to the sevice `serviceB` with a
+RPS of `RpsToB`, there are three scenairos that we care about:
+
+1. All requests responeded without any busy responses. In this case, no request
+   should be throttled. `clientA` should continue to send all requests to `serviceB`. 
+2. If a busy response is received, `clientA` should throttle its sent requests to `serviceB`.
```
anson627: this is similar in many ways to TCP congestion control, busy response is like packet drop due to congestion, throttling rate is like packets window size:
https://en.wikipedia.org/wiki/TCP_congestion-avoidance_algorithm

CUBIC is used by default in the latest Linux kernel, in which the window is a cubic function of time since the last congestion event:
https://en.wikipedia.org/wiki/CUBIC_TCP

my feeling is cutting by half upon congestion might be too aggressive, and increasing rate by 5% might be not enough for the traffic to ramp up quickly again.

docs/busy-throttling.md
```diff
@@ -0,0 +1,26 @@
...
+   We use an exponential throttling rate, i.e., for every busy response received,
+   we set the RPS limit to `RpsToB / 2`. This will
+   actively cut down half of the request volume from `clientA` to `serviceB`.
+3. When requests to `serviceB` is under throttling, there should be a gradual increase
+   of the RPS limit to allow some requests coming through. Such a recover rate should
+   be linear. For example, the RPS limit can increase at a 5% rate per second.
```
kriskowal: This should be an additive increase (add a constant, instead of multiplying by 1.05). We will have to tune recovery times for common scenarios. Or just see what happens with MIMD with the simulations :)

kriskowal:
Great. From the simulation, I would like to see how this behaves for an endpoint with an unknown RPS limit, beyond which it sends busy frames. Maybe even go on and see how it adapts to changes in a dynamic RPS limit. I think that would be enough validation to proceed.

kriskowal:
Perhaps also consider the case of a caller that maintains a steady RPS for a long while without trouble and then the RPS that the service can maintain drops. How long does the algorithm take to adapt? How long does the algorithm take to recover?

docs/throttling.py
```diff
@@ -0,0 +1,183 @@
...
+    return current(start, time, style)
+
+def get_rps_limit(rps_limit, busy):
+    limit = 0
+    busy = 20 if busy > 20 else busy
+    if recover_value > 0:
+        limit = (rps_limit + recover_value) / math.pow(2, busy);
+    elif recover_rate > 0:
+        incr = rps_limit * recover_rate
+        limit = (rps_limit + incr) / math.pow(2, busy);
```
kriskowal: `math.pow(2, busy)` is the same as `2 ** busy` in Python.

docs/throttling.py
```diff
@@ -0,0 +1,183 @@
...
+    elif style == 'linear_grow':
+        return start + 100 * time
+    elif style == 'linear_decrease':
+        return start - 100 * time
+    elif style == 'exponential_grow':
+        return start * math.pow(2, x)
+    elif style == 'exponential_decrease':
+        return start / math.pow(2, x)
+    else:
+        raise Exception('wrong style!')
```
kriskowal: Interesting that linear is time dependent and exponential is not.

docs/throttling.py
```diff
@@ -0,0 +1,183 @@
...
+    busy = 20 if busy > 20 else busy
+    if recover_value > 0:
+        limit = (rps_limit + recover_value) / math.pow(2, busy);
+    elif recover_rate > 0:
+        incr = rps_limit * recover_rate
+        limit = (rps_limit + incr) / math.pow(2, busy);
+    else:
+        raise Exception('wrong recover_value || recover_rate!')
+
+    limit = limit if limit <= default_rps_limit else default_rps_limit
```
kriskowal: Can use min and max functions for this

docs/busy-throttling.md
```diff
@@ -0,0 +1,185 @@
...
+
+1. All requests responeded without any busy responses. In this case, no request
+   should be throttled. `clientA` should continue to send all requests to `serviceB`. 
+2. If a busy response is received, `clientA` should throttle its sent requests to `serviceB`.
+   We use an exponential throttling rate, i.e., for every busy response received,
+   we set the RPS limit `RpsLimitToB` to `RpsToB / 2`. This will
+   actively cut down half of the request volume from `clientA` to `serviceB`.
+3. When requests to `serviceB` is under throttling, there should be a gradual increase
+   of the RPS limit to allow some requests coming through. Based on the simulation, we
+   set the recover rate as 3, i.e., for every second, we set `RpsLimitToB` = `RpsLimitToB * 3` .
```
kriskowal: I think you mean `+ 3` here.

docs/throttling.py
ShanniLi: good point!

docs/busy-throttling.md
ShanniLi: No, I meant * 3. The simulation result is actually quite good.
+ 3 is not good enough for recovery. On other hand, +200 makes the throttling flappy.

docs/throttling.py
ShanniLi: nice catch. It was a typo. I actually didn't run this pattern. :)

docs/busy-throttling.md
```diff
@@ -0,0 +1,185 @@
...
+there is a busy response, requests will get throttled in an exponential rate. On the other 
+hand, the recover rate makes sure that the throttling condition can be gradually lifted when
+the service's throughput is improved.  
+
+## Simulation
+We use a simulation program to validate the design. The program simulates time as a incremental counter of
+seconds. For each second, there will be a customizable amout of out requests and busy responses. It is 
+important to make sure that the number of busy responses never exceeds the number of actual out requests.
+
+First, We let the simulation run for 10 seconds (i.e., 10 increment of time counter) with various traffic
```
jcorbin: That seems an awfully limited test scenario, I'd like to at least see adding a "normal" starting phase before the troubled time of busies.

Even better would be to simulate with a slowly varying (e.g. over the course of an hour(s) or even up to a day) and rate-of-request triggered busies in the endpoint.

Raynos:
This is definitely something we want.

If we have consensus on this proposal for throttling then lets merge it!

ShanniLi:
@Raynos, we have uneven distribution of worker load right now. Will this feature interfere with retries on busy frames?

Raynos:
We may need to rethink this feature in terms of even-ness of load and being specific to peers instead of service name.

ShanniLi:
Another way may just be checking if the busy frame is due to node busy or service busy. We can do retry for first and throttling for second.

Raynos:
There is no way to check that. There are three types of busy ATM.

Rate limit : serviceName busy
Rate limit : hyperbahn worker busy
Too busy : this tchannel instance is busy because event loop / ram etc.

ShanniLi:
Do we want to add a subtype for busy frames? Having one type isn't really helpful.
On the other hand, I guess this PR (#1348) may help the issue of uneven distributed worker load.

PR #1160, opened 2015-08-14
Update http error mapping doc

anson627:
r: @ShanniLi @jcorbin @Raynos 
cc: @gshu @Matt-Esch

### DIFF
--- a/docs/error-mapping-tchannel-http.md
+++ b/docs/error-mapping-tchannel-http.md
@@ -24,11 +24,11 @@ The proposed mapping is:
 | code   | name                 | http status code/message
 | ------ | -------------------- | -------------------------
 | `0x01` | timeout              | 504 Gateway Timeout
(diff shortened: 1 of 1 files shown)


jcorbin:
lgtm, however I would suggest varying the status message meaningfully fwiw.  I.e. "500 TChannel Network Error" "500 TChannel Protocol Error" "500 Cancelled" etc...

ShanniLi:
lgtm. :+1: @jcorbin, specific messages will be more clear.

anson627:
agree. The concern is assigning different message to same status code seems to be inconsistent with the standard.

jcorbin:
Not so @anson627 , http status messages have always had only "suggested defaults" only; returning semantically meaningful status messages has always been A Thing in http (an/the entire point of separating out standard codes, is so that you can change the message ;-) ).

anson627:
done with specific messages.

PR #1162, opened 2015-08-14
Stop double building responses in connection error path

jcorbin:
So previously the "double build" check path merely emitted an error, and then proceeded to potentially build a double response anyhow (if the error event didn't turn out to be fatal).

r @kriskowal @Raynos

### DIFF
--- a/node/connection_base.js
+++ b/node/connection_base.js
@@ -146,7 +146,7 @@ function buildResponse(req, options) {
     var self = this;
 
     if (req.res && req.res.state !== States.Initial) {
-        self.errorEvent.emit(self, errors.ResponseAlreadyStarted({
+        req.errorEvent.emit(req, errors.ResponseAlreadyStarted({
             state: req.res.state,
             reason: 'buildResponse called twice',
             codeString: req.res.codeString,
--- a/node/connection_base.js
+++ b/node/connection_base.js
@@ -152,7 +152,15 @@ function buildResponse(req, options) {
             codeString: req.res.codeString,
             responseMessage: req.res.message
         }));
+        return req.res;
     }
+
+    return self._buildResponse(req, options);
+};
+
+TChannelConnectionBase.prototype._buildResponse =
+function _buildResponse(req, options) {
+    var self = this;
 
     options.channel = self.channel;
     options.inreq = req;
--- a/node/test/index.js
+++ b/node/test/index.js
@@ -37,6 +37,7 @@ require('./max_pending.js');
 require('./tchannel.js');
 require('./regression-inOps-leak.js');
 require('./regression-listening-on-used-port.js');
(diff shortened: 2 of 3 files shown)


kriskowal:
Test please

jcorbin:
Great call @kriskowal, as it turns out this innocuous fix eliminated a potential for the chronic orphaned popOutReq info log.  Here's the one that the regression test provoked before this fix:

```
    TCHANNEL INFO: popOutReq received for unknown or lost id ~ { context:
       { error:
          { [TchannelUnexpectedError: nope]
            type: 'tchannel.unexpected',
            message: 'nope',
            isErrorFrame: true,
            codeName: 'UnexpectedError',
            errorCode: 5,
            originalId: 2,
            name: 'TchannelUnexpectedError',
            fullType: 'tchannel.unexpected' },
         id: 2,
         info: 'got error frame for unknown id' },
      socketRemoteAddr: 'localhost:56844',
      direction: 'out' }
```

Plus I eliminated the closure allocation in the early return branch for cleanliness.

node/connection_base.js
```diff
@@ -146,14 +146,22 @@ function buildResponse(req, options) {
     var self = this;
 
     if (req.res && req.res.state !== States.Initial) {
-        self.errorEvent.emit(self, errors.ResponseAlreadyStarted({
+        req.errorEvent.emit(req, errors.ResponseAlreadyStarted({
```
jcorbin: Plus, having to test this pointed out that this is the better errorEvent rather than on connection itself.

jcorbin:
I'm currently working on computing T1:
- currently we have: `FIX T2 EVENT_CHANGE T3`
- I want `T1 FIX T2 EVENT_CHANGE T3` to even better understand original master behavior before this PR

jcorbin:
Okay, I managed to get T1, here's the break down:
- master has two nearly identical error logs plus a popOutReq info log
- after return early, we now see 3 error logs, since the subsequent `res.send*`s happen on the same, correctly stateful, res
- using a `req.errorEvent` rather than `conn.errorEvent` reduces the two identical logs to one clearer log, since this test case uses the "build first, then already done; build second, try to send error" pattern

However this last change does introduce an, arguably desirable, change of behavior: if the handler instead "bulid first.... fumble around, don't send anything, build second, now send"... this used to be okay, and papered over; however now the "build second" call will send an UnexpectedError response to the incoming request, not allowing the broken handler to respond.

Now that I realize this implication, I'll add another case to the regression test to clarify.

jcorbin:
Nevermind, the semantic change isn't A Thing (tm)... since the "double check" logic is only "is it out of initial state?".  So the change would've only been A Thing for streaming handlers, which isn't a huge stability concern yet, so not separating.

kriskowal:
Thanks, this looks solid.

PR #1164, opened 2015-08-15
Add error logging in as-http example

anson627:
Update as-http example with error logging
Useful for reproducing and debugging as-http/tchannel error
@ShanniLi @Raynos @davewhat

### DIFF
--- a/node/examples/http_egress.js
+++ b/node/examples/http_egress.js
@@ -87,7 +87,7 @@ function onHTTPListening() {
 }
 
 function onHTTPRequest(hreq, hres) {
-    asHTTP.forwardToTChannel(svcChan, hreq, hres, {}, function(error) {
+    asHTTP.forwardToTChannel(svcChan, hreq, hres, {}, function onComplete(error) {
         if (error) {
             console.log('forward to tchannel failed', error);
         }
--- a/node/examples/http_egress.js
+++ b/node/examples/http_egress.js
@@ -87,5 +87,9 @@ function onHTTPListening() {
 }
 
 function onHTTPRequest(hreq, hres) {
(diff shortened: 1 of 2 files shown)


node/examples/http_egress.js
```diff
@@ -87,5 +87,9 @@ function onHTTPListening() {
 }
 
 function onHTTPRequest(hreq, hres) {
-    asHTTP.forwardToTChannel(svcChan, hreq, hres, {}, function() { return true; });
+    asHTTP.forwardToTChannel(svcChan, hreq, hres, {}, function(error) {
```
ShanniLi: it would be nice to also add name to the function

node/examples/http_ingress.js
```diff
@@ -86,5 +86,9 @@ function onRequest(inreq, outres) {
     asHTTP.forwardToHTTP(svcChan, {
         host: destHost,
         port: destPort,
-    }, inreq, outres, function() { return true; });
+    }, inreq, outres, function(error) {
```
ShanniLi: it would be nice to also add name to the function

ShanniLi:
lgtm with unblocking comments

node/examples/http_egress.js
```diff
@@ -87,5 +87,9 @@ function onHTTPListening() {
...
+        if (error) {
+            console.log('forward to tchannel failed', error);
```
Raynos: Never console.log. Always self.logger

anson627: This is example code. Do we want logtron here?

Raynos: Oh sorry; let's use `console.error` then.

Issue #1166, opened 2015-08-17
wanted: ability to listen on existing socket/fd


benfleis:
In order to implement zero-downtime process rollover, I would like to have to the ability to listen on an existing fd, a la `server.listen({ fd: 0 })` API [1]. I have a WIP diff for node/js, which will shortly be added.

[1] https://nodejs.org/api/net.html#net_server_listen_handle_callback

jcorbin:
Can you please lay out details on one or more use cases?  I think I understand how you intend to use this, but I want to be sure.

Raynos:
The main use case is passing a listening fd from one server to another with a low amount of downtime.

You want to be able to start a new instance of a worker and only close the old listening socket once the new instance has listened. This is useful if your process takes seconds to start.

jcorbin:
I get that @raynos, but I still would like more clarity on an example of how this gets used in practice.

benfleis:
@jcorbin 

This is how I intend to actually use it. (crank might be replaced with a supervisord plugin module, but that has other risks.)

- supervisord spawns [crank](https://github.com/pusher/crank) process
- crank maintains an open, listenable socket/fd on configured port
- crank spawns ringpop-app

On upgrade:
- untar new package
- signal to crank, which in turn
- spawns new ringpop-app with handoff FD, after which
- old ringpop-app is killed

Make sense?

Crank could be socketmaster (which is time based), or for nodejs, I already have a cluster-based approach to doing the same.

PR #1167, opened 2015-08-17
add support to listen on socket/fd, via opts argument to .listen.

benfleis:
addresses https://github.com/uber/tchannel/issues/1166

Forgot to note that the tests are being worked on.

### DIFF
--- a/node/channel.js
+++ b/node/channel.js
@@ -152,6 +152,7 @@ function TChannel(options) {
     // Filled in by the listen call:
     self.host = null;
     self.requestedPort = null;
+    self.listenFd = null;
 
     // Filled in by listening event:
     self.hostPort = null;
@@ -453,20 +454,37 @@ TChannel.prototype.makeSubChannel = function makeSubChannel(options) {
     return chan;
 };
 
-TChannel.prototype.listen = function listen(port, host, callback) {
+TChannel.prototype.listen = function listen(port, host, opts, callback) {
     // Note:
     // - 0 is a valid port number, indicating that the system must assign an
     //   available ephemeral port
     // - 127.0.0.1 is a valid host, primarily for testing
+    // - opts supports:
+    //   {fd: X} - listen will attach to the existing socket, and host/port
+    //   remain required, but become only descriptive. See also:
+    //   https://nodejs.org/api/net.html#net_server_listen_handle_callback
     var self = this;
+    if (opts && typeof opts === 'function') {
+        callback = opts;
+        opts = {};
+    }
+    opts = opts || {};
     assert(!self.listened, 'TChannel can only listen once');
     assert(typeof host === 'string', 'TChannel requires host argument');
     assert(typeof port === 'number', 'TChannel must listen with numeric port');
+    assert(!('fd' in opts) || typeof opts.fd === 'number', 'TChannel listen opts.fd must be numeric');
     assert(host !== '0.0.0.0', 'TChannel must listen with externally visible host');
     self.listened = true;
     self.requestedPort = port;
     self.host = host;
-    self.getServer().listen(port, host, callback);
+    self.listenFd = opts.fd;
+
+    if (self.listenFd >= 0) {
+        assert(port !== 0, 'TChannel.listen given a FD must have a defined port');
+        self.getServer().listen({ fd: self.listenFd }, callback);
+    } else {
+        self.getServer().listen(port, host, callback);
+    }
 };
 
 TChannel.prototype.register = function register(name, options, handler) {
@@ -493,7 +511,11 @@ TChannel.prototype.register = function register(name, options, handler) {
 TChannel.prototype.address = function address() {
     var self = this;
     if (self.serverSocket) {
-        return self.serverSocket.address() || null;
+        if (self.listenFd !== null && self.listenFd !== undefined) {
+            return { port: self.port, family: 'IPv4', address: self.address };
+        } else {
+            return self.serverSocket.address() || null;
+        }
     } else if (self.topChannel) {
         return self.topChannel.address();
     } else {
--- a/node/channel.js
+++ b/node/channel.js
@@ -454,30 +454,30 @@ TChannel.prototype.makeSubChannel = function makeSubChannel(options) {
     return chan;
 };
 
-TChannel.prototype.listen = function listen(port, host, opts, callback) {
+TChannel.prototype.listen = function listen(port, host, options, callback) {
     // Note:
     // - 0 is a valid port number, indicating that the system must assign an
     //   available ephemeral port
     // - 127.0.0.1 is a valid host, primarily for testing
-    // - opts supports:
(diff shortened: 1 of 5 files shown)


node/channel.js
```diff
@@ -453,20 +454,37 @@ TChannel.prototype.makeSubChannel = function makeSubChannel(options) {
...
 
-TChannel.prototype.listen = function listen(port, host, callback) {
+TChannel.prototype.listen = function listen(port, host, opts, callback) {
     // Note:
     // - 0 is a valid port number, indicating that the system must assign an
     //   available ephemeral port
     // - 127.0.0.1 is a valid host, primarily for testing
+    // - opts supports:
+    //   {fd: X} - listen will attach to the existing socket, and host/port
+    //   remain required, but become only descriptive. See also:
```
kriskowal: This does seem reasonable. Would it be more reasonable to implement `relisten()` and carry the fd and obtained address from a prior listen instead?

node/channel.js
```diff
@@ -453,20 +454,37 @@ TChannel.prototype.makeSubChannel = function makeSubChannel(options) {
     return chan;
 };
 
-TChannel.prototype.listen = function listen(port, host, callback) {
+TChannel.prototype.listen = function listen(port, host, opts, callback) {
```
kriskowal: s/opts/options/ (we have been moving toward using options consistently, please bear with the inconsistencies that remain)

node/channel.js
```diff
@@ -453,20 +454,37 @@ TChannel.prototype.makeSubChannel = function makeSubChannel(options) {
...
     var self = this;
+    if (opts && typeof opts === 'function') {
+        callback = opts;
+        opts = {};
+    }
+    opts = opts || {};
     assert(!self.listened, 'TChannel can only listen once');
     assert(typeof host === 'string', 'TChannel requires host argument');
     assert(typeof port === 'number', 'TChannel must listen with numeric port');
+    assert(!('fd' in opts) || typeof opts.fd === 'number', 'TChannel listen opts.fd must be numeric');
```
kriskowal: Please say `typeof opts.fd === 'number'`. We avoid `in` since it deoptimizes the surrounding function.

node/channel.js
Raynos: The use case here passing the `fd` from the command line and sharing it across two processes for zero-downtime restarts.

node/channel.js
```diff
@@ -493,7 +511,11 @@ TChannel.prototype.register = function register(name, options, handler) {
 TChannel.prototype.address = function address() {
     var self = this;
     if (self.serverSocket) {
-        return self.serverSocket.address() || null;
+        if (self.listenFd !== null && self.listenFd !== undefined) {
+            return { port: self.port, family: 'IPv4', address: self.address };
```
kriskowal: self.address is a method. Might mean `host`. family is not necessarily IPv4. It might be best to carry the address return value over from a prior listen.

node/channel.js
Raynos: `fd` is optional so we should check for `opts.fd === undefined` instead.

Would appreciate benchmark demonstrating the `in` truly does de-optimize.

node/channel.js
Raynos: Probably should be `address: self.host`

node/channel.js
```diff
@@ -493,7 +511,11 @@ TChannel.prototype.register = function register(name, options, handler) {
 TChannel.prototype.address = function address() {
     var self = this;
     if (self.serverSocket) {
-        return self.serverSocket.address() || null;
+        if (self.listenFd !== null && self.listenFd !== undefined) {
```
Raynos: `if (typeof self.listenFd === 'number')` would be simpler.

node/channel.js
kriskowal: Acknowledged. Disregard.

node/channel.js
kriskowal: Strike that. I was thinking of for/in.

kriskowal:
review completed, please address bug in address(). All other feedback advisory.

node/channel.js
benfleis: ack

node/channel.js
benfleis: eek! ack.

node/channel.js
benfleis: ack

node/channel.js
benfleis: converted `in` to `opts.fd === undefined`.

Raynos:
lgtm; feel free to merge.

node/channel.js
```diff
@@ -453,20 +454,37 @@ TChannel.prototype.makeSubChannel = function makeSubChannel(options) {
...
     var self = this;
+    if (options && typeof options === 'function') {
+        callback = options;
+        options = {};
+    }
+    options = options || {};
     assert(!self.listened, 'TChannel can only listen once');
     assert(typeof host === 'string', 'TChannel requires host argument');
     assert(typeof port === 'number', 'TChannel must listen with numeric port');
+    assert(options.fd === undefined || typeof options.fd === 'number', 'TChannel listen options.fd must be numeric');
```
jwolski: check for NaN too?

benfleis: As a generic question, how far do we go here? Now that I look another time, I would say that it should actually check >= 0, and that it's an integer. @Raynos @jwolski  thoughts?  (a la `fd | 0 === fd, fd >= 0`?)

jwolski: good question. i think the level of validation is partially subjective. It looks like the `port` check above doesn't account for NaN. and you may ask, "why should i?". that'd be a good question to ask too ;). in my mind:

1) make sure listen() doesn't blow up with the input that is provided to it.
2) make sure there's reasonable behavior for the developer developing against the TChannel API
3) don't go farther than you need to go to protect said developer from shooting themselves in the foot.

jcorbin: `typeof X === 'number' && X >= 0` should suffice here... and if someone passes a non-integer, well they'll get an exception anyhow when we try to pass it to listen:
- typeof undefined !== 'number'
- ditto for typeof null
- NaN annihilates all positive binary relations on numbers (i.e. X R NaN === false for R in <=, <, >, >=, ...)

In fact, simply the typeof should be fine, since NaN should blow up the underlying api anyhow, which is as good an assertion imo.

benfleis: switching to a hybrid solution, since undefined is legit, as is a positive number:

  `options.fd === undefined || (typeof options.fd === 'number' && options.fd >= 0)`

seems the most sane. And agreed on `NaN` annilhilation, which is why I like the >= test as well.

jwolski:
lgtm

Raynos:
lgtm.

jcorbin:
This broke tests on master, reverted; unable to require the child process module in travis at least, despite running locally fro me.

PR #1169, opened 2015-08-17
Improve call req validation error handling paths

jcorbin:
r @kriskowal @ShanniLi 

Especial feedback to the part where I emit a req error event now, rather than assert.

### DIFF
--- a/node/test/max-call-overhead.js
+++ b/node/test/max-call-overhead.js
@@ -81,16 +81,23 @@ allocCluster.test('request() with large arg1', {
         serviceName: 'server'
     });
 
+    var arg1 = '';
+    for (var i = 0; i < 16 * 1024 + 1; i++) {
+        arg1 += 'a';
+    }
+
     subTwo.waitForIdentified({
         host: one.hostPort
     }, function onIdentified(err) {
         assert.ifError(err);
 
-        var arg1 = '';
-        for (var i = 0; i < 16 * 1024 + 1; i++) {
-            arg1 += 'a';
-        }
+        // TODO: O(pinions) -- is this really what we want?
+        assert.throws(doRequest, /arg1 length \d+ is larger than the limit \d+/);
 
+        assert.end();
+    });
+
+    function doRequest() {
         subTwo.request({
             serviceName: 'server',
             hasNoParent: true,
@@ -99,18 +106,20 @@ allocCluster.test('request() with large arg1', {
                 'cn': 'wat'
             }
         }).send(arg1, 'a', 'b', onResponse);
-    });
+    }
 
     function onResponse(err, resp, arg2, arg3) {
-        assert.ok(err);
-        assert.equal(err.type, 'tchannel.arg1-over-length-limit');
-        assert.equal(err.message,
-            'arg1 length 16385 is larger than the limit 16384'
-        );
+        assert.fail("shouldn't get a response callback");
 
-        assert.equal(null, resp);
+        // assert.ok(err);
+        // assert.equal(err.type, 'tchannel.arg1-over-length-limit');
+        // assert.equal(err.message,
+        //     'arg1 length 16385 is larger than the limit 16384'
+        // );
 
-        assert.end();
+        // assert.equal(null, resp);
+
+        // assert.end();
     }
 });
 
--- a/node/test/max-call-overhead.js
+++ b/node/test/max-call-overhead.js
@@ -91,11 +91,8 @@ allocCluster.test('request() with large arg1', {
     }, function onIdentified(err) {
         assert.ifError(err);
 
-        // TODO: O(pinions) -- is this really what we want?
-        assert.throws(doRequest, /arg1 length \d+ is larger than the limit \d+/);
-
-        assert.end();
-    });
+        doRequest();
+});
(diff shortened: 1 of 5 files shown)


node/v2/handler.js
```diff
@@ -630,36 +630,33 @@ TChannelV2Handler.prototype.verifyCallRequestFrame =
...
-    } else if (!req.headers || !req.headers.as) {
-        self.logger.error('Expected "as" header to be set for request', {
-            arg1: req.endpoint,
-            callerName: req.headers && req.headers.cn,
-            remoteName: self.remoteName,
-            serviceName: req.serviceName,
-            socketRemoteAddr: self.connection.socketRemoteAddr
-        });
+    if (!req.headers || !req.headers.as) {
+        if (self.requireAs) {
```
Raynos: The whole point of `requireAs` was to make it a programmer error.

Making an outgoing call without a `cn` header should statically throw rather then give you an error in a callback.

jcorbin:
Understood @raynos, I walked this PR back to throwing errors, and we'll have a "should it through" discussion in another PR.  In this PR now all I do is normalize arg1 vs header validation error handling.

node/test/max-call-overhead.js
```diff
@@ -99,18 +106,20 @@ allocCluster.test('request() with large arg1', {
...
     function onResponse(err, resp, arg2, arg3) {
-        assert.ok(err);
-        assert.equal(err.type, 'tchannel.arg1-over-length-limit');
-        assert.equal(err.message,
-            'arg1 length 16385 is larger than the limit 16384'
-        );
+        assert.fail("shouldn't get a response callback");
 
-        assert.equal(null, resp);
+        // assert.ok(err);
```
ShanniLi: remove the commented code?

jcorbin: Maybe, it's here in case we to the TODO above about making this a non-throwing path again.

ShanniLi:
:+1: Like the idea to have more immediate/clear errors.

node/test/max-call-overhead.js
```diff
@@ -81,16 +81,23 @@ allocCluster.test('request() with large arg1', {
...
     subTwo.waitForIdentified({
         host: one.hostPort
     }, function onIdentified(err) {
         assert.ifError(err);
 
-        var arg1 = '';
-        for (var i = 0; i < 16 * 1024 + 1; i++) {
-            arg1 += 'a';
-        }
+        // TODO: O(pinions) -- is this really what we want?
```
kriskowal: O(bservation): There will be cases of computed arg1, possibly including some HTTP translation layer, in which case it should be an operational error, not a hard stop and die.

node/test/max-call-overhead.js
```diff
@@ -99,18 +106,20 @@ allocCluster.test('request() with large arg1', {
...
-    });
+    }
 
     function onResponse(err, resp, arg2, arg3) {
-        assert.ok(err);
-        assert.equal(err.type, 'tchannel.arg1-over-length-limit');
-        assert.equal(err.message,
-            'arg1 length 16385 is larger than the limit 16384'
-        );
+        assert.fail("shouldn't get a response callback");
```
kriskowal: lintme

node/test/max-call-overhead.js
```diff
@@ -99,18 +106,20 @@ allocCluster.test('request() with large arg1', {
...
+        // assert.ok(err);
+        // assert.equal(err.type, 'tchannel.arg1-over-length-limit');
+        // assert.equal(err.message,
+        //     'arg1 length 16385 is larger than the limit 16384'
+        // );
 
-        assert.end();
+        // assert.equal(null, resp);
+
+        // assert.end();
```
kriskowal: I recommend restoring this behavior.

jcorbin:
Although, @Raynos will likely shed a tear, not throwing an error does seem less fragile going forward, as it eliminates a potentially fatal crash for Hyperbahn and maybe even xlate that would be out of its control.

In the case where there has been no callback / error event listener given, it will revert to a thrown error.

node/v2/handler.js
kriskowal: Did we not implement a version of this that could be safely employed at Hyperbahn ingress?

node/v2/handler.js
```diff
@@ -630,39 +633,34 @@ TChannelV2Handler.prototype.verifyCallRequestFrame =
...
+                arg1: req.endpoint,
+                remoteName: self.remoteName,
+                serviceName: req.serviceName,
+                socketRemoteAddr: self.connection.socketRemoteAddr
+            });
+        }
     }
 
-    return true;
+    return null;
```
kriskowal: Thank you.

kriskowal:
:ok:

Raynos:
@jcorbin 

You should throw an error. it's impossible to hit this in the relay. We enforce cn & as to exist on incoming request so they should always exist on out request.

Please stick to making programming errors be programming errors. Premature FUD of avoiding an uncaught exception is a waste of time. When we get into a bad state we should shout loudly.

Remember that hyperbahn never crashes on uncaught exceptions.

PR #1171, opened 2015-08-18
Add lbpool option in as-http

anson627:
without lbpool, I got 1000+ RPS, along with 10%+ timeout and EADDRNOTAVAIL errors
with lbpool (keep_alive = true), I got 1700+ RPS, no errors
so lbpool is awesome, except it doesn't expose readable stream
@mranney @jcorbin @Raynos @ShanniLi

### DIFF
--- a/node/as/http.js
+++ b/node/as/http.js
@@ -163,13 +163,14 @@ TChannelHTTP.prototype.sendResponse = function send(buildResponse, hres, body, c
     }
     var arg2 = arg2res.value;
     if (body) {
-        callback(null);
-        return buildResponse({
+        buildResponse({
             streamed: false,
             headers: {
                 as: 'http'
             }
         }).sendOk(arg2, body);
+        callback(null);
+        return;
     }
 
     return buildResponse({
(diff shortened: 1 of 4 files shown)


anson627:
Also tried ForeverAgent, it's not as nice as lbpool.

node/as/http.js
```diff
@@ -154,10 +158,19 @@ TChannelHTTP.prototype.sendResponse = function send(buildResponse, hres, callbac
         var toBufferErr = errors.HTTPResArg2toBufferError(arg2res.err, {
             head: head
         });
-        callback(toBufferErr, null, null);
+        callback(toBufferErr);
         return null;
     }
     var arg2 = arg2res.value;
+    if (body) {
+        callback(null);
```
ShanniLi: move callback to after the send

node/as/http.js
```diff
@@ -265,7 +303,7 @@ TChannelHTTP.prototype.forwardToHTTP = function forwardToHTTP(tchannel, options,
     function onResponse(inres) {
         if (!sent) {
             sent = true;
-            outres.sendResponse(inres);
+            outres.sendResponse(inres, null);
```
ShanniLi: null is not necesssary

ShanniLi:
Do you have data for large payload, e.g., ~1MB?

anson627:
not yet, I'm working on the get benchmark support for as-http. I guess the perf sucks since we are not streaming the response.

node/as/http.js
anson627: I thought we prefer to be explicit and avoiding the undefined

ShanniLi:
lgtm

PR #1172, opened 2015-08-18
Fix out reqres opts

jcorbin:
Carved out of #1163 to fix immediate pain towards #1135 rather than Doing All The Things first.

r @kriskowal


kriskowal:
:ok:

PR #1174, opened 2015-08-19
(v1.x) add support to listen on socket/fd, via options argument to

benfleis:
.listen. conceptually identical to same patch for master (v2.x).

addresses https://github.com/uber/tchannel/issues/1166

### DIFF
--- a/index.js
+++ b/index.js
@@ -97,8 +97,19 @@ require('util').inherits(TChannel, require('events').EventEmitter);
 
 // Decoulping config and creation from the constructor.
 // This also allows us to better unit test the code as the test process
-// is not blocked by the listening connections
-TChannel.prototype.listen = function () {
+// is not blocked by the listening connections.
+// - options supports:
+//   {fd: X} - listen will attach to the existing socket, and host/port
+//   remain required, but become only descriptive. See also:
+//   https://nodejs.org/api/net.html#net_server_listen_handle_callback
+TChannel.prototype.listen = function (options) {
+	options = options || {};
+	if (options && typeof options !== 'object') {
+		throw new Error('listen options must be an object');
+	}
+	if (options.fd !== undefined && !(options.fd >= 0)) {
+		throw new Error('listen options.fd must be positive integer');
+	}
 	if (!this.serverSocket) {
 		throw new Error('Missing server Socket.');
 	}
@@ -109,7 +120,12 @@ TChannel.prototype.listen = function () {
 		throw new Error('Missing server port.');
 	}
 
-	this.serverSocket.listen(this.port, this.host);
+	if (options.fd >= 0) {
+		this.listenFd = options.fd;
+		this.serverSocket.listen({ fd: options.fd });
+	} else {
+		this.serverSocket.listen(this.port, this.host);
+	}
 };
 
 
--- a/test/listen-on-fd-child.js
+++ b/test/listen-on-fd-child.js
@@ -18,13 +18,20 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 // THE SOFTWARE.
 
-'use strict';
+var TChannel = require('../index.js');
 
-require('./safe-quit.js');
-require('./timeouts.js');
-require('./send.js');
-require('./register.js');
-require('./identify.js');
-require('./tchannel.js');
-require('./listen-on-fd.js');
-require('./regression-inOps-leak.js');
+// just in case, always die.
+setTimeout(process.exit.bind(null, 0), 5000);
+process.on('disconnect', process.exit);
+
+process.on('message', doListen);
+
+function doListen(port, socket) {
+    var serverOptions = {host: '127.0.0.1', port: parseInt(port), listening: false};
+    var server = new TChannel(serverOptions);
+
+    server.register('endpoint', function(x, y, z, cb) { cb(null, 'ok'); });
+    server.on('socketClose', process.exit.bind(process));
+    server.serverSocket.on('listening', process.send.bind(process, port));
+    server.listen({ fd: socket.fd });
+}
--- a/test/listen-on-fd.js
+++ b/test/listen-on-fd.js
@@ -1,0 +1,56 @@
+// Copyright (c) 2015 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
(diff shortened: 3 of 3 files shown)


jwolski:
lgtm

test/tchannel.js
```diff
@@ -185,3 +187,22 @@ test('make socket: should throw for invalid destination', function t(assert) {
...
 //});
+
+
+test('make listen-on-fd socket: should work', function t(assert) {
+  var socket = new net.createServer();
+  socket.listen();
+
+  var server = new TChannel(serverOptions);
+  server.register('endpoint', function(_x, _y, _z, cb) { cb(null, 'ok'); });
+  server.listen(socket._handle.fd);
```
jwolski: this looks a bit weird. doesn't this need to be passed as: `{fd: socket._handle.fd}` ?

jwolski:
actually, one comment about the test you defined. looks wrong. maybe I'm missing something

test/tchannel.js
benfleis: your are correct. looking into it.

test/tchannel.js
```diff
@@ -185,3 +187,22 @@ test('make socket: should throw for invalid destination', function t(assert) {
...
+  socket.close();
+
+  var client = new TChannel(clientOptions);
+  client.send({host: serverName, timeout: 100}, 'endpoint', null, null, function(err, res1) {
```
Raynos: serverName ?

benfleis: defined @ top of file. it's factually incorrect, given that the port will change in my in progress corrected version, but the ref was valid.

jcorbin:
Please don't fork and open PRs from off repository, push feature branches to uber/tchannel and PR form there please; that way reviewer can easily check out the branch without extra work.

benfleis:
@jcorbin Good point, I was thinking about keeping the mess down, but had forgotten that  inconvenience. Will do in the future.

benfleis:
@jwolski @raynos @jcorbin

Updated with proper test, and tweaked option validation. Can it get a little look-see?

Raynos:
looks good.

test/tchannel.js
```diff
@@ -18,6 +18,8 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 // THE SOFTWARE.
 
+var net = require('net');
```
jwolski: is this necessary?

benfleis: nope, will clean that out too.

jwolski:
cool! looks good. left one comment about potentially unneeded requires clause

test/listen-on-fd-child.js
```diff
@@ -0,0 +1,41 @@
...
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+var TChannel = require('../index.js');
+
+var serverOptions = {host: '127.0.0.1', port: 4040, listening: false};
+var serverName = serverOptions.host + ':' + serverOptions.port;
```
jwolski: this isnt' used

test/listen-on-fd-child.js
```diff
@@ -0,0 +1,41 @@
...
+
+// just in case, always die.
+setTimeout(process.exit.bind(null, 0), 5000);
+process.on('disconnect', process.exit);
+
+process.on('message', doListen);
+
+function doListen(port, socket) {
+    serverOptions.port = parseInt(port);
+    var serverName = serverOptions.host + ':' + serverOptions.port;
```
jwolski: this isn't used

test/listen-on-fd-child.js
```diff
@@ -0,0 +1,41 @@
...
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+var TChannel = require('../index.js');
+
+var serverOptions = {host: '127.0.0.1', port: 4040, listening: false};
```
jwolski: is it necessary to define up here instead of in `doListen()`

jwolski:
left a few more cleanup-type comments

index.js
```diff
@@ -109,7 +120,12 @@ TChannel.prototype.listen = function () {
 		throw new Error('Missing server port.');
 	}
 
-	this.serverSocket.listen(this.port, this.host);
+	if (options.fd >= 0) {
+		this.listenFd = options.fd;
```
jwolski: `listenFd` isn't actually used anywhere. Is this more for bookkeeping/debuggability?

test/listen-on-fd-child.js
benfleis: fixed

test/listen-on-fd-child.js
benfleis: fixed

test/listen-on-fd-child.js
benfleis: could be there, this was copied from another test, and I left it above. since it's a single test, local is good.

index.js
benfleis: I followed the model of other debuggability stuff I found in the v2 code; it has no immediate use, and I'm happy to axe if it needed. I thought that it may well be useful in live debugging, since it's the only surviving indication that a channel was spawned via FD instead of normal listen.

Raynos: leaving it here is a good idea :)

jwolski: agreed. that it is good to leave. just wanted to make sure that was the intention.

jwolski:
looks swell to me. squash and pull the trigger when ready.

PR #1177, opened 2015-08-19
WIP: Inline PassThrough, and kill the runtime require

jcorbin:
cc @anson627 this should cut out the runtime requires from your flame graphs


Raynos:
Screw this. just `require('stream')`

Let's just drop readable-stream. Importing from node core is far simpler.

jcorbin:
Okay, so we'll choose either this PR or #1179, let's see.

jcorbin:
Closing but leaving the branch in place in case we need it.

jcorbin:
So amusingly this branch ended up being ~5% faster (less latency) than #1179, so I might resurrect it in spirit by just writing a direct PassThrough replacement in one class...

PR #1179, opened 2015-08-19
Drop pinned readable-stream dep

jcorbin:
Alternative to #1177


jcorbin:
Anyhow, since this passes tests, going to land for simplicity's sake, and then we can resurrect #1177 if it turns out not to fix our woes.

PR #1180, opened 2015-08-19
Hyperbahn client: read bootstrap file from disk

ShanniLi:
r @kriskowal @blampe @anson627 
cc @jcorbin 

Talked to @blampe, we still want to keep the hostPortList option to make the transition smooth.

### DIFF
--- a/node/hyperbahn/index.js
+++ b/node/hyperbahn/index.js
@@ -21,6 +21,7 @@
 'use strict';
 
 var assert = require('assert');
+var fs = require('fs');
 var TypedError = require('error/typed');
 var WrappedError = require('error/wrapped');
 var timers = require('timers');
@@ -91,10 +92,19 @@ function HyperbahnClient(options) {
         'Must pass in top level tchannel');
     assert(options.tchannel.tracer, 'Top channel must have trace enabled');
     assert(options.serviceName, 'must pass in serviceName');
-    assert(Array.isArray(options.hostPortList),
-        'Must pass in hostPortList');
+    if (Array.isArray(options.hostPortList)) {
+        self.hostPortList = options.hostPortList;
+    } else if (typeof options.hostPortFile === 'string') {
+        var hostContent = safeSyncRead(options.hostPortFile);
+        assert(!hostContent.error, 'Failed to read the hostPortFile');
+        var hostJson = safeJSONParse(hostContent.fileContents);
+        assert(!hostJson.error, 'Failed to parse the hostPortFile');
+        assert(Array.isArray(hostJson.json), 'The hostPortFile must contain a hostPortList');
+        self.hostPortList = hostJson.json;
+    } else {
+        assert(false, 'Must pass in hostPortList or hostPortFile');
+    }
 
-    self.hostPortList = options.hostPortList;
     self.serviceName = options.serviceName;
     self.callerName = options.callerName || options.serviceName;
     self.tchannel = options.tchannel;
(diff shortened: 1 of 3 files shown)


node/hyperbahn/index.js
```diff
@@ -425,3 +435,35 @@ HyperbahnClient.prototype.destroy = function destroy() {
...
+        error = err;
+    }
+
+    return {
+        fileContents: fileContents,
+        error: error
+    };
+}
+
+function safeJSONParse(text) {
```
jcorbin: we already have a safe-json-parse dependency for this function, use it

node/hyperbahn/index.js
```diff
@@ -425,3 +435,35 @@ HyperbahnClient.prototype.destroy = function destroy() {
...
+    var fileContents;
+    var error;
+
+    try {
+        fileContents = fs.readFileSync(filePath, 'utf8');
+    } catch (err) {
+        error = err;
+    }
+
+    return {
```
jcorbin: you may as well use a [err, result] tuple array here, for consistency with safe-json-parse as mentioned below

node/hyperbahn/index.js
```diff
@@ -425,3 +435,35 @@ HyperbahnClient.prototype.destroy = function destroy() {
     timers.clearTimeout(self._advertisementTimer);
     timers.clearTimeout(self.advertisementTimeoutTimer);
 };
+
+function safeSyncRead(filePath) {
+    var fileContents;
```
jcorbin: please initialize these to null rather than setting object fields to undefined implicitly

node/test/hyperbahn/advertise.js
```diff
@@ -66,4 +67,45 @@ function runTests(HyperbahnCluster) {
             assert.end();
         }
     });
+
+    HyperbahnCluster.test('can advertise using hostPortFile', {
+        size: 5
+    }, function t(cluster, assert) {
+        var bob = cluster.remotes.bob;
+        var hostPortFile = '/tmp/host-special.json';
```
jcorbin: yeah... that's not going to turn out well, use some sort of standard temp file name generator please to avoid clobbering between concurrent test runs and such

node/test/hyperbahn/advertise.js
```diff
@@ -66,4 +67,45 @@ function runTests(HyperbahnCluster) {
...
+
+            assert.equal(result.head, null);
+
+            // Because of duplicates in a size 5 cluster we know
+            // that we have at most 5 kValues
+            assert.ok(result.body.connectionCount <= 5,
+                'expect to have at most 5 advertise results');
+
+            client.destroy();
+            if (fs.existsSync(hostPortFile)) {
```
jcorbin: let's please put this cleanup in an `assert.once('end', function cleanup() { ... })` rigght after L76 above where we create it.

anson627:
:+1:

node/hyperbahn/index.js
```diff
@@ -91,10 +92,19 @@ function HyperbahnClient(options) {
         'Must pass in top level tchannel');
     assert(options.tchannel.tracer, 'Top channel must have trace enabled');
     assert(options.serviceName, 'must pass in serviceName');
-    assert(Array.isArray(options.hostPortList),
-        'Must pass in hostPortList');
+    if (Array.isArray(options.hostPortList)) {
+        self.hostPortList = options.hostPortList;
+    } else if (typeof options.hostPortFile === 'string') {
+        var hostContent = safeSyncRead(options.hostPortFile);
+        assert(!hostContent.error, 'Failed to read the hostPortFile');
```
Raynos: This approach is ok for now; however in the future we need to an async read.

We actually want to keep polling the file within the advertise timeout period.

jcorbin: Last discussion with @mranney ended with deciding that we don't in fact want to poll, watch, or in any way do anything fancy with the seed list file; let's discuss further offline if you have more to add to that conversation.

Raynos: It's more about polling until the file exists. If we corrupt the file or oops puppet or anything bad happens I want processes that restarted during that period to keep trying.

I.e. if you have the file; cool story. If you dont; keep trying until we fix the files.

PR #1181, opened 2015-08-19
Eliminate double send mutation

jcorbin:
Idea inspired / uncovered by #1145:
- we shouldn't mutate the out res in a double send scenario
- this also allows an easy(ier) way for the handler to implement its own error if appropriate, rather than having a hardcoded log spam for double sends in core tchannel
- a handler could even choose to implement a "rollback" hook for already sent responses... but that's crazy down the road idea...

If we like this idea, I can add similar checks to as/json and as/http, and get it past tests.

r @kriskowal @rf @ShanniLi @Raynos

### DIFF
--- a/node/as/json.js
+++ b/node/as/json.js
@@ -233,12 +233,13 @@ TChannelJSON.prototype.register = function register(
                     'Could not JSON stringify');
             }
 
-            res.headers.as = 'json';
-            res.setOk(respObject.ok);
-            res.send(
-                stringifyResult.value.head,
-                stringifyResult.value.body
-            );
+            if (res.setOk(respObject.ok)) {
+                res.headers.as = 'json';
+                res.send(
+                    stringifyResult.value.head,
+                    stringifyResult.value.body
+                );
+            }
         }
     }
 };
--- a/node/as/thrift.js
+++ b/node/as/thrift.js
@@ -161,12 +161,13 @@ function register(channel, name, opts, handle) {
                     'Could not serialize thrift');
             }
 
-            res.headers.as = 'thrift';
-            res.setOk(thriftRes.ok);
-            res.send(
-                stringifyResult.value.head,
-                stringifyResult.value.body
-            );
+            if (res.setOk(thriftRes.ok)) {
+                res.headers.as = 'thrift';
+                res.send(
+                    stringifyResult.value.head,
+                    stringifyResult.value.body
+                );
+            }
         }
     }
 };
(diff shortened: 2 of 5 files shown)


node/streaming_out_response.js
```diff
@@ -86,10 +86,12 @@ StreamingOutResponse.prototype.setOk = function setOk(ok) {
         self.errorEvent.emit(self, errors.ResponseAlreadyStarted({
             state: self.state
         }));
+        return false;
```
kriskowal: Regardless, we should return here indeed. Definitely should not change an already sent response.

node/as/thrift.js
```diff
@@ -161,12 +161,13 @@ function register(channel, name, opts, handle) {
                     'Could not serialize thrift');
             }
 
-            res.headers.as = 'thrift';
-            res.setOk(thriftRes.ok);
-            res.send(
-                stringifyResult.value.head,
-                stringifyResult.value.body
-            );
+            if (res.setOk(thriftRes.ok)) {
```
kriskowal: I don’t think we can rely on consumers to do this. res.send needs a similar guard. Shame that they’ll plow ahead overwriting the headers though.

kriskowal:
:ok:

node/as/json.js
```diff
@@ -233,12 +233,13 @@ TChannelJSON.prototype.register = function register(
                     'Could not JSON stringify');
             }
 
-            res.headers.as = 'json';
-            res.setOk(respObject.ok);
-            res.send(
-                stringifyResult.value.head,
-                stringifyResult.value.body
-            );
+            if (res.setOk(respObject.ok)) {
```
Raynos: would prefer a simple `if (!res.end) {` check.

Raynos:
I think this is a bad change.

You lose a lot of visibility into what exactly is going in the double response path. I had a very hard time debugging double responses in ncar itself.

I would prefer if we just reverted this entire change.

If you want to truly add gaurds then add focused gaurds

 - as layer protects you from writing to a timed out response.
 - as layer protects you from double callback.

Raynos:
I guess this change would be ok if we fixed the `setOk()` log call site.

mostly all these log options ( https://github.com/uber/tchannel/blob/master/node/out_response.js#L287-L308 ).

node/as/json.js
jcorbin: `res.end` isn't sufficient, `res.state` is the source of truth, which is what `res.setOk(...)` is checking in the first place.  In fact I don't trust that we're correctly setting `.end` whenever we move into a terminal `.state`; the `.end` property was always intended as a convenience to record a time, and shouldn't be branched on like a state indicator.

Raynos: agreed.

PR #1182, opened 2015-08-19
Add Peers method within the Registrar interface

aravindvs:
@prashantv 

based on your suggestion to make the integration with subchannels easier..

### DIFF
--- a/golang/channel.go
+++ b/golang/channel.go
@@ -216,7 +216,7 @@ type Registrar interface {
 	// Logger returns the logger for this Registrar.
 	Logger() Logger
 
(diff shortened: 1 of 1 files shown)


golang/channel.go
```diff
@@ -215,6 +215,9 @@ type Registrar interface {
 
 	// Logger returns the logger for this Registrar.
 	Logger() Logger
+
+	// Peers returns the peerlist for this Registrar
```
prashantv: nit: "Peers returns the peer list for this Registrar."
(formatting)

prashantv:
lgtm

PR #1186, opened 2015-08-20
go: fix panic caused by closed sendCh when running ringpop-go tests

prashantv:
This fixes the race conditions between receiving calls/making calls and closing the connection.

There is still one race condition left caused by timeouts. There is already a TODO on line 175 which requires a larger refactoring.


prashantv:
cc @slantin

PR #1188, opened 2015-08-20
discovery api for getting hostports by service name

ShanniLi:
r @jcorbin @kriskowal @anson627  
cc @blampe


node/hyperbahn/handler.js
```diff
@@ -24,10 +24,14 @@
 
 var assert = require('assert');
 var util = require('util');
+var fs = require('fs');
+var path = require('path');
 var setTimeout = require('timers').setTimeout;
 
 var Errors = require('../errors.js');
 var TChannelJSON = require('../as/json');
+var TChannelThrift = require('../as/thrift');
```
Raynos: Let's not use thrift until `thriftify` is gone. I want to use the new `thriftrw` only.

node/hyperbahn/handler.js
```diff
@@ -338,3 +359,108 @@ function unadvertise(service) {
     var self = this;
     self.channel.topChannel.handler.removeServicePeer(service.serviceName, service.hostPort);
 };
+
+HyperbahnHandler.prototype.handleHostportsByService =
+function handleHostportsByService(self, req, arg2, arg3, cb) {
```
Raynos: req, head, body

node/hyperbahn/handler.js
```diff
@@ -338,3 +359,108 @@ function unadvertise(service) {
...
+    var serviceName = arg3.serviceName;
```
Raynos: Cannot assume `serviceName` exists because lolthrift

node/hyperbahn/handler.js
```diff
@@ -338,3 +359,108 @@ function unadvertise(service) {
...
+    var exitNodes = self.egressNodes.exitsFor(serviceName);
+    var exitHosts = Object.keys(exitNodes);
+
+    var myHost = self.channel.hostPort;
+    if (exitHosts.indexOf(myHost) === -1) {
+        // Since Hyperbahn is fully connected to service hosts,
+        // randomly picking one suffices
+        var exit = Math.floor(Math.random() * exitHosts.length);
+        self.sendRelayAsThrift({
```
Raynos: Just inline the request here. no need to retry

node/hyperbahn/handler.js
```diff
@@ -338,3 +359,108 @@ function unadvertise(service) {
...
+        }, onFinish);
+    } else {
+        var svcchan = self.channel.topChannel.subChannels[serviceName];
+        var hosts = [];
+        if (svcchan) {
+            hosts = svcchan.peers.keys();
+        }
+        var err;
+        if (hosts.length === 0) {
+            err = NoHostError({
```
Raynos: Don't bother; just return an empty array.

node/hyperbahn/handler.js
```diff
@@ -338,3 +359,108 @@ function unadvertise(service) {
...
+                serviceName: serviceName
+            });
+        }
+
+        onFinish(null, {
```
Raynos: Don't do this stuff. Use a different path for error and not error.

node/hyperbahn/handler.js
```diff
@@ -338,3 +359,108 @@ function unadvertise(service) {
...
+    if (exitHosts.indexOf(myHost) === -1) {
+        // Since Hyperbahn is fully connected to service hosts,
+        // randomly picking one suffices
+        var exit = Math.floor(Math.random() * exitHosts.length);
+        self.sendRelayAsThrift({
+            hostPort: exitHosts[exit],
+            body: arg3,
+            inreq: req,
+            endpoint: 'Hyperbahn::hostportsByService'
+        }, onFinish);
```
Raynos: dont pass the callback to the method. Do all the body building in here.

node/hyperbahn/handler.js
```diff
@@ -338,3 +359,108 @@ function unadvertise(service) {
...
+                (
+                    codeName === 'NetworkError' ||
+                    codeName === 'Timeout'
+                )
+            ) {
+                setTimeout(tryRequest, self.relayAdRetryTime);
+            } else {
+                self.logError(err, opts, response);
+
+                callback(null, {
```
Raynos: Don't do boolean logic in the object. use proper if statements.

node/hyperbahn/hyperbahn.thrift
```diff
@@ -0,0 +1,16 @@
+exception Exception {
+    1: required string message
+}
+
+exception NoHost {
+    1: required string message
+}
+
+service Hyperbahn {
+    list<string> hostportsByService(
```
Raynos: This is not how you do thrift. Always take one argument; a struct. Always return one argument a struct.

node/hyperbahn/hyperbahn.thrift
```diff
@@ -0,0 +1,16 @@
+exception Exception {
+    1: required string message
+}
+
+exception NoHost {
```
Raynos: These exceptions are useless; they need more fields; namely `type` and `fullType` and any other meta info.

node/hyperbahn/hyperbahn.thrift
```diff
@@ -0,0 +1,16 @@
+exception Exception {
```
Raynos: A catch all exception is useless.

node/hyperbahn/hyperbahn.thrift
Raynos: The logical operation here is `discover`. Let's call this `discover`

Raynos:
Let's shelf this feature until we are ready to fully commit to thrift.

node/hyperbahn/handler.js
```diff
@@ -338,3 +359,108 @@ function unadvertise(service) {
...
+        if (hosts.length === 0) {
+            err = NoHostError({
+                serviceName: serviceName
+            });
+        }
+
+        onFinish(null, {
+            ok: !err,
+            body: err || hosts,
+            typeName: err ? 'noHost' : null
```
anson627: what is this typeName? error type?

node/hyperbahn/handler.js
```diff
@@ -338,3 +359,108 @@ function unadvertise(service) {
...
+
+HyperbahnHandler.prototype.sendRelayAsThrift =
+function sendRelayAsThrift(opts, callback) {
+    var self = this;
+
+    var attempts = 0;
+
+    tryRequest();
+
+    // TODO: move functions out to methods
```
anson627: why not move it now?

node/hyperbahn/handler.js
anson627: I guess I don't understand why thrift, even json dependency is already there.

ShanniLi: This is a task that is due next week ... And it specifically requires using thrift.

node/hyperbahn/handler.js
ShanniLi: Empty array won't show up in TChannel stats. NoHostError will.

node/hyperbahn/hyperbahn.thrift
ShanniLi: "This is not how you do thrift. Always take one argument; a struct. Always return one argument a struct."
this sounds wired. Any reason?

blampe: Extensibility in addition to weirdness around how Thrift handles function arguments. Similar in spirit to always returning JSON objects instead of lists.

node/hyperbahn/handler.js
blampe: thriftify/thriftrw is just a matter of how the objects get serialized on the wire, no? Can we not swap this out with thriftrw later?

node/hyperbahn/handler.js
Raynos: good point; seems legit.

node/hyperbahn/handler.js
Raynos: thriftify has very low test coverage and can cause hyperbahn routers to crash in edge cases.

thriftrw is almost ready. I'd rather not add a production dependency to our hyperbahn deployment with low test coverage.

ShanniLi: so, should we stick to JSON for now?

Raynos: If we cant wait then just get a JSON version out. I'd rather we wait

ShanniLi: W.r.t. thriftify, it has been used by production for months, so I wouldn't worry too much about the test coverage.

Raynos: Apache thrift has been used in production for months. 

The discovery feature isn't urgent. No need to rush. Let's just do it right. Thrift is for life, you get one shot at the interface.

ShanniLi:
Talked to @Raynos offline. Will shelf this until @kriskowal finishes the thriftrw

Raynos:
It's been 3 weeks. Thriftrw is not as close as I thought. I'll unblock this. 

We can add thriftrw in the future. Cc @kriskowal @jcorbin

jcorbin:
What is blocking thriftrw from being usable?

ShanniLi:
Sounds good. I will move forward with thriftify. I will have this PR updated today or tomorrow.

node/hyperbahn/handler.js
ShanniLi: two reasons: 1) generalizing this interface for moving other Hyperbahn endpoints to thrift. 2) retry is necessary, e.g., when k-value is being updated.

node/test/hyperbahn/hostports.js
```diff
@@ -0,0 +1,124 @@
...
+        function onResponse(err, res) {
+            assert.ok(!err, 'should be no errors');
+            assert.ok(res, 'should be a result');
+            assert.ok(!res.ok, 'result should be not ok');
+            assert.equals(res.body.message, 'no peer available for matt', 'error message as expected');
+            assert.end();
+        }
+    });
+
+    HyperbahnCluster.test.only('get host port as expected', {
```
Raynos: dont leave onlys

node/hyperbahn/handler.js
```diff
@@ -36,6 +40,22 @@ var RELAY_AD_TIMEOUT = 500;
 
 module.exports = HyperbahnHandler;
 
+// TODO: should be part of Hyperbahn error file
+var TypedError = require('error/typed');
+var NoPeersAvailable = TypedError({
+    type: 'no-peers-asvailable',
```
Raynos: s/asvailable/available/

Raynos: hyperbahn.no-peers-available

node/hyperbahn/handler.js
```diff
@@ -337,3 +365,139 @@ function unadvertise(service) {
...
+
+        res.push(obj);
+    }
+
+    return res;
+}
+
+HyperbahnHandler.prototype.discover =
+function discover(self, req, head, body, cb) {
+    var serviceName = body && body.query && body.query.serviceName;
```
Raynos: No need to do any of this if checking; 

Write tests and rely on thriftrw.

node/hyperbahn/handler.js
```diff
@@ -337,3 +365,139 @@ function unadvertise(service) {
...
+    if (typeof serviceName !== 'string' || serviceName.length === 0) {
```
Raynos: Should always be string; but do write tests. 

Do check for non empty though :)

node/hyperbahn/handler.js
Raynos: Dont retry here; the edge client will retry.

Raynos: Do not prematurely generalize.

node/hyperbahn/handler.js
```diff
@@ -337,3 +365,139 @@ function unadvertise(service) {
...
+                body: {
+                    peers: hosts
+                }
+            });
+        }
+    }
+};
+
+HyperbahnHandler.prototype.sendRelayAsThrift =
+function sendRelayAsThrift(opts, callback) {
```
Raynos: Lets drop the retries.

node/test/hyperbahn/hostports.js
```diff
@@ -0,0 +1,124 @@
...
+                    serviceName: 'matt'
+                }
+            },
+            onResponse
+        );
+        function onResponse(err, res) {
+            assert.ok(!err, 'should be no errors');
+            assert.ok(res, 'should be a result');
+            assert.ok(!res.ok, 'result should be not ok');
+            assert.equals(res.body.message, 'no peer available for matt', 'error message as expected');
```
Raynos: assert res.body.serviceName

Raynos: How do we know which error it is ? Is there a `res.typeName` ?

node/test/hyperbahn/hostports.js
```diff
@@ -0,0 +1,124 @@
...
+                },
+                check
+            );
+        }
+
+        function check(err, res) {
+            assert.ok(!err, 'should be no errors');
+            assert.ok(res, 'should be a result');
+            assert.ok(res.ok, 'result should be ok');
+            assert.equals(res.body[0], bob.channel.hostPort, 'should get the expected hostPort');
```
Raynos: This cant be right at all.

The response should have a peers field and stuff.

node/test/hyperbahn/hostports.js
```diff
@@ -0,0 +1,124 @@
...
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+'use strict';
+
+var DebugLogtron = require('debug-logtron');
```
Raynos: We need more tests.

Let's add a bunch of tests for thrift edge cases.

 - test sending a request with a malformed thrift IDL where we send an empty body. Assert BadRequest
 - test sending a request with a malformed thrift IDL where we send a body with a query without the serviceName field. Assert BadRequest
 - test sending a request with correct thrift IDL where we send an empty string service Name. Assert NotOk application error
 - test sending a request with a malformed thrift IDL which has zero exceptions defined; Send a request with empty string serviceName. Assert that the client doesn't blow up and gets back a not ok response.

ShanniLi:
All tests written with failing tests commented out. See issues:
https://github.com/uber/thriftrw/issues/54
https://github.com/uber/thriftrw/issues/55

node/hyperbahn/handler.js
```diff
@@ -337,3 +365,119 @@ function unadvertise(service) {
...
 };
+
+function convertHosts(hosts) {
+    var res = [];
+    for (var i = 0; i < hosts.length; i++) {
+        var strs = hosts[i].split(':');
+        var obj = {
+            port: parseInt(strs[1])
+        };
+        strs = strs[0].replace('localhost', '127.0.0.1').split('.');
```
Raynos: There should not be any localhost in here...

node/hyperbahn/handler.js
```diff
@@ -337,3 +365,119 @@ function unadvertise(service) {
...
+
+        res.push(obj);
+    }
+
+    return res;
+}
+
+HyperbahnHandler.prototype.discover =
+function discover(self, req, head, body, cb) {
+    var serviceName = body.query.serviceName;
```
Raynos: This is only save to do once we fixed the thriftrw bugs.

node/hyperbahn/handler.js
```diff
@@ -337,3 +365,119 @@ function unadvertise(service) {
...
+                cn: self.callerName
+            },
+            retryLimit: 0,
+            parent: opts.inreq
+        }), opts.endpoint, null, opts.body, onResponse);
+    }
+
+    function onResponse(err, response) {
+        if (response) {
+            return callback(null, {
```
Raynos: just callback(null, response)

node/hyperbahn/handler.js
```diff
@@ -337,3 +365,119 @@ function unadvertise(service) {
...
+    if (exitHosts.indexOf(myHost) === -1) {
+        // Since Hyperbahn is fully connected to service hosts,
+        // randomly picking one suffices
+        var exit = Math.floor(Math.random() * exitHosts.length);
+        self.sendRelayAsThrift({
+            hostPort: exitHosts[exit],
+            body: body,
+            inreq: req,
+            endpoint: 'Hyperbahn::discover'
+        }, cb);
```
Raynos: Don't do this.

Add a real handler here.

node/hyperbahn/handler.js
```diff
@@ -337,3 +365,119 @@ function unadvertise(service) {
...
+        if (response) {
+            return callback(null, {
+                ok: response.ok,
+                body: response.body,
+                typeName: response.typeName
+            });
+        }
+
+        self.logError(err, opts, response);
+        callback(null, {
```
Raynos: This is not valid at all. Just callback(err)

node/hyperbahn/handler.js
```diff
@@ -337,3 +365,119 @@ function unadvertise(service) {
...
+    function onResponse(err, response) {
+        if (response) {
+            return callback(null, {
+                ok: response.ok,
+                body: response.body,
+                typeName: response.typeName
+            });
+        }
+
+        self.logError(err, opts, response);
```
Raynos: Let's not log here but instead of log in the other method.

node/hyperbahn/handler.js
```diff
@@ -337,3 +365,119 @@ function unadvertise(service) {
...
+    function onIdentified(err) {
+        if (err) {
+            return onResponse(err);
+        }
+
+        self.tchannelThrift.send(self.channel.request({
+            host: opts.hostPort,
+            serviceName: 'hyperbahn',
+            trace: false,
+            timeout: self.relayAdTimeout,
```
Raynos: Wrong timeout; use a new timeout.

node/hyperbahn/handler.js
Raynos: Instead of passing in `cb` you want to do something like

function handleForward(err, resp) {
    if (resp) {
        return cb(null, resp)
    }

    logAThingy();
    sendErrorFrame(errors.classify(err), err.message)
}
```

File: node/test/hyperbahn/hostports.js [hunk 29]
```diff
@@ -0,0 +1,267 @@
...
+            if (err) {
+                assert.end(err);
+            }
+            assert.ok(!res.ok, 'should be not ok');
+            // TODO: the res.body.message is missing, need to assert on that
+            assert.end();
+        }
+    });
+
+    // HyperbahnCluster.test('malformed thrift IDL: an empty body', {
```
Raynos: We will want to fix these bugs before landing.

node/hyperbahn/handler.js
ShanniLi: Not sure I understand: sendErrorFrame(errors.classify(err), err.message)?
1) what is sendErrorFrame?
2) errors.classify is for tchannel error only. The error here can also be NoPeersAvailable or InvalidServiceName.

ShanniLi: kind of understand what you mean. I will send out an update.

node/test/hyperbahn/hostports.js
ShanniLi: These are minor issues of thriftrw and only occur under bad IDLs (very uncommon scenarios). They are also orthogonal to this PR. Can we land this PR first? :)

node/hyperbahn/handler.js
Raynos: This statement should be save to have as `var serviceName = body.query.serviceName` again.

node/hyperbahn/handler.js
```diff
@@ -337,3 +369,128 @@ function unadvertise(service) {
...
+    }
+
+    var exitNodes = self.egressNodes.exitsFor(serviceName);
+    var exitHosts = Object.keys(exitNodes);
+
+    var myHost = self.channel.hostPort;
+    if (exitHosts.indexOf(myHost) === -1) {
+        // Since Hyperbahn is fully connected to service hosts,
+        // randomly picking one suffices
+        var exit = Math.floor(Math.random() * exitHosts.length);
```
Raynos: Wait; why are we picking a random peer ? 

We can just make a request on the sub channel for the serviceName and sub channel will pick a random peer for us.

We can also skip `waitForIdentified` that way and use retries on other random peers :D

node/hyperbahn/handler.js
```diff
@@ -337,3 +369,128 @@ function unadvertise(service) {
...
+                ok: true,
+                body: {
+                    peers: hosts
+                }
+            });
+        }
+    }
+};
+
+HyperbahnHandler.prototype.sendRelayAsThrift =
```
Raynos: We should not need this helper if we make a retryable request directly on the sub channel.

node/test/hyperbahn/hostports.js
```diff
@@ -0,0 +1,271 @@
...
+            {
+                query: {
+                    serviceName: ''
+                }
+            },
+            onResponse
+        );
+        function onResponse(err, res) {
+            assert.ok(err, 'should be error');
+            assert.equals(err.type, 'tchannel-thrift-handler.parse-error.body-failed',
```
Raynos: Can we add an assertion that this is a response parsing failure ?

node/test/hyperbahn/hostports.js
```diff
@@ -0,0 +1,271 @@
...
+        var badThrift = new TChannelAsThrift({source: badSource});
+        badThrift.send(request,
+            'Hyperbahn::discover',
+            null,
+            {query: {}},
+            onResponse
+        );
+        function onResponse(err, res) {
+            assert.ok(err, 'should be error');
+            assert.equals(err.type, 'tchannel.bad-request', 'error should be bad request');
```
Raynos: Can we add an assertion on the `err.message` to show why it failed ?

node/test/hyperbahn/hostports.js
```diff
@@ -0,0 +1,271 @@
...
+        var badThrift = new TChannelAsThrift({source: badSource});
+        badThrift.send(request,
+            'Hyperbahn::discover',
+            null,
+            {},
+            onResponse
+        );
+        function onResponse(err, res) {
+            assert.ok(err, 'should be error');
+            assert.equals(err.type, 'tchannel.bad-request', 'error should be bad request');
```
Raynos: Can we add an assertion on the `err.message` to show why it failed ?

node/test/hyperbahn/hostports.js
```diff
@@ -0,0 +1,271 @@
...
+            );
+        }
+
+        function check(err, res) {
+            if (err) {
+                assert.end(err);
+            }
+            assert.ok(res, 'should be a result');
+            assert.ok(res.ok, 'result should be ok');
+            console.log(res.body.peers[0]);
```
Raynos: no console.log

node/test/hyperbahn/hostports.js
```diff
@@ -0,0 +1,271 @@
...
+
+module.exports = runTests;
+
+if (require.main === module) {
+    runTests(require('../lib/hyperbahn-cluster.js'));
+}
+
+function covertHost(host) {
+    var res = '';
+    console.log(host.ip.ipv4);
```
Raynos: no console.log

ShanniLi: haha, my bad. :)

node/hyperbahn/handler.js
ShanniLi: Good point!

node/test/hyperbahn/hostports.js
ShanniLi: Will add. However, such validation should really be in thriftrw ...

node/hyperbahn/handler.js
```diff
@@ -337,3 +370,98 @@ function unadvertise(service) {
...
+            if (err) {
+                self.channel.logger.error('Failed to call discover API on exit node', {
+                    error: err,
+                    serviceName: serviceName
+                });
+                return cb(err, null);
+            }
+
+            // Need to reconstruct the error
+            if (resp.body.name === 'ThriftException') {
```
Raynos: This feels liks a really bad check. Can we do `resp.typeName` ?

node/hyperbahn/handler.js
```diff
@@ -337,3 +370,98 @@ function unadvertise(service) {
...
+    if (exitHosts.indexOf(myHost) === -1) {
+        // Since Hyperbahn is fully connected to service hosts,
+        // any exit node suffices.
+        svcchan = self.channel.topChannel.handler.getOrCreateServiceChannel(serviceName);
+        self.tchannelThrift.send(svcchan.request({
+            serviceName: 'hyperbahn',
+            headers: {
+                cn: 'hyperbahn'
+            },
+            parent: req
```
Raynos: Let's set a `timeout` and `timeoutPerAttempt`; consider 5000 & 500 for defaults.

Let's also set `trace: false`

Raynos:
Looking good. #soclose

node/as/thrift.js
```diff
@@ -323,7 +322,9 @@ TChannelAsThrift.prototype._parse = function parse(opts) {
         if (bodyRes.value && opts.ok) {
             bodyRes.value = bodyRes.value.success;
         } else if (bodyRes.value && !opts.ok) {
-            bodyRes.value = onlyProperty(bodyRes.value);
+            var typeName = onlyKey(bodyRes.value);
+            bodyRes.value = bodyRes.value[typeName];
+            bodyRes.value.typeName = typeName;
```
Raynos: dont mutate `bodyRes.value`. Just set `bodyRes.typeName`

Raynos:
Looks good. Let's get this in after tests pass.

Raynos:
#sick

PR #1189, opened 2015-08-20
Open source hyperbahn.

Raynos:
There is a lot here because hyperbahn is a lot.

All the tests are here and they pass. The Makefile
is here and it works.

Our closed source hyperbahn will just be a bunch
of integration with uber specific infrastructure
like alerting and deploying.

This depends on a closed source code review of
the internal hyperbahn before getting merged.

r: @jcorbin @rf @kriskowal

### DIFF
--- a/hyperbahn/bin/admin.js
+++ b/hyperbahn/bin/admin.js
@@ -15,6 +15,8 @@ var fs = require('fs');
 function main() {
     /*eslint no-process-exit: 0*/
     var argv = minimist(process.argv.slice(2));
+    argv.hostfile = argv.H || argv.hostfile;
+
     var admin = Admin(argv);
     admin.run(onAdmin);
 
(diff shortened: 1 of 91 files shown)


hyperbahn/bin/admin.js
```diff
@@ -0,0 +1,458 @@
...
+
+function Admin(options) {
+    if (!(this instanceof Admin)) {
+        return new Admin(options);
+    }
+
+    var self = this;
+
+    self.hosts = options.hosts || getHostPortsForProd(
+        '/etc/uber/hyperbahn/ringpop-v2.json'
```
abhinav: ?

Raynos: This should use config instead... hmm.

anson627:
why not standalone hyperbahn repo? Does it mean git clone, npm install, npm test etc. are going to be slower?

jcorbin:
The plan is to split it out once the existing hyperbahn folder is
completed, which is what this PR does

Raynos:
We need to consolidate `hyperbahn` and `node/hyperbahn` into one folder. 

This is a multi-step progress where we try to break as few open branches as possible; one step at a time.

Raynos:
People can review at their leisure. This is going in. I'm going to release the lock on these thousands of lines of code by migrating it to their new home.

PR #1193, opened 2015-08-20
as/http: drop usage of PassThrough

jcorbin:
While implementing @anson627 's ask in #1040, I realized that we don't even need PassThrough at all in as/http.


kriskowal:
excellent

anson627:
:+1:

PR #1194, opened 2015-08-20
go: Add frame pools that use sync.Pool and channels

prashantv:
The new pools are NOT used, the default is still to rely on the heap + GC.

This change adds pool that we can test before making them the default.

### DIFF
--- a/golang/frame_pool_b_test.go
+++ b/golang/frame_pool_b_test.go
@@ -1,0 +1,82 @@
+package tchannel_test
+
+import (
+	"math/rand"
+	"sync"
+	"sync/atomic"
+	"testing"
+
+	. "github.com/uber/tchannel/golang"
+)
+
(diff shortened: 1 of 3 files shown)


golang/frame_pool_b_test.go
```diff
@@ -0,0 +1,82 @@
...
+import (
+	"math/rand"
+	"sync"
+	"sync/atomic"
+	"testing"
+
+	. "github.com/uber/tchannel/golang"
+)
+
+// Copyright (c) 2015 Uber Technologies, Inc.
```
breerly: Seems odd putting the license here... is that common in Go or something?

prashantv: Good point, the license goes after package but before imports. Then I think goimports has been adding imports to above the license, and I didn't notice. Will fix in a separate PR, thanks.

PR #1197, opened 2015-08-21
relay_handler: fix for no streamed arg1

jcorbin:
So v2.6.1 regressed relay_handler, because the only test against it is in Hyperbahn...

r @kriskowal @Raynos


kriskowal:
This is perhaps the right time to migrate that test.

jcorbin:
Yup, in a later PR.

PR #1205, opened 2015-08-21
Upgrade ringpop and tchannel

jcorbin:


hyperbahn/package.json
```diff
@@ -35,13 +35,13 @@
     "raynos-replr": "0.2.5-port-0-support",
     "readable-stream": "1.0.33-2",
     "ready-signal": "1.2.0",
-    "ringpop": "10.0.0-beta6",
+    "ringpop": "^10.3.0",
```
Raynos: Don't use carets for hyperbahn. Only fixed dependencies.

jcorbin:
Fixed to exact deps

PR #1206, opened 2015-08-22
go: Make json Headers map[string]string

prashantv:
Since json.Context and thrift.Context both use map[string]string, use a common tchannel.headerCtx to provide implementations for both.


breerly:
nice

PR #1207, opened 2015-08-22
Copy meta.thrift and remove thriftIDL

prashantv:
Move the meta.thrift file out of `node/as` since it is common to all languages.

Removed the thriftIDL method till we figure out the correct solution and add library support to implement it.

r: @ShanniLi @Raynos 
cc: @blampe


prashantv:
Updated: Copied the file instead of removing the node version.

ShanniLi:
lgtm

PR #1211, opened 2015-08-24
Clarify stale vs init timeouts

jcorbin:
- also dropped a (perhaps common) closure allocation in else branch of `Operations#checkLastTimeoutTime`-
- added more log context
- started the `conn.extendLogInfo` pattern

r @kriskowal


kriskowal:
Looks good to me.

PR #1212, opened 2015-08-24
as/http: add support for array values (e.g. set-cookie)

jcorbin:


davewhat:
LGTM

jcorbin:
We should add (more) headers test coverage to test/as-http, but meanwhile this doesn't fail the existing test @anson627 .

PR #1217, opened 2015-08-25
Hyperbahn: upgrade ringpop to 10.3.0 (again)

jcorbin:
Now that we've ship the test_ringpop_10.3 branch, and not been able to reproduce the problems that caused us to back it out.

r @kriskowal


kriskowal:
:ok:

PR #1218, opened 2015-08-25
Proposal for routing to named instances

kriskowal:
This captures an idea that @mranney communicated for enabling health checks to specific instances through Hyperbahn.

r @jsu1212 @jcorbin @Raynos

### DIFF
--- a/docs/health.md
+++ b/docs/health.md
@@ -1,0 +1,30 @@
+
+# Instance tracking for health checks
+
+To facilitate health checks to named workers through Hyperbahn, workers may
+advertise both the service name (e.g., `tcollector`) and a instance name (e.g.,
+`tcollector01-1-dc1`).
+The advertisement is broadcast to every affine Hyperbahn relay for that
+service.
+
+With partial affinity, each relay is responsible for a proportional subset of
+the advertising service workers.
+Relays determine the relay responsible for a worker by sorting the known relay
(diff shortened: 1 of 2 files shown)


prashantv:
I thought tags might be a nicer solution rather than doing something specifically for routing to an instance.

junchaowu:
We need to discuss more about that. This is also a problem I am trying to solve right now. The instance name for some python service are same among some instance. For example, the "service A" have 10 processes are naming "serviceA"

anson627:
@junchaowu interesting! How come different service instances have same name? aren't they named svc-0, svc-1, svc-2 etc.? I mean how deployment system or service controller would allow that.

junchaowu:
This is a legacy issue from python uwsgi world. Uwsgi spawns up multi-process of instances and all of them listen to same port.

anson627:
thanks @junchaowu! good to know.

kriskowal:
@prashantv Yes, tags instead of instance name is the more general approach. You also mentioned using vector clocks. Let’s work on another iteration of this proposal. Each service would need to regenerate an index[tag]:set<peer> based on the most advanced ad/unad per worker.

Raynos:
I like this proposal as is.

It's really simple and can be shipped.

We are missing only one thing. Let's change init request and init response to add `instance_name` as an optional header so that it's done at init time and not at advertise time.

PR #1220, opened 2015-08-25
Update rate limiting stats

prashantv:
Make it clear that `target-service` is not the literal string but a variable.

r: @ShanniLi


ShanniLi:
lgtm

PR #1221, opened 2015-08-25
as/http improve http header parsing and buffering

anson627:
this is a follow-up of #1212.

HTTP header set-cookie should always be array (even it only has one element), max-forwards should be first win, and the rest of them should be comma-separated list.

how node 10 manages http header:
https://github.com/joyent/node/blob/d13d7f74d794340ac5e126cfb4ce507fe0f803d5/lib/_http_incoming.js#L142

RFC:
https://www.ietf.org/rfc/rfc2616.txt

r: @jcorbin @kriskowal @ShanniLi 
cc: @davewhat @vipulaneja


jcorbin:
lgtm anyhow :100:

PR #1222, opened 2015-08-25
Relay refactor

jcorbin:
- continues the `foo.extendLogInfo(info)` pattern towards more consistent logging
- pulls key parts out of RelayRequest, to be driven by RelayHandler, to survive in the upcoming lazy path
- drops branched closure allocations
- moves closure allocations to constructor, rather than maybe called methods

r @kriskowal 

cc @jwolski: @Raynos tells me that Ringpop's been using this API, and that this change will require an update in its forwarding path.  We don't consider this a public API nor a breaking change, so major won't be bumping... however I will give you a minor bump, it won't be a 2.6.x.  Note also the crucial change from your perspective is https://github.com/uber/tchannel/commit/c8be0373216a6e0813f49f2d42c1b54a0ad33fb9, which I could ultimately leave out of this branch, and punt the peer selection work until a later lazy branch, let me know how you want to handle this.

### DIFF
--- a/node/in_request.js
+++ b/node/in_request.js
@@ -69,6 +69,24 @@ inherits(TChannelInRequest, EventEmitter);
 
 TChannelInRequest.prototype.type = 'tchannel.incoming-request';
 
+TChannelInRequest.prototype.extendLogInfo = function extendLogInfo(info) {
+    var self = this;
+
+    // TODO: add:
+    // - request id?
+    // - tracing id?
+    // - other?
+    info.inRemoteAddr = self.remoteAddr;
+    info.serviceName = self.serviceName;
+    if (self.endpoint !== null) {
+        info.arg1 = self.endpoint;
+    } else {
+        info.arg1 = String(self.arg1);
+    }
+
+    return info;
+};
+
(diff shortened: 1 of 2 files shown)


node/in_request.js
```diff
@@ -69,6 +69,24 @@ inherits(TChannelInRequest, EventEmitter);
...
+TChannelInRequest.prototype.extendLogInfo = function extendLogInfo(info) {
+    var self = this;
+
+    // TODO: add:
+    // - request id?
+    // - tracing id?
+    // - other?
+    info.inRemoteAddr = self.remoteAddr;
+    info.serviceName = self.serviceName;
+    if (self.endpoint !== null) {
```
kriskowal: Perhaps `info.arg1 = null` up front to improve probability of producing a consistent hidden class.

kriskowal:
:ok:

jwolski:
@jcorbin Go ahead and do your thing (merged as is). I'll freeze the TChannel dependency in Ringpop and make sure to update the Ringpop code that depends on this API when we bump the TChannel minor.

jcorbin:
@wolski this will ship in 2.7.0

jwolski:
noted

PR #1223, opened 2015-08-26
Add lazy support to v2/{call,cont,header,checksum,error_response}

jcorbin:
Add lazy RW support to key v2 parts.

r @kriskowal @rf


jcorbin:
Now with Moar Tests ™, please review.

kriskowal:
Good enough for now—my hope is a single specialized R that extracts all properties of interest in one pass, and a W that updates exactly the points of interest (id, tracing).

PR #1225, opened 2015-08-26
Fix lbpool encoding

anson627:
r: @joshua @ShanniLi @kriskowal 
cc: @davewhat @vipulaneja 
lbpool by default use utf8 as encoding, which is not valid when a binary response is present.

https://github.com/Voxer/lb_pool/blob/6f53c5c4ad8b6e0dd6899fb9a9f35b710c8300de/pool_endpoint_request.js#L137


node/as/http.js
```diff
@@ -311,7 +311,9 @@ TChannelHTTP.prototype.forwardToHTTP = function forwardToHTTP(tchannel, options,
 
 TChannelHTTP.prototype._forwardToLBPool = function _forwardToLBPool(options, inreq, outres, callback) {
     var self = this;
-
+    options = extend(options, {
```
jcorbin: Extend is slow, especially on an every-request method like this; we've abolished all (most?) usage of it from production code paths.  Please instead:
```
if (!options) { options = {}; }
options.encoding = null;
```

Mutating the passed options directly should be okay, but if you'd really rather not look at existing tchannel options handling code paths for inspiration.

PR #1226, opened 2015-08-26
go: Add support for setting shard key on calls

prashantv:
This change refactors how contexts are built and where TChannel parameters are stored. We use a single context Value to store all TChannel parameters to avoid creating redundant contexts and avoiding O(n) lookup (where n is the number rof values).

There is still WIP to figure out how best to deal with CallOptions, as now we have two methods:
 - directly when doing BeginCall
 - as part of the Context


prashantv:
cc @junchaowu

PR #1228, opened 2015-08-26
go: Advertise multiple services hosted on the same channel

prashantv:
When calling Advertise, a client can (optionally) pass subchannels to advertise with Hyperbahn. Currently, all services must be defined when doing the first Advertise call.

cc: @cjbradfield

### DIFF
--- a/golang/hyperbahn/client.go
+++ b/golang/hyperbahn/client.go
@@ -32,8 +32,9 @@ import (
 
 // Client manages Hyperbahn connections and registrations.
 type Client struct {
-	tchan *tchannel.Channel
-	opts  ClientOptions
+	tchan    *tchannel.Channel
+	services []string
+	opts     ClientOptions
 }
 
 // FailStrategy is the strategy to use when registration fails maxRegistrationFailures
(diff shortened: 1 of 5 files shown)


golang/hyperbahn/client.go
```diff
@@ -119,9 +120,21 @@ func addPeer(ch *tchannel.Channel, hostPort string) {
...
+	for _, s := range otherServices {
+		c.services = append(c.services, s.ServiceName())
+	}
+}
+
 // Advertise advertises the service with Hyperbahn, and returns any errors on initial advertisement.
+// Advertise can register multiple services hosted on the same endpoint.
 // If the advertisement succeeds, a goroutine is started to re-advertise periodically.
-func (c *Client) Advertise() error {
+func (c *Client) Advertise(otherServices ...tchannel.Registrar) error {
```
cjbradfield: So, now if you have no subchannels, you still have to pass in the channel you used when you created the client, right?

prashantv: Nope, it's backwards compatible, so if you have no subchannels, you just do `client.Advertise()` and it will just register the top channel.

cjbradfield:
looks good to me then!

PR #1229, opened 2015-08-26
Fix and improve ServiceDispatchHandler#removeServicePeer

jcorbin:
r @kriskowal @anson627

### DIFF
--- a/node/hyperbahn/service_proxy.js
+++ b/node/hyperbahn/service_proxy.js
@@ -296,7 +296,12 @@ function refreshServicePeer(serviceName, hostPort) {
 ServiceDispatchHandler.prototype.removeServicePeer =
 function removeServicePeer(serviceName, hostPort) {
     var self = this;
+
     var svcchan = self.channel.subChannels[serviceName];
+    if (!svcchan) {
+        return;
+    }
+
     var peer = svcchan.peers.get(hostPort);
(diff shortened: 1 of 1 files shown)


jcorbin:
The first commit fixes the uncaught we saw in prod.

The second two fix the semantics to be (more) correct.

kriskowal:
:shipit:

node/hyperbahn/service_proxy.js
```diff
@@ -296,12 +296,31 @@ function refreshServicePeer(serviceName, hostPort) {
...
+    var subChanKeys = Object.keys(self.channel.subChannels);
+    for (var i = 0; i < subChanKeys; i++) {
+        var subChan = self.channel.subChannels[subChanKeys[i]];
+        if (subChan.peers.get(hostPort)) {
+            anyOtherSubChan = true;
+            break;
+        }
+    }
+
+    if (!anyOtherSubChan) {
```
anson627: so if there is any other sub channel, we don't do anything, meaning traffic will still go through. my original intention is to stop forwarding once we de-advertise.

jcorbin: good point... guess what: the original never did that either, upcoming PR to fix... if only there was a unit test for this module ;-)

PR #1230, opened 2015-08-26
ServiceDispatchHandler: restore removing of service peers

jcorbin:
Out of #1229


kriskowal:
:ok:

anson627:
we do have a unit test:
https://github.com/uber/tchannel/blob/master/node/test/hyperbahn/unadvertise.js
but I guess the test cases are not well round to capture these cases.

PR #1233, opened 2015-08-27
go: Add TraceReporter which reports TCollector

junchaowu:
@prashantv  add zipkin trace submit


golang/channel.go
```diff
@@ -140,10 +142,24 @@ func NewChannel(serviceName string, opts *ChannelOptions) (*Channel, error) {
 		connectionOptions: opts.DefaultConnectionOptions,
 		log:               logger.WithFields(LogField{"service", serviceName}),
 		statsReporter:     statsReporter,
-		traceReporter:     traceReporter,
 		handlers:          &handlerMap{},
 		subChannels:       &subChannelMap{},
 	}
+
+	var traceReporter TraceReporter
+
```
prashantv: nit: avoid too many blank lines, and avoid branching where possible. A more idiomatic way to write this would be:
```go
traceReporter := opts.TraceReporter
if opts.TraceReporterFactory != nil {
  traceReporter = opts.TraceReporterFactory(ch)
}
if traceReporter == nil {
  traceReporter = NullReporter
}
ch.SetTraceReporter(traceReporter)
```

golang/channel.go
```diff
@@ -157,6 +173,11 @@ func NewChannel(serviceName string, opts *ChannelOptions) (*Channel, error) {
 	return ch, nil
 }
 
+// SetTraceReporter sets TraceReporter to the Channel instance.
```
prashantv: This shouldn't be exported. In fact, I'd avoid a method for this, and set `ch.traceReporter` directly in the constructor. This can only be set during construction - if this is called outside of construction, it would end up in a data race (since there may be code reading traceReporter but someone could be trying to write to it at the same time).

golang/messages.go
```diff
@@ -152,7 +152,7 @@ const (
 	FailureDomain TransportHeaderName = "fd"
 
 	// ShardKey header value is used by ringpop to deliver calls to a specific tchannel instance.
-	ShardKey TransportHeaderName = "sk"
+	ShardKey TransportHeaderName = "shardKey"
```
prashantv: The shard key transport header key is `sk` according to the protocol docs,
https://github.com/uber/tchannel/blob/master/docs/protocol.md#user-content-transport-headers

Please revert this change.

golang/trace/zipkin_tracereporter.go
```diff
@@ -0,0 +1,100 @@
...
+		SetShardKey(Base64Encode(span.TraceID())).Build()
+	defer cancel()
+
+	thriftSpan := BuildZipkinSpan(span, annotations, binaryAnnotations, name, endpoint)
+	// client submit
+	client.Submit(ctx, thriftSpan)
+}
+
+// BuildZipkinSpan builds zipkin span based on tchannel span.
+func BuildZipkinSpan(span tc.Span, annotations []tc.Annotation, binaryAnnotations []tc.BinaryAnnotation, name string, endpoint *tc.Endpoint) *tcollector.Span {
```
prashantv: Does this method need to be exported? Your `_test.go` file can access unexported methods from the same package, so you could unexport this but still use it and test it.

golang/trace/zipkin_tracereporter.go
```diff
@@ -0,0 +1,100 @@
...
+	"github.com/uber/tchannel/golang/trace/thrift/gen-go/tcollector"
+)
+
+// ZipkinTraceReporter is a trace reporter that submits trace spans in to zipkin trace server.
+type ZipkinTraceReporter struct {
+	tchannel *tc.Channel
+}
+
+// Report method will submit trace span to tcollector server.
+func (r *ZipkinTraceReporter) Report(
```
prashantv: Should spans be reported synchronously? Do we need to block our tchannel call till tcollector gets the span?

(We can change this in a separate change, I just want to make sure we're thinking about how this could be done in the background)

golang/trace/zipkin_tracereporter.go
```diff
@@ -0,0 +1,100 @@
+package trace
```
prashantv: Add package documentation in the form
```go
// Package trace [...]
```

golang/trace/zipkin_tracereporter.go
prashantv: Same for all the methods below

golang/trace/zipkin_tracereporter.go
```diff
@@ -0,0 +1,100 @@
...
+			Value:     (string)(ann.Key),
+		}
+	}
+	return zipkinAnns
+}
+
+// InetAton converts string Ipv4 to uint32
+func InetAton(ip string) uint32 {
+	ipByte := net.ParseIP(ip).To4()
+	var ipInt uint32
```
prashantv: This can be simplified by using the inbuilt `binary` package:
```go
func InetAton(ip string) uint32 {
	ipBytes := net.ParseIP(ip).To4()
	return binary.BigEndian.Uint32(ipBytes)
}
```

golang/trace/zipkin_tracereporter.go
```diff
@@ -0,0 +1,100 @@
...
+// ZipkinTraceReporter is a trace reporter that submits trace spans in to zipkin trace server.
+type ZipkinTraceReporter struct {
+	tchannel *tc.Channel
+}
+
+// Report method will submit trace span to tcollector server.
+func (r *ZipkinTraceReporter) Report(
+	span tc.Span, annotations []tc.Annotation, binaryAnnotations []tc.BinaryAnnotation, name string, endpoint *tc.Endpoint) {
+
+	thriftClient := thrift.NewClient(r.tchannel, "tcollector", nil)
```
prashantv: You don't want to create a NewClient for every single Report. You should create one on creation and use the same client throughout.

golang/tracereporter.go
```diff
@@ -64,7 +64,7 @@ func NewBinaryAnnotation(key string, value interface{}) BinaryAnnotation {
 
 // TraceReporter is the interface used to report Trace spans.
 type TraceReporter interface {
-	Report(span Span, annotations []Annotation, binaryAnnotations []BinaryAnnotation)
+	Report(span Span, annotations []Annotation, binaryAnnotations []BinaryAnnotation, name string, endpoint *Endpoint)
```
prashantv: We should be much more explicit about what these new parameters are - let's do a separate diff with just the interface change. We need to name / type them better, `name` and `endpoint` are too vague.

golang/trace/zipkin_tracereporter.go
junchaowu: Let's sync more on this one.

golang/tracing.go
```diff
@@ -34,7 +34,14 @@ var (
 	traceRng = NewRand(time.Now().UnixNano())
 )
 
-// Span represents Zipkin-style span
+// Endpoint represents Zipkin-style endpoint.
```
prashantv: let's also remove this till since it's unused right now.

golang/channel.go
```diff
@@ -140,10 +142,19 @@ func NewChannel(serviceName string, opts *ChannelOptions) (*Channel, error) {
 		connectionOptions: opts.DefaultConnectionOptions,
 		log:               logger.WithFields(LogField{"service", serviceName}),
 		statsReporter:     statsReporter,
-		traceReporter:     traceReporter,
 		handlers:          &handlerMap{},
 		subChannels:       &subChannelMap{},
 	}
+
+	traceReporter := opts.TraceReporter
+	if opts.TraceReporterFactory != nil {
```
prashantv: can you create a test to make sure that `TraceReporterFactory` is preferred over `TraceReporter`

golang/trace/zipkin_tracereporter.go
```diff
@@ -0,0 +1,116 @@
...
+	return binary.BigEndian.Uint32(ipBytes)
+}
+
+// base64Encode encodes uint64 with base64 StdEncoding.
+func base64Encode(data uint64) string {
+	return base64.StdEncoding.EncodeToString(uInt64ToBytes(data))
+}
+
+// UInt64ToBytes converts uint64 to bytes.
+func uInt64ToBytes(i uint64) []byte {
```
prashantv: This should be called `uint64ToBytes` (and make sure the comment is updated to reflect the new name)

golang/trace/zipkin_tracereporter.go
```diff
@@ -0,0 +1,116 @@
...
+	var buf = make([]byte, 8)
```
prashantv: you can use the short variable declaration form: `buf := make([]byte, 8)`

golang/trace/zipkin_tracereporter.go
```diff
@@ -0,0 +1,116 @@
...
+
+// ZipkinTraceReporter is a trace reporter that submits trace spans in to zipkin trace server.
+type ZipkinTraceReporter struct {
+	tchannel *tc.Channel
+	client   tcollector.TChanTCollector
+}
+
+// NewZipkinTraceReporter returns a zipkin trace reporter that submits span to tcollector service.
+func NewZipkinTraceReporter(ch *tc.Channel) *ZipkinTraceReporter {
+	thriftClient := thrift.NewClient(ch, "tcollector", nil)
```
prashantv: let's declare a const at the top for tcollector:
```go
const tcollectorServiceName = "tcollector"
```

golang/trace/zipin_tracereporter_test.go
```diff
@@ -0,0 +1,190 @@
...
+		span := *tchannel.NewRootSpan()
+		name := "test"
+		annotations := RandomAnnotations()
+		thriftSpan := buildZipkinSpan(span, annotations, nil, name, host)
+		thriftSpan.BinaryAnnotations = []*gen.BinaryAnnotation{}
+		ret := &gen.Response{Ok: true}
+		args.s.On("Submit", ctxArg(), thriftSpan).Return(ret, nil)
+		got, err := args.c.Submit(ctx, thriftSpan)
+		require.NoError(t, err)
+		assert.Equal(t, ret, got)
```
prashantv: add a string about what failed, so the above 2 lines would be something like:
```go
require.NoError(t, err, "Submit failed")
assert.Equal(t, ret, got, "Submit response mismatch")
```

golang/trace/zipin_tracereporter_test.go
```diff
@@ -0,0 +1,190 @@
...
+			Port:        8888,
+			ServiceName: "test",
+		},
+		Name:        name,
+		Id:          uInt64ToBytes(span.SpanID()),
+		ParentId:    uInt64ToBytes(span.ParentID()),
+		Annotations: buildZipkinAnnotations(annotations),
+		Debug:       false,
+	}
+	reflect.DeepEqual(thriftSpan, expectedSpan)
```
prashantv: reflect.DeepEqual doesn't do any tests - it just returns a bool. You want:
```go
assert.Equal(t, expectedSpan, thriftSpan, "Span mismatch")
```

golang/trace/zipin_tracereporter_test.go
```diff
@@ -0,0 +1,190 @@
...
+}
+
+func TestInetAton(t *testing.T) {
+	assert.Equal(t, inetAton("1.0.0.1"), (uint32)(16777217))
```
prashantv: You don't need the parens around `uint32`. Also you should use a value like "1.2.3.4" since you want to make sure that the endianness being used is correct.
```go
assert.Equal(t, inetAton("1.2.3.4"), uint32(16909060))
```

golang/trace/zipin_tracereporter_test.go
```diff
@@ -0,0 +1,190 @@
...
+func TestUInt64ToBytes(t *testing.T) {
+	assert.Equal(t, uInt64ToBytes(54613478251749257), []byte("\x00\xc2\x06\xabK$\xdf\x89"))
+}
+
+func TestBase64Encode(t *testing.T) {
+	assert.Equal(t, base64Encode(12711515087145684), "AC0pDj1TitQ=")
+}
+
+func RandomAnnotations() []tchannel.Annotation {
+	anns := make([]tchannel.Annotation, 4)
```
prashantv: You can use an array initializer. You should also use different times for the different annotations (better coverage):
```go
baseTime := time.Date(2015, 1, 2, 3, 4, 5, 6, time.UTC)
return []tchannel.Annotation{
  {
    Key: tchannel.AnnotationKeyClientReceive,
    Timestamp: baseTime,
  },
  {
    Key: tchannel.AnnotationKeyClientSend,
    Timestamp: baseTime.Add(time.Second),
  },
  {
    Key: tchannel.AnnotationKeyServerReceive,
    Timestamp: baseTime.Add(2 * time.Second),
  },
[...]
}
```

golang/trace/zipin_tracereporter_test.go
```diff
@@ -0,0 +1,190 @@
...
+		Key:       tchannel.AnnotationKeyServerSend,
+		Timestamp: time.Now(),
+	}
+
+	return anns
+}
+
+type testArgs struct {
+	s *mocks.TChanTCollector
+	c gen.TChanTCollector // ?
```
prashantv: remove comment at the end

golang/trace/zipin_tracereporter_test.go
```diff
@@ -0,0 +1,190 @@
...
+	}
+
+	server := thrift.NewServer(tchan)
+	server.Register(gen.NewTChanTCollectorServer(h))
+
+	tchan.Serve(listener)
+	return tchan, listener, nil
+}
+
+func getClient(dst string) (gen.TChanTCollector, error) {
```
prashantv: You don't want to use the Thrift client here - you're just testing the Thrift logic instead of you ZipkinTraceReporter. You need to make your client a ZipkinTraceReporter that you call Submit on and make sure it's calling the server with the expected arguments.

golang/tracing.go
junchaowu: keep for now.

golang/trace/zipin_tracereporter_test.go
junchaowu: I'd like to use TChannel so it will test whether it can serialize into wire and then deserialize back. In python, we have seem errors from that.

prashantv: If you use ZipkinTraceReporter, you'll test both - right now, there's no validation that Submit actually sends the span to the tcollector service. You should use `NewZipkinTraceReporter(ch)` so it'll still report using Thrift serialization.

golang/tracereporter.go
```diff
@@ -64,7 +64,7 @@ func NewBinaryAnnotation(key string, value interface{}) BinaryAnnotation {
 
 // TraceReporter is the interface used to report Trace spans.
 type TraceReporter interface {
-	Report(span Span, annotations []Annotation, binaryAnnotations []BinaryAnnotation)
+	Report(span Span, annotations []Annotation, binaryAnnotations []BinaryAnnotation) (bool, *error)
```
prashantv: Document the return values. Name them (purely for documentation) and add a method doc to the interface with what the `bool` is.

prashantv: Actually `*error` is a weird thing to return - pretty much all go code returns `error`.

golang/trace/zipin_tracereporter_test.go
```diff
@@ -0,0 +1,187 @@
...
+			Port:        8888,
+			ServiceName: "test",
+		}
+		span := *tchannel.NewRootSpan()
+		name := "test"
+		annotations := RandomAnnotations()
+		thriftSpan := buildZipkinSpan(span, annotations, nil, name, host)
+		thriftSpan.BinaryAnnotations = []*gen.BinaryAnnotation{}
+		ret := &gen.Response{Ok: true}
+		args.s.On("Submit", ctxArg(), thriftSpan).Return(ret, nil)
```
prashantv: nit: probably add a blank line above this to separate the data / setup from the actual call and verification

prashantv:
lgtm once the interface has documentation and returns `error` instead of `*error`. You almost never want to use pointers to interface (and `error` is an interface).

golang/trace/zipkin_tracereporter.go
```diff
@@ -0,0 +1,121 @@
...
+func (r *ZipkinTraceReporter) Report(
+	span tc.Span, annotations []tc.Annotation, binaryAnnotations []tc.BinaryAnnotation) error {
+	ctx, cancel := tc.NewContextBuilder(time.Second).
+		SetShardKey(base64Encode(span.TraceID())).Build()
+	defer cancel()
+
+	// FIXME remove this dummy endpoint.
+	endpoint := &tc.Endpoint{Ipv4: "127.0.0.1", Port: 8888, ServiceName: "test"}
+	thriftSpan := buildZipkinSpan(span, annotations, binaryAnnotations, "test", endpoint)
+	// client submit
```
prashantv: remove comment (since the code is pretty clear)

PR #1234, opened 2015-08-27
Hyperbahn: improve handleRelay path

jcorbin:
- log endpoint
- simpler logic
- fix argument ordering: callback goes last

r @kriskowal


anson627:
:ok:

kriskowal:
:ok:

PR #1237, opened 2015-08-28
Fix late initialization of circuits

kriskowal:
Previously, initializing circuits by flipping the circuit breaker on remotely would result in an uncaught exception since the circuits object would not have been initialized.

r @jcorbin

### DIFF
--- a/node/hyperbahn/service_proxy.js
+++ b/node/hyperbahn/service_proxy.js
@@ -51,14 +51,13 @@ function ServiceDispatchHandler(options) {
     self.permissionsCache = options.permissionsCache;
     self.serviceReqDefaults = options.serviceReqDefaults || {};
 
-    self.circuitsEnabled = options.circuitsConfig && options.circuitsConfig.enabled || false;
-    self.circuits = self.circuitsEnabled ? new Circuits({
-        timeHeap: self.channel.timeHeap,
-        timers: self.channel.timers,
-        random: self.random,
-        egressNodes: self.egressNodes,
-        config: options.circuitsConfig
-    }) : null;
+    self.circuitsEnabled = false;
+    self.circuitsConfig = options.circuitsConfig;
+    self.circuits = null;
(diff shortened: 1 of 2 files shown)


node/hyperbahn/service_proxy.js
```diff
@@ -548,6 +539,25 @@ function destroy() {
...
+        timeHeap: self.channel.timeHeap,
+        timers: self.channel.timers,
+        random: self.random,
+        egressNodes: self.egressNodes,
+        config: self.circuitsConfig
+    });
+
+    self.circuits.circuitStateChangeEvent.on(onCircuitStateChange);
+
+    function onCircuitStateChange(stateChange) {
```
anson627: why bother to have a closure? can we do self.circuits.circuitStateChangeEvent.on(self.onCircuitStateChange); ?

kriskowal: The closure is necessary to ensure that `this` within `self.onCircuitStateChange` is bound to `self` (the dispatch handler). The same can be accomplished with `self.onCircuitStateChange.bind(self)`, but is a known performance gotcha.

jcorbin: let's move the closure allocation back to the constructer, using a `self.boundOnCircuitsChange = ...;` deal

jcorbin:
lgtm, better than what's on master now; would like to see the closures kept down.

Do you plan to address the (zero vs remote) config TODO in this PR or a future one?

kriskowal:
@jcorbin Not going to address the zero vs remote comment. Will revise. zeroconfig is not in the picture, but there is a static-config (an @raynos original) from which we get the initial circuitsConfig. It looks to me like remoteConfig will emit an update event and call through to enableCircuits when it gets the *initial* remote configuration. So there’s no bug there.

jcorbin:
lgtm still, 99/100, for :100: don't churn the closure function

PR #1238, opened 2015-08-28
Better req/res logging

jcorbin:
- more consistent log fields
- don't log about response timeout errors on already timed out requests


kriskowal:
Hi, I’m @kriskowal and I :approve: this change.

anson627:
:+1: for fixing multiple timeout errors.

PR #1239, opened 2015-08-28
Add lazy handling support to frame handler and connection

jcorbin:
r @kriskowal @Raynos

### DIFF
--- a/node/test/lazy_conn_handler.js
+++ b/node/test/lazy_conn_handler.js
@@ -1,0 +1,111 @@
+// Copyright (c) 2015 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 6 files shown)


node/test/lazy_conn_handler.js
```diff
@@ -0,0 +1,111 @@
...
+        }
+        assert.equal(res.value, 'bob', 'expected called service name');
+
+        res = frame.bodyRW.lazy.readArg1(frame);
+        if (res.err) {
+            throw res.err;
+        }
+        assert.deepEqual(res.value, Buffer('such'), 'expected called arg1');
+
+        this.sendCallResponseBody(frame.id, new v2.CallResponse(
```
kriskowal: this/self. Also, don’t see an implementation of this method here or on master.

kriskowal:
Apart from the missing method revealed by Travis, this looks like progress to me—looking forward to the next step.

kriskowal:
:ok:

PR #1241, opened 2015-08-28
RelayRequest: fix log augmentation

jcorbin:
r @kriskowal


kriskowal:
This looks like a good idea. We should do this.

PR #1243, opened 2015-08-29
Improve errors mapping between tchannel and http

anson627:
based on:
https://github.com/uber/tchannel/blob/master/docs/error-mapping-tchannel-http.md
r: @jcorbin @kriskowal @ShanniLi 
cc: @davewhat @vipulaneja

### DIFF
--- a/node/as/http.js
+++ b/node/as/http.js
@@ -229,6 +229,7 @@ TChannelHTTP.prototype.forwardToTChannel = function forwardToTChannel(tchannel, 
 };
 
 TChannelHTTP.prototype._sendHTTPError = function _sendHTTPError(hres, error) {
+    hres.setHeader('Content-Type', 'text/plain');
     var codeName = errors.classify(error);
     switch (codeName) {
         case 'Cancelled':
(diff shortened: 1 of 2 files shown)


node/as/http.js
```diff
@@ -240,6 +228,37 @@ TChannelHTTP.prototype.forwardToTChannel = function forwardToTChannel(tchannel,
...
+        case 'ProtocolError':
+            hres.writeHead(500, 'TChannel Protocol Error');
+            break;
+        case 'NetworkError':
+            hres.writeHead(500, 'TChannel Network Error');
+            break;
+        default:
+            hres.writeHead(500, 'Internal Server Error');
+    }
+    hres.end(error.message);
```
jcorbin: you could make this a bit more useful / idiomatic:
```
hres.write(util.format("%s %s\n", codeFromAbove, messFromAbove));
hres.end(util.format("%s: %s\n", err.type, err.message)); // maybe you don't want to expose the type, but it's be nice for transparency.
```

Should set response content type to text/plain (bonus points if you do a text/html version, but maybe that's a bit much ;-) ).

jcorbin:
lgtm, idea for improvement

ShanniLi:
lgtm

node/as/http.js
anson627: 'Content-Type' is set to 'text/plain'.

The error message is good enough for the body, since the header already has status code and status message.

PR #1244, opened 2015-08-29
Support enabling circuits for a single test service

kriskowal:


jcorbin:
lgtm

Raynos:
lgtm

Raynos:
Wait; you cannot make this change in one commit. 

Hyperbahn has a tchannel dependency that is not linked by default.

jcorbin:
subtree will split it just fine, but for least mind bending, yes it should be two PRs until we breakout / recollect all the hyperbahn code.

Raynos:
Agreed; we should get the service proxy into the top level hyperbahn folder and out of tchannel.

There is a blocker on this that we need to rewrite relay-network / hyperbahn-cluster without it.

kriskowal:
Landing the first part.

PR #1245, opened 2015-08-31
Add dual writing of per-worker & per-server stats

Raynos:
This branch adds a DualStatsd change. This is simply
a client that writes to two clients under the hood.

I would expect us to be 10% slower with this
change, however this is worth getting per-worker stats.

I will work on proving the performance of our per-worker
dashboard and once that's done we can cut over to it fully.

In this PR I also cleaned up a bunch of remote-config info
log noice in the tests that was frustrating me.

r: @jcorbin @rf


anson627:
hyperbahn test is failing, concern about 10% slower.

rf:
put in that license otherwise lgtm

Raynos:
@anson627 It's 10% slower in terms of CPU utilization not latency. Should not effect xlate; just effects capacity planning.

PR #1247, opened 2015-08-31
add operation as part of begincall interface.

junchaowu:
@prashantv

### DIFF
--- a/golang/outbound.go
+++ b/golang/outbound.go
@@ -84,12 +84,8 @@ func (c *Connection) beginCall(ctx context.Context, serviceName string, callOpti
 		TimeToLive: timeToLive,
 	}
 	call.statsReporter = c.statsReporter
-	call.createStatsTags(c.commonStatsTags)
+	call.createStatsTags(c.commonStatsTags, callOptions, operation)
 	call.log = c.log.WithFields(LogField{"Out-Call", requestID})
-
-	if callOptions.Format != HTTP {
-		call.commonStatsTags["target-endpoint"] = string(operation)
-	}
 
 	// TODO(mmihic): It'd be nice to do this without an fptr
 	call.messageForFragment = func(initial bool) message {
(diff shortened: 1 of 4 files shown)


golang/peer.go
```diff
@@ -211,7 +211,10 @@ func (p *Peer) BeginCall(ctx context.Context, serviceName string, operationName
 		return nil, err
 	}
 
-	call, err := conn.beginCall(ctx, serviceName, callOptions)
+	if callOptions == nil {
+		callOptions = defaultCallOptions
+	}
+	call, err := conn.beginCall(ctx, serviceName, callOptions, operationName)
```
prashantv: can you delete the check for `callOptions == nil` in `calloptions.go`

golang/outbound.go
```diff
@@ -32,7 +32,7 @@ import (
 const maxOperationSize = 16 * 1024
 
 // beginCall begins an outbound call on the connection
-func (c *Connection) beginCall(ctx context.Context, serviceName string, callOptions *CallOptions) (*OutboundCall, error) {
+func (c *Connection) beginCall(ctx context.Context, serviceName string, callOptions *CallOptions, operation string) (*OutboundCall, error) {
```
prashantv: where are you calling writeOperation in this method?

golang/outbound.go
```diff
@@ -87,6 +87,10 @@ func (c *Connection) beginCall(ctx context.Context, serviceName string, callOpti
 	call.createStatsTags(c.commonStatsTags)
 	call.log = c.log.WithFields(LogField{"Out-Call", requestID})
 
+	if callOptions.Format != HTTP {
+		call.commonStatsTags["target-endpoint"] = string(operation)
```
prashantv: can you pass the `callOptions` and `operation` into createStatsTags and move this commonStatsTag setter into that method

junchaowu: done.

golang/peer.go
junchaowu: deleted

golang/outbound.go
prashantv: don't see this change - is it pushed?

junchaowu: damn. not commit.

golang/outbound.go
```diff
@@ -170,13 +175,18 @@ func (call *OutboundCall) Response() *OutboundCallResponse {
...
 		"target-service": call.callReq.Service,
 	}
 	for k, v := range connectionTags {
 		call.commonStatsTags[k] = v
 	}
+
+	if callOptions.Format != HTTP {
+		call.commonStatsTags["target-endpoint"] = string(operation)
+	}
+
```
prashantv: nit: delete blank, go style normally avoids new lines after a `{` and before a `}`

prashantv:
lgtm

PR #1250, opened 2015-08-31
bump timeouts to 50 to avoid flaps

Raynos:
r: @kriskowal @rf

### DIFF
--- a/hyperbahn/test/admin/rate-limiter.js
+++ b/hyperbahn/test/admin/rate-limiter.js
@@ -59,7 +59,7 @@ allocCluster.test('disable rate limiter and forward', {
 
         bob.clientChannel.request({
             serviceName: 'steve',
-            timeout: 100
+            timeout: 50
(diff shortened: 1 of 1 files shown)


hyperbahn/cut.sh
```diff
@@ -67,5 +67,5 @@ git push origin master --tags
 git push -f origin dev_hyperbahn
 
 git archive --prefix=package/ --format tgz dev_hyperbahn >package.tgz
-${NPM:-npm} publish package.tgz --tag "${NPM_TAG:-alpha}"
+${NPM:-npm} publish package.tgz --tag "${NPM_TAG:-latest}"
```
kriskowal: this will make a sad @jcorbin.

kriskowal: please remove so we can green stamp the rest

Raynos: No. Hyperbahn should always be published as latest.

kriskowal: Separable PR.

PR #1252, opened 2015-09-01
Return error when no service name provided

i:
Empty service name causes issues with hyperbahn.
@prashantv

### DIFF
--- a/golang/channel.go
+++ b/golang/channel.go
@@ -35,6 +35,10 @@ import (
 var (
 	errAlreadyListening  = errors.New("channel already listening")
 	errInvalidStateForOp = errors.New("channel is in an invalid state for that operation")
+
+	// ErrNoServiceName is returned when no service name is provided when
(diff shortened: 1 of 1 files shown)


golang/channel.go
```diff
@@ -35,6 +35,8 @@ import (
 var (
 	errAlreadyListening  = errors.New("channel already listening")
 	errInvalidStateForOp = errors.New("channel is in an invalid state for that operation")
+
+	ErrNoServiceName = errors.New("No service name provided")
```
prashantv: Can you also add a comment since the variable is exported:
```go
// ErrNoServiceName is returned when creating a channel with no service name.
// A service name is required for [...]
```
Also, make error lowercase ("no service name provided") - all go errors are supposed to be lower case.

i:
@prashantv, added the changes.

prashantv:
Thanks @i

PR #1253, opened 2015-09-01
add TargetEndpoint struct to save addition tracing infor

junchaowu:
@prashantv

### DIFF
--- a/golang/inbound.go
+++ b/golang/inbound.go
@@ -79,7 +79,6 @@ func (c *Connection) handleCallReq(frame *Frame) bool {
 	call.AddBinaryAnnotation(BinaryAnnotation{Key: "cn", Value: callReq.Headers[CallerName]})
 	call.AddBinaryAnnotation(BinaryAnnotation{Key: "as", Value: callReq.Headers[ArgScheme]})
 	call.AddAnnotation(AnnotationKeyServerReceive)
-
 	response := new(InboundCallResponse)
 	response.mex = mex
 	response.conn = c
@@ -91,7 +90,12 @@ func (c *Connection) handleCallReq(frame *Frame) bool {
 	response.messageForFragment = func(initial bool) message {
 		if initial {
 			call.AddAnnotation(AnnotationKeyServerSend)
-			call.Report(callReq.Tracing, c.traceReporter)
+			targetEndpoint := TargetEndpoint{
+				HostPort:    c.localPeerInfo.HostPort,
+				ServiceName: call.serviceName,
+				Name:        string(call.Operation()),
+			}
+			call.Report(callReq.Tracing, targetEndpoint, c.traceReporter)
 
 			callRes := new(callRes)
 			callRes.Headers = response.headers
--- a/golang/inbound.go
+++ b/golang/inbound.go
@@ -93,7 +93,7 @@ func (c *Connection) handleCallReq(frame *Frame) bool {
 			targetEndpoint := TargetEndpoint{
 				HostPort:    c.localPeerInfo.HostPort,
 				ServiceName: call.serviceName,
-				Name:        string(call.Operation()),
+				Operation:   string(call.Operation()),
 			}
 			call.Report(callReq.Tracing, targetEndpoint, c.traceReporter)
 
--- a/golang/trace/zipin_tracereporter_test.go
+++ b/golang/trace/zipin_tracereporter_test.go
@@ -44,15 +44,14 @@ func TestZipkinTraceReporterFactory(t *testing.T) {
 }
 
 func TestBuildZipkinSpan(t *testing.T) {
-	host := &tchannel.Endpoint{
-		Ipv4:        "127.0.0.1",
-		Port:        8888,
+	endpoint := tchannel.TargetEndpoint{
+		HostPort:    "127.0.0.1:8888",
 		ServiceName: "test",
+		Name:        "test",
 	}
 	span := *tchannel.NewRootSpan()
-	name := "test"
 	annotations := RandomAnnotations()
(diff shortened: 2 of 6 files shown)


golang/inbound.go
```diff
@@ -91,7 +90,12 @@ func (c *Connection) handleCallReq(frame *Frame) bool {
 	response.messageForFragment = func(initial bool) message {
 		if initial {
 			call.AddAnnotation(AnnotationKeyServerSend)
-			call.Report(callReq.Tracing, c.traceReporter)
+			targetEndpoint := TargetEndpoint{
```
prashantv: to be consistent, let's either create `targetEndpoint` either inside `messageForFragment` or outside of it in both `inbound.go` and `outbound.go`

golang/trace/zipin_tracereporter_test.go
```diff
@@ -44,15 +44,14 @@ func TestZipkinTraceReporterFactory(t *testing.T) {
 }
 
 func TestBuildZipkinSpan(t *testing.T) {
-	host := &tchannel.Endpoint{
-		Ipv4:        "127.0.0.1",
-		Port:        8888,
+	endpoint := tchannel.TargetEndpoint{
+		HostPort:    "127.0.0.1:8888",
 		ServiceName: "test",
```
prashantv: use different values for `ServiceName` and `Name` so that you know you're not accidentally swapping values around.

golang/tracing.go
```diff
@@ -34,11 +34,11 @@ var (
...
 
-// Endpoint represents Zipkin-style endpoint.
-type Endpoint struct {
-	Ipv4        string
-	Port        int32
+// TargetEndpoint represents target server endpoint information.
+type TargetEndpoint struct {
+	HostPort    string
 	ServiceName string
+	Name        string
```
prashantv: Call this `Operation` to be consistent with the `BeginCall` methods

prashantv:
We should also add tests for TraceReporting - either as part of this change or as a separate PR. I can help pair to for those tests.

golang/inbound.go
junchaowu: Not sure what you mean. You probably need to give me a code walkthrough. Otherwise I will have no idea about those things.

golang/tracing.go
junchaowu: I'd do opposite way. Operation doesn't bring proper meaning at all.

golang/inbound.go
prashantv: I just mean that in this diff, you are declaring `targetEndpoint` in 2 places:
1. In this file, right before you call `call.Report` which is inside of `messageForFragment` function.
2. In the `outbound.go` file, you do it outside of `messageForFragment`.

The logic of those 2 functions is similar, and they should look similar, and the variable should be declared in a consistent location. For now, I would suggest that `targetEndpoint` should be created right before you do `call.Report`.

golang/tracing.go
prashantv: `Operation` may not be great, but it's consistent. We should call this `Operation` here, and can do a separate diff with renaming everything. `Name` is too vague, and I want to avoid it. Maybe we can call it `Method`, or `MethodName`. Either way, that should be done as a separate diff, as `Operation` is used all over the place, and I don't want to introduce a new terminology, I'd rather update them all at once.

junchaowu: Yes, that's what I mean change all of them.

golang/inbound.go
junchaowu: oh..

golang/trace/zipin_tracereporter_test.go
junchaowu: done

golang/trace/zipkin_tracereporter.go
```diff
@@ -64,18 +64,20 @@ func (r *ZipkinTraceReporter) Report(
 }
 
 // buildZipkinSpan builds zipkin span based on tchannel span.
-func buildZipkinSpan(span tc.Span, annotations []tc.Annotation, binaryAnnotations []tc.BinaryAnnotation, name string, endpoint *tc.Endpoint) *tcollector.Span {
+func buildZipkinSpan(span tc.Span, annotations []tc.Annotation, binaryAnnotations []tc.BinaryAnnotation, targetEndpoint tc.TargetEndpoint) *tcollector.Span {
```
breerly: This is a constructor? I believe the convention is `NewSpan`

golang/tracing.go
```diff
@@ -34,11 +34,11 @@ var (
 	traceRng = NewRand(time.Now().UnixNano())
 )
 
-// Endpoint represents Zipkin-style endpoint.
-type Endpoint struct {
-	Ipv4        string
-	Port        int32
+// TargetEndpoint represents target server endpoint information.
+type TargetEndpoint struct {
+	HostPort    string
```
breerly: This should be it's own type, else we're going to have these littered everywhere.

breerly: To clarify, i mean pulling `HostPort` into something like:

```golang
type HostPort string

type TargetEndpoint struct {
    Hostport
    // ...
}
```

breerly: Potentially move it to a `glossary.go` so that it can be reused, probably same with `ServiceName`

golang/trace/zipkin_tracereporter.go
prashantv: I like to use `NewTypeName` when `TypeName` is defined by us and it is the constructor. In this case, it's not *the* constructor since it's a Thrift type, so I don't mind the build name here.

prashantv:
lgtm. like @breerly's suggestions for stronger typing, we should do those in separate diffs.

PR #1254, opened 2015-09-01
Fix flappy Hyperbahn connection reset test

kriskowal:
This test flapped due to varying specificity about why the ad timed out.


kriskowal:
r @raynos @jcorbin

Raynos:
lgtm

PR #1255, opened 2015-09-01
Add simulation of weighted average stat gossip

kriskowal:
Clearing out my attic / working copy…

This is a simulation that illustrates that we could some day gossip stats like RPS and pending request count among affine nodes, particularly for improving the fidelity of the upcoming SLA feature, by eliminating the assumption that the cluster has homogenous load and performance.

For a rainy day.


Raynos:
Can we move simulations into `docs/simulations/*.py`

Otherwise lgtm. #shipit

PR #1256, opened 2015-09-01
TChannel: fix subtle edge case for listening sub channels

jcorbin:
Found while working on #1239:
- we have an assumption / invariant that conn.channel is the root channel
- this is guaranteed for outgoing ones, since peer.channel has this invariant, and peers create outgoing connections
- however channels create incoming connections, and weren't making the same guarantee

So for consistency, do this; I'll need this in #1239 since only the root channel will have the cached "should all connections be handled lazily" bool, not any of the sub channels.

r @kriskowal @Raynos

### DIFF
--- a/node/channel.js
+++ b/node/channel.js
@@ -323,7 +323,8 @@ TChannel.prototype.onServerSocketConnection = function onServerSocketConnection(
     }
 
     var socketRemoteAddr = sock.remoteAddress + ':' + sock.remotePort;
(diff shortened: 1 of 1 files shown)


Raynos:
There are probably other bugs associated with listening on sub channels.

This PR looks fine but I would still not recommend listening on sub channels.

jcorbin:
Indeed @raynos, I don't recommend it either, however I do want to be safe against this assumption / invariant in case someone does :-)

PR #1257, opened 2015-09-01
Hyperbahn client should only destroy in HardFail mode

anson627:
When advertisement timeout or other failures happens, hyperbahn client will destroy itself even it's not in HardFail mode. And if advertise is called again, it will trigger AlreadyDestroyed error and return, so hyperbahn client will never be able to get to advertised state again.

r: @ShanniLi @Raynos @vipulaneja

### DIFF
--- a/node/hyperbahn/index.js
+++ b/node/hyperbahn/index.js
@@ -160,6 +160,9 @@ function HyperbahnClient(options) {
         self.advertisementTimeoutTime = advertisementTimeout || 5000;
     } else {
         self.advertisementTimeoutTime = 0;
+        self.logger.info('HyperbahnClient advertisement timeout disabled', {
+            service: self.serviceName
+        });
     }
 }
 
--- a/node/hyperbahn/index.js
+++ b/node/hyperbahn/index.js
@@ -159,7 +159,7 @@ function HyperbahnClient(options) {
     if (self.hardFail) {
         self.advertisementTimeoutTime = advertisementTimeout || 5000;
     } else {
-        self.advertisementTimeoutTime = advertisementTimeout;
+        self.advertisementTimeoutTime = 0;
     }
 }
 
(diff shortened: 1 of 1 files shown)


kriskowal:
:100:

Raynos:
This feels wrong to me. I feel like `advertiseFailure()` should never be called when hardFail is false.

Raynos:
:-1: If you set hardFail: false your supposed to set advertiseTimeout: 0

kriskowal:
@raynos seems like an elaborate dance for a single effect. If that’s the case, perhaps advertiseTimeout: 0 should be implied by hardFail: false.

anson627:
@Raynos what if I don't want hardFail (be able to re-use hyperbahn client), but I still want to know (log/stats) when advertisement is timing out. This is a valid use case.

Raynos:
@anson627 There are logs & stats everywhere already. Both in hyperbahn client and in tchannel.

You dont need this change. advertisementTimeout purely means "if hyperbahn client cannot finish an advertisement even after retries then fail."

Raynos:
@kriskowal advertisementTimeout: undefined is the default value.

ShanniLi:
@Raynos, not for the !hardFail case:
https://github.com/uber/tchannel/blob/hyperbahn-client-timeout/node/hyperbahn/index.js#L159
https://github.com/uber/tchannel/blob/hyperbahn-client-timeout/node/hyperbahn/index.js#L301

anson627:
OK. To avoid confusing users, I would suggest always setting advertisementTimeout to 0 when HardFail equals false.

Raynos:
@anson627 Good idea; let's do it.

ShanniLi:
lgtm

node/hyperbahn/index.js
```diff
@@ -304,10 +304,6 @@ function advertise(opts) {
             },
             self.advertisementTimeoutTime
         );
-    } else {
-        self.logger.info('HyperbahnClient advertisement timeout disabled', {
```
Raynos: Do not remove this log.

anson627: put the log back, but move it to constructor, so it only log once, but not per advertise call.

Raynos: Ah cool :)

Raynos:
lgtm.

PR #1259, opened 2015-09-02
Add non-streaming option for forwardToTChannel in as-http

anson627:
As discussed, leave options - streamed to forwardToTChannel caller instead of force streaming, so that tchannel can retry by turning off request streaming.
r: @Raynos @jcorbin @ShanniLi 
cc: @vipulaneja

### DIFF
--- a/node/as/http.js
+++ b/node/as/http.js
@@ -127,7 +127,7 @@ TChannelHTTP.prototype.sendRequest = function send(treq, hreq, options, callback
         }
         function arg2Ready(err, arg2) {
             if (err) {
-                callback(err, null, null);
+                callback(err, null, null, null);
             } else {
                 readArg2(tres, arg2);
             }
@@ -143,7 +143,7 @@ TChannelHTTP.prototype.sendRequest = function send(treq, hreq, options, callback
             var fromBufferErr = errors.HTTPReqArg2fromoBufferError(arg2res.err, {
                 arg2: arg2
             });
-            callback(fromBufferErr, null, null);
+            callback(fromBufferErr, null, null, null);
         } else if (tres.streamed) {
             callback(null, arg2res.value, tres.arg3, null);
         } else {
--- a/node/as/http.js
+++ b/node/as/http.js
@@ -20,7 +20,6 @@
 
 'use strict';
 
-var assert = require('assert');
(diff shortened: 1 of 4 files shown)


node/as/http.js
```diff
@@ -94,11 +93,32 @@ TChannelHTTP.prototype.sendRequest = function send(treq, hreq, options, callback
...
 
     treq.headers.as = 'http';
-    return treq.sendStreams(arg1, arg2, hreq, onResponse);
+    if (treq.streamed) {
+        return treq.sendStreams(arg1, arg2, hreq, onStreamResponse);
+    }
+    getRawBody(hreq, {
+        length: hreq.headers['content-length'],
+        limit: '1mb'
+      }, onRawBody);
```
jcorbin: please fix indent

node/as/http.js
```diff
@@ -94,11 +93,32 @@ TChannelHTTP.prototype.sendRequest = function send(treq, hreq, options, callback
...
+
+    function onRawBody(err, body) {
+        if (err) {
+            return err;
```
jcorbin: shouldn't you call the callback with the error?

node/package.json
```diff
@@ -33,6 +33,7 @@
     "hexer": "^1.4.5",
     "json-stringify-safe": "^5.0.0",
     "lru-cache": "^2.6.4",
+    "raw-body": "^2.1.2",
```
jcorbin: @Raynos can you please weigh in on the provenance of this new 3rd party dep?

jcorbin:
rebase on latest master to fix the jshint failure

node/package.json
anson627: I actually got the idea of using raw-body from @Raynos

anson627:
will do.

node/as/http.js
anson627: will do

node/package.json
jcorbin: cool, just wanted to check :-)

jcorbin:
lgtm; especially since it's tested

anson627:
@jcorbin now lbpool version tests are exercising non-streaming path (both request and response), while non-lbpool version tests are on the streaming path. I will keep in mind about adding more tests.

PR #1266, opened 2015-09-03
V2 handler cleanup

jcorbin:
r @ShanniLi @kriskowal @Raynos

### DIFF
--- a/node/v2/handler.js
+++ b/node/v2/handler.js
@@ -641,7 +641,10 @@ function sendCallResponseFrame(res, flags, args) {
         return;
     }
 
-    self.validateCallResponseFrame(res);
+    var err = self.validateCallResponseFrame(res);
+    if (err) {
+        return err;
+    }
 
     var code = res.ok ? v2.CallResponse.Codes.OK : v2.CallResponse.Codes.Error;
     var req = res.inreq;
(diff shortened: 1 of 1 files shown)


node/v2/handler.js
```diff
@@ -700,25 +686,15 @@ TChannelV2Handler.prototype.sendCallRequestContFrame = function sendCallRequestC
         self.errorEvent.emit(self, errors.SendCallReqContBeforeIdentifiedError());
         return;
     }
-    var reqBody = new v2.CallRequestCont(flags, req.checksum.type, args);
-    var result = self._sendCallBodies(req.id, reqBody, req.checksum);
-    req.checksum = result.checksum;
 
-    var req0 = self.connection.ops.getOutReq(req.id);
```
jcorbin: @ShanniLi can you please explain why this was added in the first place?

ShanniLi: req is a request continue frame that doesn't have serviceName etc. req0 is the first request sent out that has the info.

ShanniLi:
not sure we can drop the req lookup; otherwise, serviceName, cn, etc will be missing in stats for continuous requests.

node/v2/handler.js
jcorbin: no, it's not a frame object, it's the outgoing request object, and should be identical (same object instance: req === req0).

jcorbin:
Object identical, not just similar properties, but exact same object.

kriskowal:
:ok:

PR #1267, opened 2015-09-03
v2/handler: use res.code directly

jcorbin:
So that we can ever support as modules defining their own non-zero code space.

r @kriskowal @Raynos


kriskowal:
:ok:

PR #1268, opened 2015-09-03
Health check support in TChannelAsThrift

ShanniLi:

### DIFF
--- a/node/as/thrift.js
+++ b/node/as/thrift.js
@@ -21,6 +21,8 @@
 'use strict';
 
 var assert = require('assert');
+var fs = require('fs');
+var path = require('path');
 var bufrw = require('bufrw');
 var Result = require('bufrw/result');
 var thriftify = require('thriftify');
@@ -53,8 +55,29 @@ function TChannelAsThrift(opts) {
     self.logParseFailures = typeof logParseFailures === 'boolean' ?
         logParseFailures : true;
 
-    // only used in request()
     self.channel = opts.channel;
+
+    self.healthCheckCallback = opts.healthCheckCallback;
+    assert(!self.healthCheckCallback || typeof self.healthCheckCallback === 'function',
+        'healthCheckCallback must be a function');
+    assert(!self.healthCheckCallback || self.channel,
+        'channel must be provided with healthCheckCallback');
+
+    if (self.healthCheckCallback) {
+        fs.readFile(path.join(__dirname, 'meta.thrift'), 'utf8', registerHealthCheck);
+    }
+
+    function registerHealthCheck(err, source) {
+        if (err) {
+            self.channel.logger.error('failed to read meta.thrift file', {
+                error: err
+            });
+            return;
+        }
+
+        var metaSpec = thriftify.parseSpec(source);
+        self.register(self.channel, 'Meta::health', self, health, metaSpec);
+    }
 }
 
 function TChannelThriftRequest(options) {
@@ -63,6 +86,17 @@ function TChannelThriftRequest(options) {
     self.channel = options.channel;
     self.reqOptions = options.reqOptions;
     self.tchannelThrift = options.tchannelThrift;
+}
+
+function health(self, req, head, body, callback) {
+    var status = self.healthCheckCallback();
+    return callback(null, {
+        ok: true,
+        body: {
+            ok: status.ok,
+            message: status.message
+        }
+    });
 }
 
 TChannelThriftRequest.prototype.send =
@@ -88,7 +122,7 @@ TChannelAsThrift.prototype.request = function request(reqOptions) {
 };
 
 TChannelAsThrift.prototype.register =
-function register(channel, name, opts, handle) {
+function register(channel, name, opts, handle, spec) {
     var self = this;
 
     if (!self.logger) {
@@ -110,7 +144,8 @@ function register(channel, name, opts, handle) {
             head: inHeadBuffer,
             body: inBodyBuffer,
             endpoint: name,
-            direction: 'in.request'
+            direction: 'in.request',
+            spec: spec
         });
 
         if (parseResult.err) {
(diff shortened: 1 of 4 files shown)


node/as/thrift.js
```diff
@@ -53,8 +55,29 @@ function TChannelAsThrift(opts) {
     self.logParseFailures = typeof logParseFailures === 'boolean' ?
         logParseFailures : true;
 
-    // only used in request()
     self.channel = opts.channel;
+
+    self.healthCheckCallback = opts.healthCheckCallback;
+    assert(!self.healthCheckCallback || typeof self.healthCheckCallback === 'function',
```
Raynos: healthCheckCallback must be optional for back compat.

node/as/thrift.js
```diff
@@ -53,8 +55,29 @@ function TChannelAsThrift(opts) {
     self.logParseFailures = typeof logParseFailures === 'boolean' ?
         logParseFailures : true;
 
-    // only used in request()
     self.channel = opts.channel;
+
+    self.healthCheckCallback = opts.healthCheckCallback;
```
Raynos: Let's call this `self.isHealthy = opts.isHealthy`. This is not a callback so cant be called Callback.

node/as/thrift.js
Raynos: Oh I see; got it, cool :)

node/as/thrift.js
```diff
@@ -65,6 +88,17 @@ function TChannelThriftRequest(options) {
     self.tchannelThrift = options.tchannelThrift;
 }
 
+function health(self, req, head, body, callback) {
+    var status = self.healthCheckCallback();
```
Raynos: Let's add `assert('ok' in status, 'Status must have ok field')` and `assert('message' in status, 'Status must have message field')`

Raynos:
Please also update the docs folder.

ShanniLi:
@Raynos confused. which doc file?

node/as/thrift.js
ShanniLi: Will add ok. However, message is optional in thrift.

node/as/thrift.js
```diff
@@ -65,6 +88,17 @@ function TChannelThriftRequest(options) {
...
+    return callback(null, {
```
anson627: return sth return by a callback passed in by caller? what's the use case here?

node/as/thrift.js
Raynos: Good point; `assert(!status.ok || 'message' in status)`

Raynos:
@ShanniLi docs/thrift.md

node/as/thrift.js
ShanniLi: used to provide a health check wrapper for your service, so your service don't have to worry about meta.thrift, etc.

anson627:
rebase master to pass Jenkins? otherwise lgtm

node/as/thrift.js
anson627: I see. The doc make it clear.

node/as/thrift.js
```diff
@@ -65,6 +88,21 @@ function TChannelThriftRequest(options) {
     self.tchannelThrift = options.tchannelThrift;
 }
 
+function health(self, req, head, body, callback) {
+    var status = self.healthCheckCallback();
+    assert(status && typeof status.ok === 'boolean', 'status must have ok field');
+    assert(status && (!status.message || typeof status.message === 'string'),
```
Raynos: if status.ok === false then there must be a message.

node/docs/as-thrift.md
```diff
@@ -91,7 +91,13 @@ It can be passed options.
     `logger.warn()`. If you do not want these log statements you
     can set the option to `false`
  - `opts.channel` The channel used to make requests on. This
-    option is required if you want to use the `.request()` method
+    option is required if you want to use the `.request()` method or
+    `opts.healthCheckCallback` is passed in.
+ - `opts.healthCheckCallback` The callback function used to return the
```
Raynos: Let's pick a better name. This is not a callback. Consider `isHealthy`

node/test/health.js
```diff
@@ -0,0 +1,170 @@
...
+
+    var client = cluster.channels[1].subChannels.server;
+
+    tchannelAsThrift.send(client.request({
+        serviceName: 'server',
+        hasNoParent: true
+    }), 'Chamber::echo', null, {
+        value: 10
+    }, function onResponse(err, res) {
+        assert.ifError(err);
```
kriskowal: please use the `if (err) return assert.end(err)` pattern. Failing tests should not crash the test runner.

node/test/health.js
```diff
@@ -0,0 +1,170 @@
...
+        source: source,
+        logParseFailures: false
+    });
+
+    healthThrift.send(client.request({
+        serviceName: 'server',
+        hasNoParent: true
+    }), 'Meta::health', null, {}, function onResponse(err, res) {
+        if (err) {
+            assert.end(false);
```
kriskowal: `assert.end(err)`.

node/test/health.js
```diff
@@ -0,0 +1,170 @@
...
+    });
+
+    var options = {
+        isOptions: true
+    };
+
+    var health = opts.good ? healthGood : healthBad;
+
+    var tchannelAsThrift = new TChannelAsThrift({
+        source: opts.thriftText || globalThriftText,
```
kriskowal: Please name this `thriftSource` or simply `source`. Don’t mix `text` and `source` for the same semantics.

node/as/thrift.js
```diff
@@ -53,8 +55,29 @@ function TChannelAsThrift(opts) {
...
         logParseFailures : true;
 
-    // only used in request()
     self.channel = opts.channel;
+
+    self.healthCheckCallback = opts.healthCheckCallback;
+    assert(!self.healthCheckCallback || typeof self.healthCheckCallback === 'function',
+        'healthCheckCallback must be a function');
+    assert(!self.healthCheckCallback || self.channel,
+        'channel must be provided with healthCheckCallback');
```
kriskowal: Let’s be sure to note that this is a backward incompatible change in the migration docs, and make a major version bump, or not make a breaking change.

ShanniLi: it is actually backward compatible

Raynos:
lgtm.

PR #1270, opened 2015-09-03
Graceful call draining

jcorbin:
Implements graceful draining support for both incoming and outgoing calls.

### DIFF
--- a/node/channel.js
+++ b/node/channel.js
@@ -50,6 +50,7 @@ var TChannelServices = require('./services');
 var TChannelStatsd = require('./lib/statsd');
 var RetryFlags = require('./retry-flags.js');
 var TimeHeap = require('./time_heap');
+var CountedReadySignal = require('ready-signal/counted');
 
 var TracingAgent = require('./trace/agent');
 
@@ -220,6 +221,11 @@ function TChannel(options) {
     self.listened = false;
     self.listening = false;
     self.destroyed = false;
+    self.draining = false;
+
+    // set when draining (e.g. graceful shutdown)
+    self.drainReason = '';
+    self.drainExempt = null;
 
     var trace = typeof self.options.trace === 'boolean' ?
         self.options.trace : true;
@@ -300,6 +306,41 @@ TChannel.prototype.setLazyHandling = function setLazyHandling(enabled) {
     }
 };
 
+TChannel.prototype.drain = function drain(reason, exempt, callback) {
+    var self = this;
+
+    // TODO: we could do this by defaulting and/or forcing you into an
+    // exemption function that exempting anything not matching the given sub
+    // channel's service name; however there are many other complications to
+    // consider to implement sub channel draining, so for now:
+    assert(!self.topChannel, 'sub channel draining not supported');
+    assert(!self.draining, 'channel already draining');
+
+    if (callback === undefined) {
+        callback = exempt;
+        exempt = null;
+    }
+
+    self.draining = true;
+    self.drainReason = reason;
+    self.drainExempt = exempt;
+
+    var drained = CountedReadySignal(1);
+    drained(callback);
+    self.eachConnection(drainEachConn);
+    process.nextTick(drained.signal);
+    self.logger.info('draining channel', {
+        hostPort: self.hostPort,
+        reason: self.drainReason,
+        count: drained.counter
+    });
+
+    function drainEachConn(conn) {
+        drained.counter++;
+        conn.drain(self.drainReason, self.drainExempt, drained.signal);
+    }
+};
+
 TChannel.prototype.getServer = function getServer() {
     var self = this;
     if (self.serverSocket) {
@@ -362,6 +403,10 @@ TChannel.prototype.onServerSocketConnection = function onServerSocketConnection(
     var socketRemoteAddr = sock.remoteAddress + ':' + sock.remotePort;
     var chan = self.topChannel || self;
     var conn = new TChannelConnection(chan, sock, 'in', socketRemoteAddr);
+
+    if (self.draining) {
+        conn.drain(self.drainReason, self.drainExempt, null);
+    }
 
     conn.errorEvent.on(onConnectionError);
 
--- a/node/connection_base.js
+++ b/node/connection_base.js
@@ -126,6 +126,12 @@ function connBaseRequest(options) {
     // options.checksumType = options.checksum;
 
     var req = self.buildOutRequest(options);
+    if (self.draining && (
+            !self.drainExempt || !self.drainExempt(req)
+        )) {
+        req.drained = true;
+        req.drainReason = self.drainReason;
+    }
     self.ops.addOutReq(req);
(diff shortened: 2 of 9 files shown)


jcorbin:
need to sort out excepting certain services, e.g. ringpop

jcorbin:
Wrote initial test, covers the server case, should finish out test for client case too.  However the test is currently provoking "mismatched onReqDone callback" warn logs, unsure if pre-existing condition, or bug in new code.

node/channel.js
```diff
@@ -300,6 +306,41 @@ TChannel.prototype.setLazyHandling = function setLazyHandling(enabled) {
...
 
+TChannel.prototype.drain = function drain(reason, exempt, callback) {
+    var self = this;
+
+    // TODO: we could do this by defaulting and/or forcing you into an
+    // exemption function that exempting anything not matching the given sub
+    // channel's service name; however there are many other complications to
+    // consider to implement sub channel draining, so for now:
+    assert(!self.topChannel, 'sub channel draining not supported');
+    assert(!self.draining, 'channel already draining');
```
anson627: return if self.draining?

node/connection_base.js
```diff
@@ -72,6 +76,41 @@ TChannelConnectionBase.prototype.setLazyHandling = function setLazyHandling() {
...
+        if (self.ops.hasDrained()) {
+            process.nextTick(callback);
+        } else {
+            self.ops.drainEvent.on(callback);
+        }
+    }
+};
+
+TChannelConnectionBase.prototype.undrain =
+function undrain() {
```
anson627: what's the purpose of undrain? is connection re-usable after draining?

Raynos:
After we merge this and before we move to production.

Let's add `drain()` to the HyperbahnCluster so that all our end to end tests drain instead of terminating connections. We can also audit the Hyperbahn tests for any "extra" setTimeout before assert.end() that were added because we didnt have drain().

node/connection_base.js
```diff
@@ -99,6 +144,14 @@ TChannelConnectionBase.prototype.handleCallRequest = function handleCallRequest(
     req.remoteAddr = self.remoteName;
     self.ops.addInReq(req);
 
+    if (self.draining && (
```
Raynos: This is in the wrong place ? We add the inReq and then check draining.

Let's avoid adding the InReq.

node/operations.js
```diff
@@ -208,6 +216,46 @@ Operations.prototype.addInReq = function addInReq(req) {
...
+    }
+};
+
+Operations.prototype._isCollDrained = function _isCollDrained(coll) {
+    var self = this;
+
+    /* jshint forin:false */
+    for (var id in coll) {
+        var op = coll[id];
+        if (!(op instanceof OperationTombstone) &&
```
Raynos: Let's check `op.drained` as well.

It's possible to add an outgoing request in the "drained" state to this operation.

node/operations.js
```diff
@@ -208,6 +216,46 @@ Operations.prototype.addInReq = function addInReq(req) {
     return req;
 };
 
+Operations.prototype.hasDrained = function hasDrained() {
+    var self = this;
+
+    if (self.pending.in === 0 &&
+        self.pending.out === 0) {
+        return true;
+    } else if (self.drainExempt &&
```
Raynos: We want to check _isCollDrained even if no drainExempt because `outReq.drained`

node/test/chan_drain.js
```diff
@@ -0,0 +1,374 @@
...
+    function runTest(err, gotClients) {
+        if (err) {
+            finish(err);
+            return;
+        }
+        clients = gotClients;
+
+        finishCount = 2;
+        assert.timeoutAfter(50);
+        collectParallel(clients.a, sendOne, sendsDone);
```
Raynos: sendsDone takes a bunch of arguments.

node/test/chan_drain.js
```diff
@@ -0,0 +1,374 @@
...
+            serviceName: service,
+            headers: {
+                as: 'raw',
+                cn: service + 'Client'
+            }
+        }
+    });
+}
+
+function checkSendsDone(name, check, callback) {
```
Raynos: nitpick: Not a fan of this indirection.

node/test/chan_drain.js
```diff
@@ -0,0 +1,374 @@
...
+        }
+        for (var i = 0; i < res.length; i++) {
+            var desc = name + ' res[' + i + ']: ';
+            check(desc, res[i], i);
+        }
+        callback();
+    }
+}
+
+function checkFinish(assert, err, cluster, finishCount) {
```
Raynos: This might be nicer as `assert.once('end', checkFinish);` that just does the assertCleanState() stuff.

node/test/chan_drain.js
```diff
@@ -0,0 +1,374 @@
...
+            finish(err);
+            return;
+        }
+        clients = gotClients;
+
+        finishCount = 2;
+        assert.timeoutAfter(50);
+        collectParallel(clients.a, sendOne, sendsDone);
+
+        setTimeout(testdown, 1);
```
Raynos: The guaranteed ordering feature of node core timers (which makes it slow) is what makes this 1ms vs 5ms code work.

node/test/chan_drain.js
Raynos: Never mind. sendsDone vs checkSendsDone.

node/test/chan_drain.js
```diff
@@ -0,0 +1,374 @@
...
+
+        function checkASend(desc, res, i) {
+            assert.ifError(res.err, desc + 'no error');
+            assert.equal(res.value && String(res.value.arg3), 'mess' + (i + 1),
+                         desc + 'expected arg3');
+        }
+
+        function checkBSend(desc, res, i) {
```
Raynos: Why do Bs fail but A does not fail.

Raynos: Because B is exempt which means we terminated them harshly.

node/test/chan_drain.js
```diff
@@ -0,0 +1,374 @@
...
+        assert.equal(finishCount, 2, 'requests have not finished');
+
+        finishCount++;
+        server.drain('testdown', drainExemptB, drained);
+
+        finishCount++;
+        collectParallel(clients.a, sendOne, checkSendsDone('service:a', checkADecline, finish));
+
+        finishCount++;
+        collectParallel(clients.b, sendOne, checkSendsDone('servire:b', checkBRes, finish));
```
Raynos: nitpick: s/servire/service/

node/channel.js
jcorbin: no it's a programmer error to drain twice currently

node/connection_base.js
jcorbin: Not in wrong place, it has to be added so that sendError pop can work ;-)

Raynos:
Let's just get this in. lgtm.

PR #1273, opened 2015-09-04
bug fix: TChannel.prototype.register should support different handlerTypes

ShanniLi:


Raynos:
legit.

PR #1276, opened 2015-09-04
go: Make TCollectorTraceReporter not block.

### LINKED ISSUE #1269: go: TCollector TraceReporter should Report in the background
Currently, the TCollector TraceReporter reports synchronously. We should move this to the background, so reporting just puts requests into a channel.

As part of this change, we may be able to use multisubmits, as the background goroutine might just wait some duration when it gets a request, and then send all pending requests at once.

junchaowu:
Fixes #1269 

@prashantv

### DIFF
--- a/golang/trace/zipin_tracereporter_test.go
+++ b/golang/trace/zipin_tracereporter_test.go
@@ -106,7 +106,7 @@ func RandomAnnotations() []tchannel.Annotation {
 
 type testArgs struct {
 	s *mocks.TChanTCollector
-	c *ZipkinTraceReporter
+	c tchannel.TraceReporter
 }
 
 func ctxArg() mock.AnythingOfTypeArgument {
@@ -127,15 +127,12 @@ func TestSubmit(t *testing.T) {
 		ret := &gen.Response{Ok: true}
 
 		called := make(chan struct{})
-		args.s.On("Submit", ctxArg(), thriftSpan).Return(ret, nil).Run(func(arg mock.Arguments) {
+		args.s.On("Submit", ctxArg(), thriftSpan).Return(ret, nil).Run(func(_ mock.Arguments) {
 			close(called)
 		})
-		got, err := args.s.Submit(ctx, thriftSpan)
-		require.NoError(t, err)
-		assert.Equal(t, ret, got)
 		args.c.Report(span, annotations, nil, endpoint)
 
-		// wait for the Report to get called
+		// wait for the server's Submit to get called
 		select {
 		case <-time.After(time.Second):
 			t.Fatal("Submit not called")
@@ -186,7 +183,7 @@ func setupServer(h *mocks.TChanTCollector) (*tchannel.Channel, net.Listener, err
 	return tchan, listener, nil
 }
 
-func getClient(dst string) (*ZipkinTraceReporter, error) {
+func getClient(dst string) (tchannel.TraceReporter, error) {
 	tchan, err := tchannel.NewChannel("client", &tchannel.ChannelOptions{
 		Logger: tchannel.SimpleLogger,
 	})
--- a/golang/trace/zipin_tracereporter_test.go
+++ b/golang/trace/zipin_tracereporter_test.go
@@ -106,7 +106,7 @@ func RandomAnnotations() []tchannel.Annotation {
 
 type testArgs struct {
 	s *mocks.TChanTCollector
-	c tchannel.TraceReporter
+	c *ZipkinTraceReporter
 }
 
 func ctxArg() mock.AnythingOfTypeArgument {
@@ -126,9 +126,21 @@ func TestSubmit(t *testing.T) {
 		thriftSpan.BinaryAnnotations = []*gen.BinaryAnnotation{}
 		ret := &gen.Response{Ok: true}
 
-		args.s.On("Submit", ctxArg(), thriftSpan).Return(ret, nil)
-		err := args.c.Report(span, annotations, nil, endpoint)
-		assert.NoError(t, err)
+		called := make(chan struct{})
+		args.s.On("Submit", ctxArg(), thriftSpan).Return(ret, nil).Run(func(arg mock.Arguments) {
+			close(called)
+		})
+		got, err := args.s.Submit(ctx, thriftSpan)
+		require.NoError(t, err)
+		assert.Equal(t, ret, got)
+		args.c.Report(span, annotations, nil, endpoint)
+
+		// wait for the Report to get called
+		select {
+		case <-time.After(time.Second):
+			t.Fatal("Submit not called")
+		case <-called:
+		}
 	})
 }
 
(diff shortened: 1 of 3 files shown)


golang/trace/zipin_tracereporter_test.go
```diff
@@ -126,9 +126,21 @@ func TestSubmit(t *testing.T) {
 		thriftSpan.BinaryAnnotations = []*gen.BinaryAnnotation{}
 		ret := &gen.Response{Ok: true}
 
-		args.s.On("Submit", ctxArg(), thriftSpan).Return(ret, nil)
-		err := args.c.Report(span, annotations, nil, endpoint)
-		assert.NoError(t, err)
+		called := make(chan struct{})
+		args.s.On("Submit", ctxArg(), thriftSpan).Return(ret, nil).Run(func(arg mock.Arguments) {
```
prashantv: since you don't use `arg`, you should use `_` as the name:
```go
func(_ mock.Arguments) {
  close(called)
}
```

golang/trace/zipin_tracereporter_test.go
```diff
@@ -126,9 +126,21 @@ func TestSubmit(t *testing.T) {
...
+			close(called)
+		})
+		got, err := args.s.Submit(ctx, thriftSpan)
```
prashantv: why are you calling Submit on the server? You should only be testing `args.c.Report`

golang/trace/zipin_tracereporter_test.go
```diff
@@ -126,9 +126,21 @@ func TestSubmit(t *testing.T) {
...
+		require.NoError(t, err)
+		assert.Equal(t, ret, got)
+		args.c.Report(span, annotations, nil, endpoint)
+
+		// wait for the Report to get called
```
prashantv: nit: s/wait for the Report to get called/wait for Submit to get called/

golang/trace/zipin_tracereporter_test.go
```diff
@@ -174,7 +186,7 @@ func setupServer(h *mocks.TChanTCollector) (*tchannel.Channel, net.Listener, err
 	return tchan, listener, nil
 }
 
-func getClient(dst string) (tchannel.TraceReporter, error) {
+func getClient(dst string) (*ZipkinTraceReporter, error) {
```
prashantv: This shouldn't need to change

golang/trace/zipkin_tracereporter.go
```diff
@@ -34,35 +34,69 @@ import (
...
 	// client submit
 	// ignore the response result because TChannel shouldn't care about it.
 	_, err := r.client.Submit(ctx, thriftSpan)
 	return err
 }
 
+func (r *ZipkinTraceReporter) zipkinSpanWorker() {
+	for {
+		data := <-r.c
+		err := r.zipkinReport(&data)
```
prashantv: nit: inline the `err := r.zipkinReport` call with the `if`.

golang/trace/zipkin_tracereporter.go
```diff
@@ -34,35 +34,69 @@ import (
...
 func (r *ZipkinTraceReporter) Report(
-	span tc.Span, annotations []tc.Annotation, binaryAnnotations []tc.BinaryAnnotation, targetEndpoint tc.TargetEndpoint) error {
+	span tc.Span, annotations []tc.Annotation, binaryAnnotations []tc.BinaryAnnotation, targetEndpoint tc.TargetEndpoint) {
+	data := zipkinData{
+		Span:              span,
+		Annotations:       annotations,
+		BinaryAnnotations: binaryAnnotations,
+		TargetEndpoint:    targetEndpoint,
+	}
+	r.c <- data
```
prashantv: If the buffer is full, we might want to drop the trace. You can use `select` to detect whether a channel is full when inserting into it.

golang/trace/zipkin_tracereporter.go
```diff
@@ -34,35 +34,72 @@ import (
...
+	thriftSpan := buildZipkinSpan(data.Span, data.Annotations, data.BinaryAnnotations, data.TargetEndpoint)
 	// client submit
 	// ignore the response result because TChannel shouldn't care about it.
 	_, err := r.client.Submit(ctx, thriftSpan)
 	return err
 }
 
+func (r *ZipkinTraceReporter) zipkinSpanWorker() {
+	for {
+		data := <-r.c
```
prashantv: I forgot to mention earlier, you should probably use a range loop over the channel:
```go
for data := range r.c {
  [...]
}
```

That way, if we close the channel, it will stop the goroutine.

golang/trace/zipkin_tracereporter.go
```diff
@@ -34,35 +34,72 @@ import (
...
+	data := zipkinData{
+		Span:              span,
+		Annotations:       annotations,
+		BinaryAnnotations: binaryAnnotations,
+		TargetEndpoint:    targetEndpoint,
+	}
+
+	select {
+	case r.c <- data:
+	default:
```
prashantv: do we want to log when dropping packets because our buffer is full?

junchaowu: yes we should do that. I will add it.

golang/trace/zipkin_tracereporter.go
```diff
@@ -34,35 +34,78 @@ import (
...
 	// client submit
 	// ignore the response result because TChannel shouldn't care about it.
 	_, err := r.client.Submit(ctx, thriftSpan)
 	return err
 }
 
+func (r *ZipkinTraceReporter) zipkinSpanWorker() {
+	for data := range r.c {
+		if err := r.zipkinReport(&data); err != nil {
+			r.tchannel.Logger().Infof("Zipkin Span submit failed. Get error: %v", err)
```
prashantv: use `r.logger` here instead of `r.tchannel.Logger()`

prashantv:
lgtm once `r.logger` is used in `zipkinSpanWorker`.

golang/trace/zipkin_tracereporter.go
junchaowu: I forget to push some code.

golang/trace/zipkin_tracereporter.go
```diff
@@ -34,35 +34,78 @@ import (
...
-	thriftSpan := buildZipkinSpan(span, annotations, binaryAnnotations, targetEndpoint)
+	thriftSpan := buildZipkinSpan(data.Span, data.Annotations, data.BinaryAnnotations, data.TargetEndpoint)
 	// client submit
 	// ignore the response result because TChannel shouldn't care about it.
 	_, err := r.client.Submit(ctx, thriftSpan)
 	return err
 }
 
+func (r *ZipkinTraceReporter) zipkinSpanWorker() {
+	for data := range r.c {
```
yurishkuro: it's preferable for the client to buffer multiple spans and report a batch. Also, with buffering, certain entries can be combined, such as multiple annotations for the same span. See Node's implementation, for example.

golang/trace/zipkin_tracereporter.go
yurishkuro: instead of logging, I would suggest incrementing a metrics counter. There is not much benefit in polluting the app's log with these entries, a metrics counter will capture the same info.

golang/trace/zipkin_tracereporter.go
prashantv: Multiple annotations for a span are buffered by TChannel before it calls `Report` already btw. We can also report multiple independent spans using `multi_submit`, which was only implemented in the last day afaik.

PR #1277, opened 2015-09-07
Fix chan.setLazyHandling wrt incoming connections

jcorbin:
Bug uncovered when writing #1270, plus generalize connection walking for use by drain work.


kriskowal:
This does seem like a good place to draw the line in the sand between inlining all the loops and generalizing the iterator. Looks good to me.

jcorbin:
Indeed, this iterator is designed to be called off the hot path, and both this and the future use case are just that.

PR #1278, opened 2015-09-08
Misc cleanups ahead of #1270

jcorbin:
Assorted improvements and minor changes ahead of #1270:
- hookup callbacks first in streaming out req
- fix a small issue with the "mismatched onReqDone callback" log fields
- allow allocCluster auto empty assertion to be disabled (so that test case can do its own call with a manually crafted expectation)
- other misc cleanups


kriskowal:
Each of these looks fine. :+1:

PR #1279, opened 2015-09-08
Improve hyperbahn client host port list parsing

### LINKED ISSUE #1275: Hyperbahn client hides error when failing to read hostPortFile
The only thing reported is:

```
{"log":"AssertionError: Failed to read the hostPortFile\n","stream":"stderr","time":"2015-09-04T18:04:31.825560731Z"}
```
Which isn't very helpful.

anson627:
To fix #1275
Host port file reading and parsing error need be exposed to users.

@Willyham @ShanniLi @Raynos

### DIFF
--- a/node/hyperbahn/index.js
+++ b/node/hyperbahn/index.js
@@ -95,15 +95,15 @@ function HyperbahnClient(options) {
     assert(options.serviceName, 'must pass in serviceName');
     if (Array.isArray(options.hostPortList)) {
         self.hostPortList = options.hostPortList;
-    } else if (typeof options.hostPortList === 'string') {
-        var tuple = safeSyncRead(options.hostPortList);
+    } else if (typeof options.hostPortFile === 'string') {
+        var tuple = safeSyncRead(options.hostPortFile);
         assert(!tuple[0], 'Read host port list failed with ' + tuple[0]);
         tuple = safeJsonParse(tuple[1]);
         assert(!tuple[0], 'Parse host port list failed with ' + tuple[0]);
         assert(Array.isArray(tuple[1]), 'Host port list in the file is not array');
         self.hostPortList = tuple[1];
     } else {
-        assert(false, 'hostPortList must be either an array or a file path');
+        assert(false, 'Must pass in hostPortList as array or hostPortFile as string');
     }
 
     self.serviceName = options.serviceName;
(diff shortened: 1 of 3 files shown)


node/test/hyperbahn/advertise.js
```diff
@@ -88,7 +88,7 @@ function runTests(HyperbahnCluster) {
         var client = new HyperbahnClient({
             serviceName: 'hello-bob',
             callerName: 'hello-bob-test',
-            hostPortFile: hostPortFile,
+            hostPortList: hostPortFile,
```
ShanniLi: why this change?

ShanniLi:
I think @Willyham meant to expose the file read error, which is different from this PR. Not sure merging hostPortFile into hostPortList is a good idea. hostPortList by its name is a list/array

anson627:
Exposing error is part of this PR. The other part is to simplify the interface, having two parameter for the same thing make it hard to use. For example, xlate has hostPortFile in config file, if I want to overwrite it by passing --config hostPortFile [localhost:21300]. I need write code like this: if (Array.isArray(hostPortFile)), call constructor with hostPortList, otherwise, call with hostPortFile.

node/test/hyperbahn/advertise.js
anson627: this change is to simplify the constructor interface by reducing two parameters to one:
if (Array.isArray(hostPortList)) {
use it
} else if (typeof hostPortList is string) {
read and parse json file
} else {
throw error
}

As a user, I can get hostPort list from config or command line, it can be a file path or an array of strings. I just want to pass it into constructor, without worrying about which parameter to use.

Raynos:
:-1: This is confusing.

Let's have type stability and two arguments.

Raynos:
Please split this into two commits or two PRs.

The improved error messages are :+1:

anson627:
Will do the split.

node/hyperbahn/index.js
```diff
@@ -97,13 +97,13 @@ function HyperbahnClient(options) {
         self.hostPortList = options.hostPortList;
     } else if (typeof options.hostPortFile === 'string') {
         var tuple = safeSyncRead(options.hostPortFile);
-        assert(!tuple[0], 'Failed to read the hostPortFile');
+        assert(!tuple[0], 'Read host port list failed with ' + tuple[0]);
```
ShanniLi: This is string + Error. Can you paste a sample how it will look like?

ShanniLi:
one comment. Otherwise, lgtm.

node/hyperbahn/index.js
anson627: one sample:
Read host port list failed with Error: ENOENT, no such file or directory \'~~~~~\'

PR #1262, opened 2015-09-02
Implement a benchmark for the HyperbahnWorker

Raynos:
This contains both tchannel & hyperbahn changes.
I will break this out into two PRs.

The main change is to refactor TChannel benchmarks to 
be more overwritable.

I've added a Hyperbahn benchmark.

You can see the resulting flames here:

 - TChannel: https://people.uberinternal.com/~jakev/tchannel-master-flame.svg
 - Hyperbahn: https://people.uberinternal.com/~jakev/hyperbahn-master-flame.svg

r: @rf @jcorbin @kriskowal

PR #1280, opened 2015-09-08
Improve TChannel benchmarks for easier sub classing

Raynos:
These are a set of improvements to the TChannel
benchmarks so that they can be imported from the 
Hyperbahn repository.

See #1262 for Hyperbahn changes.

r: @jcorbin @rf @kriskowal


jcorbin:
lgtm... sad more prod weight, but it's necessary

PR #1282, opened 2015-09-08
Use thriftrw

Raynos:
This PR replaces thriftify with thriftrw.

There is no interface change other then an addition
of a `strict` boolean.

Technically this is a breaking change as we now support
a stricter subset of Thrift IDL files and enforce
the required field.

However thrift is marked as unstable in the documentation
so we will just bump minor.

r: @jcorbin @kriskowal @rf

### DIFF
--- a/node/as/thrift.js
+++ b/node/as/thrift.js
@@ -25,7 +25,7 @@ var fs = require('fs');
 var path = require('path');
 var bufrw = require('bufrw');
 var Result = require('bufrw/result');
-var thriftify = require('thriftify');
+var thriftrw = require('thriftrw');
 
 var errors = require('../errors.js');
 
@@ -43,7 +43,10 @@ function TChannelAsThrift(opts) {
     assert(opts && typeof opts.source === 'string',
         'must pass source as an argument');
 
-    self.spec = thriftify.parseSpec(opts.source);
+    self.spec = new thriftrw.Thrift({
+        source: opts.source,
+        strict: opts.strict
+    });
 
     // Pulled off of things in `.register` and `.send` rather than passed in
     self.logger = null;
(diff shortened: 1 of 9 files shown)


node/as/thrift.js
```diff
@@ -279,7 +284,7 @@ function TChannelThriftResponse(response, parseResult) {
     if (response.ok) {
         self.body = parseResult.body;
     } else {
-        self.body = errors.ReconstructedError(parseResult.body);
+        self.body = parseResult.body;
```
jcorbin: y u if/else still? get rid of the onion

kriskowal: Looks like the same code regardless of response.ok. Simplify.

node/test/as-thrift.js
```diff
@@ -129,9 +129,12 @@ allocCluster.test('send and receive a not ok', {
         assert.ifError(err);
 
         assert.ok(!res.ok);
+
+        console.log('?', res.body);
```
jcorbin: did you mean to leave these?

node/test/as-thrift.js
```diff
@@ -155,6 +158,9 @@ allocCluster.test('send and receive a typed not ok', {
...
```
kriskowal: elide

node/test/as-thrift.js
kriskowal: elide

node/test/anechoic-chamber.thrift
```diff
@@ -30,9 +30,9 @@ exception NoEchoTypedError {
 }
 
 service Chamber {
-    i32 echo(0: required i32 value) throws (
+    i32 echo(1: required i32 value) throws (
         1: NoEchoError noEcho
         2: NoEchoTypedError noEchoTyped
     )
-    i32 echo_big(0: required list<string> value)
+    i32 echo_big(1: required list<string> value)
```
kriskowal: Note to future: convert to thriftCase.

kriskowal:
:ok: with recommended changes.

PR #1283, opened 2015-09-09
Improve conn resetAll error handling

jcorbin:
r @kriskowal @Raynos

PR #1286, opened 2015-09-09
WIP: Emit in req reset errors

jcorbin:
This was the problematic part of #1283 before being split out; the problem is that adding this induces a non-deterministic uncaught exception (suspect race condition).

TODO:
- [ ] debug and fix the bug, possibly in a separate PR

### DIFF
--- a/node/connection.js
+++ b/node/connection.js
@@ -527,9 +527,11 @@ TChannelConnection.prototype.resetAll = function resetAll(err) {
     //   own outgoing work, which is hard to cancel. By setting this.closing, we make sure
     //   that once they do finish that their callback will swallow the response.
     inOpKeys.forEach(function eachInOp(id) {
-        self.ops.popInReq(id);
-        // TODO: support canceling pending handlers
-        // TODO report or handle or log errors or something
+        var req = self.ops.popInReq(id);
+        if (!req) {
+            return;
+        }
+        req.emitError(makeReqError(req));
     });
 
     // for all outgoing requests, forward the triggering error to the user callback


jcorbin:
To repro the uncaught:
```
$ while node test/hyperbahn/forward.js; do done
...

TAP version 13
# advertise and forward
HYPERBAHNCLIENT INFO: HyperbahnClient advertisement timeout disabled ~ { service: 'steve' }
ok 1 header is null
ok 2 got a body
ok 3 should be equal
ok 4 null
ok 5 should be equal

/Users/joshua/Uber/tchannel/node/lib/event_emitter.js:140
    throw err;
          ^
TchannelResponseAlreadyDoneError: cannot send error frame, response already done in state: 3
    at Object.createError [as ResponseAlreadyDone] (/Users/joshua/Uber/tchannel/node/node_modules/error/typed.js:31:22)
    at V2OutResponse.sendError (/Users/joshua/Uber/tchannel/node/out_response.js:199:43)
    at TChannelV2Handler.onReqError (/Users/joshua/Uber/tchannel/node/v2/handler.js:908:17)
    at DefinedEvent.onReqError [as listener] (/Users/joshua/Uber/tchannel/node/v2/handler.js:891:14)
    at DefinedEvent.emit (/Users/joshua/Uber/tchannel/node/lib/event_emitter.js:86:14)
    at TChannelConnection.buildResponse (/Users/joshua/Uber/tchannel/node/connection_base.js:145:24)
    at RelayRequest.buildResponse [as buildRes] (/Users/joshua/Uber/tchannel/node/connection_base.js:136:21)
    at RelayRequest.createOutResponse (/Users/joshua/Uber/tchannel/node/relay_handler.js:217:24)
    at RelayRequest.onError (/Users/joshua/Uber/tchannel/node/relay_handler.js:265:15)
    at Array.onError [as 1] (/Users/joshua/Uber/tchannel/node/relay_handler.js:110:14)
```

PR #1287, opened 2015-09-09
Fix for upgraded thriftrw

jcorbin:
Travis being down yesterday masked this failure.

r @kriskowal @Raynos

### DIFF
--- a/hyperbahn/endpoints/index.js
+++ b/hyperbahn/endpoints/index.js
@@ -74,7 +74,7 @@ function setupEndpoints(clients, services) {
     /*eslint no-unused-vars: 0*/
     // thrift health endpoint
     var tchannelAsThrift = new TChannelAsThrift({
-        source: '',
+        source: '// lul',
         channel: clients.hyperbahnChannel,
         isHealthy: require('./health').isHealthy
     });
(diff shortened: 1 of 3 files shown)


hyperbahn/endpoints/index.js
```diff
@@ -74,7 +74,7 @@ function setupEndpoints(clients, services) {
     /*eslint no-unused-vars: 0*/
     // thrift health endpoint
     var tchannelAsThrift = new TChannelAsThrift({
-        source: '',
+        source: '// lul',
```
kriskowal: https://github.com/uber/thriftrw/issues/51

kriskowal:
there is a stamp. the stamp is green.

node/tcollector/tcollector.thrift
```diff
@@ -2,9 +2,9 @@
 // `Endpoint`, will be the server being hit by the call.
 
 struct Endpoint {
-    1: i32 ipv4
-    2: i32 port
-    3: string serviceName
```
Raynos: You cannot do this. Revert

Raynos: This is a breaking change.

Raynos: Just set strict: false

Raynos:
Verified this change in python & node. Its safe. Going to change the flashlight/tcollector backend.

PR #1289, opened 2015-09-09
Add peer no-preference selection strategy

ShanniLi:
r @jcorbin @kriskowal @Raynos 

Tested with Ringpop and observed the 2x connection issue gone.

### DIFF
--- a/node/peer.js
+++ b/node/peer.js
@@ -29,6 +29,7 @@ var TChannelConnection = require('./connection');
 var errors = require('./errors');
 var Request = require('./request');
 var PreferOutgoing = require('./peer_score_strategies.js').PreferOutgoing;
+var NoPreference = require('./peer_score_strategies.js').NoPreference;
 
 var DEFAULT_REPORT_INTERVAL = 1000;
 
@@ -62,7 +63,13 @@ function TChannelPeer(channel, hostPort, options) {
         );
     }
 
-    self.setScoreStrategy(PreferOutgoing);
+    self.preferOutgoing = self.options.preferOutgoing;
+
+    if (self.preferOutgoing) {
+        self.setScoreStrategy(PreferOutgoing);
+    } else {
+        self.setScoreStrategy(NoPreference);
+    }
 
     function onReport() {
         if (!self.hostPort) {
--- a/node/peer_score_strategies.js
+++ b/node/peer_score_strategies.js
@@ -21,6 +21,7 @@
 'use strict';
 
 module.exports.PreferOutgoing = PreferOutgoing;
+module.exports.NoPreference = NoPreference;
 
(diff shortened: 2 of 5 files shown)


anson627:
so xlate egress should specify PreferOutgoing, since it only make outbound calls?

node/peer_score_strategies.js
```diff
@@ -74,3 +75,100 @@ PreferOutgoing.prototype.getScore = function getScore() {
...
+            }
+            /* falls through */
+        case PreferOutgoing.UNCONNECTED:
+            /* falls through */
+        case PreferOutgoing.FRESH_OUTGOING:
+            return 0.1 + random * 0.3;
+        case PreferOutgoing.READY_OUTGOING:
+            return 0.4 + random * 0.6;
+    }
+};
```
kriskowal: This PreferOutgoing is a copy of PreferOutgoing from above.

kriskowal:
We’ll also need preferIncoming for service channels speaking to Hyperbahn, so they favor connections from their affine nodes. The eventual API should be `preferConnectionDirection: 'incoming' | 'outgoing' | null` to avoid weird cases like `preferIncoming: true, preferOutgoing: true`. Other than that, this looks solid.

node/peer.js
```diff
@@ -62,7 +63,13 @@ function TChannelPeer(channel, hostPort, options) {
         );
     }
 
-    self.setScoreStrategy(PreferOutgoing);
+    self.preferOutgoing = self.options.preferOutgoing;
```
Raynos: let's default to `preferOutgoing` and manually mark the ringpop sub channel as preferNone.

ShanniLi: preferOutgoing cannot be the default, since service side should preferIncoming. In general, no preference cannot cause any issue. preferOutgoing & preferIncoming are more like optimization.

ShanniLi: @kriskowal, yes, I will add a preferIncoming in a separate PR.

ShanniLi:
@anson627 xlate egress is a hyperbahn service, right? If so, it should prefer incoming.

node/peer.js
Raynos: We should default to preferOutgoing for back compat.

Raynos: Fine lets prefer neither by default.

Raynos:
We should add some tests.

```
test('prefer outgoing creates new conn even if incoming');
test('prefer any re-uses incoming conn and does not open outgoing conn');
```

Raynos:
The main aim of tests is:

 - to assert that for prefer outgoing there are two connection objects between two tchannels.
 - to assert that for prefer any there is only one connection between two tchannels.

Raynos:
lgtm

PR #1291, opened 2015-09-09
TChannelOutResponse: stop sending after connection is closed

jcorbin:
r @kriskowal @Raynos

### DIFF
--- a/node/out_response.js
+++ b/node/out_response.js
@@ -196,7 +196,8 @@ TChannelOutResponse.prototype.sendCallResponseContFrame = function sendCallRespo
 TChannelOutResponse.prototype.sendError = function sendError(codeString, message) {
     var self = this;
 
-    if (self.inreq.connection.closing) {
+    if (self.inreq.connection && // because selfpeer/connection
+        self.inreq.connection.closing) {
         self.logger.info('ignoring outresponse.sendError on a closed connection', {
             codeString: codeString,
             errorMessage: message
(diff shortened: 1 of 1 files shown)


node/out_response.js
```diff
@@ -195,6 +195,21 @@ TChannelOutResponse.prototype.sendCallResponseContFrame = function sendCallRespo
 
 TChannelOutResponse.prototype.sendError = function sendError(codeString, message) {
     var self = this;
+
+    // if the request has already error with a similar error, stop
+    if (self.inreq.err &&
+        errors.classify(self.inreq.err) === codeString) {
```
Raynos: Should logger.info()

node/out_response.js
```diff
@@ -337,6 +352,12 @@ TChannelOutResponse.prototype.send = function send(res1, res2) {
...
     }
 
+    // TODO: may be spam, consider dropping
+    if (self.inreq.connection.closing) {
+        self.logger.info('ignoring outresponse.send on a closed connection', self.extendLogInfo({}));
+        return;
+    }
+
     self.sendCallResponseFrame([self.arg1, res1, res2], true);
     self.emitFinish();
```
kriskowal: These early returns above preclude this emitFinish(). Do all of these cases imply that emitFinish() has already been sent? Perhaps we should emitFinish() up top if it hasn’t been emitted yet.

Also, callers of sendError may be expecting it to return self consistently.

Raynos:
lgtm.

node/out_response.js
```diff
@@ -195,6 +195,15 @@ TChannelOutResponse.prototype.sendCallResponseContFrame = function sendCallRespo
 
 TChannelOutResponse.prototype.sendError = function sendError(codeString, message) {
     var self = this;
+
+    if (self.inreq.connection.closing) {
+        self.logger.info('ignoring outresponse.sendError on a closed connection', {
+            codeString: codeString,
+            errorMessage: message
+        });
+        return;
```
kriskowal: should return self here?

kriskowal: We can get here without being in the done or error state. This should also emit finish before returning.

kriskowal:
:ok:

PR #1292, opened 2015-09-09
Larch reservoir sampled logger

rf:
@Raynos @jcorbin @kriskowal 

To continue the discussion on the commit:

1. Should we do straight reservoir sampling or do a scored-record sampling approach for the first impl? **For v0 we'll do straight reservoir sampling just to see what affect it has on resource usage when generating a lot of error logs.**
2. Is violation of log ordering a problem? I think its not; the logs will be mis ordered per 50ms chunk and only if there are more than 500 logs in that chunk. 
3. My weird copy serialize optimization in `record.js`. It uses an object with the data that doesn't change and copies the meta fields into it. My benchmarks showed that to be faster, but I could remove the optimization for clarity and instead just copy all fields into a new object.

### DIFF
--- a/hyperbahn/clients/logger.js
+++ b/hyperbahn/clients/logger.js
@@ -100,7 +100,8 @@ function createLogger(options) {
     });
 
     var logger = LarchLogger({
-        backends: [larchBackend]
+        backends: [larchBackend],
+        statsd: options.statsd
     });
 
     return logger;
--- a/hyperbahn/clients/logger.js
+++ b/hyperbahn/clients/logger.js
@@ -21,8 +21,12 @@
 'use strict';
 
 var os = require('os');
-var Logger = require('logtron');
+var Logtron = require('logtron');
 var process = require('process');
+
+var Larch = require('../lib/larch/larch');
+var LarchLogtronBackend = require('../lib/larch/logtron-backend');
+var LarchReservoirBackend = require('../lib/larch/reservoir-backend');
 
 var Levels = {
     TRACE: 10,
@@ -38,7 +42,7 @@ module.exports = createLogger;
 
 // inline createLogger for now because yolo
 function createLogger(options) {
-    return Logger({
+    var logtronLogger = new Logtron({
         meta: {
             team: options.team,
             project: options.project,
@@ -76,7 +80,7 @@ function createLogger(options) {
             }
         },
         statsd: options.statsd,
-        backends: Logger.defaultBackends({
+        backends: Logtron.defaultBackends({
             kafka: options.kafka,
             logFile: options.logFile,
             console: options.console,
@@ -88,4 +92,16 @@ function createLogger(options) {
         }),
         transforms: []
     });
+
+    // The backend for larch is a reservoir sampler that logs to a logtron
+    // backend which redirects to a Logtron instance
+    var larchBackend = LarchReservoirBackend({
+        backend: LarchLogtronBackend(logtronLogger)
+    });
+
+    var logger = Larch({
+        backends: [larchBackend]
+    });
+
+    return logger;
 }
--- a/hyperbahn/clients/logger.js
+++ b/hyperbahn/clients/logger.js
@@ -21,12 +21,12 @@
 'use strict';
 
 var os = require('os');
-var Logtron = require('logtron');
+var Logger = require('logtron');
 var process = require('process');
 
-var Larch = require('../lib/larch/larch');
-var LarchLogtronBackend = require('../lib/larch/logtron-backend');
-var LarchReservoirBackend = require('../lib/larch/reservoir-backend');
+var LarchLogger = require('../lib/larch/larch');
+var LogtronLogBackend = require('../lib/larch/logtron-backend');
+var ReservoirLogBackend = require('../lib/larch/reservoir-backend');
 
 var Levels = {
     TRACE: 10,
@@ -42,7 +42,7 @@ module.exports = createLogger;
 
 // inline createLogger for now because yolo
 function createLogger(options) {
-    var logtronLogger = new Logtron({
+    var logtronLogger = new Logger({
         meta: {
             team: options.team,
             project: options.project,
@@ -80,7 +80,7 @@ function createLogger(options) {
             }
         },
         statsd: options.statsd,
-        backends: Logtron.defaultBackends({
+        backends: Logger.defaultBackends({
             kafka: options.kafka,
             logFile: options.logFile,
             console: options.console,
@@ -95,11 +95,11 @@ function createLogger(options) {
 
     // The backend for larch is a reservoir sampler that logs to a logtron
     // backend which redirects to a Logtron instance
-    var larchBackend = LarchReservoirBackend({
-        backend: LarchLogtronBackend(logtronLogger)
+    var larchBackend = ReservoirLogBackend({
+        backend: LogtronLogBackend(logtronLogger)
     });
 
-    var logger = Larch({
+    var logger = LarchLogger({
         backends: [larchBackend]
     });
 
--- a/hyperbahn/lib/larch/base-backend.js
+++ b/hyperbahn/lib/larch/base-backend.js
@@ -47,7 +47,7 @@ function BaseBackend(options) {
 
     assert(
         typeof self.destroy === 'function',
-        '`bootstrap` method of BaseBackend must be overridden by function'
+        '`destroy` method of BaseBackend must be overridden by function'
     );
 
     assert(
--- a/hyperbahn/lib/larch/base-backend.js
+++ b/hyperbahn/lib/larch/base-backend.js
@@ -23,6 +23,8 @@
 var assert = require('assert');
 var collectParallel = require('collect-parallel/array');
 
+var Errors = require('./errors');
+
 module.exports = BaseBackend;
 
 function BaseBackend(options) {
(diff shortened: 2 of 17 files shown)


Raynos:
@rf Let's do straight sampling for v0.

We want to verify that sampling makes error frames faster with our benchmarks.

I'm working on https://github.com/uber/tchannel/pull/1262

Raynos:
I think we can keep the ordering semantics per worker somehow. But if we lose it is not a problem. Let's just get the `ts` correct on the LogRecord so we can re-order if needed.

hyperbahn/lib/larch/larch.js
```diff
@@ -0,0 +1,123 @@
...
+        if (done >= self.backends.length) {
+            cb();
+        }
+    }
+};
+
+Larch.prototype.bootstrap = function bootstrap(cb) {
+    var self = this;
+
+    collectParallel(self.backends, bootstrapBackend, cb);
```
Raynos: collect parallel returns (null, Array<{ err, value }>). This is probably not what you wanted.

hyperbahn/lib/larch/base-backend.js
```diff
@@ -0,0 +1,75 @@
...
+};
+
+BaseBackend.prototype.destroy = function bootstrap(cb) {
+    cb();
+};
+
+BaseBackend.prototype.logMany = function logMany(records, cb) {
+    var self = this;
+
+    collectParallel(records, eachRecord, cb);
```
Raynos: again collectParallel probably does not do what you want it to do.

hyperbahn/lib/larch/larch.js
```diff
@@ -0,0 +1,123 @@
...
+
+    var record = new Record(level, msg, meta, null);
+
+    var i;
+    for (i = 0; i < self.backends.length; i++) {
+        self.backends[i].log(record, backendDone);
+    }
+
+    var done = 0;
+    function backendDone() {
```
Raynos: please do error handling here.

hyperbahn/lib/larch/record.js
```diff
@@ -0,0 +1,67 @@
...
+'use strict';
+
+var os = require('os');
+var process = require('process');
+var stringify = require('json-stringify-safe');
+
+module.exports = Record;
+
+function Record(level, msg, meta, time) {
+    this.meta = (meta === null || meta === void 0) ? null : meta;
```
Raynos: s/void 0/undefined/

hyperbahn/lib/larch/record.js
```diff
@@ -0,0 +1,67 @@
...
+            this.data[i] = this.meta[i];
+        }
+
+        this.hasBeenMerged = true;
+    }
+
+    return this.data;
+};
+
+function RecordData(level, msg, meta, time) {
```
Raynos: let's add `this.fields = null` here.

hyperbahn/lib/larch/record.js
```diff
@@ -0,0 +1,67 @@
...
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+'use strict';
+
+var os = require('os');
+var process = require('process');
+var stringify = require('json-stringify-safe');
```
Raynos: We dont want this. Let's just make cyclic objects a programmer error.

hyperbahn/lib/larch/record.js
```diff
@@ -0,0 +1,67 @@
...
+    if (this.serialized !== null) {
+        this.serialized = stringify(this.toJSON());
+    }
+
+    return this.serialized;
+};
+
+Record.prototype.toJSON = function toJSON() {
+    var i;
+    if (!this.hasBeenMerged) {
```
Raynos: instead of this merging stuff just do `this.data.fields = this.meta`

hyperbahn/lib/larch/record.js
rf: That did cause a problem in production in Arbiter once. Built in stringify will just spin and blow the stack.

hyperbahn/lib/larch/reservoir-backend.js
```diff
@@ -0,0 +1,134 @@
...
+    self.backend.logMany(copy);
+
+    self.records.length = 0;
+    self.count = 0;
+};
+
+ReservoirBackend.prototype.log = function log(record, cb) {
+    var self = this;
+
+    if (typeof cb === 'function') {
```
Raynos: lets call cb at end of function

hyperbahn/lib/larch/record.js
rf: That significantly changes the output of the logger; I was trying to make it match logtron. If this doesn't matter I'll do `this.data.fields = this.meta`; I like that better anyway. Then there's just one class.

hyperbahn/lib/larch/reservoir-backend.js
```diff
@@ -0,0 +1,134 @@
...
+util.inherits(ReservoirBackend, BaseBackend);
+
+ReservoirBackend.rangeRand = function rand(lo, hi) {
+    return Math.random() * (hi - lo) + lo;
+};
+
+ReservoirBackend.prototype.flush = function flush(records) {
+    var self = this;
+
+    var copy = records.slice(0);
```
Raynos: If we care about ordering we could re-order here.

hyperbahn/lib/larch/reservoir-backend.js
```diff
@@ -0,0 +1,134 @@
...
+    if (typeof cb === 'function') {
+        cb();
+    }
+
+    self.count += 1;
+
+    if (self.records.length < self.size) {
+        self.records.push(record);
+    } else {
+        var probability = self.rangeRand(1, self.count);
```
Raynos: 0, self.count - 1 ?

Raynos:
Lets get some tests in here too :)

hyperbahn/lib/larch/record.js
Raynos: ffffff---

Your right. Let's just not do anything clever and just match logtron.

Let's also not add all those new fields to `meta` and just pass meta, level, etc to logtron.

Raynos: i.e. all the new fields in `this.data` are opinion thats out of scope. 

Let's just drop `this.data` and only put fields on `this` that are necessary for LogtronBackend.

hyperbahn/lib/larch/record.js
Raynos: we still want to have a programmer error if theres cyclic data. Cyclic data is banned.

rf: Yeah but a stack overflow means you can't even see where the error originated. I ended up spinning up a smartos vm so I could look at a core file in mdb. At the very least it should throw an exception, but it shouldn't just throw an unhandleable exception that's a pain to debug.

Raynos: ok; lets use json-stringify-safe for now.

hyperbahn/clients/logger.js
```diff
@@ -21,9 +21,13 @@
 'use strict';
 
 var os = require('os');
-var Logger = require('logtron');
+var Logtron = require('logtron');
```
kriskowal: I think it would be good to retain the principle that brand names from npm only infect the first component of the top-level module identifier. I would keep this line as-is. Keep larch in the module identifier in anticipation of breaking it into an npm package, but: `var ReservoirLogBackend = require('../lib/larch/reservoir-backend');` etc. This will make the purpose more obvious to the casual observer.

hyperbahn/lib/larch/record.js
```diff
@@ -0,0 +1,59 @@
...
+
+Record.prototype.serialize = function serialize() {
+    if (this.serialized !== null) {
+        this.serialized = stringify(this.toJSON());
+    }
+
+    return this.serialized;
+};
+
+Record.prototype.toJSON = function toJSON() {
```
Raynos: This method is not used in production right ? Let's just get rid of it.

hyperbahn/lib/larch/reservoir-backend.js
Raynos: Let's emit stats here like counter `larch.flushed.{level}`

hyperbahn/lib/larch/reservoir-backend.js
```diff
@@ -0,0 +1,145 @@
...
+
+    self.count += 1;
+
+    if (self.records.length < self.size) {
+        self.records.push(record);
+    } else {
+        var probability = self.rangeRand(0, self.count);
+        if (probability < self.size) {
+            // evict and replace
+            self.records[probability] = record;
```
Raynos: let's emit stats here like counter `larch.dropped.{level}`

Raynos:
Let's add stats then ship.

Raynos:
Fix lint.

Raynos:
Let's also get a stat for everything we flush

In production I want to see the distribution of logs that are dropped and flushed by levelName.

Raynos:
legit. please fix travis though.

PR #1296, opened 2015-09-09
Improve errors.js to be source of truth

Raynos:
I moved the classification of should retry logic into
the errors.js file.

This will help teach other people what the consequences and
meaning of code names are for the TChannel library in
terms of retries & circuit breaking.

r: @kriskowal @rf


rf:
cool and neat A++++

kriskowal:
:100:

PR #1297, opened 2015-09-09
default emitConnectionMetrics to false

Raynos:
We turn it off in production for performance reasons. Having
it on by default burns 10% of the process CPU with peer
reporting logic.

Until we've optimized connection metrics it should be off
by default.

r: @kriskowal @rf


rf:
A+ :v:

kriskowal:
:ok:

PR #1299, opened 2015-09-10
use 127.0.0.1 instead of localhost in alloc-cluster

ShanniLi:
r @Raynos @kriskowal


node/test/as-http.js
```diff
@@ -316,6 +317,7 @@ function allocHTTPBridge(opts) {
         cluster.egressChan.peers.add(cluster.ingressServer.hostPort);
 
         var addr = cluster.httpEgress.address();
+        console.log(addr);
```
Raynos: remove dead log.

Raynos:
lets remove console.log; otherwise lgtm

PR #1300, opened 2015-09-11
Extend log info for service proxy

kriskowal:


node/hyperbahn/service_proxy.js
```diff
@@ -136,13 +132,10 @@ function rateLimit(req, buildRes) {
     // apply rate limiter
     if (self.rateLimiter.shouldRateLimitTotalRequest(req.serviceName)) {
         var totalLimit = self.rateLimiter.totalRequestCounter.rpsLimit;
-        self.logger.info('hyperbahn node is rate-limited by the total rps limit', {
-            rpsLimit: totalLimit,
-            cn: req.headers && req.headers.cn,
-            serviceName: req.serviceName,
-            endpoint: req.endpoint,
-            remoteAddr: req.remoteAddr
```
kriskowal: This property is preserved by the existing requestRemoteAddr introduced by req.extendLogInfo, but renamed. I introduce socketRemoteAddr as well, as produced manually above, by adding that property to req.extendLogInfo below.

kriskowal:
r @Raynos (in advance of CI passing—need to get master non-flappy again)

Raynos:
legit.

PR #1301, opened 2015-09-11
Improve time series test and add more cases.

Raynos:
This PR was meant to reproduce a memory leak / latency leak
with 512kb payloads, however it does not reproduce the issue.

I've refactored the time-series test into a re-usable
TimeSeriesCluster() helper. I've added three cases for 4kb,
48kb and 512kb requests.

These are tests that take around 3 minutes and are not meant
to be part of the test suite but instead run as re-usable
recipes to verify how Hyperbahn operates under load.

r: @rf @kriskowal


hyperbahn/test/time-series/_time-series-cluster.js
```diff
@@ -0,0 +1,720 @@
+// Copyright (c) 2015 Uber Technologies, Inc.
```
kriskowal: Let’s not start with _prefix file names.

hyperbahn/test/time-series/_time-series-cluster.js
```diff
@@ -0,0 +1,720 @@
...
+
+var startOfFile = Date.now();
+
+TimeSeriesCluster.test = tapeCluster(tape, TimeSeriesCluster);
+
+TimeSeriesCluster.buildString = function buildString(size) {
+    var tempArray = [];
+
+    for (var i = 0; i < size; i++) {
+        tempArray.push('A');
```
kriskowal: use string concatenation. consider string or buffer doubling.

```js
var str = 'A';
while (str.length < size) {
    str = str + str;
}
str = str.slice(0, size);
```

hyperbahn/test/time-series/_time-series-cluster.js
```diff
@@ -0,0 +1,720 @@
...
+        self.clusterOptions.remoteConfig['rateLimiting.enabled'] = false;
+    }
+    self.clusterOptions.namedRemotes = self.namedRemotes;
+    if (!('hyperbahn.circuits' in self.clusterOptions.remoteConfig)) {
+        self.clusterOptions.remoteConfig['hyperbahn.circuits'] = {
+            period: 100,
+            maxErrorRate: 0.5,
+            minRequests: 5,
+            probation: 5,
+            enabled: false
```
kriskowal: let's not replicate the default values here. should be sufficient to enabled: false, period: 100 since those are the only overrides.

hyperbahn/test/time-series/_time-series-cluster.js
```diff
@@ -0,0 +1,720 @@
...
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
```
kriskowal: We should start writing at least a sentence in each file to explain its purpose.

hyperbahn/test/time-series/_time-series-cluster.js
```diff
@@ -0,0 +1,720 @@
...
+    self.clientTimeout = opts.clientTimeout;
+    self.endpointToRequest = opts.endpointToRequest;
+
+    self.clientBatchDelay = opts.clientBatchDelay || 100;
+    self.numBatchesInBucket = opts.numBatchesInBucket || 5;
+    self.clientRequestsPerBatch = opts.clientRequestsPerBatch || 15;
+    self.serverInstances = opts.serverInstances || 10;
+    self.clientInstances = opts.clientInstances || 5;
+    self.requestBody = opts.requestBody || '';
+    self.requestBodyChunks = opts.requestBodyChunks || [];
```
kriskowal: Please describe the semantics of each of these options in single-sentence comments.

hyperbahn/test/time-series/_time-series-cluster.js
```diff
@@ -0,0 +1,720 @@
...
+        streamed: true,
+        timeout: self.clientTimeout,
+        headers: {
+            'as': 'raw',
+            'cn': self.clientServiceName
+        }
+    };
+    self.streamWriteDelay = 10;
+
+    self.freeList = new Array(self.requestVolume);
```
kriskowal: I don’t believe there’s a win to constructing an array with an initial size. You don’t have to do this to avoid push function calls.

kriskowal:
:ok: All comments advisory.

hyperbahn/test/time-series/_time-series-cluster.js
Raynos: thats not how remoteConfig works. there is no shallow or things.

hyperbahn/test/time-series/_time-series-cluster.js
Raynos: I just wanted stable memory.

Raynos:
@kriskowal 

Wrote documentation and moved file :)

PR #1302, opened 2015-09-11
add default implementation of meta health endpoint

### LINKED ISSUE #1288: go: Simplify Meta service endpoint registration
Design interface for services to easily register a Meta endpoint - they should only need to specify a function that returns a HealthResult.

junchaowu:
Fixes #1288 

@prashantv


jinpingv:
Is there any plan to add one for Python/tornado?

2015年9月10日星期四，junchaowu <notifications@github.com> 写道：

> @prashantv <https://github.com/prashantv>
> ------------------------------
> You can view, comment on, or merge this pull request online at:
>
>   https://github.com/uber/tchannel/pull/1302
> Commit Summary
>
>    - add default implementation of meta health endpoint
>
> File Changes
>
>    - *A* golang/thrift/gen-go/meta/constants.go
>    <https://github.com/uber/tchannel/pull/1302/files#diff-0> (18)
>    - *A* golang/thrift/gen-go/meta/meta.go
>    <https://github.com/uber/tchannel/pull/1302/files#diff-1> (354)
>    - *A* golang/thrift/gen-go/meta/ttypes.go
>    <https://github.com/uber/tchannel/pull/1302/files#diff-2> (150)
>    - *A* golang/thrift/health.go
>    <https://github.com/uber/tchannel/pull/1302/files#diff-3> (30)
>    - *A* golang/thrift/health_test.go
>    <https://github.com/uber/tchannel/pull/1302/files#diff-4> (87)
>    - *A* golang/thrift/meta.thrift
>    <https://github.com/uber/tchannel/pull/1302/files#diff-5> (8)
>    - *M* golang/thrift/server.go
>    <https://github.com/uber/tchannel/pull/1302/files#diff-6> (27)
>    - *A* golang/thrift/tchan-meta.go
>    <https://github.com/uber/tchannel/pull/1302/files#diff-7> (82)
>
> Patch Links:
>
>    - https://github.com/uber/tchannel/pull/1302.patch
>    - https://github.com/uber/tchannel/pull/1302.diff
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/uber/tchannel/pull/1302>.
>


-- 

-Ping

junchaowu:
I will add it immediately.

golang/thrift/health.go
```diff
@@ -0,0 +1,51 @@
...
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package thrift
+
+import "github.com/uber/tchannel/golang/thrift/gen-go/meta"
+
+// HealthFunc is function interface for health check endpoint.
+// Customized health endpoint has to follow this interface.
+type HealthFunc func(ctx Context) (bool, *string)
```
prashantv: Let's add names to the return variables so people know, and document those return variables. Let's leave the message as string, since they can just return empty string.
```go
// HealthFunc is the interface for custom health endpoints.
// ok is whether the service health is OK, and message is optional additional information for the health result.
type HealthFunc func(ctx Context) (ok bool, message string)
```

golang/thrift/health.go
```diff
@@ -0,0 +1,51 @@
...
+}
+
+// newHealthHandler return a new HealthHandler instance.
+func newHealthHandler() *healthHandler {
+	return &healthHandler{}
+}
+
+// Health returns true as default Health endpoint.
+func (h *healthHandler) Health(ctx Context) (r *meta.HealthStatus, err error) {
+	if h.handler != nil {
```
prashantv: let's remove the branch here. instead, have a function 
```go
func defaultHealth(ctx Context) (*meta.HealthStatus, error) {
  return &meta.HealthStatus{Ok: true}, nil
}
```

and then `newHealthHandler` can just do
```go
return &healthHandler{handler: defaultHealth}
```

This function can then call h.handler unconditionally

golang/thrift/health.go
```diff
@@ -0,0 +1,51 @@
...
+	handler HealthFunc
+}
+
+// newHealthHandler return a new HealthHandler instance.
+func newHealthHandler() *healthHandler {
+	return &healthHandler{}
+}
+
+// Health returns true as default Health endpoint.
+func (h *healthHandler) Health(ctx Context) (r *meta.HealthStatus, err error) {
```
prashantv: remove named parameters `r` and `err` here since they're unnecessary

golang/thrift/health_test.go
```diff
@@ -0,0 +1,104 @@
...
+func TestDefaultHealth(t *testing.T) {
+	withMetaSetup(t, func(ctx Context, c tchanMeta) {
+		ret, err := c.Health(ctx)
+		assert.Equal(t, ret.Ok, true)
+		assert.NoError(t, err)
+		assert.Nil(t, ret.Message)
+	}, nil)
+}
+
+func withMetaSetup(t *testing.T, f func(ctx Context, c tchanMeta), healthHandler HealthFunc) {
```
prashantv: it's nicer to make the `func` argument the last argument, so that your calls have all the parameters before the test code.

golang/thrift/health_test.go
```diff
@@ -0,0 +1,104 @@
...
+	thriftClient := NewClient(tchan, "meta", nil)
+	return newTChanMetaClient(thriftClient), nil
+}
+
+func customHealth(ctx Context) (bool, *string) {
+	message := "from me"
+	return false, &message
+}
+
+func TestCustomHealth(t *testing.T) {
```
prashantv: Put the `customHealth` and `TestCustomHealth` functions above the setup code.

golang/thrift/health_test.go
```diff
@@ -0,0 +1,104 @@
...
+	withMetaSetup(t, func(ctx Context, c tchanMeta) {
+		ret, err := c.Health(ctx)
+		assert.Equal(t, ret.Ok, false)
```
prashantv: Check err before the others, and only check the ret if err is nil. Also for all your `assert` calls (in this whole diff), add a message about what failed. e.g. this would become:
```go
if assert.NoError(t, err, "Health endpoint failed") {
  assert.Equal(t, ret.Ok, false, "Health status mismatch")
  assert.Equal(t, ret.Message, thrift.String("from me"), "Health message mismatch")
}
```
Note that I don't dereference the message, since if it's `nil`, it will crash the test, whereas the above test will just fail.

golang/thrift/server.go
```diff
@@ -61,6 +67,11 @@ func (s *Server) Register(svr TChanServer) {
 	}
 }
 
+// RegisterHealthHandler registers User health endpoint handler into TChannel.
```
prashantv: nit: User does not need to be capitalized. I'd probably make the message:
RegisterHealthHandler uses the user-specified function f for the Health endpoint.

golang/thrift/health.go
```diff
@@ -0,0 +1,52 @@
...
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package thrift
+
+import "github.com/uber/tchannel/golang/thrift/gen-go/meta"
+
+// HealthFunc is the interface for custom health endpoints.
+// ok is whether the service health is OK, and message is optional additional information for the health result.
+type HealthFunc func(ctx Context) (ok bool, message *string)
```
prashantv: Don't think message needs to be a `*string`. Just `string` is fine, since we don't care about the difference between empty message and missing message (both are the same as far as we care)

in the `Health` handler, if message is blank string, then we shouldn't set it in the `HealthStatus`. (We should test for this too).

golang/thrift/health_test.go
```diff
@@ -0,0 +1,107 @@
...
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"github.com/uber/tchannel/golang"
+)
+
+func TestDefaultHealth(t *testing.T) {
+	withMetaSetup(t, nil, func(ctx Context, c tchanMeta) {
+		ret, err := c.Health(ctx)
+		if assert.NoError(t, err, "Health endpoint failed") {
+			assert.Equal(t, ret.Ok, true, "Health status mismatch")
```
prashantv: use `assert.True` instead of `assert.Equal` here

golang/thrift/health_test.go
```diff
@@ -0,0 +1,107 @@
...
+		return nil, nil, err
+	}
+
+	listener, err := net.Listen("tcp", ":0")
+	if err != nil {
+		return nil, nil, err
+	}
+
+	server := NewServer(tchan)
+	if healthHandler != nil {
```
prashantv: Rather than do this here, I like the Registration to be explicit in `TestCustomHealth` - tests can serve as examples of how to use it.  The `setup` is easy to ignore with all the noise. Let's rewrite the setup logic to:
 - not take a healthHandler argument, but instead it returns the thrift.Server
 - the `thrift.Server` should be passed to the test function
 - `TestCustomHealth` should use the passed `thrift.Server`, and call `RegisterHealthHandler` expicitly

golang/thrift/health.go
junchaowu: in thrift gen code, the message interface is *string.

prashantv:
lgtm, thanks for addressing all my feedback!

PR #1304, opened 2015-09-11
Peer selection strategy: prefer incoming

ShanniLi:
r @kriskowal @Raynos 
cc @jcorbin 

Next steps:
1. update Hyperbahn service channel to prefer outgoing.
2. update Hyperbahn client to prefer incoming.

### DIFF
--- a/node/peer.js
+++ b/node/peer.js
@@ -30,6 +30,7 @@ var errors = require('./errors');
 var Request = require('./request');
 var PreferOutgoing = require('./peer_score_strategies.js').PreferOutgoing;
 var NoPreference = require('./peer_score_strategies.js').NoPreference;
+var PreferIncoming = require('./peer_score_strategies.js').PreferIncoming;
 
 var DEFAULT_REPORT_INTERVAL = 1000;
 
@@ -64,9 +65,12 @@ function TChannelPeer(channel, hostPort, options) {
     }
 
     self.preferOutgoing = self.options.preferOutgoing;
+    self.preferIncoming = self.options.preferIncoming;
 
     if (self.preferOutgoing) {
         self.setScoreStrategy(PreferOutgoing);
+    } else if (self.preferIncoming) {
+        self.setScoreStrategy(PreferIncoming);
     } else {
         self.setScoreStrategy(NoPreference);
     }
@@ -158,13 +162,22 @@ TChannelPeer.prototype.close = function close(callback) {
     }
 };
 
-TChannelPeer.prototype.getInConnection = function getInConnection() {
+TChannelPeer.prototype.getInConnection = function getInConnection(preferIdentified) {
     var self = this;
+    var candidate = null;
     for (var i = 0; i < self.connections.length; i++) {
         var conn = self.connections[i];
-        if (!conn.closing) return conn;
+        if (conn.closing) continue;
+        if (!preferIdentified) return conn; // user doesn't care, take first incoming
+        if (conn.remoteName) return conn; // user wanted an identified channel, and we found one
+        if (!candidate) candidate = conn; // we'll fallback to returning this if we can't find an identified one
     }
-    return null;
+    return candidate;
+};
+
+TChannelPeer.prototype.getIdentifiedInConnection = function getIdentifiedInConnection() {
+    var self = this;
+    return self.getInConnection(true);
 };
 
 TChannelPeer.prototype.getOutConnection = function getOutConnection(preferIdentified) {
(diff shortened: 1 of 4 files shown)


node/peer.js
```diff
@@ -402,6 +421,16 @@ TChannelPeer.prototype.outPendingWeightedRandom = function outPendingWeightedRan
...
 
+TChannelPeer.prototype.inPendingWeightedRandom = function inPendingWeightedRandom() {
+    // See outPendingWeightedRandom for explanation
+    var self = this;
+    var pending = self.pendingIdentified + self.countInPending();
+    var max = Math.pow(0.5, pending);
+    var min = max / 2;
+    var diff = max - min;
+    return min + diff * self.random();
+};
```
kriskowal: Please find a way to not make a duplicate of this function.

node/root_peers.js
```diff
@@ -38,6 +39,10 @@ function TChannelRootPeers(channel, options) {
     self.allocPeerEvent = self.defineEvent('allocPeer');
     self.peerOptions = self.options.peerOptions || {};
     self.preferOutgoing = self.options.preferOutgoing;
+    self.preferIncoming = self.options.preferIncoming;
+    assert(!self.preferOutgoing || !self.preferIncoming,
+        'preferOutgoing and preferIncoming cannot be both set');
+
```
kriskowal: In earnest, `options.preferConnectionDirection: "incoming" / "outgoing" / null` will DRY up much more easily, regardless of whether we eventually DRY up the three score strategies into one.

node/test/peer_strategies.js
```diff
@@ -146,12 +257,15 @@ allocCluster.test('prefer outgoing creates new conn even if incoming', {
     }
 });
 
-function setupEcho(channel, serviceName) {
+function setupEcho(channel, serviceName, cb) {
```
kriskowal: Rename `cb` to `onEchoSent` to clarify that this is not a callback but handler that will be called after every response.

node/test/peer_strategies.js
```diff
@@ -92,6 +93,116 @@ allocCluster.test('prefer any re-uses incoming conn and does not open outgoing c
...
+        assert.equals(bobCount.outCount, 1, 'bob should have 1 outgoing connection');
+        assert.equals(steveCount.outCount, 0, 'steve should not have outgoing connections');
+        assert.equals(steveCount.inCount, 1, 'steve should have 1 incoming connection');
+        assert.equals(steve2Count.outCount, 1, 'steve2 should have 1 outgoing connections');
+        assert.equals(steve2Count.inCount, 0, 'steve2 should not have incoming connections');
+
+        subBob.request({
+            serviceName: 'steve',
+            hasNoParent: true
+        }).send('echo', 'a', 'b', function noop() {});
```
kriskowal: The additional callback on setupEcho should be unnecessary since you can use done for the echo response handler.

node/test/peer_strategies.js
```diff
@@ -92,6 +93,116 @@ allocCluster.test('prefer any re-uses incoming conn and does not open outgoing c
...
+    }
+
+    function done() {
```
kriskowal: Be sure to thread done(err) into assert.end(err)

kriskowal:
Summary: please address feedback but you may punt on DRYing up Strategies except for the peer weighting algorithm. Please factor the common part of that function into a single implementation to avoid having to synchronize the duplicates.

kriskowal:
:ok:

node/peer.js
```diff
@@ -205,7 +216,13 @@ TChannelPeer.prototype.countConnections = function countConnections(direction) {
 // ensures that a connection exists
 TChannelPeer.prototype.connect = function connect(outOnly) {
     var self = this;
-    var conn = self.getIdentifiedOutConnection();
+    var conn = null;
+    if (self.preferConnectionDirection === 'in' && !outOnly) {
```
Raynos: This `outOnly` flag feels broken now. Where is it used ?

node/peer.js
```diff
@@ -395,13 +412,29 @@ TChannelPeer.prototype.outPendingWeightedRandom = function outPendingWeightedRan
...
+        countPending = self.countOutPending();
+    }
+    var pending = self.pendingIdentified + countPending;
     var max = Math.pow(0.5, pending);
     var min = max / 2;
     var diff = max - min;
     return min + diff * self.random();
 };
 
+TChannelPeer.prototype.outPendingWeightedRandom = function outPendingWeightedRandom() {
```
Raynos: these methods dont help much; just inline them.

Raynos: These methods don't help much. Just inline them.

node/peer.js
```diff
@@ -413,6 +446,17 @@ TChannelPeer.prototype.countOutPending = function countOutPending() {
     return pending;
 };
 
+TChannelPeer.prototype.countInPending = function countInPending() {
```
Raynos: Consider generalizing into `countPending(direction)`

Raynos:
lgtm.

node/peer.js
ShanniLi: It is still used. I feel like the second is unnecessary though.
1) https://github.com/uber/tchannel/blob/master/node/hyperbahn/service_proxy.js#L279
2) https://github.com/uber/tchannel/blob/master/node/peer_score_strategies.js#L67

PR #1307, opened 2015-09-12
Verify that busy frames can be smoothed over

Raynos:
This uses the TimeSeriesCluster to create a test case that
generates 25 req/s per HyperbahnWorker.

We set the rateLimit to 10 req/s for one out of 10 Hyperbahn
workers and then assert that zero requests fail due to retries.

We also assert that the rate limiter actually kicked in.

r: @kriskowal @rf

cc @anson627 This proofs that retries on Busy solve our problem.
We still need to debug why xlate didn't do retries.

### DIFF
--- a/hyperbahn/app.js
+++ b/hyperbahn/app.js
@@ -103,7 +103,6 @@ Application.prototype.bootstrap = function bootstrap(cb) {
 
     setupEndpoints(self.clients, self.services);
 
-    self.hookupSignals();
     self.clients.bootstrap(onClientsReady);
 
     function onClientsReady(err) {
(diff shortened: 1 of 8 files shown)


anson627:
awesome! Then I can port TimeSeriesCluster to xlate and do the same test.

Raynos:
@anson627 put me on the review for that. That's something I can help with.

kriskowal:
Please leave some notes for future-@raynos et al, in comments or markdown, for the usage of this demo and, in particular, what makes the first bucket special.

Raynos:
@kriskowal will do.

Raynos:
This PR will need to be rebased in github.com/uber/hyperbahn

PR #1308, opened 2015-09-13
WIP: Naive relay

Raynos:
This is a from scratch tchannel relay to find out where the performance ceiling for node is.

I've used @prashantv benchmark runner to determine the QPS per core and found this could do about 20-22k req/s compared to go's 35-37k req/s.

The semantics of this relay are to take a list of destinations from the command line and round robin call requests for all services to those destinations. The only action it does on forwarding is rewriting of ids.

We may want to re-use some of the tricks here in mainline hyperbahn like the FrameParser.

cc @kriskowal @jcorbin @rf 

P.S: There is no value in merging this. Just wanted to share the performance experiment.


Raynos:
RIP

PR #1309, opened 2015-09-14
allow reservoir to be live configured

rf:
@Raynos 

the way I return both parts of the logger is pretty gross, if you can think of a cleaner way to do that it would be appreciated

### DIFF
--- a/hyperbahn/clients/index.js
+++ b/hyperbahn/clients/index.js
@@ -48,7 +48,7 @@ var RemoteConfig = require('./remote-config.js');
 module.exports = ApplicationClients;
 
 function ApplicationClients(options) {
-    /*eslint max-statements: [2, 40] */
+    /*eslint max-statements: [2, 50] */
     if (!(this instanceof ApplicationClients)) {
         return new ApplicationClients(options);
     }
@@ -81,7 +81,7 @@ function ApplicationClients(options) {
 
     if (options.seedClients.logger) {
         self.logger = options.seedClients.logger;
-        self.reservoir = null;
+        self.logReservoir = null;
     } else {
         var loggerParts = createLogger({
             team: config.get('info.team'),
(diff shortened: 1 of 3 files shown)


hyperbahn/clients/index.js
```diff
@@ -300,6 +307,24 @@ ApplicationClients.prototype.onRemoteConfigUpdate = function onRemoteConfigUpdat
     self.updateExemptServices();
     self.updateRpsLimitForServiceName();
     self.updateKValues();
+    self.updateLogReservoirSize();
+};
+
+ApplicationClients.prototype.updateReservoir = function updateReservoir() {
+    var self = this;
+    if (self.logReservoir) {
```
Raynos: `self.reservoir` ?

Raynos:
Too much complexity here.

Just always use the reservior and have it always enable; lets toggle the flushInterval & size by flipr though.

rf:
okay

hyperbahn/clients/logger.js
```diff
@@ -104,5 +104,8 @@ function createLogger(options) {
         statsd: options.statsd
     });
 
-    return logger;
+    return {
```
Raynos: consider

```
logger.reservoir = larchBackend;
return logger
```

Raynos:
legit.

hyperbahn/clients/index.js
```diff
@@ -300,6 +307,18 @@ ApplicationClients.prototype.onRemoteConfigUpdate = function onRemoteConfigUpdat
     self.updateExemptServices();
     self.updateRpsLimitForServiceName();
     self.updateKValues();
+    self.updateReservoir();
+};
+
+ApplicationClients.prototype.updateReservoir = function updateReservoir() {
+    var self = this;
+    if (self.logReservoir) {
+        var size = self.remoteConfig.get('log.reservoir.size', 100);
```
Raynos: Wait we made a mistake here; we do not want the default to be `100`; let's bump the default to `1000`

PR #1312, opened 2015-09-15
Setting size > 20 fails for unknown reasons

Raynos:
For some reason locally on linux setting size > 20 fails with
a ton of socket closed.

Don't know why. Going to bump it down to 15 to unbreak master.


rf:
fine :+1:

PR #1313, opened 2015-09-15
Remove init timeout connection noise

Raynos:
Any given connection that is closed before it finishes
the init req/res handshake lies about it timing out.

This log should only happen if it actually timed out.


rf:
sounds good :+!:

PR #1314, opened 2015-09-15
log socket-closed errors

Raynos:
If we do not log these errors then various tests or production
systems will just be silent on socket failures.

We need to know about the other end closing the socket to
reason about why our code hangs or production systems fail.

r: @jcorbin @kriskowal @rf


rf:
ok :+1:

PR #1318, opened 2015-09-15
Fix hostports tests to work with hyperbahn

Raynos:
The hyperbahn cluster is slightly different. This gets the
tests to pass from Hyperbahn itself

r: @jcorbin @rf @shannili

### DIFF
--- a/hyperbahn/test/hyperbahn-client.js
+++ b/hyperbahn/test/hyperbahn-client.js
@@ -31,7 +31,7 @@ require('./hyperbahn-client/egress-nodes.js');
 require('./hyperbahn-client/forward.js')(TestCluster);
 require('./hyperbahn-client/advertise.js')(TestCluster);
 require('./hyperbahn-client/unadvertise.js')(TestCluster);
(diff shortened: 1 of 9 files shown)


node/test/hyperbahn/hostports.js
```diff
@@ -207,6 +211,7 @@ function runTests(HyperbahnCluster) {
                 'error message should be a parsing failure');
 
             var items = cluster.logger.items();
+            console.log('items', items);
```
kriskowal: console

kriskowal:
This screams `getOrCreateSubChannel` or simply `getChannel` with the rest implied. cc @jcorbin

jcorbin:
yes please, let's add a 'getOrMake' (or s/make/create/ to taste).

ShanniLi:
+1 for getOrMake. However, it feels optional for this diff.

lgtm.

Raynos:
Disagree with getOrMake.

This is just a temporary mess because two versions of TestCluster.

PR #1319, opened 2015-09-15
use rate limiter counter for more detailed traffic stats

ShanniLi:

### DIFF
--- a/node/hyperbahn/service_proxy.js
+++ b/node/hyperbahn/service_proxy.js
@@ -124,6 +124,9 @@ ServiceDispatchHandler.prototype.rateLimit =
 function rateLimit(req, buildRes) {
     var self = this;
 
+    // stats all traffic using rate limiter's counter
+    self.rateLimiter.incrementServiceCounter(req.headers.cn + '~~' + req.serviceName);
+
     var isExitNode = self.isExitFor(req.serviceName);
     if (isExitNode) {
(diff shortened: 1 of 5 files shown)


node/hyperbahn/service_proxy.js
```diff
@@ -124,6 +124,9 @@ ServiceDispatchHandler.prototype.rateLimit =
 function rateLimit(req, buildRes) {
     var self = this;
 
+    // stats all traffic using rate limiter's counter
+    self.rateLimiter.incrementServiceCounter(req.headers.cn + '~~' + req.serviceName);
```
Raynos: I think it's bad to make this a service counter. 

I think its better to have a new bucket in the rate limiter thats not "ServiceCounter" but is instead "EdgeCounter"

Raynos: It will also be good to have different stats for service-rps and edge-rps

Raynos: We also shouldnt rate limit edges. 

This is a lot of coupling and nuance here.

Raynos:
This needs more work and less coupling.

node/hyperbahn/service_proxy.js
ShanniLi: the counter is not rate limited. I will make changes to decouple.

Raynos:
lgtm.

PR #1322, opened 2015-09-16
Fix flappy drain test

### LINKED ISSUE #1317: Node: Flappy drain test
```
TCHANNEL INFO: draining channel ~ { hostPort: '127.0.0.1:53391', reason: 'testdown', count: 7 }
  not ok 3423 unexpected channel[1] peer[0] conn[0]
⨯ immediate drain      operator: fail
      at: eachConn (/home/travis/build/uber/tchannel/node/test/lib/alloc-cluster.js:87:32)
    ...
  not ok 3425 unexpected channel[2] peer[0] conn[0]
    ---
      operator: fail
      at: eachConn (/home/travis/build/uber/tchannel/node/test/lib/alloc-cluster.js:87:32)
    ...
  not ok 3427 unexpected channel[3] peer[0] conn[0]
    ---
      operator: fail
      at: eachConn (/home/travis/build/uber/tchannel/node/test/lib/alloc-cluster.js:87:32)
    ...
```

cc @jcorbin @kriskowal

anson627:
To fix #1317
@jcorbin @Raynos @ShanniLi

### DIFF
--- a/node/test/chan_drain.js
+++ b/node/test/chan_drain.js
@@ -249,6 +249,7 @@ allocCluster.test('drain server with a few incoming (with exempt service)', {
 
     function drained() {
         assert.pass('drain happened');
+        waitForConnRemoved(6, cluster, finish);
         server.close(closed);
     }
 
@@ -278,8 +279,6 @@ allocCluster.test('drain server with a few incoming (with exempt service)', {
                 msg: 'ignoring outresponse.send on a closed connection'
             }, 'expected zero or more sends after close');
         }
-
-        setTimeout(finish, 1);
     }
 
     function finish(err) {
@@ -750,8 +749,8 @@ function waitForConnRemoved(count, cluster, callback) {
     });
 
     function removed() {
-        if (--count === 0) {
-            callback();
+        if (--count <= 0) {
+            callback(null);
         }
     }
 }
--- a/node/test/chan_drain.js
+++ b/node/test/chan_drain.js
@@ -44,6 +44,7 @@ allocCluster.test('immediate drain', {
 
     function drained() {
         assert.pass('immediate drain happened');
+        waitForConnRemoved(6, cluster, finish);
         server.close(closed);
     }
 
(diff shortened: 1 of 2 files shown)


node/test/chan_drain.js
```diff
@@ -739,11 +734,24 @@ function checkFinish(assert, err, cluster, finishCount) {
...
+function waitForConnRemoved(count, cluster, callback) {
+    cluster.channels.forEach(function eachChannel(chan) {
+        var peers = chan.peers.values();
+        peers.forEach(function eachPeer(peer) {
+            peer.on('removeConnection', removed);
+        });
+    });
+
+    function removed() {
+        if (--count === 0) {
```
jcorbin: please defensive: `<= 0`

node/test/chan_drain.js
```diff
@@ -739,11 +734,24 @@ function checkFinish(assert, err, cluster, finishCount) {
...
+            callback();
```
jcorbin: please explicit null arg

node/test/chan_drain.js
```diff
@@ -739,11 +734,24 @@ function checkFinish(assert, err, cluster, finishCount) {
...
-    }
 
-    if (finishCount <= 0) {
         assert.end();
     }
 
     return finishCount;
 }
+
+function waitForConnRemoved(count, cluster, callback) {
```
jcorbin: is there any chance we could just compute the count rather than statically set it? maybe something like:
```
count = 1
  for each chan
    for each peer:
      count += peer.connection.length
      peer.on(..., removed)
removed() // this is to match the "1 to start" above, and ensures that we won't erroneously fire the callback while still iterating chan/peers
```

jcorbin:
great catch, some comments for improvement

node/test/chan_drain.js
anson627: I actually tried it first, the last two test cases timeout, because only 2 out of 6 connections are expected to be removed.

jcorbin: ahh okay, that makes sense I suppose, never mind then :-)

jcorbin:
lgtm :+1:

PR #1324, opened 2015-09-16
more validation for TChannelAsThrift.request

ShanniLi:

### DIFF
--- a/node/as/thrift.js
+++ b/node/as/thrift.js
@@ -119,6 +119,10 @@ TChannelAsThrift.prototype.request = function request(reqOptions) {
     var self = this;
 
     assert(self.channel, 'channel is required for thrift.request()');
+    assert(reqOptions &&
+        reqOptions.type !== 'tchannel.request' &&
+        reqOptions.type !== 'tchannel.outgoing-request',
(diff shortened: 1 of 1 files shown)


anson627:
:+1:

node/as/thrift.js
```diff
@@ -119,6 +119,10 @@ TChannelAsThrift.prototype.request = function request(reqOptions) {
     var self = this;
 
     assert(self.channel, 'channel is required for thrift.request()');
+    assert(reqOptions &&
+        reqOptions.type !== 'tchannel.request' &&
```
jcorbin: why is `reqOptions.type` even a thing? what are you protecting against that we saw here? it seems to me like we're simply trying to say "does it look like they passed a request object in instead of an options object"... and when is "type" ever a valid option?

Raynos: The following typo was hard to debug.

```
tchannelThrift.request(tchannel.request({ ... }), arg1, arg2, arg3, cb)
```

We wanted to do 

```
tchannelThrift.send(tchannel.request({ ... }), arg1, arg2, arg3, cb)
```

Raynos:
lgtm.

PR #1325, opened 2015-09-16
prefer incoming connections for client channels

ShanniLi:


anson627:
:+1:

jcorbin:
lgtm, we should track getting this feature in other languages too

Raynos:
lgtm

PR #1326, opened 2015-09-16
prefer outgoing connections for hyperbahn service channels

ShanniLi:
cc @jcorbin 

PreferOutgoing.getScore should not create new outgoing connections; otherwise, it gets into race condition that ends up creating multiple out connections. That's why we observed 4x connection increase instead of 2x in Ringpop. On the other hand, it is not necessary anymore.

### DIFF
--- a/node/hyperbahn/service_proxy.js
+++ b/node/hyperbahn/service_proxy.js
@@ -225,6 +225,10 @@ function createServiceChannel(serviceName) {
         options.requestDefaults = self.serviceReqDefaults[serviceName];
     }
 
+    if (mode === 'exit') {
+        options.preferConnectionDirection = 'out';
+    }
+
     var svcchan = self.channel.makeSubChannel(options);
     svcchan.serviceProxyMode = mode; // duck: punched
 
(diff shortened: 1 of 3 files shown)


anson627:
:+1:

node/hyperbahn/service_proxy.js
```diff
@@ -225,6 +225,10 @@ function createServiceChannel(serviceName) {
         options.requestDefaults = self.serviceReqDefaults[serviceName];
     }
 
+    if (mode === 'exit') {
+        options.preferConnectionDirection = 'out';
```
jcorbin: remind me please, the default is now no-preference? that is what would make most sense for internal forwarding nodes imo.

node/peer_score_strategies.js
```diff
@@ -64,9 +64,6 @@ PreferOutgoing.prototype.getScore = function getScore() {
     self.lastTier = tier;
     switch (tier) {
         case PreferOutgoing.ONLY_INCOMING:
-            if (!self.peer.channel.destroyed) {
-                self.peer.connectTo();
```
jcorbin: this looks like a separate change from the "use it in hyperbahn intent", please split it out.

node/hyperbahn/service_proxy.js
ShanniLi: This is for service channels, i.e., egress nodes connecting to service processes.

ShanniLi:
Turned into two separate PRs.

PR #1330, opened 2015-09-16
prefer outgoing connections for hyperbahn service channels

ShanniLi:
cc @jcorbin

split from #1326


anson627:
:+1:

PR #1331, opened 2015-09-16
PreferOutgoing.getScore should not create new outgoing connections

ShanniLi:
cc @jcorbin

split from #1326
PreferOutgoing.getScore should not create new outgoing connections; otherwise, it gets into race condition that ends up creating multiple out connections. That's why we observed 4x connection increase instead of 2x in Ringpop. On the other hand, it is not necessary anymore.

### DIFF
--- a/node/peer_score_strategies.js
+++ b/node/peer_score_strategies.js
@@ -64,9 +64,6 @@ PreferOutgoing.prototype.getScore = function getScore() {
     self.lastTier = tier;
     switch (tier) {
         case PreferOutgoing.ONLY_INCOMING:
-            if (!self.peer.channel.destroyed) {
-                self.peer.connectTo();
-            }
             /* falls through */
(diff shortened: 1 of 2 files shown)


anson627:
:+1:

node/peer_score_strategies.js
```diff
@@ -64,9 +64,6 @@ PreferOutgoing.prototype.getScore = function getScore() {
     self.lastTier = tier;
     switch (tier) {
         case PreferOutgoing.ONLY_INCOMING:
-            if (!self.peer.channel.destroyed) {
-                self.peer.connectTo();
```
Raynos: This breaks preferOutGoing ?

ShanniLi:
no, it should still work for the service channels. That's the whole point of removing it. I don't think it is necessary anymore. :)

PR #1332, opened 2015-09-16
support Meta::thriftIDL in TChannelAsThrift

ShanniLi:
r @Raynos @anson627 
cc @blampe

### DIFF
--- a/node/as/thrift.js
+++ b/node/as/thrift.js
@@ -77,10 +77,12 @@ function TChannelAsThrift(opts) {
             return;
         }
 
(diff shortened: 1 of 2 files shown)


Raynos:
https://github.com/uber/tchannel/blob/thrift-idl/thrift/meta.thrift

thriftIDL is not a thing; need to add it to the meta.thrift too.

ShanniLi:
talked to @Raynos, will create a separate PR for https://github.com/uber/tchannel/blob/thrift-idl/thrift/meta.thrift

PR #1335, opened 2015-09-17
update the thrift meta file for thriftIDL

ShanniLi:
r @prashantv @Raynos @junchaowu 
cc @blampe 

@Raynos suggested us to move forward implementing a default handler for thriftIDL ... Node example is here:  #1332

### DIFF
--- a/thrift/meta.thrift
+++ b/thrift/meta.thrift
@@ -5,4 +5,5 @@ struct HealthStatus {
 
 service Meta {
     HealthStatus health()
+    string thriftIDL()
 }
--- a/thrift/meta.thrift
+++ b/thrift/meta.thrift
@@ -5,7 +5,9 @@ struct HealthStatus {
 
 struct ThriftIDLs {
     // map: filename -> contents
-    1: required map<string, string> idls 
+    1: required map<string, string> idls
+    // the entry IDL that imports others
+    2: required string entryPoint
 }
 
 service Meta {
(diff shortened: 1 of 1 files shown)


prashantv:
As I mentioned earlier, can we please think about the interface that is appropriate here
 - `map<string,string>` is a lot more future proof for multi-Thrift services than `string`
 - any non-trivial return should be a `struct`

Raynos:
We can always have `oneThriftIdl` and `manyThriftIdl`

Or we can fix it right now.

prashantv:
Let's fix it now, it's not hard to make a new struct with a map.

Raynos:
Cool let's fix it.

ShanniLi:
Meta is implemented per service, right?

Raynos:
The map is to future proof for import statements.

prashantv:
There is a single `Meta` service for each Hyperbahn service. One hyperbahn service can expose multiple Thrift services though.

We also may have multiple Thrift files due to imports.

ShanniLi:
I see. Makes sense.

ShanniLi:
So, something like the following?
struct ThriftIDLs {
    1: required map<string,string> idls 
}

ThriftIDLs thriftIDL()

prashantv:
yep, I think the map should be `map<string, string> idls`, but I think that's good enough for now

And we should add a comment for the map explaining what idls is (map from filename -> contents)

Raynos:
@ShanniLi 

We need also `2: required entryPoint` to tell tcurl which file is the root that imports the others. Otherwise hacks

ShanniLi:
@Raynos  fixed.

prashantv:
lgtm

PR #1341, opened 2015-09-17
Assert only one version of tchannel exists

Raynos:
The TChannel node library should not be used multiple
times as that's a bad idea.

r: @jcorbin @kriskowal

### DIFF
--- a/node/channel.js
+++ b/node/channel.js
@@ -25,7 +25,7 @@ var version = require('./package.json').version;
 if (typeof global.tchannelVersion === 'string' &&
     version !== global.tchannelVersion
 ) {
-    assert(true,
+    assert(false,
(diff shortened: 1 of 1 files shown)


node/channel.js
```diff
@@ -21,6 +21,19 @@
 'use strict';
 
 var assert = require('assert');
+var version = require('./package.json').version;
+if (typeof global.tchannelVersion === 'string' &&
+    version !== global.tchannelVersion
+) {
+    assert(true,
```
jcorbin: wat, don't you mean assert false, or throw?

Raynos: wat; assert(false lol

kriskowal:
:ok:

PR #1345, opened 2015-09-18
Drop tcurl and tcap deps

jcorbin:


kriskowal:
Outlived necessity :+1:

Raynos:
@jcorbin

This should break tests. Please fix child process test >:3

PR #1346, opened 2015-09-18
peer selection should favor least outgoing reqs

Raynos:
We had a bug in PreferIncoming where we "generalized" the direction
too much.

We should always favor a peer with least outgoing because we want to
measure and pick the peer with the lowest P99 latency which we are doing
by least pending requests (i.e. more pending requests === bigger latency)

r: @jcorbin @kriskowal

cc @shannili

### DIFF
--- a/node/peer.js
+++ b/node/peer.js
@@ -390,7 +390,7 @@ TChannelPeer.prototype.makeOutConnection = function makeOutConnection(socket) {
     return conn;
 };
 
-TChannelPeer.prototype.pendingWeightedRandom = function pendingWeightedRandom(direction) {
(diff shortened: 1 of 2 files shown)


kriskowal:
:ok:

jcorbin:
Good catch, I must've missed it while reviewing @ShanniLi 's recent work that we even added a direction argument down the pending weight path... this was always intended to be out only, pending in doesn't mean much from a peer selection standpoint.

jcorbin:
also: lgtm, land it once tests green

ShanniLi:
:+1:

PR #1347, opened 2015-09-18
Take busy frames into account for peer selection

Raynos:
This will penalize a peer that has busy frames by
giving him more pending counts and selecting him less.

Once that peer stops returning busy frames or the
tombstones time out we will start selecting him again.

This allows Hyperbahn to route requests around a busy
worker (same with xlate) without having to have retries
on.

In the long term this will make less requests fail
when you have 100 exit nodes and 4 of them are busy.

r: @jcorbin @kriskowal @rf

cc @anson627

### DIFF
--- a/node/operations.js
+++ b/node/operations.js
@@ -50,7 +50,8 @@ function Operations(opts) {
     };
     self.pending = {
         in: 0,
-        out: 0
+        out: 0,
+        busy: 0
     };
     self.lastTimeoutTime = 0;
 }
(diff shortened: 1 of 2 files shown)


kriskowal:
supremo

anson627:
:+1:

jcorbin:
Good start... but don't we actually want to have this be a wider class of "any retryable error" and not just busy?

Raynos:
@jcorbin 

Agreed; Starting with busy; not sure which other error frames to penalize; maybe Declined.

Raynos:
This will need to rebased on github.com/uber/tchannel-node

PR #1339, opened 2015-09-17
subChannel should be able to set peer preference

ShanniLi:
r @jcorbin @Raynos @kriskowal @anson627

PR #1348, opened 2015-09-18
subChannel should be able to set peer preference

ShanniLi:
r @jcorbin @Raynos @kriskowal 
cc @anson627

This is a new PR from #1339 based on suggestions.

### DIFF
--- a/node/peer.js
+++ b/node/peer.js
@@ -65,14 +65,8 @@ function TChannelPeer(channel, hostPort, options) {
         );
     }
 
-    self.preferConnectionDirection = self.options.preferConnectionDirection;
-    if (self.preferConnectionDirection === 'out') {
-        self.setScoreStrategy(PreferOutgoing);
-    } else if (self.preferConnectionDirection === 'in') {
-        self.setScoreStrategy(PreferIncoming);
-    } else {
-        self.setScoreStrategy(NoPreference);
-    }
+    var direction = self.options.preferConnectionDirection || 'any';
+    self.setPreferConnectionDirection(direction);
 
     function onReport() {
         if (!self.hostPort) {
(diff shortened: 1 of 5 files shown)


node/peer.js
```diff
@@ -65,14 +65,8 @@ function TChannelPeer(channel, hostPort, options) {
...
 
-    self.preferConnectionDirection = self.options.preferConnectionDirection;
-    if (self.preferConnectionDirection === 'out') {
-        self.setScoreStrategy(PreferOutgoing);
-    } else if (self.preferConnectionDirection === 'in') {
-        self.setScoreStrategy(PreferIncoming);
-    } else {
-        self.setScoreStrategy(NoPreference);
-    }
+    var direction = self.options.preferConnectionDirection || 'any';
```
kriskowal: 'any' is nice.

node/peer.js
```diff
@@ -95,6 +89,24 @@ function TChannelPeer(channel, hostPort, options) {
 
 inherits(TChannelPeer, EventEmitter);
 
+TChannelPeer.prototype.setPreferConnectionDirection = function setPreferConnectionDirection(direction) {
+    var self = this;
+    if (self.preferConnectionDirection === direction) {
+        return;
+    }
```
kriskowal: Given that @jcorbin has suggested that each peer should only have on strategy since that is the norm in current practice, perhaps we should assert that all subchannels using that subchannel agree on a single strategy, or warn if they conflict. We should definitely not allow subchannels to thrash on strategy.

ShanniLi: Yes, that is commonly the case except for a few corner cases we have to support. For example, an advertisement request will initiate a peer as 'any' direction. If it happens to reach the exit node for the service, this peer should be changed to outgoing later.

node/test/peer_strategies.js
```diff
@@ -207,6 +207,67 @@ allocCluster.test('prefer incoming selects the incoming peer when outgoing is pr
     }
 });
 
+allocCluster.test('subChannel: prefer incoming selects the incoming peer when outgoing is present' , {
+    numPeers: 3,
+    skipEmptyCheck: true,
+    preferConnectionDirection: 'out'
```
jcorbin: interesting, how does this allocCluster test option work? I think this should be under something like `channelOptions` if you actually want it, but shouldn't be having an effect as is?

ShanniLi: If channelOptions is not provided, opts is used.
https://github.com/uber/tchannel/blob/master/node/test/lib/alloc-cluster.js#L58

Raynos:
This will need to be rebased on github.com/uber/tchannel-node

PR #1354, opened 2015-09-18
relay latency stat

rf:
@Raynos @jcorbin 

Did I do it right? This is just v0 of this concept; v1 would be a stat for each of: in-out, out-service, out-in, in-client.

### DIFF
--- a/node/channel.js
+++ b/node/channel.js
@@ -132,6 +132,8 @@ function TChannel(options) {
     self.connectionsClosedStat = self.defineCounter('tchannel.connections.closed');
     // self.connectionsBytesRcvdStat = self.defineCounter('connections.bytes-recvd');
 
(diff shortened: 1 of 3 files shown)


node/channel.js
```diff
@@ -132,6 +132,8 @@ function TChannel(options) {
     self.connectionsClosedStat = self.defineCounter('tchannel.connections.closed');
     // self.connectionsBytesRcvdStat = self.defineCounter('connections.bytes-recvd');
 
+    self.relayLatencyStat = self.defineTiming('tchannel.relay.latency');
```
Raynos: This is the bad pattern. use `emitFastStat`

Raynos:
#LEGIT.

Raynos:
#DOIT

jcorbin:
nice

PR #1355, opened 2015-09-18
Hyperbahn unadvertise should not return error when it succeeds

anson627:
Most of time remote socket closes immediately after unadvertise request is made, and shows up as a reset error, which actually means success.

@ShanniLi @Raynos @jcorbin

### DIFF
--- a/node/hyperbahn-client.js
+++ b/node/hyperbahn-client.js
@@ -380,10 +380,8 @@ function unadvertise(opts) {
     timers.clearTimeout(self._advertisementTimer);
     timers.clearTimeout(self.advertisementTimeoutTimer);
     self.latestAdvertisementResult = null;
-    self.state = States.UNADVERTISED;
-    self.emit('unadvertised');
     function unadvertiseInternalCb(error, result) {
-        if (error) {
+        if (error && error.type !== 'tchannel.connection.reset') {
             self.logger.warn('HyperbahnClient: unadvertisement failure', {
                 error: error,
                 serviceName: self.serviceName,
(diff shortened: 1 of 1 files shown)


ShanniLi:
lgtm

node/hyperbahn-client.js
```diff
@@ -380,10 +380,8 @@ function unadvertise(opts) {
     timers.clearTimeout(self._advertisementTimer);
     timers.clearTimeout(self.advertisementTimeoutTimer);
     self.latestAdvertisementResult = null;
-    self.state = States.UNADVERTISED;
-    self.emit('unadvertised');
     function unadvertiseInternalCb(error, result) {
-        if (error) {
+        if (error && error.type !== 'tchannel.connection.reset') {
```
Raynos: :-1: wrong fix.

Please fix Hyperbahn backend to drain connection rather then close.

node/hyperbahn-client.js
```diff
@@ -391,6 +389,8 @@ function unadvertise(opts) {
             });
             return;
         }
+        self.state = States.UNADVERTISED;
```
Raynos: This is a good change though.

node/hyperbahn-client.js
anson627: I have a plan to implement the drain, but the drain only applies to request whose service name is not hyperbahn, right?

anson627: when hyperbahn backend receive unad request, it will close and delete the peer, which causes all connections of the peer to close. And if connection reset error response comes back earlier (sometimes but not always) than unad success response, client side will see this error. Draining unad request will turn success response into declined error, which again show up as an error at client side.

Raynos: we can just do `peer.drainAllRequests()` should not be hard.

Raynos:
This PR will need to be rebase on github.com/uber/tchannel-node

PR #1356, opened 2015-09-19
test for relay latency stat

rf:
@jcorbin @Raynos 

Simple test for the relay latency stat


node/test/relay-latency-stat.js
```diff
@@ -0,0 +1,101 @@
...
+        assert.equal(String(arg2), 'foo', 'expected arg2');
+        assert.equal(String(arg3), 'bar', 'expected arg3');
+
+        client.close();
+
+        process.nextTick(checkStat);
+
+        function checkStat() {
+            assert.ok(
+                findStat(stats, 'tchannel.relay.latency', 'timing'),
```
Raynos: would be better if we just `assert.deepEqual(stats, [ /* all the things */ ])`

rf: There's a ton of stats that are emitted and I'm not sure we can trust ordering -- inbound.request.size, inbound.calls.recvd, outbound.request.size, outbound.calls.per-attempt-latency, etc. Probably good to verify all of those too but I think its out of scope of this pr

Raynos: I think we can trust the order. 

We really want to assert about ALL those stats.

Mostly to create a failing test for lazy relay because lazy relay does not have stats.

node/test/fixtures/relay-stats.js
```diff
@@ -0,0 +1,187 @@
...
+    var parts = value.split(':');
+    assert.ok(parts.length === 2, 'value ' + value + ' splits into two parts');
+    assert.ok(parts[0] === '127.0.0.1', value + ' is a lo host:port');
+    assert.ok(
+        parseInt(parts[1]).toString() === parts[1],
+        value + ' has number port'
+    );
+}
+
+var fixture = module.exports.fixture = [
```
Raynos: let's move this fixture array into the other test file like all the other request-stats tests etc.

You can keep the generic `verify` function here if you want.

Raynos:
lgtm.

Raynos:
Also fix tests that are broke.

Raynos:
shipping it.

Raynos:
relay respects ttl flapped (lol)

Raynos:
Cant reproduce flap locally :(

PR #1358, opened 2015-09-21
Use local version of tcollector.thrift

Raynos:
We should not reach out into tchannel. In the future we can
use idl-store to fetch this file.

r: @jcorbin @rf


jcorbin:
:+1:

rf:
:+1:

PR #1359, opened 2015-09-21
do not spam log config file change in loadSync()

Raynos:
This gets really noisy in the tests and is not useful.

r: @jcorbin @rf


jcorbin:
:soon:

PR #1361, opened 2015-09-21
Fix test/relay: less flaps because smaller number

Raynos:
This should make travis better.

r: @jcorbin @rf @shannili


rf:
:+1: perfect

PR #1362, opened 2015-09-21
test: fix addrinuse flap

rf:
@Raynos

### DIFF
--- a/hyperbahn/clients/index.js
+++ b/hyperbahn/clients/index.js
@@ -54,7 +54,7 @@ function ApplicationClients(options) {
     }
 
     var self = this;
-    var config = options.config;
+    var config = self.config = options.config;
 
     // We need to move away from myLocalIp(); this fails in weird
     // ways when moving around and changing wifi networks.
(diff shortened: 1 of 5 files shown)


Raynos:
This is not enough.

We also have to fix TestCluster.

hyperbahn/clients/index.js
```diff
@@ -54,7 +54,7 @@ function ApplicationClients(options) {
     }
 
     var self = this;
-    var config = options.config;
+    var config = self.config = options.config;
```
Raynos: do not want `self.config` pattern.

hyperbahn/clients/index.js
```diff
@@ -270,13 +256,43 @@ function bootstrap(cb) {
...
+
+    self.setupChannel(setupDone);
 
-    function onListen() {
-        if (self.autobahnHostPortList) {
-            self.ringpop.bootstrap(self.autobahnHostPortList, cb);
-        } else {
-            process.nextTick(cb);
-        }
+    function setupDone() {
```
Raynos: handle error ?

hyperbahn/test/lib/test-cluster.js
```diff
@@ -92,36 +90,39 @@ inherits(TestCluster, EventEmitter);
...
+
+        for (i = 0; i < self.apps.length; i++) {
             self.hostPortList[i] = self.apps[i].hostPort;
         }
 
+        self.ringpopHosts = self.hostPortList;
+
+        for (i = 0; i < self.apps.length; i++) {
+            self.apps[i].clients.autobahnHostPortList = self.hostPortList;
+            self.apps[i].clients.setupRingpop(ringpopBootstrapped.signal);
```
Raynos: callback vs signal.

Raynos:
Two things:

 - `config.get()` should only be called in constructor.
 - Let's not confuse Callback vs OnceListener functions. OnceListener functions are part of ReadySignal. We should handle errors at the right places and consistently name functions with `onFoo` vs `foo` where the former does not take an `err` and the latter does.

We should also always explicitly call `cb(null)`.

Raynos:
lgtm if passed.

PR #1366, opened 2015-09-21
specify registry in cut.sh

rf:
@raynos

So we don't accidentally try to publish to internal mirror

### DIFF
--- a/node/cut.sh
+++ b/node/cut.sh
@@ -67,5 +67,5 @@ git push origin master --tags
 git push -f origin dev_node
 
 git archive --prefix=package/ --format tgz dev_node >package.tgz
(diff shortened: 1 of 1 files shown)


kriskowal:
:ok:

node/cut.sh
```diff
@@ -67,5 +67,5 @@ git push origin master --tags
 git push -f origin dev_node
 
 git archive --prefix=package/ --format tgz dev_node >package.tgz
-${NPM:-npm} publish package.tgz --tag "${NPM_TAG:-alpha}"
+${NPM:-npm} publish --registry=http://registry.npmjs.org/ package.tgz --tag "${NPM_TAG:-alpha}"
```
jcorbin: probably https

Raynos:
legit.

PR #1376, opened 2015-09-23
go: Simplify trace tests and verify buildZipkinSpan

prashantv:
Seeing trace reports that don't seem to make sense, so I added tests to verify that `buildZipkinSpan` is creating timestamps correctly.

No functional changes in this change.

### DIFF
--- a/golang/trace/zipin_tracereporter_test.go
+++ b/golang/trace/zipin_tracereporter_test.go
@@ -21,7 +21,6 @@
 package trace
 
 import (
-	"net"
 	"testing"
 	"time"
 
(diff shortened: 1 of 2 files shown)


golang/trace/zipin_tracereporter_test.go
```diff
@@ -163,30 +230,21 @@ func withSetup(t *testing.T, f func(ctx thrift.Context, args testArgs)) {
...
-	}
-
-	listener, err := net.Listen("tcp", ":0")
-	if err != nil {
-		return nil, nil, err
+		return nil, err
 	}
 
 	server := thrift.NewServer(tchan)
 	server.Register(gen.NewTChanTCollectorServer(h))
```
junchaowu: do we need to call newServer again? what's the testutils.NewServer

prashantv: `testutils.NewServer` just sets up a listening tchannel (no thrift setup or anything). It is mostly used to get rid of some boilerplate and has disabled logging by default, but you can enable it by adding `--connectionLog`. That way tests are quiet and I can enable logging when debugging tests.

PR #1383, opened 2015-09-25
Add binary annotation support.

junchaowu:
@prashantv @rf

### DIFF
--- a/golang/trace/zipin_tracereporter_test.go
+++ b/golang/trace/zipin_tracereporter_test.go
@@ -343,7 +343,7 @@ func TestBuildBinaryAnnotation(t *testing.T) {
 }
 
 func TestBuildBinaryAnnotationsWithEmptyList(t *testing.T) {
-	result, err := buildBinaryAnnotations([]tchannel.BinaryAnnotation{})
+	result, err := buildBinaryAnnotations(nil)
 	assert.NoError(t, err, "Failed to build binary annotations.")
 	assert.Equal(t, len(result), 0, "BinaryAnnotations should be empty.")
 }
--- a/golang/trace/zipin_tracereporter_test.go
+++ b/golang/trace/zipin_tracereporter_test.go
@@ -337,16 +337,33 @@ func TestBuildBinaryAnnotation(t *testing.T) {
 	tests := generateBinaryAnnotationsTestCase()
 	for _, tt := range tests {
 		result, err := buildBinaryAnnotation(tt.annotation)
-		assert.NoError(t, err)
+		assert.NoError(t, err, "Failed to build binary annotations.")
 		assert.Equal(t, tt.expected, result, "BinaryAnnotation is mismatched.")
 	}
 }
 
-func TestBuildBinaryAnnotations(t *testing.T) {
+func TestBuildBinaryAnnotationsWithEmptyList(t *testing.T) {
+	result, err := buildBinaryAnnotations([]tchannel.BinaryAnnotation{})
+	assert.NoError(t, err, "Failed to build binary annotations.")
+	assert.Equal(t, len(result), 0, "BinaryAnnotations should be empty.")
+}
+
+func TestBuildBinaryAnnotationsWithMultiItems(t *testing.T) {
 	tests := generateBinaryAnnotationsTestCase()
+	var binaryAnns []tchannel.BinaryAnnotation
+	var expectedAnns []*gen.BinaryAnnotation
 	for _, tt := range tests {
-		result, err := buildBinaryAnnotations([]tchannel.BinaryAnnotation{tt.annotation})
-		assert.NoError(t, err)
-		assert.Equal(t, tt.expected, result[0], "BinaryAnnotation is mismatched.")
+		binaryAnns = append(binaryAnns, tt.annotation)
+		expectedAnns = append(expectedAnns, tt.expected)
 	}
+	result, err := buildBinaryAnnotations(binaryAnns)
+	assert.NoError(t, err, "Failed to build binary annotations.")
+	assert.Equal(t, expectedAnns, result, "BinaryAnnotation is mismatched.")
+}
+
+func TestBuildBinaryAnnotationsWithError(t *testing.T) {
+	_, err := buildBinaryAnnotations(
+		[]tchannel.BinaryAnnotation{{Key: "app", Value: []bool{false}}},
+	)
+	assert.Error(t, err, "An Error was expected.")
 }
--- a/golang/trace/zipin_tracereporter_test.go
+++ b/golang/trace/zipin_tracereporter_test.go
@@ -51,7 +51,8 @@ func TestBuildZipkinSpan(t *testing.T) {
 	}
 	span := *tchannel.NewRootSpan()
 	_, annotations := RandomAnnotations()
-	thriftSpan := buildZipkinSpan(span, annotations, nil, endpoint)
+	binaryAnnotations := []tchannel.BinaryAnnotation{{Key: "cn", Value: "string"}}
+	thriftSpan := buildZipkinSpan(span, annotations, binaryAnnotations, endpoint)
 
 	expectedSpan := &gen.Span{
 		TraceId: uint64ToBytes(span.TraceID()),
@@ -60,11 +61,12 @@ func TestBuildZipkinSpan(t *testing.T) {
 			Port:        8888,
 			ServiceName: "testServer",
 		},
-		Name:        "test",
-		Id:          uint64ToBytes(span.SpanID()),
-		ParentId:    uint64ToBytes(span.ParentID()),
-		Annotations: buildZipkinAnnotations(annotations),
-		Debug:       false,
+		Name:              "test",
+		Id:                uint64ToBytes(span.SpanID()),
+		ParentId:          uint64ToBytes(span.ParentID()),
+		Annotations:       buildZipkinAnnotations(annotations),
+		BinaryAnnotations: buildBinaryAnnotations(binaryAnnotations),
+		Debug:             false,
 	}
 
 	assert.Equal(t, thriftSpan, expectedSpan, "Span mismatch")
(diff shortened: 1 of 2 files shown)


golang/trace/zipin_tracereporter_test.go
```diff
@@ -60,11 +61,12 @@ func TestBuildZipkinSpan(t *testing.T) {
...
-		Id:          uint64ToBytes(span.SpanID()),
-		ParentId:    uint64ToBytes(span.ParentID()),
-		Annotations: buildZipkinAnnotations(annotations),
-		Debug:       false,
+		Name:              "test",
+		Id:                uint64ToBytes(span.SpanID()),
+		ParentId:          uint64ToBytes(span.ParentID()),
+		Annotations:       buildZipkinAnnotations(annotations),
+		BinaryAnnotations: buildBinaryAnnotations(binaryAnnotations),
+		Debug:             false,
```
prashantv: can you delete all the Debug lines - they will default to False anyway, and I don't think they're relevant

golang/trace/zipin_tracereporter_test.go
```diff
@@ -267,3 +269,63 @@ func BenchmarkBuildThrift(b *testing.B) {
...
+	ii32 := int32(5)
+	ii16 := int16(5)
+	i := 5
+	b := false
+	f32 := float32(5.0)
+	f64 := float64(5.0)
+	bs := []byte{4, 3, 2}
+
+	tests := []struct {
+		banns    tchannel.BinaryAnnotation
```
prashantv: let's just  call this `annotation`. `banns` looks weird.

golang/trace/zipin_tracereporter_test.go
```diff
@@ -267,3 +269,63 @@ func BenchmarkBuildThrift(b *testing.B) {
 		buildZipkinSpan(span, annotations, nil, endpoint)
 	}
 }
+
+func TestBuildBinaryAnnotation(t *testing.T) {
+	s := "testString"
+	ii64 := int64(5)
```
prashantv: prefer using different values for the different variables in tests

golang/trace/zipin_tracereporter_test.go
```diff
@@ -267,3 +269,63 @@ func BenchmarkBuildThrift(b *testing.B) {
...
+			tchannel.BinaryAnnotation{Key: "bytes", Value: bs},
+			&gen.BinaryAnnotation{Key: "bytes", BytesValue: bs, AnnotationType: gen.AnnotationType_BYTES},
+		},
+	}
+
+	for _, tt := range tests {
+		result := buildBinaryAnnotation(tt.banns)
+		assert.Equal(t, tt.expected, result, "BinaryAnnotation is mismatched.")
+	}
+
```
prashantv: delete blank

golang/trace/zipin_tracereporter_test.go
```diff
@@ -267,3 +269,63 @@ func BenchmarkBuildThrift(b *testing.B) {
...
+	f32 := float32(5.0)
+	f64 := float64(5.0)
+	bs := []byte{4, 3, 2}
+
+	tests := []struct {
+		banns    tchannel.BinaryAnnotation
+		expected *gen.BinaryAnnotation
+	}{
+		{
+			tchannel.BinaryAnnotation{Key: "string", Value: s},
```
prashantv: usually you don't skip the struct field name in the anonymous struct, so it should look something like:

```go
tests := []struct {
  annotation tchannel.BinaryAnnotation
  expected *gen.BinaryAnnotation
}{
  {
    annotation: tchannel.BinaryAnnotation{...},
    expected: &gen.BinaryAnnotation{...},
  },
[...]
}
```

golang/trace/zipin_tracereporter_test.go
```diff
@@ -267,3 +269,63 @@ func BenchmarkBuildThrift(b *testing.B) {
 		buildZipkinSpan(span, annotations, nil, endpoint)
 	}
 }
+
```
prashantv: add tests for `TestBuildBinaryAnnotations` (you can use `buildBinaryAnnotation` in that test)

golang/trace/zipkin_tracereporter.go
```diff
@@ -116,20 +116,71 @@ func buildZipkinSpan(span tc.Span, annotations []tc.Annotation, binaryAnnotation
...
-		Annotations: buildZipkinAnnotations(annotations),
-		Debug:       false,
+		TraceId:           uint64ToBytes(span.TraceID()),
+		Host:              &host,
+		Name:              targetEndpoint.Operation,
+		Id:                uint64ToBytes(span.SpanID()),
+		ParentId:          uint64ToBytes(span.ParentID()),
+		Annotations:       buildZipkinAnnotations(annotations),
+		BinaryAnnotations: buildBinaryAnnotations(binaryAnnotations),
+		Debug:             false,
```
prashantv: delete - the default is false anyway

golang/trace/zipkin_tracereporter.go
```diff
@@ -116,20 +116,71 @@ func buildZipkinSpan(span tc.Span, annotations []tc.Annotation, binaryAnnotation
...
 	}
 
 	return &thriftSpan
 }
 
+func buildBinaryAnnotation(ann tc.BinaryAnnotation) *tcollector.BinaryAnnotation {
+	bann := &tcollector.BinaryAnnotation{Key: ann.Key}
+	switch v := ann.Value.(type) {
+	default:
```
prashantv: let's put the default at the very bottom, and log something if we get to that point.

golang/trace/zipkin_tracereporter.go
```diff
@@ -116,20 +116,71 @@ func buildZipkinSpan(span tc.Span, annotations []tc.Annotation, binaryAnnotation
...
+		bann.DoubleValue = &temp
+	case float64:
+		bann.AnnotationType = tcollector.AnnotationType_DOUBLE
+		bann.DoubleValue = &v
+	}
+	return bann
+}
+
+func buildBinaryAnnotations(anns []tc.BinaryAnnotation) []*tcollector.BinaryAnnotation {
+	zipkinAnns := make([]*tcollector.BinaryAnnotation, len(anns))
```
prashantv: avoid `zipkinAnns` - you have `buildZipkinAnnotations` which refers to the normal `Annotation`. Using that same name here is confusing.

Maybe `tAnns` or something.

golang/trace/zipkin_tracereporter.go
```diff
@@ -116,20 +116,71 @@ func buildZipkinSpan(span tc.Span, annotations []tc.Annotation, binaryAnnotation
...
+	default:
+	case bool:
+		bann.AnnotationType = tcollector.AnnotationType_BOOL
+		bann.BoolValue = &v
+	case int64:
+		bann.AnnotationType = tcollector.AnnotationType_I64
+		bann.IntValue = &v
+	case int32:
+		bann.AnnotationType = tcollector.AnnotationType_I32
+		var temp int64
```
prashantv: just do
```go
temp := int64(v)
bann.IntValue = &temp
```
and the same for all the int types.

golang/trace/zipin_tracereporter_test.go
```diff
@@ -267,3 +271,82 @@ func BenchmarkBuildThrift(b *testing.B) {
...
+	tests := generateBinaryAnnotationsTestCase()
+	for _, tt := range tests {
+		result, err := buildBinaryAnnotation(tt.annotation)
+		assert.NoError(t, err)
+		assert.Equal(t, tt.expected, result, "BinaryAnnotation is mismatched.")
+	}
+}
+
+func TestBuildBinaryAnnotations(t *testing.T) {
+	tests := generateBinaryAnnotationsTestCase()
```
prashantv: You don't want to test single lists here. Here is what I would suggest:
 - test the empty list case
 - test the muliple elements case (pass all the annotations, and verify the expected)
 - add error message to `assert.NoError`

golang/trace/zipin_tracereporter_test.go
```diff
@@ -267,3 +271,82 @@ func BenchmarkBuildThrift(b *testing.B) {
...
+		},
+		{
+			annotation: tchannel.BinaryAnnotation{Key: "bytes", Value: bs},
+			expected:   &gen.BinaryAnnotation{Key: "bytes", BytesValue: bs, AnnotationType: gen.AnnotationType_BYTES},
+		},
+	}
+}
+
+func TestBuildBinaryAnnotation(t *testing.T) {
+	tests := generateBinaryAnnotationsTestCase()
```
prashantv: Add a test for when it's expected to return an error

golang/trace/zipin_tracereporter_test.go
```diff
@@ -267,3 +271,82 @@ func BenchmarkBuildThrift(b *testing.B) {
...
+	_ii32 := int64(6)
+	ii16 := int16(7)
+	_ii16 := int64(7)
+	i := 8
+	_i := int64(8)
+	b := false
+	f32 := float32(5.0)
+	_f32 := float64(5.0)
+	f64 := float64(6.0)
+	_f64 := float64(6.0)
```
prashantv: you shouldn't need this since you `f64`

golang/trace/zipkin_tracereporter.go
```diff
@@ -116,18 +120,75 @@ func buildZipkinSpan(span tc.Span, annotations []tc.Annotation, binaryAnnotation
...
+	case float32:
+		bann.AnnotationType = tcollector.AnnotationType_DOUBLE
+		temp := float64(v)
+		bann.DoubleValue = &temp
+	case float64:
+		bann.AnnotationType = tcollector.AnnotationType_DOUBLE
+		temp := float64(v)
+		bann.DoubleValue = &temp
+	default:
+		return bann, errors.New("Unrecognized date type.")
```
prashantv: The error case should:
 - return nil
 - mention the type that couldn't be converted
 - use lower case error strings

```
return nil, fmt.Errorf("unrecognized data type: %T", v)
```

golang/trace/zipin_tracereporter_test.go
```diff
@@ -267,3 +271,99 @@ func BenchmarkBuildThrift(b *testing.B) {
...
+	tests := generateBinaryAnnotationsTestCase()
+	for _, tt := range tests {
+		result, err := buildBinaryAnnotation(tt.annotation)
+		assert.NoError(t, err, "Failed to build binary annotations.")
+		assert.Equal(t, tt.expected, result, "BinaryAnnotation is mismatched.")
+	}
+}
+
+func TestBuildBinaryAnnotationsWithEmptyList(t *testing.T) {
+	result, err := buildBinaryAnnotations([]tchannel.BinaryAnnotation{})
```
prashantv: can pas `nil` instead of `[]tchannel.BinaryAnnotation{}`

prashantv:
lgtm, just one comment above.

PR #1388, opened 2015-10-02
go: Don't log warning on EOF

prashantv:


golang/connection.go
```diff
@@ -552,7 +553,11 @@ func (c *Connection) SendSystemError(id uint32, span *Span, err error) error {
 
 // connectionError handles a connection level error
 func (c *Connection) connectionError(err error) error {
-	c.log.Warnf("Connection error: %v", err)
+	if err == io.EOF {
+		c.log.Debugf("Connection got EOF")
```
junchaowu: is the EOF thing same as connection closed

prashantv: Yep, just means connection closed. This happens a lot when using tcurl for health checks - since the tcurl process dies, and so it's expected.

junchaowu: so why not just change the error message to connection closed. When user see this EOF, they may confuse.

prashantv: I wanted to make it explicit, because when connection.Close is called, or channel starts graceful close, it logs things like "Channel Close". This way, it's explicit and kept separate from that Close

PR #1390, opened 2015-10-08
use typedef filename in meta.thrift

ShanniLi:
r @prashantv @kriskowal @Raynos @junchaowu @abhinav @blampe ...

per suggestion from @kriskowal

### DIFF
--- a/thrift/meta.thrift
+++ b/thrift/meta.thrift
@@ -3,9 +3,11 @@ struct HealthStatus {
     2: optional string message
 }
 
+typedef string filename
+
 struct ThriftIDLs {
     // map: filename -> contents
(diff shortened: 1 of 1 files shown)


thrift/meta.thrift
```diff
@@ -3,9 +3,11 @@ struct HealthStatus {
...
 }
 
+typedef string filename
+
 struct ThriftIDLs {
     // map: filename -> contents
-    1: required map<string, string> idls
+    1: required map<filename, string> idls
     // the entry IDL that imports others
     2: required string entryPoint
```
kriskowal: required filename entryPoint

prashantv:
lgtm

PR #1391, opened 2015-10-14
Fix JSON-as-TChannel arg2 docs

willsalz:
Please see: https://github.com/uber/tchannel/pull/1206 and https://github.com/uber/tchannel-python/pull/187. I'm going out on a limb here, but I'm assuming that Node is not implemented in the same way…

cc @prashantv @breerly @junchaowu @Raynos

### DIFF
--- a/docs/json.md
+++ b/docs/json.md
@@ -19,7 +19,7 @@ failure.
 For both `call req` and `call res`
 
 - `arg1` must be the method name as defined by [`arg1`][]
-- `arg2` must be the application headers encoded as JSON
+- `arg2` must be the application headers in the format `nh:2 (k~2 v~2){nh}`
 - `arg3` must be the application response as defined by [`arg3`][]
 
 ### `arg1`


Raynos:
:-1: If it's going to be a headers it should probably be JSON encoded headers, i.e. valid JSON but restricted to only a shallow object of key value pairs where they are strings.

prashantv:
+1 to @Raynos's suggestion. Go treats headers as `map<string,string>` but encoded with JSON.

Raynos:
cc @breerly 

So we have the following:

 - node arg2 is any valid JSON
 - go arg2 is a map<string, string> encoded as JSON
 - pythong arg2 is a map<string, string> encoded as `nh:2 (k~2 v~2){nh}`
 - java arg2 is a map<string, string> encoded as JSON
 - tchannel spec; arg2 is a map<string, string> encoded as JSON

At this point we need to make a binary breaking change somewhere. I want to avoid making as many binary breaking changes as possible.

hyperbahn uses as=json for the advertise protocol so we cannot break hyperbahn.

I recommend that we do the following:

 - fix node so that arg2 is a map<string, string> encoded as JSON
 - fix python so that arg2 is a map<string, string> encoded as JSON

Raynos:
Looks like @abhinav is working on updating python.

abhinav:
RIP

PR #1395, opened 2015-10-22
Clarify the term "message"

kriskowal:
@ShanniLi observes that the Java implementation appears to confuse the terms Frame and Message. This amendment to the protocol document clarifies the distinction.


kriskowal:
The Node.js implementation uses the term "operation" or "req/res" in place of "message". Python uses "message" properly.

r @mranney

ShanniLi:
:+1:

mranney:
I like this clarification.

Raynos:
lgtm.

PR #1396, opened 2015-10-27
Document tchannel service to be implemented by client libraries

prashantv:
I imagine introspection APIs would also live under this tchannel service.

r @blampe @Raynos @kriskowal @jcorbin

### DIFF
--- a/docs/meta.md
+++ b/docs/meta.md
@@ -1,0 +1,9 @@
+
+# TChannel Service
+
+Client libraries should register a default service that will be provide meta
+information about the client library and internal state. This default service
+should be registered under as "tchannel" which can be used without knowing
+the service name of the application hosting the endpoint.
+
+The Thrift schema for this service is specified in [meta.thrift](../thrift/meta./thrift).
--- a/thrift/meta.thrift
+++ b/thrift/meta.thrift
@@ -12,7 +12,13 @@ struct ThriftIDLs {
     2: required filename entryPoint
 }
 
+struct VersionInfo {
+  1: required string language
+  2: required string version
+}
+
 service Meta {
     HealthStatus health()
     ThriftIDLs thriftIDL()
+    VersionInfo versionInfo()
 }


thrift/tchannel.thrift
```diff
@@ -0,0 +1,8 @@
+struct ClientLibrary {
```
blampe: Why not in meta?

Raynos: Here's an idea that might help with Meta vs TChannel

Meta is something that either:

 - The application developer can customize
 - Contains optional things.

The TChannel api is something that will be:

 - Implemented by tchannel without user customization
 - Does not contain optional features at all.

For example:

 - Meta health is customized by users so belongs in Meta
 - Meta thriftIdl only applies to thrift encoding so belongs in Meta
 - TChannel library version is not customized and is mandatory so belongs in tchannel introspection.

thrift/tchannel.thrift
```diff
@@ -0,0 +1,8 @@
...
+  1: required string language
+  2: required string version
+}
+
+service TChannel {
+  ClientLibrary clientLibrary()
```
Raynos: Consider `VersionInfo versionInfo()`

Raynos:
:+1: I dont care about naming that much...

thrift/tchannel.thrift
prashantv: Was thinking about that - if one Channel is registering on multiple Hyperbahn services, meta is registered on each of the service names (and the user may want custom health checks for each of those).

This seems more like global TChannel state that only needs to be registered once. However, it doesn't hurt to put this in meta, so happy to move it there if that's preferred.

E.g. it's the difference between exposing these endpoints (if it's part of Meta):
```
s1 Meta::health
s1 Meta::clientLibrary
s2 Meta::health
s2 Meta::clientLIbrary
tchannel Meta::health
tchannel Meta::clientLibrary
```
vs exposing these endpoints
```
s1 Meta::health
s2 Meta::health
tchannel Meta::clientLibrary
```

thrift/tchannel.thrift
blampe: +1

thrift/tchannel.thrift
blampe: Ok my only concern then is that we (humans) would basically have to address workers individually to discover a service's version info?

Raynos: Endoints are cheap. Might as well put everything in meta and expose meta on all service names including tchannel.

prashantv:
Renamed to `versionInfo` and moved to `Meta`.

PR #1402, opened 2015-11-12
Proposal: streaming bit field

Raynos:
To improve our timeout implementations I would like to add an "is streaming" bit flag so we can improve the timeout semantics in Hyperbahn.

 - Without streaming bit, the timeout is from the first call request frame until the last call response cont frame
 - With streaming bit, the timeout is from the first call request frame until the first call response frame

I'll be sending a broader document to define the semantics of Timeouts in more edge cases but getting this PR out to add the new bit field to the protocol as it's slightly orthogonal.

i.e. we can start sending this bit field before we fix the Timeouts.

r: @jcorbin @prashantv @mranney 

cc: @junchaowu @blampe @kriskowal

### DIFF
--- a/docs/protocol.md
+++ b/docs/protocol.md
@@ -885,8 +885,8 @@ the "more fragments remain" flag was not set.
 
 ## Streaming
 
-A call request or call response can be marked as streaming. Streaming is separated from Fragmentation.
+Any request may be fragmented into multiple frames. For non-streaming requests, this can occur for arguments that exceed the size of a single frame, but must ultimately be captured by a single contiguous chunk of memory on both the sender and receiver. Streaming, paired with a streaming argument scheme, implies that the arguments may continue indefinitely and can be processed in chunks. Streaming requests have alternate timeout semantics.
 
-When a request/response is only fragmented the caller is splitting up a large body over multiple frames. You can generally expect the continuation frames to arrive shortly after the initial call frame.
+A non-streaming message times out if the last call response frame is not received within the ttl from the time the first frame request frame was sent.
 
-When a request/response is streamed there are no semantics or expectations about when frames arrive or what the delay between frames is.
+A streaming message times out if the first call response frame does not arrive within the ttl from the time the first frame request was sent. All streaming response handlers should send a call response frame immediately upon receiving the first call request frame, even if that response carries no content. TChannel imposes no constraints on the liveliness of streaming requests or responses thereafter.
--- a/docs/protocol.md
+++ b/docs/protocol.md
@@ -886,7 +886,7 @@ the "more fragments remain" flag was not set.
 
 ## Streaming
 
-A call request or call response can be marked as streaming. Streaming is seperated from Fragmentation.
+A call request or call response can be marked as streaming. Streaming is separated from Fragmentation.
 
 When a request/response is only fragmented the caller is splitting up a large body over multiple frames. You can generally expect the continuation frames to arrive shortly after the initial call frame.
 
--- a/docs/protocol.md
+++ b/docs/protocol.md
@@ -337,8 +337,7 @@ Used to control fragmentation. Valid flags:
 If the fragments flag isn't set, then this is the only/last frame for this
 message id.
 
-If the streaming flag isn't set, then this request has a fragmented body
-instead of a streaming body.
+If the streaming flag is set, then streaming request timeout semantics (time to first response frame) apply instead of the non-streaming semantics (time to last response frame).
 
 The streaming flag `0x02` may only be set on a `CallRequest` or `CallResponse`
 frame. If the flag is set on a `CallRequestCont` or `CallResponseCont` frame
--- a/docs/protocol.md
+++ b/docs/protocol.md
@@ -340,6 +340,10 @@ message id.
 If the streaming flag isn't set, then this request has a fragmented body
 instead of a streaming body.
 
+The streaming flag `0x02` may only be set on a `CallRequest` or `CallResponse`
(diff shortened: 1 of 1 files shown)


docs/protocol.md
```diff
@@ -875,3 +879,11 @@ the "more fragments remain" flag was not set.
 | 0x03 | 1  | flags:1=0x1, ttl:4=0x2328, tracing:24=0x1,0x2,0x3, traceflags:1=0x1, service~1=0x5"svc A", nh:1=0x1, hk~1=0x1"k", hv~1=0xA"abcdefghij", csumtype:1=0x2 csum:4=0xBEEF arg1~2=0x2<2 bytes> | sending arg1
 | 0x13 | 1  | flags:1=0x1, csumtype:1=0x2 csum:4=0xDEAD arg1~2=0x2<2 bytes> arg2~2=0x2<2 bytes> | sending arg2
 | 0x13 | 1  | flags:1=0x0, csumtype:1=0x2 csum:4=0xF00F arg2~2=0x0<0 bytes> arg3~2=0x8<8 bytes> | complete
+
+## Streaming
+
+A call request or call response can be marked as streaming. Streaming is seperated from Fragmentation.
```
prashantv: nit: typo, seperated -> separated.

prashantv:
lgtm

jcorbin:
- please add limitation that this bit can only be set on initial call req/res frames; it should be a protocol error if this bit is set on a continuation frame
- we might consider adding a (optional? required?) header for "total stream timeout"; perhaps not all streams will have an easy way to pick on tho?

Raynos:
What's the use case for the total stream timeout ? What are the semantics.

Streams sound like virtual sockets to me. Is it similar to a socket timeout ? That might have value. Do we want a total socket timeout ?

breerly:
:+1:

jcorbin:
when using a stream to implement a transaction, I could see wanted an overall timeout... however that may be better left to the application developer, let's not add the header just yet.

kriskowal:
Timeout should have the same semantics for streaming and non-streaming requests: the time from first request frame to the last response frame. However, there should be a notion of an indefinitely streaming request, or simply an enormous timeout. Liveliness is a separate concern, and would have independent values for request streams and response streams.

For large requests, pagination is an idiomatic option for breaking large responses into chunks in the confines of the ttl/retry approach to RPC.

Raynos:
Correct. the "notion of an indefinitely streaming request" is exactly this bit flag. 

The semantics of timeouts (which must always be <2 minutes) when this bit flag is set is to fail the request if the first call response frame is not returned within the ttl.

docs/protocol.md
```diff
@@ -875,3 +883,11 @@ the "more fragments remain" flag was not set.
 | 0x03 | 1  | flags:1=0x1, ttl:4=0x2328, tracing:24=0x1,0x2,0x3, traceflags:1=0x1, service~1=0x5"svc A", nh:1=0x1, hk~1=0x1"k", hv~1=0xA"abcdefghij", csumtype:1=0x2 csum:4=0xBEEF arg1~2=0x2<2 bytes> | sending arg1
 | 0x13 | 1  | flags:1=0x1, csumtype:1=0x2 csum:4=0xDEAD arg1~2=0x2<2 bytes> arg2~2=0x2<2 bytes> | sending arg2
 | 0x13 | 1  | flags:1=0x0, csumtype:1=0x2 csum:4=0xF00F arg2~2=0x0<0 bytes> arg3~2=0x8<8 bytes> | complete
+
+## Streaming
+
+A call request or call response can be marked as streaming. Streaming is separated from Fragmentation.
```
kriskowal: “Any request may be fragmented into multiple frames. For non-streaming requests, this can occur for arguments that exceed the size of a single frame, but must ultimately be captured by a single contiguous chunk of memory on both the sender and receiver. Streaming, paired with a streaming argument scheme, implies that the arguments may continue indefinitely and can be processed in chunks. Streaming requests have alternate timeout semantics.”

“A non-streaming message times out if the *last call response frame* is not received within the ttl from the time the first frame request frame was sent.”

“A streaming message times out if the *first call response frame* does not arrive within the ttl from the time the first frame request was sent. All streaming response handlers should send a call response frame immediately upon receiving the first call request frame, even if that response carries no content. TChannel imposes no constraints on the liveliness of streaming requests or responses thereafter.”

docs/protocol.md
```diff
@@ -332,10 +332,18 @@ Used to control fragmentation. Valid flags:
 | flag   | description
 | ------ | -----------
 | `0x01` | more fragments follow
+| `0x02` | is request streaming
 
 If the fragments flag isn't set, then this is the only/last frame for this
 message id.
 
+If the streaming flag isn't set, then this request has a fragmented body
+instead of a streaming body.
```
kriskowal: Fragmentation and streaming are orthogonal.

“If the streaming flag is set, then streaming request timeout semantics (time to first response frame) apply instead of the non-streaming semantics (time to last response frame).”.

kriskowal:
I’ve posted some recommended prose changes. Otherwise, I buy the approach. The streaming bit means indefinite. It does not disable the ttl, but does relax it in a way that implies handlers have to at least ack immediately.

kriskowal:
@jcorbin To your point, perhaps this timeout behavior is orthogonal to whether the request is streaming. Perhaps we should rename the flag "nottl" or "ttlack" and let it be used to relax the timeout for *some* but not all streaming requests. The existing semantics are sufficient for streaming requests with a long timeout for completion.

Raynos:
I've addressed all feedback.

PR #1404, opened 2015-11-20
Link to the right meta.thrift doc

achals:


prashantv:
lgtm, thanks for the fix

PR #1405, opened 2015-11-24
document rd transport header

Raynos:
The `xs` header will be used to route to the cross datacenter service and resolve this request to the write datacenter so we can send requests to the write master.

cc @jcorbin @mranney @blampe

### DIFF
--- a/docs/protocol.md
+++ b/docs/protocol.md
@@ -634,6 +634,7 @@ valid or not in a call req or res.  Following sections will elaborate on details
 | `se`  | Y   | N   | Speculative Execution
 | `fd`  | Y   | Y   | Failure Domain
 | `sk`  | Y   | N   | Shard key
+| `xs`  | Y   | N   | Cross scope request
 
 ### Transport Header `as` -- Arg Scheme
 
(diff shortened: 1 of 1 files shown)


Raynos:
Also cc @prashantv as we probably will add scope support to go first.

jcorbin:
Capturing our conversation:
- please let's pivot this to a "Custom Routing" (aka `cr`) header
- the value of this header MUST be a service name that will implement the custom routing
- hyperbahn then can continue making purely fast synchronous routing decisions, while allowing unbounded implementation complexity encapsulated behind the service pointed to

Raynos:
Updated document to talk about `cr` header instead.

prashantv:
Would we expect that other transport headers be used to set the namespace?

Can we update the PR name to match the new field.

Raynos:
We expect other APPLICATION headers are used to give hints for where it should be routed.

This transport header is purely about "which service implements custom routing"

docs/protocol.md
```diff
@@ -727,6 +728,16 @@ For example you may want to keep some in memory state, i.e. cache, aggregation.
...
 that has ownership for the shard key.
 
+### Transport Header `cr` -- Custom routing
+
+The custom routing request header must be set to a valid service name.
+
+When the custom routing header is set we will route to the service name
+in the `cr` header instead of the service name in the call request.
+
+It is expected that the service in the `cr` header is capable of
```
blampe: This makes it sound even more similar to `sk`. I'm having trouble seeing the difference between these two.

Raynos: `sk` contains a uuid. `cr` contains a service name.

`sk` is ignored by hyperbahn but respected by ringpop. `cr` is respected by hyperbahn but ignored by ringpop...

blampe: If we want a request to hit a service identified by `cr` so that the app can then route it based on `arg1`, isn't that the same as sending the request to `arg1 = cr` and letting the app route based on `sk`? Is the uuid constraint on `sk` explicit or just expected by ringpop?

rf:
I renamed it to `rd`

Raynos:
Good. ship it.

PR #1406, opened 2015-11-25
Frame size can only be 65535 bits because the size variable is 16 bits

ShanniLi:
cc: @blampe @jcorbin @kriskowal @mranney @prashantv @Raynos 

I believe 65535 is what we are actually using.


junchaowu:
Yes we do.

jcorbin:
Good catch, :+1:

PR #1407, opened 2015-12-01
Remove ancillary projects from components list.

breerly:
These projects are in-service to tchannel-node, listing them here convolutes the bigger pictures.


abhinav:
:ok:

PR #1409, opened 2015-12-04
document background messages

Raynos:
In the TTL documentation we talk about decrementing and
propagating the TTL.

This feature is currently unimplemented in all languages
and even if implemented might have suprising behaviour
for anyone trying to make background / async requests.

It should be possible for you to implement a low latency
handler that returns an immediate response and then makes
a background request.

I've updated the protocol document to mention background
requests and how they do not have TTL propagation.

r: @jcorbin @prashantv @blampe

### DIFF
--- a/docs/protocol.md
+++ b/docs/protocol.md
@@ -770,6 +770,10 @@ TTLs are not sent back with the response. It is up to the calling service to
 track the time spent with dependent requests and validate the incoming deadline
 before dispatching every new dependent request.
 
+An RPC message may be marked as backgrounded in the rpc library. For backgrounded
+messages we do not decrement or propagate the TTL. The TTL for all outgoing
+backgrounded messages is the specified timeout when the message is sent.
+
 ## Checksums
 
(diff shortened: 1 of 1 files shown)


docs/protocol.md
```diff
@@ -770,6 +770,10 @@ TTLs are not sent back with the response. It is up to the calling service to
 track the time spent with dependent requests and validate the incoming deadline
 before dispatching every new dependent request.
 
+An RPC message may be marked as backgrounded in the rpc library. For backgrounded
+messages we do not decrement or propagate the TTL. The TTL for all outgoing
+backgrounded messages is the specified timeout when the message is sent.
```
kriskowal: An RPC message may effect further RPC messages. By default, if a request has a parent request, its trace extends the parent's trace and its ttl is bound by the parent's ttl. The child request may opt-in for either a new trace, a new ttl, or both.

Raynos:
RIP

PR #1421, opened 2017-08-09
Add arguments to health endpoint to determine source

prashantv:
This matches the health endpoint we've defined internally in YARPC.

This change is backwards compatible since we're adding new optional
fields to the request and response.

### DIFF
--- a/idl/github.com/uber/tchannel/meta.thrift
+++ b/idl/github.com/uber/tchannel/meta.thrift
@@ -1,6 +1,23 @@
+enum State {
+    REFUSING = 0,
+    ACCEPTING = 1,
+    STOPPING = 2,
+    STOPPED = 3,
+}
+
+enum RequestType {
+    LEGACY = 0,
+    TRAFFIC = 1,
+}
+
+struct HealthRequest {
+    1: optional RequestType type
+}
+
 struct HealthStatus {
     1: required bool ok
     2: optional string message
+    3: optional State state
 }
 
 typedef string filename
@@ -22,7 +39,7 @@ struct VersionInfo {
 }
 
 service Meta {
-    HealthStatus health()
+    HealthStatus health(1: HealthRequest hr)
     ThriftIDLs thriftIDL()
     VersionInfo versionInfo()
 }
--- a/idl/github.com/uber/tchannel/meta.thrift
+++ b/idl/github.com/uber/tchannel/meta.thrift
@@ -1,23 +1,34 @@
-enum State {
+// The HealthState provides additional information when the
+// health endpoint returns !ok.
+enum HealthState {
     REFUSING = 0,
     ACCEPTING = 1,
     STOPPING = 2,
     STOPPED = 3,
 }
 
-enum RequestType {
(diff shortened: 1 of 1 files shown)


idl/github.com/uber/tchannel/meta.thrift
```diff
@@ -22,7 +39,7 @@ struct VersionInfo {
 }
 
 service Meta {
-    HealthStatus health()
+    HealthStatus health(1: HealthRequest hr)
```
breerly: Is adding this argument backwards compatible? I had thought that new params were breaking...

prashantv: Adding new parameters should be backwards compatible since parameters are optional.

cc @abhinav to confirm.

prashantv:
Thanks for the review @breerly 

> Assuming the implementation will follow in another PR.

Yeah each language that wants to implement this will need to do so in the language specific repos.

idl/github.com/uber/tchannel/meta.thrift
breerly: I think it will be wire compatible, and from this PR that is probably enough.

I know that Go will fail to compile when a new param is added, e.g. you need to manually add the param to the handlers implementing the interface.

abhinav: Yeah, this change is backwards compatible over-the-wire and with existing services which is really all that matters. The only breakage will be that your old code will fail to compile until you address the new parameter. This is completely acceptable.

abhinav: @prashantv this needs some basic docs. Without knowing that the parameter is new, I would assume that `hr` is always expected to be non-nil.

idl/github.com/uber/tchannel/meta.thrift
```diff
@@ -1,6 +1,23 @@
...
+    STOPPED = 3,
+}
+
+enum RequestType {
+    LEGACY = 0,
+    TRAFFIC = 1,
+}
+
+struct HealthRequest {
+    1: optional RequestType type
```
abhinav: What about HealthRequestType? In case we ever add another function which has different request types.

idl/github.com/uber/tchannel/meta.thrift
```diff
@@ -1,6 +1,23 @@
+enum State {
+    REFUSING = 0,
+    ACCEPTING = 1,
+    STOPPING = 2,
+    STOPPED = 3,
+}
+
+enum RequestType {
```
Raynos: can we have a comment about what legacy & traffic mean ?

prashantv: Yep, added some comments here. I actually renamed it to `PROCESS` instead of `LEGACY` which doesn't really provide any useful information.

idl/github.com/uber/tchannel/meta.thrift
prashantv: 👍

PR #1424, opened 2018-07-19
add chinese translation

1046102779:


prashantv:
Thanks for the change, but I don't think this is something we need to add, sorry.

